PR #7, opened 2021-08-22
Update lich core to support monk stamina use

OSXLich-Doug:
This update modifies Spell cast routines to detect the Feat 'mental acuity' and to enable monk casting of minor mental spells with stamina instead of mana.  It is a moderate impact update, requires testing. 

Barring logical errors or commentary from members, this PR will be merged in 7 days (29 Aug) to allow ample testing time.


OSXLich-Doug:
Added Naijin based code to extend official ID to Effects for Spells, Buffs, Debuffs and Cooldowns.

PR #13, opened 2021-09-03
Adding lich5-update

OSXLich-Doug:
This process replaces the fragile one-off mechanisms that were temporarily used in update-lich5

It is supported by a json file in the data subdir - format is important.


OSXLich-Doug:
Removing due to logic error - will create new branch / PR

PR #20, opened 2021-09-07
Reposition 5.0.18 featureset

OSXLich-Doug:
This PR repositions the 5.0.18 features.  It includes:

1) Enables monks with FEAT Mental Acuity to reliably cast from stamina, rather than mana.  It is profession future-proof in that it only checks the existence of the FEAT, not the profession
2) Updates the XMLData dialogs to conform to data as provided in XML by SIMU
3) Improves the backward compatible XMLData method `active_spells` to return a simple hash as previously coded but include all four categories of PSM3 XML as sent by SIMU
4) Updates the Effects module to consume the XML data as provided
5) Simplifies the Effects calls to permit the method `.active?` to respond to spell / effect number, spell / effect name irrespective of case used, and type Symbol for spell / effect name irrespective of case used
6) Simplifies the Effects call to method `.to_h` to conform to output provided by XMLData.active_spells, or at optional argument `complex` to display the same XML output consistent with the current SIMU XML format
7) Creates a new method `.temp_putup` expecting one argument - the effect - to permit split-second or game-delayed response of XML data to be tracked until the next SIMU XML update is received

### DIFF
--- a/lich.rbw
+++ b/lich.rbw
@@ -35,7 +35,7 @@
 # Lich is maintained by Matt Lowe (tillmen@lichproject.org)
 # Lich version 5 and higher maintained by Elanthia Online and only supports GTK3 Ruby
 
-LICH_VERSION = '5.0.16'
+LICH_VERSION = '5.0.18'
 TESTING = false
 
 if RUBY_VERSION !~ /^2|^3/
@@ -1504,7 +1504,28 @@ class XMLParser
 
   # for backwards compatability
   def active_spells
-    @dialogs["Active Spells"]
+    z = {}
+    XMLData.dialogs.sort.each do |a,b|
+      b[:names].each do |k,v|
+        case a
+        when /Active Spells|Buffs/
+          z.merge!(k => b[:data].fetch(v))
+        when /Cooldowns/
+          if z.has_key?(k)
+            z.merge!("#{k} - Recovery" => b[:data].fetch(v))
+          else
+            z.merge!(k => b[:data].fetch(v))
+  end
+        when /Debuffs/
+          if z.has_key?(k)
+            z.merge!("#{k} - Debuff" => b[:data].fetch(v))
+          else
+            z.merge!(k => b[:data].fetch(v))
+          end
+        end
+      end
+    end
+    z
   end
 
   def reset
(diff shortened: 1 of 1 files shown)


OSXLich-Doug:
This PR is now active and available for testing, defect repair and suggestions on simplification.  Barring any noted defects or required changes, it will be deployed as the current Lich5 version Friday Sep 10 at 9 AM EDT.

lich.rbw
```diff
@@ -1544,16 +1565,35 @@ class XMLParser
 
   DECADE = 10 * 31_536_000
 
+def init_psm3_dialog(kind)
```
ondreian, line 1568: this is a unnecessary, breaking change that needs to be avoided.

`Id` and `Name` can co-exist in a flat hash, if you want to maintain some sort of mapping, we should add a new hash called `id_name` or something.

lich.rbw
```diff
@@ -8736,20 +8758,40 @@ module Effects
       @dialog = dialog
     end
 
-    def to_h
-      XMLData.dialogs.fetch(@dialog, {})
+    def to_h(complex: false)
```
ondreian, line 8761: this should not be done, it should be the same as before

ondreian (changes requested):

lich.rbw
```diff
@@ -8736,20 +8758,40 @@ module Effects
...
+        expiry = to_h[:data].fetch(id_or_name, 0) if id_or_name.instance_of?(Integer)
+        expiry = to_h[:data].fetch(to_h[:names].transform_keys(&:downcase).fetch(id_or_name.downcase), 0) if id_or_name.instance_of?(String)
+        expiry = to_h[:data].fetch(to_h[:names].transform_keys(&:downcase).fetch(id_or_name.to_s.gsub(/[\_]/, ' ').split.map(&:downcase).join(' ')), 0) if id_or_name.instance_of?(Symbol)
       expiry.to_i > Time.now.to_i
+      rescue
+        false
     end
   end
 
+  def temp_putup(effect:)
```
ondreian, line 8784: This is an unnecessary addition if the breaking changes are reverted

lich.rbw
OSXLich-Doug: I'm still exploring the simplification discussion, and may opt to follow this path however (see Lich Dev discord) I have demonstrated this is not 'breaking'.  I would appreciate your review and concurrence then mark the discussion resolved purely from the 'breaking' context.

lich.rbw
OSXLich-Doug: I may support this view based on the offered simplifications - presuming I get them in and can make them work to both existing and future needs.  However, at present, this feature does not break anything.  It may require modification / extension depending on the open question of best `active_spell` display methods, but can be changed anytime including after PR merge. I would appreciate your review of this commend and cocurrence, then mark the discussion resolved purely in the context of 'should be the same' given the stated design goal.  See exchange in Lich Dev server for detail.

lich.rbw
OSXLich-Doug: If I can successfully incorporate the suggested simplifications, I will test - however, it is unclear how even the suggested simplifications will address a lag event or delay from SIMU providing the XML data.  This holds true today in 5.0.16, and I would expect holds true with any changes in 5.0.18 as it is dependent on SIMU timing, not Lich coordination.  This is likely to stay unless we decide it should be a script responsibility and not exposed in core Lich.

lich.rbw
OSXLich-Doug: With simplification suggested, this init procedure is not required.  Agreed, removing, resolved.

lich.rbw
OSXLich-Doug: I have opted to extend the `to_h` def as shown, but with a better understanding of what you are saying here, O, I have reverted the `to_h` non extended to be specific only to the data in parse_psm3_progressbar.  As long as the extension concept passes muster, this also is resolved.

lich.rbw
OSXLich-Doug: I will leave open the question around managing SIMU xml data delays (game lag), but maintain this may become something we want to include.  I do not see the direct correlation between this and breaking changes, but will continue to explore.

OSXLich-Doug:
I believe that this PR, as provided, does not meet with the design goals I had intended and does not meet O's points on breaking changes.  I have recrafted the code to align better to all expectations.  I will close this PR, and provide the updates in a subsequent PR (5.0.19 proposed)

PR #21, opened 2021-09-08
Monks psm3 active spells features

OSXLich-Doug:
This PR repositions the 5.0.18 features as 5.0.19. It includes:

1) Enables monks with FEAT Mental Acuity to reliably cast from stamina, rather than mana. It is profession future-proof in that it only checks the existence of the FEAT, not the profession (5.0.17)
2) Creates an extended XMLData dialog to conform to data as provided in XML by SIMU (5.0.18)
3) Improves the backward compatible XMLData method active_spells to return a simple hash as previously coded but include all four categories of PSM3 XML as sent by SIMU (largely the same) (5.0.19)
4) Updates the Effects module to consume the extended XML data as provided
 -o- Simplifies the Effects calls to permit the method .active? to respond to spell / effect number, spell / effect name irrespective of case used, and type Symbol for spell / effect name irrespective of case used
 -o- Simplifies the Effects call to method .to_h to conform to previous parse_psm3_progressbar output or at optional argument `ext` to display the extended XML output consistent with the current SIMU XML format

The previous submission included a `temp-putup` feature that has been removed, pending further discussions.

### DIFF
--- a/lich.rbw
+++ b/lich.rbw
@@ -35,7 +35,7 @@
 # Lich is maintained by Matt Lowe (tillmen@lichproject.org)
 # Lich version 5 and higher maintained by Elanthia Online and only supports GTK3 Ruby
 
-LICH_VERSION = '5.0.18'
+LICH_VERSION = '5.0.19'
 TESTING = false
 
 if RUBY_VERSION !~ /^2|^3/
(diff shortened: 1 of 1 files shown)


OSXLich-Doug:
Erroneous PR.  Closing to try . . .again?

Issue #35, opened 2021-09-19
Defect: Round Time Calculations in Lich


OSXLich-Doug:
Round times as supported by various waitrt and castwaitrt methods are inaccurate, leading to additional '... wait' messages which slow game response time, and leading to 'spam' approaches due to the realization that wait times often have a longer duration than the game reported roundtime expiry.

Ondreian has provided greatly simplified code for these methods that needs to be incorporated.  The root of the issue remains, however.  The root of the issue is the calculated offset of local time to server time.  This value varies widely between -1 and 1 with infrequent results outside those ranges observed.  As a result, adding the lower range offset results in a shorter waitrt or castwaitrt than is expected  increases the frequencies of the '... wait' message. Adding the upper range of the offset needlessly increases the sleep period for the method and supports the 'spam' approach competitive script authors are using.

To address, the calculated offset should be capped to a max value (recommend -.20) and should be subtracted - thereby resulting in appropriately increasing the calculated waitrt or waitcastrt to the expected value with a significantly improved performance.

OSXLich-Doug:
Code - existing method with Ondreian proposed simplifications

```
def waitrt?
  sleep checkrt
end

def checkrt
  [0, XMLData.roundtime_end.to_f - Time.now.to_f + XMLData.server_time_offset.to_f].max
end

def checkcastrt
  [0, XMLData.cast_roundtime_end.to_f - Time.now.to_f + XMLData.server_time_offset.to_f].max
end

def waitcastrt?
  sleep checkcastrt
end
```

Test results - 1000 waits with login / death / thought (and lnet) messages enabled.

```
[exec4: fastest / slowest so far: 1.3176641464233398 / 3.0029592514038086 of 1000 attempts]
[exec4: number of ...wait messages: 33]
[exec4: logons/logoffs/deaths/thoughts: 50]
[exec4: Average wait time is 2.002129964351654.]
```

OSXLich-Doug:
Code - proposed method (subtract offset)

```
def waitrt?
  sleep checkrt
end

def checkrt
  offset = [ XMLData.server_time_offset - 1.0, -0.20 ].max
  [0, XMLData.roundtime_end.to_f - Time.now.to_f - offset].max
end

def checkcastrt
  offset = [ XMLData.server_time_offset - 1.0, -0.20 ].max
  [0, XMLData.cast_roundtime_end.to_f - Time.now.to_f - offset].max
end

def waitcastrt?
  sleep checkcastrt
end
```

[...]

PR #50, opened 2021-09-27
roundtime / cast roundtime corrections

OSXLich-Doug:
Corrects issue #35


ondreian:
seems good, but same thing about version

Issue #41, opened 2021-09-20
Feat: Allow strings, symbols, integers indifferent query to Effects module


OSXLich-Doug:
Category: User / Script Author convenience

    Feat: Indifferent method to determine results from Effects module
      Impacts: lich.rbw, some existing and many future scripts.  Closes defects in poorly designed .available? methods in PSM category modules (CMan, Shield, Weapon)
      Approach: Create utility module with methods to determine request variable class (Integer, String, Symbol) and appropriately query Effects module.  Potentially create flat hash that holds mappings of ID to NAME to use as basis for normalization.
      To-do: PR for existing prototype
      Needs: testing / feedback
      Currently working: Doug
      Status:
        09/19/2021 - prototype created

OSXLich-Doug:
FYI - the feature updates are contained in the various PSM breakout.rb files (cman.rb, etc.)

@ondreian - Question: The semver will be updated in those files, but there will be no change to the lich.rbw file semver anticipated.  Is this accurate, or should a lib file semver change also be reflected in a lich.rbw semver change?

ondreian:
there should only be 1 semver, the main one in lich.rb

OSXLich-Doug:
Gotcha.  Thanks!

Issue #38, opened 2021-09-19
Feat: Item script indifferent Stow method


OSXLich-Doug:
Category: User / Script Author convenience

  Feat: Convert all stow item methods to indifferent item script response (no regex!)
  Impacts: lich.rbw and many scripts.  Closes pirate-harness and other scripted issues / defects.
  Approach: Use newly implemented stash_hands as basis to call out to indifferent stow
  To-do: prototype for Lich5 (prototype for gem purifier Puritan by Ondreian proposed)
  Needs: prototype
  Currently working: TBD
  Status:
    09/19/2021 - proposed / not prioritized

PR #51, opened 2021-09-27
Staging util and stash files for features updates

OSXLich-Doug:
Staging files for features in issues #38 and #41 

Modifying lich to take advantage of these new files will be a separate PR

### DIFF
--- a/lib/stash.rb
+++ b/lib/stash.rb
@@ -1,0 +1,170 @@
+=begin
+stash.rb: Core lich file for extending free_hands, empty_hands functions in
+  item / container script indifferent method.  Usage will ensure no regex is
+  required to be maintained.
+
+    Maintainer: Elanthia-Online
+    Original Author: Tillmen, Ondreian, others
+    game: Gemstone
+    tags: CORE, spells
+    required: Lich > 5.0.19
+    version: 1.0.0
+
+  changelog:
+    version 1.0.0
+     * Initial release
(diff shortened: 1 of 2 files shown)


lib/stash.rb
```diff
@@ -0,0 +1,170 @@
...
+    def self.find_container(param)
+      GameObj.inv.find do |container|
+        container.name =~ %r[#{param}]
+      end or fail "could not find Container[name: #{param}]"
+    end
+
+    def self.container(param)
+      container = find_container(param)
+      if container.contents.nil?
+        # opening a closed container populates the container's inventory in XML.
```
ondreian, line 30: This can also be done with a simple look, which is 1 command less (in most cases)
We would need a fallback for when it's closed

OSXLich-Doug:
Yes.  I found the screen scroll very unsightly, though, which is why I went
to the close / open strategy.  I thought about doing a silent look as well,
which may be an option.

OSXLich-Doug:
Ok, I think I'll roll with this, instead of close / open strategy.  Feedback welcome, @ondreian 

```
    def self.container(param)
      container = find_container(param)
      result = Lich::Util.quiet_command_xml("look in my #{container}", /In the .*$|That is closed\./) if container.contents.nil?
      fput "open my #{container}" if result.include?('That is closed.')
      return container
    end
```

Issue #40, opened 2021-09-20
Defect:  Lich method MOVE does not respect game SET GROUPMOVEMENT ON


OSXLich-Doug:
Background: Lich's MOVE function does not respect SET GROUPMOVEMENT ON.  As a result group leaders will frequently leave grouped members behind at various obstacles when the expectation would be for the group leader to wait until all group members could continue.
    Impacts: lich.rbw MOVE function, and may also affect mapDB procs / go2 procs (needs evaluation)
    Approach:
    To-do: evaluate mapDB / go2 for impacts.  Propose code solution.
    Needs: unknown
    Currently working: TBD
    Status:
      09/18/2021 - identified / added as defect to register

PR #55, opened 2021-09-27
Respect groupmovement cannot follow result

OSXLich-Doug:
Addresses issue #40


ondreian:
looks good, but please do not bump the version (as per Discord discussion)

PR #96, opened 2022-02-28
Update stash.rb

mrhoribu:
add support for ethereal weapon tattoos


mrhoribu:
Also changed from using `pp` to `echo` for the result of a right_hand due to profanity FE display issues when used.

PR #100, opened 2022-03-06
Infomon updates for Test and Invoker

OSXLich-Doug:
This is a significant change to Infomon that paves the direction for a future rewrite based on the Effects module.  In short, it stops looking for up / down messages from spells and simply takes in the current xmlparser active_spells output.  There are some names that are then transformed to conform with current naming conventions in spell-list.xml and various scripts to limit breaking changes.

Needs thorough testing.

### DIFF
--- a/scripts/infomon.lic
+++ b/scripts/infomon.lic
@@ -9,13 +9,15 @@
           game: Gemstone
           tags: core
       required: Lich > 4.6.10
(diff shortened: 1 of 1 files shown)


OSXLich-Doug:
Closing this.  Not merged.  Better option created.

PR #144, opened 2022-03-12
Update gameobj-data.xml

jscholly:
add sentinel as a grimswarm type


mrhoribu:
gameobj-data.xml is actually maintained over at https://github.com/elanthia-online/scripts/ and is built with a bunch of compilation and tests to ensure proper data format/integrity.

I've gone ahead and opened a PR over there https://github.com/elanthia-online/scripts/pull/619 with this issue. Thanks for catching this and submitting.

PR #147, opened 2022-03-13
Correcting for downstreamhooks that are not strings

OSXLich-Doug:
Xanlin discovery - causing havoc especially in autostarted scripts with hooks


gs4-Xanlin (approved):

gs4-Xanlin (approved):
Looks good to me.  Follows Ondreian's feedback per discord too.

Issue #75, opened 2021-10-30
Defect: Update autostart to unload various script auto-updates from Tillmen repository


OSXLich-Doug:
One of the problematic support activities is a repo auto-update on login that overwrites lich5 core script files.  Need to add logic to unwind any auto-update entry for any core lich file we distribute.

We do this already for lich.rbw, needs to include alias, autostart, infomon, narost, repository, vars, xnarost presently based on GUI edits to keep these functional

Probably should include go2 (when go2 core is updated), lnet (if lnet messaging is rationalized as in issue #44  ).

Unfortunately, has to be discrete list - users may have other scripts included to auto-update beyond the Lich5 core that we can't disrupt.

asechrest:
While perusing the Lich5 issues to see if there are any on which I could get my feet wet, I came across this one. Just wondering if this can be closed via #151 ?

mrhoribu:
This should be closed now as #151 resolved this issue and any further issues should be addressed on EO\scripts.

PR #151, opened 2022-03-13
Update repository.lic

mrhoribu:
Remove Tillmen updates to prevent issues of downgrading to older script versions.


mrhoribu:
Also removed xnarost from autostarting on new character/install.

mrhoribu:
Address concerns from #75

PR #152, opened 2022-03-13
Script CLASS empty array initializing bugfix

OSXLich-Doug:
Preventing class Script from pushing empty arguments to scripts being loaded

This change modifies class Script to prevent empty or nil arguments being passed to scripts as they're being loaded.  Scripts that do not accept arguments were unaffected by recent Lich 5.4.x changes, but those scripts that did accept arguments, but did not intend to catch NIL or EMPTY arguments showed erratic behavior.

edited for spelling

### DIFF
--- a/lich.rbw
+++ b/lich.rbw
@@ -1237,6 +1237,8 @@ class Script
       unless (args[:args].nil? || args[:args].empty?)
         @vars = [ args[:args].join(" ") ]
         @vars.concat args[:args]
+      else
+        @vars = Array.new
       end
     else
       @vars = Array.new
--- a/lich.rbw
+++ b/lich.rbw
@@ -722,9 +722,7 @@ class Script
           end
         elsif args[1].class == Hash
           options = args[1]
-          unless options[:args].nil? or options[:args].empty?
-            script_args = options[:args]
-          end
+          script_args = (options[:args] || String.new)
         else
           # fixme: error
           next nil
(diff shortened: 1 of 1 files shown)


mrhoribu:
Issue is due to fix for populating script.vars[0] with entire set of vars if Array[] is pushed to Script initialization. Unfortunately, if empty array was pushed, it would go from being empty`[]` to being `[""]`.

mrhoribu:
> Issue is due to fix for populating script.vars[0] with entire set of vars if Array[] is pushed to Script initialization. Unfortunately, if empty array was pushed, it would go from being empty`[]` to being `[""]`.

This was introduced with DragonRealms fixes from #133

mrhoribu:
For more info dragon-realms/dr-lich/issues/46

MahtraDR:
With this change. 
```
--- Lich: dependency active.                                                                                                                                                                                                                  
[dependency: ***INVALID ARGUMENTS DON'T MATCH ANY PATTERN***]                                                                                                                                                                                 
Provided Arguments: ''                                                                                                                                                                                                                        
                                                                                                                                                                                                                                              
  ;dependency [debug] [install]                                                                                                                                                                                                               
   debug                                                                                                                                                                                                                                      
   install                                                                                                                                                                                                                                    
--- Lich: dependency has exited.            
```
Without this change, dependency loads silently as expect.


```
>;autostart list                                                                                                                                                                                                                              
--- Global autostart scripts: alias, lnet, repository (args: download-updates), dependency
```

MahtraDR (approved):
Fine for the DR side.

PR #153, opened 2022-03-14
Add Spell.incant, Spell.channel, Spell.evoke

mrhoribu:
Add extra Spell options:
- [X] add Spell.force_cast(target, args, results_of_interest)
- [X] add Spell.force_channel(target, args, results_of_interest)
- [X] add Spell.force_evoke(target, args, results_of_interest)
- [X] add Spell.force_incant(args, results_of_interest)

Does not currently know what spells can be channel or evoke. Will attempt and if can't evoke/channel, switch to cast instead.

### DIFF
--- a/lib/spell.rb
+++ b/lib/spell.rb
@@ -570,14 +570,14 @@ module Games
             cast_cmd = 'cast'
           end
           unless (arg_options.nil? || arg_options.empty?)
-            if arg_options.split(" ")[0] =~ /incant|channel|evoke/
+            if arg_options.split(" ")[0] =~ /incant|channel|evoke|cast/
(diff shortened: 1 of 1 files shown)


mrhoribu:
New defs will force attempt to evoke/incant/channel if called. They do not currently respect spell-list.xml settings for incant or channel.

mrhoribu:
> New defs will force attempt to evoke/incant/channel if called. They do not currently respect spell-list.xml settings for incant or channel. Thoughts on adding a forced cast option? Maybe `Spell.force_cast` ?? So that way you can just basic cast without it respecting whatever spell-list.xml settings there are?

Added this as well.

PR #163, opened 2022-03-15
split map file by game until DR gets uids

gs4-Xanlin:
Splitting the map file into map_gs and map_dr since map_dr is loading/saving/pulling room object data where as the gemstone version doesn't.  Removed logic that utilized uids for Map.current and Map.current_or_new in map_dr.

map for gemstone: 
  - set the conditions for determining a room to be waived if the uid has a single match.  (currently always either 1 match or 0 unless duplicate rooms are added)  
  - added logic to current_or_new to add titles/descriptions to the room when found on an existing uid.
  - current_or_new will also update the uid hash now where previously that was only being done on map load.
  - removed logic that added location checks for rooms that have identical rooms based title and description.

Added checks to see if the method is defined before calling Map.last_seen_objects in XMLParser.rb and lich.rbw.


OSXLich-Doug:
Nice!  Question - we need to rename existing 5\.(?:4|5)\.(?:2|0) map.rb to map_gs.rb as well.  Manual or execute as part of this PR

(a little regex humor for ya)

PR #165, opened 2022-03-17
[lib] [lich.rb] Reducing db calls from Lich.display_lichid and Lich.display_roomid.  …

gs4-Xanlin:
…Followed the same pattern for other boolean variables in the Lich module.  Simplified logic in lich.rbw for adding ids to room titles, skips the open regex pattern search of every string when both are turned off.


ondreian (commented):
i don't think this is necessary, we should be removing these db caching mechanisms not adding more. SQLite is never going to be the slow part of this operation

PR #167, opened 2022-03-18
Second pass at DR dependency and drinfomon

OSXLich-Doug:
DR fix


OSXLich-Doug:
Closing because it looks like I stepped on something.  Will resubmit.

PR #171, opened 2022-03-18
Profanity init_str causes garbage in Genie

OSXLich-Doug:
For testing and incorporation if approved.


MahtraDR (approved):
Code looks good to me.

PR #173, opened 2022-03-18
Update log.lic to bring in DR

OSXLich-Doug:
DR uses configuration details for string output.  This file supports both GS and DR


OSXLich-Doug:
Dumping because yuch

PR #179, opened 2022-03-23
[scripts] [autostart] Skip dependency in autostarts, and instructions on how to remove it.

MahtraDR:
Simple as anything. Just skips dependency if it's additionally listed as an autostart script, and prints a note on how to remove it from the autostart liss.

```
--- dependency found in autostart list. Attempting to remove it now. If unsuccessful, it can be safely removed with ';autostart remove --global dependency'                                                                                   
                                                                                                                                                                                                                                              
--- dependency was removed from the global autostart list         
```


MahtraDR:
Taking @mrhoribu helpful advice, this now auto-removes dependency from the autostart list, with a friendly note, and backup actions if it fails.

PR #181, opened 2022-03-25
[lib] [util.rb] Add quiet_command and resolve Timeout issues

mrhoribu:
v1.2.0 (2022-03-16)
* Add Lich::Util.quiet_command to mimic XML version
* Removed usage of Timeout::timeout


mrhoribu:
Original attempt had some bad copy/pasta variables that should now be corrected.

PR #182, opened 2022-03-29
[lich.rbw] Clean lich char comma or semi-colon

MahtraDR:
Allows usage of a comma, or the semi-colon as the lich char. In the past, Genie couldn't handle semi-colons, but that's no longer the case.

I don't believe adding the comma as a second command character breaks anything...

PR #183, opened 2022-03-29
[scripts] [go2] Move from using lich_char to using clean_lich_char

MahtraDR:
In a world where `$lich_char` and `$clean_lich_char` are the same value, they can be used interchangeably. That isn't true if we want to support more than one `$lich_char` that's, I think, the original intent of separating the two. Anytime we want to reference/display the lich command character, use `$clean_lich_char` like this, and no issues arise.


MahtraDR:
Addresses the issues seen in https://github.com/elanthia-online/lich-5/pull/182

MahtraDR:
Closing this. Other approaches make more sense.

PR #187, opened 2022-04-01
[lich.rbw] Moving map-related require-relatives to accomodate Genie, and other direct-connect clients

MahtraDR:
Tested, no regressions. Gets rid of the problem where we don't detect the game without `--login` or via the launcher.

Note, with this change, lich5 works perfectly well with DR and Genie, using the existing process documented here https://github.com/GenieClient/Genie4/wiki/02.-Connecting-and-Profiles

### DIFF
--- a/lich.rbw
+++ b/lich.rbw
@@ -7011,6 +7011,7 @@ if arg = ARGV.find { |a| (a == '-g') or (a == '--game') }
     $frontend = 'unknown'
   end
 elsif ARGV.include?('--gemstone')
+  require_relative("./lib/map_gs.rb")
   if ARGV.include?('--platinum')
     $platinum = true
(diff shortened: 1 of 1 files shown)


MahtraDR:
> Would also need to add the includes in both the `--fallen` and the `--shattered` sections as well for each game in order for this to work fully.

We also need to figure out how to get it in with the launcher. I'll spend more time thinking about this today.

MahtraDR:
@OSXLich-Doug @mrhoribu How's this instead?

MahtraDR:
@OSXLich-Doug 's fix on Discord looks like a better solution than this.

OSXLich-Doug:
Closing without merge per agreement.

PR #188, opened 2022-04-03
[lib][shield.rb] Added Shield.use

ServerFailure:
^


OSXLich-Doug:
I like this submission, and thank you @ServerFailure for this.  I think this can be a useful and extensible model for all of our CMan / Feat type activities as well.  I do have a couple of questions that I'll try to resolve for myself but if insight is avaialble, would love to hear it.

1) Should we combine usage and results into one hash of hashes, for maintenance, or is separate better?
2) Is there any logical sense or useful purpose for capture groups in the regex hash?

Good stuff!

mrhoribu:
> I like this submission, and thank you @ServerFailure for this. I think this can be a useful and extensible model for all of our CMan / Feat type activities as well. I do have a couple of questions that I'll try to resolve for myself but if insight is avaialble, would love to hear it.
> 
>     1. Should we combine usage and results into one hash of hashes, for maintenance, or is separate better?
> 
>     2. Is there any logical sense or useful purpose for capture groups in the regex hash?
> 
> 
> Good stuff!

I actually like the idea of combining the hash into a single form. Makes sense. I'll poke at this and see. As for capture groups, not sure.

PR #194, opened 2022-04-13
Migrating from lich5-update script to lib

OSXLich-Doug:
Moving from `;lich5-update` script to `update.rb` in Lich5/lib

This enables us to transparently maintain the update portion without having to push a new script and request it be downloaded.

`update.rb` contains the following:
- refresh is removed / deprecated
- revert now restores the entire ecosystem from snapshot, not just lich.rbw
- script | library | data discrete file download is retained
- script discrete file download now points to EO's scripts to support removing scripts from Lich5 repo
- data discrete file download remains pointed at Lich-5 data until we make that available in EO scripts

`infomon.lic` has been updated to detect Lich5 > 5.5.0 and register the `;lich5-update --<request>` commands to point to `update.rb`.  It also has a shorthand method `;l5u -<request>`; mostly because I got tired of typing while testing.

### DIFF
--- a/lib/update.rb
+++ b/lib/update.rb
@@ -1,0 +1,332 @@
+## Let's have fun updating Lich5!
+
+module Lich
+  module Util
+    module Update
+      require 'json'
+      require 'open-uri'
+
+      @current = LICH_VERSION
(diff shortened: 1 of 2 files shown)


OSXLich-Doug:
Closing no merge - will initiate new PR with Lich core updates rather than infomon updates per Tysong suggestion.

PR #201, opened 2022-04-15
[lib][gui-login.rb] and gui-saved-login.rb update for booleans

gs4-Xanlin:


OSXLich-Doug:
I appreciate the attempt.  The problem is the first time a current 5.5.0 or prior player runs this, those settings will be lost.  I was working to accept true (new method) or 'on' (lich 5.5.0 or prior), and false or 'off', to ensure the user didn't have to reset these values.

And the GUI switch selector defaults to 'on' or 'off', so I have to work that in on the saved login file, as well.

gs4-Xanlin:
When it reads the value from the database, it turns on/true/yes to true.  (per the previous commit to lich.rb)

OSXLich-Doug:
Yes, I just needed to adjust the behavior in the gui *.rb files to convert over to boolean, since they were expecting a return of 'on' or 'off'.  Work's done, I'll be posting a PR and then merging both of these.

PR #238, opened 2022-05-31
[update.rb] Updating to stdlib functions

### LINKED ISSUE #237: Windows 11 Potential Bug
`update.rb` and Windows 11 21H2 seem not to work well.

Issue is centered on `system("tar -xf #{File.join(TEMP_DIR, filename)}.zip")`

Appears to fail, and subsequent commands to open the unzipped folder all fail.  Needs exploration.

OSXLich-Doug:
As requested - interested to find out if this will work now on Win 11 with a space or spaces in the directory names as well as Linux platforms.

Fix #237

### DIFF
--- a/lib/update.rb
+++ b/lib/update.rb
@@ -163,7 +163,7 @@ module Lich
             file.write open("#{@zipfile}").read
           end
 
-          Dir.mkdir("#{TEMP_DIR}/#{filename}")
+          Dir.mkdir(File.join(TEMP_DIR, filename))
           Gem::Package.new("").extract_tar_gz(File.open(File.join(TEMP_DIR, "#{filename}.tar.gz"), "rb"), "#{TEMP_DIR}/#{filename}")
           new_target = Dir.children(File.join(TEMP_DIR, filename))
           FileUtils.cp_r(File.join(TEMP_DIR, filename, new_target[0]), TEMP_DIR)
--- a/lib/update.rb
+++ b/lib/update.rb
@@ -5,6 +5,8 @@ module Lich
     module Update
       require 'json'
       require 'open-uri'
+      require 'rubygems/package'
+      require 'zlib'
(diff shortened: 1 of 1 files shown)


lib/update.rb
```diff
@@ -157,33 +159,16 @@ def self.download_update
...
-            FileUtils.remove_dir(File.join(TEMP_DIR, filename))
-            FileUtils.mv(File.join(TEMP_DIR, new_target[0]), File.join(TEMP_DIR, filename))
-          end
-          ## These five lines may be deleted after zipfiles are created with .gitattributes
-          cleanup_dir = Dir.children(File.join(TEMP_DIR, filename))
-          FileUtils.remove_dir(File.join(TEMP_DIR, filename, "lich")) if cleanup_dir.include?("lich")
-          FileUtils.rm(File.join(TEMP_DIR, filename, "netlify.toml")) if cleanup_dir.include?("netlify.toml")
-          FileUtils.rm(File.join(TEMP_DIR, filename, "README.adoc")) if cleanup_dir.include?("README.adoc")
-          FileUtils.rm(File.join(TEMP_DIR, filename, "data", "update-lich5.json")) if Dir.children(File.join(TEMP_DIR, filename, "data")).include?("update-lich5.json")
+          Dir.mkdir("#{TEMP_DIR}/#{filename}")
```
ondreian, line 166: this should use `File.join`

ondreian (changes requested):
mostly ok, but just one place we should be using `File.join` still

lib/update.rb
OSXLich-Doug: So. . . 

`Dir.mkdir(File.join(TEMP_DIR, filename))`

I'll test it out - no issue with the change.

mrhoribu (approved):

PR #240, opened 2022-06-03
[global_defs.rb] Remove genie special bonus carriage return+newline

MahtraDR:
@jingounchained have been coordinating on this for awhile now, and with the release of https://github.com/GenieClient/Genie4/releases/tag/4.0.2.4 the bonus newlines are no longer needed for genie. 

Should remove the weird-looking double spacing issues in genie+lich.

In concert with https://github.com/rpherbig/dr-scripts/pull/5818 as well


asechrest:
I still get double-spacing in Genie with just this change (and the one to `common` on DR Lich). Example:

image

lib/global_defs.rb
```diff
@@ -1593,7 +1593,7 @@ def respond(first = "", *messages)
     end
     messages.flatten.each { |message| str += sprintf("%s\r\n", message.to_s.chomp) }
     str.split(/\r?\n/).each { |line| Script.new_script_output(line); Buffer.update(line, Buffer::SCRIPT_OUTPUT) }
-    str.gsub!(/\r?\n/, "\r\n") if $frontend == 'genie'
+    # str.gsub!(/\r?\n/, "\r\n") if $frontend == 'genie'
     if $frontend == 'stormfront' || $frontend == 'genie'
```
asechrest, line 1597: However if I comment out the check for `genie` as front end in this line, the double spacing is gone, but the nice horizontal formatting is then affected:

image

jingounchained: The Genie team is looking at the mono font issue as a problem on our side. We would prefer to consume what's sent by Lich the same as any other FE does, rather than have exceptions coded into Lich.

asechrest:
Update:

The Genie devs have helped handle this and the current test build appears to work great - mono font holds and no phantom newlines are introduced.

asechrest:
Update:

The Genie devs have helped handle this and the current test build appears to work great - mono font holds and no phantom newlines are introduced.

PR #244, opened 2022-06-05
[lich.rbw] More XML fixes for janky simu XML

MahtraDR:
Cleaning up a whole category of rooms with broken XML, and making our logging a bit more verbose so we capture newlines, and carriage returns, since that sucked to diagnose without those characters displaying...

### DIFF
--- a/lich.rbw
+++ b/lich.rbw
@@ -2270,7 +2270,7 @@ module Games
               @@last_recv = Time.now
               @@_buffer.update($_SERVERSTRING_) if TESTING
               begin
-                $cmd_prefix = String.new if $_SERVERSTRING_ =~ /^\034GSw/  
+                $cmd_prefix = String.new if $_SERVERSTRING_ =~ /^\034GSw/
                 ## Clear out superfluous tags
                 $_SERVERSTRING_ = $_SERVERSTRING_.gsub("<pushStream id=\"combat\" /><popStream id=\"combat\" />","")
                 $_SERVERSTRING_ = $_SERVERSTRING_.gsub("<popStream id=\"combat\" /><pushStream id=\"combat\" />","")
--- a/lich.rbw
+++ b/lich.rbw
@@ -2411,8 +2411,11 @@ module Games
                   rescue
                     unless $!.to_s =~ /invalid byte sequence/
                       # Fixes invalid XML with nested single quotes in it such as:
+                      # From DR intro tips
                       # <link id='2' value='Ever wondered about the time you've spent in Elanthia?  Check the PLAYED verb!' cmd='played' echo='played' />
-                      while data = $_SERVERSTRING_.match(/'([^=]*'[^=]*)'/)
+                      # From GS
+                      # <d cmd='forage Imaera's Lace'>Imaera's Lace</d>, <d cmd='forage stalk burdock'>stalk of burdock</d>
+                      while data = $_SERVERSTRING_.match(/'([^=>]*'[^=>]*)'/)
                         Lich.log "Invalid nested single quotes XML tags detected: #{$_SERVERSTRING_.inspect}"
                         $_SERVERSTRING_.gsub!(data[1], data[1].gsub!(/'/, '&apos;'))
                         Lich.log "Invalid nested single quotes XML tags fixed to: #{$_SERVERSTRING_.inspect}"
--- a/lich.rbw
+++ b/lich.rbw
@@ -2297,13 +2297,6 @@ module Games
                   Lich.log "Extraneous closed tag fixed: #{$_SERVERSTRING_.inspect}"
                 end
 
-                # Remove empty push/popbold tags
-                if $_SERVERSTRING_ == /<pushBold\/><popBold\/>/
-                  Lich.log "Empty push/popbold tags detected: #{$_SERVERSTRING_.inspect}"
-                  $_SERVERSTRING_.gsub!("<pushBold/><popBold/>", '')
-                  Lich.log "Empty push/popbold tags removed: #{$_SERVERSTRING_.inspect}"
-                end
-
                 # "<component id='room objs'>  You also see a granite altar with several candles and a water jug on it, and a granite font.\r\n"
                 # "<component id='room extra'>Placed around the interior, you see: some furniture and other bits of interest.\r\n
                 # Followed by in a new line.
--- a/lich.rbw
+++ b/lich.rbw
@@ -2271,6 +2271,25 @@ module Games
               @@_buffer.update($_SERVERSTRING_) if TESTING
               begin
                 $cmd_prefix = String.new if $_SERVERSTRING_ =~ /^\034GSw/
+          
+                # These rooms are in the Crossing temple in DR
+                # Fixes xml with \r\n in the middle of it like:
+                # <component id='room exits'>Obvious paths: clockwise, widdershins.\r\n
+                # <compass></compass></component>\r\n
+                # We close the first line and in the next segment, we remove the trailing bits
+                # Because we can only match line by line, this couldn't be fixed in one matching block...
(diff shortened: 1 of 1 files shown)


MahtraDR:
Resolves https://github.com/elanthia-online/lich-5/issues/249

MahtraDR:
> Wouldn't it be better to catch the error, get the next line, and retry parsing with both lines together?
> 
> The section you're modifying is begging for some refactoring, but if I understand it well...
> 
> You could also do something like:
> 
> ```ruby
> $_SERVERSTRING_.match(/^<component [^>]*>(.*)/) { |m|
>   if not m.captures[1].include? "</component>"
>     $_SERVERSTRING_ += @@socket.gets
>   end
> }
> ```
> 
> That would also fix this one I came across.
> 
> Sorry if any of that is not Ruby, I've never written or read it before last week.

We parse one line at a time. You might be able to do it that way, but then you need to capture each line individually, and re-parse both together. I thought the approach in this PR was simpler, and more targeted.

sproctor:
> We parse one line at a time.

This is happening before the parsing. You could think of it as a way to say "Don't parse yet, there's more stuff coming". I should probably be quiet until I've actually tested my changed though. ;)

It's fair to say that depending on where in the chain this statement is, the next line won't get the earlier adjustments applied, but I think that's ok, since in these rare cases we know it doesn't need them.

My off-topic opinion: The stuff that comes downstream is so not XML that using an XML parser isn't the best. I've written a simple parser that handles this without any hacks. https://github.com/sproctor/warlock3/tree/master/stormfront/src/main/antlr

Then your hacks aren't special cases, they're just how things are done. You can easily balance tags at the end of a line or at the close tag for a parent tag. Unbalanced close tags can just be ignored. For this particular case, that makes more sense. Having a <compass> tag inside a <component> makes no sense at all. I don't know what the code was to make that happen on their end.

lich.rbw
```diff
@@ -2282,21 +2288,29 @@ module Games
...
-                ## Fix combat wrapping components - Why, DR, Why?
-                $_SERVERSTRING_ = $_SERVERSTRING_.gsub("<pushStream id=\"combat\" /><component id=","<component id=")
-                # $_SERVERSTRING_ = $_SERVERSTRING_.gsub("<pushStream id=\"combat\" /><prompt ","<prompt ")
+                # "<component id='room objs'>  You also see a granite altar with several candles and a water jug on it, and a granite font.\r\n"
+                # "<component id='room extra'>Placed around the interior, you see: some furniture and other bits of interest.\r\n
+                # Followed by in a new line.
+                # "</component>\r\n"
+                if $_SERVERSTRING_ =~ /^<component id='room (?:objs|extra)'>[^<]*(?!<\/component>)\r\n/
+                  Lich.log "Open-ended room objects component id tag: #{$_SERVERSTRING_.inspect}"
+                  $_SERVERSTRING_.gsub!("\r\n", "</component>")
```
sproctor, line 2306: Should the ending be preserved? "</component>\r\n"

lich.rbw
```diff
@@ -2397,8 +2411,11 @@ module Games
                   rescue
                     unless $!.to_s =~ /invalid byte sequence/
                       # Fixes invalid XML with nested single quotes in it such as:
+                      # From DR intro tips
                       # <link id='2' value='Ever wondered about the time you've spent in Elanthia?  Check the PLAYED verb!' cmd='played' echo='played' />
-                      while data = $_SERVERSTRING_.match(/'([^=]*'[^=]*)'/)
+                      # From GS
+                      # <d cmd='forage Imaera's Lace'>Imaera's Lace</d>, <d cmd='forage stalk burdock'>stalk of burdock</d>
+                      while data = $_SERVERSTRING_.match(/'([^=>]*'[^=>]*)'/)
```
sproctor, line 2418: The adjustment with double-quotes should probably be updated as well.

lich.rbw
MahtraDR: No, it's not in every line, and breaks nothing to remove it.

lich.rbw
MahtraDR: Thanks for commenting, and making suggestions, btw.

I had intentionally left this line alone, since I haven't see the problem occur with the single quotes before. In theory it breaks nothing to add it, but without a problem to address (for all the other cases, I have example lines that get fixed, either from DR, or from GS) I rather not add the change. 

All changes so far I have tested robustly, across 9 DR characters, and GS testers have verified the others. Which is to say... with a good level of testing/exposure to far, we haven't run into this specific problem with single quotes.

MahtraDR:
> > We parse one line at a time.
> 
> This is happening before the parsing. You could think of it as a way to say "Don't parse yet, there's more stuff coming". I should probably be quiet until I've actually tested my changed though. ;)
> 
> It's fair to say that depending on where in the chain this statement is, the next line won't get the earlier adjustments applied, but I think that's ok, since in these rare cases we know it doesn't need them.
> 
> My off-topic opinion: The stuff that comes downstream is so not XML that using an XML parser isn't the best. I've written a simple parser that handles this without any hacks. https://github.com/sproctor/warlock3/tree/master/stormfront/src/main/antlr
> 
> Then your hacks aren't special cases, they're just how things are done. You can easily balance tags at the end of a line or at the close tag for a parent tag. Unbalanced close tags can just be ignored. For this particular case, that makes more sense. Having a tag inside a makes no sense at all. I don't know what the code was to make that happen on their end.

I agree in principle. I am not willing to change parsers just yet, but after this next round of merges, I can give your parser a try.

sproctor:
I just spent a few minutes looking at Ruby bindings for ANTLR, doesn't looks popular, but there's one option. https://github.com/camertron/antlr4-native-rb

If you're interested, I could probably throw together something in rex/racc which would be a more Ruby style option.

If you want a bloopers reel of my simu parsers:
https://github.com/sproctor/warlock2/blob/master/bundles/cc.warlock.core.stormfront/src/main/cc/warlock/core/stormfront/internal/stormfront.jj - JavaCC parser for SF protocol
https://github.com/sproctor/warlock-gtk/blob/master/src/wizard_protocol_lexer.l - YACC (or bison) lexer for Wizard protocol. That one was simple enough to not need a parser... oh not really. That's embarrassing to read. Enjoy!

MahtraDR:
> I just spent a few minutes looking at Ruby bindings for ANTLR, doesn't looks popular, but there's one option. https://github.com/camertron/antlr4-native-rb
> 
> If you're interested, I could probably throw together something in rex/racc which would be a more Ruby style option.
> 
> If you want a bloopers reel of my simu parsers: https://github.com/sproctor/warlock2/blob/master/bundles/cc.warlock.core.stormfront/src/main/cc/warlock/core/stormfront/internal/stormfront.jj - JavaCC parser for SF protocol https://github.com/sproctor/warlock-gtk/blob/master/src/wizard_protocol_lexer.l - YACC (or bison) lexer for Wizard protocol. That one was simple enough to not need a parser... oh not really. That's embarrassing to read. Enjoy!

If you write something up which can be tested, I'm happy to give it a go. I'm all linux for lich/DR, which seems to not be an issue for you, but wanted to note. :)

jherig (approved):
Tested, things appear to be working as advertised. I have some reservations about parsing XML on a line by line basis, rather than on tag completion, but that might be a major architecture change (there are no rules against having <CR><LF> in the midst of tags, only that white space should be preserved).

MahtraDR:
> Tested, things appear to be working as advertised. I have some reservations about parsing XML on a line by line basis, rather than on tag completion, but that might be a major architecture change (there are no rules against having <CR><LF> in the midst of tags, only that white space should be preserved).

Thanks for testing. 

Yeah it would be a major architectural change to parse tags (and simu still has I terminated tags anyway...)

So within the scope of this PR, we have this. 😀

PR #247, opened 2022-07-06
Add ability to launch custom commands on Linux without Wine

sproctor:
I'm attempting to support running with Warlock.

For this to work, Lich needs to be able to launch native Linux binaries.

I removed some duplicated Wine initialization code that was causing some complaints.

I re-enabled the commented out section to allow custom commands to be added, and removed the check for wine when launching custom commands. If someone needs it, they can still prefix the command with wine.

When I was confused and reading the source, I cleaned up some messy handling of client command-line flags. Stormfront/Genie use the same host/port, the Wizard is the odd one out there, though I haven't looked at Avalon.


OSXLich-Doug:
@Ahtias - this one needs your review, I don't have a linux box to test on.

PR #248, opened 2022-07-11
[lich.rbw] Add noun adjustment for Ball and Chain

MahtraDR:
The stream says 
```xml
<right exist="13116574" noun="ball and chain">spiked ball and chain</right><clearStream id='inv' ifClosed=''/><pushStream id='inv'/>Your worn items are:
```
But it's a ball.

### DIFF
--- a/lich.rbw
+++ b/lich.rbw
@@ -3607,6 +3607,7 @@ module Games
         @noun = noun
         @noun = 'lapis' if @noun == 'lapis lazuli'
         @noun = 'hammer' if @noun == "Hammer of Kai"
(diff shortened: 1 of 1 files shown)


MahtraDR:
Resolves https://github.com/rpherbig/dr-scripts/issues/6123

bwfisher82:
How does the GS4 ball and chain item come across in the XML? Just  wondering if this change could possibly cause issues over there.

MahtraDR:
> How does the GS4 ball and chain item come across in the XML? Just wondering if this change could possibly cause issues over there.

If it's an issue, we can just add an if game is DR, not fussed.

PR #251, opened 2022-08-09
Removed defunct symbolic link and add swap files to ignore

athias:


mrhoribu (approved):
don't see any issue here.

PR #255, opened 2022-10-07
[feat.rb] stamina cost fix

mrhoribu:
Fix stamina cost for dispel magic from 0 -> 30.


OSXLich-Doug (approved):
Good here - include in 5.6.2

PR #268, opened 2022-11-19
[gameobj.rb] Move GameObj to library and add GameObj.type_cache

gs4-Xanlin:
Move GameObj to library and add GameObj.type_cache


sandersch (approved):
This seems reasonable to me. I had some minor concerns that adding caching may introduce some new issues, but nothing here seems problematic. It seems like this code is reinitializing the cache when data gets reloaded which should avoid stale cache entries.

PR #271, opened 2022-12-01
Support alternate entry.dat with --entryday option

jymamon:
Allows user to override the entry.dat file used with "--entrydat {filename}"


jymamon:
I really dislike that github defaults my user branch back to the original repo instead of my test fork.

PR #274, opened 2022-12-14
Create Bounty API

sandersch:
### Overview

Adds a new Bounty top-level class that provides an API for bounty information. Primary interface is `Bounty.current` which returns a `Bounty::Task` with detailed info.

#### Top-level Bounty API

Generally I use this myself like:
```ruby
b = Bounty.current
b.type
b.town
b.requirements[:area]
```

I also added other methods to the top level bounty class that create a new Task instance from the current bounty:
```ruby
Bounty.type
Bounty.requirements
Bounty.town
Bounty.any?
Bounty.none?
Bounty.done?
```

#### Bounty::Task

* **Type:**

Bounty States

* **Requirements:** Details specific to this task e.g. the area for the bounty and how many creatures to kill or herbs to forage 
* **Town:** This is useful for determining which NPC to visit for gem/skin/herbs and guard turn ins. Also particularly for child rescue tasks in hunting grounds like Castle Varunar which can be assigned from multiple towns. There are also some area name descriptions like "grasslands" that are duplicated in different towns where you may need this to determine the right hunting area.

### Tests

This also adds a test suite for `Bounty::Parser` with a variety of examples from many areas and most towns as well as a GitHub action to run them in CI.


lib/bounty.rb
```diff
@@ -0,0 +1,250 @@
+class Bounty
+  def self.current
+    Task.new(Parser.parse(checkbounty))
+  end
+
+  # Delegate class methods to current instance of bounty
+  [:status, :type, :requirements, :town, :any?, :none?, :done?].each do |attr|
```
sandersch, line 7: There is a lot of ruby going on here. This is code I originally wrote a long time ago and might not write in such a meta way so let me know if it would be helpful to simplify this and expand the method definitions to make them explicit.

spec/bounty_parser_spec.rb
```diff
@@ -0,0 +1,522 @@
+require 'bounty'
+
+class Bounty
+  describe Parser, "#parse" do
+    it "can tell when we don't have a task" do
+      bounty = described_class.parse "You are not currently assigned a task."
```
sandersch, line 6: I am not sure this is the best thing to return for this case. Returning a bare `nil` make it hard to tell the different between no bounty assigned and unable to parse a bounty. I have gone back and forth with making task and/or status nil so curious if other people have input on how this example should behave before we release a new API.

sandersch: Updated this to return `:none` based on feedback from Ondreian which feels better.

sandersch:
I feel like is in a good spot and plan on merging it in the near future.

LeBigNasty:
Any thoughts on adding a class for :group? including group cooldown?

sandersch:
> Any thoughts on adding a class for :group? 

`Bounty::Task` has a `help?` method (there are also `assist?` and `group?` aliases for it) that will determine whether your current task is one that has been shared with you by someone else.

You can also use `Bounty.lnet(person)` to get the `Bounty::Task` for that person through LNet using the existing LNet bounty hook.

> including group cooldown?

IMHO this is out of scope for this API and should be handled by something else (probably ;infomon). `checkbounty` is the source of truth for this API and returns the same value when you don't have a task regardless of your cooldown statuses.

LeBigNasty:
I think for ;ebounty, we would move away from relaying on lnet to share tasks or perform any function.

PR #276, opened 2023-01-12
chore: update .ruby-version to 2.7.7

mrhoribu:
Bump to latest v2.7.7 of Ruby


sproctor:
The patch version isn't important and can be dropped. "2.7" works just as well and won't need to be updated in the future until 3.x. Unless there's some automation I'm not aware of.

PR #278, opened 2023-01-13
[Core Lich] Ruby 3 compatibility

MahtraDR:
This has been well tested on my end, but I haven't kept it up to date with any core changes since I created the original branch.

Also, for posterity, ... 
image


lich.rbw
```diff
@@ -100,12 +100,6 @@ class Numeric
 end
 
 class String
-  @@elevated_untaint = proc { |what| what.orig_untaint }
```
rpherbig, line 103: 💡 There's still a lot of 'elevated' variable naming floating around. Not functional, but probably could be renamed in the future to make more sense?

lich.rbw
```diff
@@ -2306,7 +2205,7 @@ module Games
                   $_SERVERSTRING_.gsub!("\r\n", "</component>")
                   Lich.log "Open-ended room objects component id tag fixed to: #{$_SERVERSTRING_.inspect}"
                 end
-                # "</component>\r\n"                
+                # "</component>\r\n"
```
rpherbig, line 2208: 💡 Delete?

rpherbig (approved):
Seems reasonable to me!

lich.rbw
```diff
@@ -3236,7 +3132,6 @@ module Games
     end
 
     require_relative("./lib/spell.rb")
-    require_relative("./lib/bounty.rb")
```
mrhoribu, line 3239: Looks like this was accidentally removed on PR to merge into existing code base. Need to add back in.

MahtraDR: Re-added. Thank you!

lich.rbw
MahtraDR: Removed. Thank you!

lich.rbw
MahtraDR: I agree. That just makes it a bigger change for now.

MahtraDR:
> Seems reasonable to me!

Thank you for taking the time to review this. I appreciate it.

MahtraDR:
This version brings us up to compatibility with Ruby 3.2.

PR #284, opened 2023-02-02
[lib/wine.rb] De-duplicate Wine module

sproctor:
This is a re-work of a previous PR. It's a pretty simple change so it should be easy to review/merge


.ruby-version
```diff
@@ -1 +1 @@
-2.7.2
\ No newline at end of file
+2.7
```
sproctor, line 1: The patch version isn't important, I've removed it so it can run with 2.7.7

MahtraDR:
How do I use lich with wine? I don't even have wine installed...

Having said this, this PR doesn't break anything, which is a major positive! :)

sproctor:
If you have wine installed, it will detect it and use it to launch the front-ends.

The logic for it is a bit flawed. Each client should be marked with which OS it runs on, then lich can take the appropriate action to run that client. On linux that would mean using wine to run wrayth, or using darling to run avalon. I have no experience with darling, just mentioning it as an idea. That's a bit of a tangent to this PR though.

PR #285, opened 2023-02-02
[lib/gui-manual-login.rb] Enable custom commands

sproctor:
This PR uncomments the checkbox and text entries for entering custom client paths. I'm not sure why it was disabled.

It also uses placeholder-text, available since GTK 3.2, to set the placeholder, which really improves the UX.

It stops executing the custom command with wine.

And it fixes a bug with the working directory logic when it's an empty string.

Ideally, there should be a flag with custom commands to enable wine on linux/mac. Until then, it's better to keep wine disabled. The user can add the wine prefix if necessary.


lich.rbw
```diff
@@ -5186,14 +5186,14 @@ main_thread = Thread.new {
         launcher_cmd = launcher_cmd.tr('/', "\\") if (RUBY_PLATFORM =~ /mingw|win/i) and (RUBY_PLATFORM !~ /darwin/i)
       end
       begin
-        if custom_launch_dir
+        unless custom_launch_dir.empty?
```
sproctor, line 5189: Can someone check this and make sure it's correct? It works for me, but I'm not sure if custom_launch_dir can be null and if there would be an issue in that case.

MahtraDR: `custom_launch_dir` can be nil. You could just have the check be `unless (custom_launch_dir.empty? ||  custom_launch_dir.nil?` ?

MahtraDR:
This seems to work as suggested too.

lich.rbw
sproctor: ok, I added the nil? check. Pretty sure the `nil?` needs to come before the `empty?`

MahtraDR: If they're in parentheses it doesn't matter.

sproctor: I don't think that is correct. I've read/written very little Ruby, but I don't know another language that works like that and my simple test says ruby doesn't either. This doesn't work:
```ruby
foo = nil
if (foo.empty? || foo.nil?)
  puts "works"
end
```
The parentheses do nothing. If you reverse the order, this works:
```ruby
foo = nil
if foo.nil? || foo.empty?
  puts "works"
end
```
It works as expected. The `||` short circuits and only evaluates the `nil?`. If you put them in the other order, the `empty?` is executed first and gives an undefined method error.

I don't think the parentheses do anything in this case, because there's no ambiguity to the order of operations.

asechrest: Note: in Ruby, `and` and `&&` are not the same, and neither are `or` or `||'. The difference is operator precedence. See: https://mixandgo.com/learn/ruby/boolean-operator-precedence

I'm not familiar with the code, here, but if we're structuring our `||` check with short-circuiting governing the success of the check, that suggests we're unsure of the data type for `custom_launch_dir`. Is that the case? Because a `.empty` check on `nil` doesn't make sense in Ruby. Is `custom_launch_dir` initialized as a string?

sproctor: custom_launch_dir is the either a string or nil. It's initialized at line 4604 `if custom_launch_dir = @launch_data.find { |opt| opt =~ /CUSTOMLAUNCHDIR=/ }`

asechrest: If the variable can be `nil` or a string, I think your check works. If I were writing the original code, I'd initialize it such that I knew the data type. But I'm not an expert.

MahtraDR: Damn, thanks for the lesson. That makes sense that the `.empty?` is undefined if the variable is nil.

sproctor: Well, now's as good a time as any. How would you write the initialization? We're already messing with this code, so we might as well make it easy to read while we're at it. Let's ignore for a moment that I had to refer to this line by a number greater than 4000.

On another topic, is anyone planning to refactor this to make it more readable/maintainable? Having 2 huge source files containing almost all of the logic is awkward. I'm really an outsider here. It's not obvious where any discussion/planning happens.

MahtraDR: Now is a good time to refactor this. Making it its own file like you have here is good. We can get it tested easy enough, then it can go into Staging for general testing. It shouldn't impact a lot of people... not too many linux users around.

MahtraDR:
Since you're here... How about just removing `Stormfront` references, and using just `Wrayth`?

sproctor:
There's a lot of other references to Stormfront that I'm not going to mess with. I'll have access to a windows computer to do some more testing next month, until then, I just changed this label.

PR #288, opened 2023-02-05
chore(all): RFC - Add dependencies to Gemfile

sproctor:
I'd like some input here. sqlite3 is a hard dependency, so it can definitely be added here. gtk3 is a hard dependency on Linux, but appears to not be on windows. I think using bundler to install the gem is the way to go. Are there many people using Lich on Windows without GTK3?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,10 +14,5 @@ jobs:
         with:
           ruby-version: ${{ matrix.ruby }}
           bundler-cache: true
-      - name: Install ruby gem dependencies with bundler
-        run: |
-          gem install bundler
-          bundle config path vendor/bundle
-          bundle install --jobs 4 --retry 3
       - name: Run RSpec tests
         run: bundle exec rspec
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,6 +9,8 @@ jobs:
         ruby: ['2.7', '3.0', '3.1', '3.2']
     name: Run tests on Ruby ${{ matrix.ruby }}
     steps:
+      - name: Install GTK3
+        run: sudo apt-get install -y libgtk-3-dev
       - uses: actions/checkout@v3
       - uses: ruby/setup-ruby@v1
         with:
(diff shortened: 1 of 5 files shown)


vtcifer:
> I'd like some input here. sqlite3 is a hard dependency, so it can definitely be added here. gtk3 is a hard dependency on Linux, but appears to not be on windows. I think using bundler to install the gem is the way to go. Are there many people using Lich on Windows without GTK3?

Can you clarify what you mean by hard dependency?  I run lich on linux without gtk _kind of_. I run in docker containers that are built with gtk (all the necessary packages, then the gem is installed, then the packages are all purged - this keeps image size WAY down).  So while gtk gem is there, it never really gets used, and in fact if something tries to use it, it fails.

MahtraDR:
I think of sqlite as the only real hard dependency. 

Gtk is entirely optional.

MahtraDR:
> I'd like some input here. sqlite3 is a hard dependency, so it can definitely be added here. gtk3 is a hard dependency on Linux, but appears to not be on windows. I think using bundler to install the gem is the way to go. Are there many people using Lich on Windows without GTK3?

What makes you say GTK3 is a hard dependency on linux? I play without GTK on linux...

sproctor:
If you try to run Lich without the gtk3 gem installed on Linux it fails with a message saying you need to install it.

MahtraDR:
> If you try to run Lich without the gtk3 gem installed on Linux it fails with a message saying you need to install it.

`unset $DISPLAY`

ondreian:
I think this is a very good proposal, if for now it will only be utilized by devs.  The current weird way we distribute Lich means we cannot actually take advantage of a Gemfile for the most part, but it might make for a much better DevEx

strnglp:
+1

I just ran into this while trying to package Lich for NixOS. In your latest RC there is a Gemfile so I took a crack at it and ran into a couple of issues.

1. There's something funky going on with the sqlite3 definition, probably related to the linux specific suffix but I don't know for sure. I'm not really a Ruby dev, I'd defer to you all to ensure it's specified correctly.
2. When I hacked in the correct sqlite3 I started running into this:
`require': cannot load such file -- mini_portile2 (LoadError)`

I'm not sure but that seems like a dependency that isn't specified either in your Gemfile or in sqlite3's (which would surprise me).

Based on the timing of these comments and the newer nature of your Gemfiles I assume that it's still a WIP but wanted to express some interest in sorting this out so I can help get the package setup on Nixpkgs.

PR #292, opened 2023-02-10
[lib/gui-manual-login.rb] Fixes issue of using wrong variable for custom_launch_dir during save

vtcifer:
The text of the @custom_launch_dir combo box is incorrectly being used to overwrite the whole object itself when being saved, instead of fetching that text and saving it to a new local variable which is actually used during the save

Until PR #285 is merged, this won't have much effect, as setting custom launch commands aren't currently accessible from the gui.


mrhoribu (approved):
Nice catch of wrong variable name. Good for staging branch acceptance!

Issue #280, opened 2023-01-18
Tabbed GUI layout doesn't show instance (at least test vs prime)


vtcifer:
When using tabbed layout on gui, the instance (test vs prime) doesn't show up properly:

Non tabbed:
![image](https://user-images.githubusercontent.com/12104547/213213556-49542bb4-ef8f-43ca-96ba-5dd6613f3d99.png)

Tabbed: 
![image](https://user-images.githubusercontent.com/12104547/213213592-33edeb9e-b9de-40d1-8a3e-dde800dc25c3.png)

OSXLich-Doug:
Thank you.  I'll see if I can address this in a future release.

mrhoribu:
Resolved via #293

PR #293, opened 2023-02-10
[lib/gui-saved-login.rb] Fix mismatch on tabbed GUI and instance names

vtcifer:
- Add all DR game codes for tabbed layout gui
 - Add specific code for GS3 instead of relying on falling through to else
 - Add unknown in case not matching any current known gamecodes.


vtcifer:
Closes #280

mrhoribu:
closes #280

PR #301, opened 2023-02-20
BUG: Ruby 3.2.1 requires change to start SIMU FE

OSXLich-Doug:
Changing up the system call to start up SIMU FE under Ruby 3.2 (why did this change?!)


OSXLich-Doug:
Nope, trying again.

PR #302, opened 2023-02-20
BUGFIX: Update for Ruby 3.2 to start SIMU FE

OSXLich-Doug:
converting system call to spawn (may not work on non Ruby 3.2)


mrhoribu (approved):
This works flawlessly to launch Wrayth/WizardFE without issue.

PR #306, opened 2023-02-24
[update.rb] BUGFIX: to allow subfolders in LIB

OSXLich-Doug:
This PR permits the `--snapshot` function to detect a subfolder in lich5/lib, and properly back it up.  It does not attempt to recurse the lich5/lib tree, and therefore is limited to one level down.


mrhoribu (approved):
Works fine for recursive directories.

mrhoribu:
Should now allow subfolder recursive snapshot and new updates in the LIB folder only. Data/Scripts and program file do not currently support the function and will need to be processed later if subfolder usage is used for such.

PR #309, opened 2023-02-24
[init.rb] Allow --no-gui on all platforms

MahtraDR:
As per title.


lib/init.rb
```diff
@@ -653,7 +653,7 @@ def Wine.registry_puts(key, value)
   exit
 end
 
-if ((RUBY_PLATFORM =~ /mingw|win/i) and (RUBY_PLATFORM !~ /darwin/i) and !ARGV.include?('--no-gui')) or ENV['DISPLAY']
+(if (RUBY_PLATFORM =~ /mingw|win/i) and (RUBY_PLATFORM !~ /darwin/i) or ENV['DISPLAY']) and !ARGV.include?('--no-gui')
```
MahtraDR, line 656:
```suggestion
if ((RUBY_PLATFORM =~ /mingw|win/i) and (RUBY_PLATFORM !~ /darwin/i) or ENV['DISPLAY']) and !ARGV.include?('--no-gui')

```

MahtraDR:
Using this now, works as expected. It's very nice to be able to have X forwarded, and not interfering with lich. When I want gtk stuff, I log out, and log back in without --no-gui. Instead of having to kill my ssh connection, and reconnect with X enabled, and set DISPLAY manually...

PR #311, opened 2023-02-27
[lich.rbw] Add support to specify IP to listen on in detachable client mode

vtcifer:
Enables greater flexibility in detached client mode, especially where this is deployed via docker or other containers.  No need to mess with iptables / firewalld / etc within the container.

### DIFF
--- a/lich.rbw
+++ b/lich.rbw
@@ -4261,9 +4261,12 @@ else
   hosts_dir = nil
 end
 
+detachable_client_host = '127.0.0.1'
 detachable_client_port = nil
 if arg = ARGV.find { |a| a =~ /^\-\-detachable\-client=[0-9]+$/ }
   detachable_client_port = /^\-\-detachable\-client=([0-9]+)$/.match(arg).captures.first
+elsif arg = ARGV.find { |a| a =~ /^\-\-detachable\-client=((?:\d{1,3}\.){3}\d{1,3}):([0-9]{1,5})$/ }
+  detachable_client_host, detachable_client_port = /^\-\-detachable\-client=((?:\d{1,3}\.){3}\d{1,3}):([0-9]{1,5})$/.match(arg).captures
 end
 
 if argv_options[:sal]
@@ -5080,7 +5083,7 @@ main_thread = Thread.new {
     detachable_client_thread = Thread.new {
       loop {
         begin
-          server = TCPServer.new('127.0.0.1', detachable_client_port)
+          server = TCPServer.new(detachable_client_host, detachable_client_port)
           char_name = ARGV[ARGV.index('--login')+1].capitalize
           Frontend.create_session_file(char_name, server.addr[2], server.addr[1])
 


MahtraDR:
Good idea this.

MahtraDR:
What about just having a toggle argument instead? `--bind-all` toggles `detachable_client_host` to `0.0.0.0` instead of localhost? Although as is here, we could bind to some other private address. In my use cases, the toggle would always suffice, which is why I suggest it.

vtcifer:
I feel like this is the better pattern.  It supports the usecase for containers, using `--detachable-client=0.0.0.0:<port>` and what others might want (specific IPs).

mrhoribu:
Dare I be the guy that asks... what about IPv6 support?

vtcifer:
I've specifically turned off ipv6 on my systems because of how annoying it can be wrt things like USB-OTG and RNDIS, and Genie (which doesn't support IPv6 at ALL).  I'm not against adding support IPv6, but I think that is a bigger edge case.

MahtraDR:
If you keep the IP:PORT mappings in separate arguments, you can easily use `Resolv::AddressRegex` natively, without having actually craft a regex for it.

```ruby
require 'resolv'

"192.168.1.1"   =~ Resolv::AddressRegex ? true : false #=> true
"192.168.1.500" =~ Resolv::AddressRegex ? true : false #=> false
"ff02::1"    =~ Resolv::AddressRegex ? true : false #=> true
"ff02::1::1" =~ Resolv::AddressRegex ? true : false #=> false
```
You also get IPv4, and IPv6 in the same match.

MahtraDR:
I can test ipv6 binding no problem.

lich.rbw
```diff
@@ -4261,9 +4261,12 @@ else
   hosts_dir = nil
 end
 
+detachable_client_host = '127.0.0.1'
 detachable_client_port = nil
 if arg = ARGV.find { |a| a =~ /^\-\-detachable\-client=[0-9]+$/ }
   detachable_client_port = /^\-\-detachable\-client=([0-9]+)$/.match(arg).captures.first
+elsif arg = ARGV.find { |a| a =~ /^\-\-detachable\-client=((?:\d{1,3}\.){3}\d{1,3}):([0-9]{1,5})$/ }
+  detachable_client_host, detachable_client_port = /^\-\-detachable\-client=((?:\d{1,3}\.){3}\d{1,3}):([0-9]{1,5})$/.match(arg).captures
 end
```
MahtraDR, line 4270:
```suggestion
detachable_client_host = '127.0.0.1'
if arg = ARGV.find { |a| a =~ /^\-\-detachable\-client\-bind\-ip=(.*)$/ }
  require 'resolv'
  passed_string = /^\-\-detachable\-client\-bind\-ip=(.*)$/.match(arg).captures.first
  if Resolv::AddressRegex =~ passed_string
    Lich.log("Binding detachable client to #{passed_string}")
    detachable_client_host = passed_string
  else
    Lich.log("#{passed_string} is not a valid IP address")
    $stdout.puts ("#{passed_string} is not a valid IP address")
    exit
  end
end

detachable_client_port = nil
if arg = ARGV.find { |a| a =~ /^\-\-detachable\-client=[0-9]+$/ }
  detachable_client_port = /^\-\-detachable\-client=([0-9]+)$/.match(arg).captures.first
end
```

MahtraDR: This works with IPv4, and IPv6, exits if the ip passed isn't valid.

PR #312, opened 2023-02-27
[dr][lich.rbw] Some CLI clean up related to DR specific options

vtcifer:
Still WIP.


MahtraDR:
@mrhoribu This is good, but `$platinum` is only used in two rooms, and we can drop all those references in this file if these two rooms are edited in the mapDB
```
./data/GSIV/map-1646865346.json:    "10329": ";e if $platinum; move 'go furrier'; else; move 'go warehouse'; end"
./data/GSIV/map-1646865346.json:    "10396": ";e if $platinum; move 'go shop'; else; move 'go building'; end"
```

mrhoribu:
> $platinum

Probably should just update those to use `XMLData.game=='GSPlat'` and then could strip out the $platinum value from `lich.rbw`

MahtraDR:
I've tidied up this section, and simplified it a bit.

We no longer conflate game with client (except where we should, apparently, such as for GS+Wrayth? Weird.)

Made the `$platinum` check a lot simpler.

Moved FE selection to its own if block, and added genie.

mrhoribu:
> I've tidied up this section, and simplified it a bit.
> 
> We no longer conflate game with client (except where we should, apparently, such as for GS+Wrayth? Weird.)
> 
> Made the `$platinum` check a lot simpler.
> 
> Moved FE selection to its own if block, and added genie.

Part of me wonders how much of this code block is even used anymore? Does it override what it pulls in from `lib/lich.rb`. Also I'm thinking that the game code section can be cleaned up quite a bit. Granted, I'd love to add in a `--test` option for GS4 but haven't poked around enough to know what the port # is to add it for the GS4 side. But here's the cleaned up code block:
```ruby
if arg = ARGV.find { |a| (a == '-g') or (a == '--game') }
  game_host, game_port = ARGV[ARGV.index(arg) + 1].split(':')
  game_port = game_port.to_i
elsif ARGV.include?('--gemstone')
  game_host = 'gs4.simutronics.net'
  if $platinum
    game_port = 10124
  else
    game_port = 10024
  end
elsif ARGV.include?('--shattered')
  game_host = 'gs4.simutronics.net'
  game_port = 10324
elsif ARGV.include?('--fallen')
  game_host = 'dr.simutronics.net'
  game_port = 11324
elsif ARGV.include?('--dragonrealms')
  game_host = 'dr.simutronics.net'
  if $platinum
    game_port = 11124
  elsif ARGV.include?('--test')
    game_port = 11624
  else    # DR PRime
    game_port = 11024
  end
else
  game_host, game_port = nil, nil
  Lich.log "info: no force-mode info given"
end
```
As apparently all those various hostnames for GS4 all resolve back to chimera.simutronics.net and all the various DR hostnames resolve back to hydra.simutronics.net. So should be able to just use the two referenced here. Does make me wonder if we should clean up the hostnames that are being used in `lib\lich.rb` to do the same.

MahtraDR:
> > I've tidied up this section, and simplified it a bit.
> > We no longer conflate game with client (except where we should, apparently, such as for GS+Wrayth? Weird.)
> > Made the `$platinum` check a lot simpler.
> > Moved FE selection to its own if block, and added genie.
> 
> Part of me wonders how much of this code block is even used anymore for hostname/port? Does it override what it pulls in from `lib/lich.rb`. Meaning, I currently connect via `ruby $LICH_BIN --login $CHAR --detachable-client=$port --without-frontend` and don't specify the game at all and just let Lich choose from the saved entries and what it's chosen there.
> 
> Also, if we're adding more FE's to the list, we should probably add `--profanity` and `--illthorn` to the list too. Granted I believe ProfanityFE sets the variable to itself upon launching, so not sure if it's needed for it's case. Likely same for Illthorn.
> 
> Also I'm thinking that the game code section can be cleaned up quite a bit. Granted, I'd love to add in a `--test` option for GS4 but haven't poked around enough to know what the port # is to add it for the GS4 side. But here's the cleaned up code block:
> 
> ```ruby
> if arg = ARGV.find { |a| (a == '-g') or (a == '--game') }
>   game_host, game_port = ARGV[ARGV.index(arg) + 1].split(':')
>   game_port = game_port.to_i
> elsif ARGV.include?('--gemstone')
>   game_host = 'gs4.simutronics.net'
>   if $platinum
>     game_port = 10124
>   else
>     game_port = 10024
>   end
> elsif ARGV.include?('--shattered')
>   game_host = 'gs4.simutronics.net'
>   game_port = 10324
> elsif ARGV.include?('--fallen')
>   game_host = 'dr.simutronics.net'
>   game_port = 11324
> elsif ARGV.include?('--dragonrealms')
>   game_host = 'dr.simutronics.net'
>   if $platinum
>     game_port = 11124
>   elsif ARGV.include?('--test')
>     game_port = 11624
>   else    # DR PRime
>     game_port = 11024
>   end
> else
>   game_host, game_port = nil, nil
>   Lich.log "info: no force-mode info given"
> end
> ```
> 
> As apparently all those various hostnames for GS4 all resolve back to chimera.simutronics.net and all the various DR hostnames resolve back to hydra.simutronics.net. So should be able to just use the two referenced here. Does make me wonder if we should clean up the hostnames that are being used in `lib\lich.rb` to do the same.

This sounds very reasonable to me. 

I don't see this PR as the be all, end all. More work can be done still. This one just moves the needle forward a bit in the right direction.

vtcifer:
Closing as masssivly stale.  Will probably try to take a stab at this at some point in the future.

PR #318, opened 2023-03-07
chore: infomon script to library conversion

OSXLich-Doug:
First proto release for infomon in lib / rb files


lib/infomon.rb
```diff
@@ -0,0 +1,515 @@
...
+
+    respond 'Calling need psm update' if $infomon_debug
+    $infomon_values['need_psm_update'] = true if $infomon_values['need_psm_update'].nil?
+
+    #
+    # Load spell info
+    #
+    unless Spell.load
+      echo 'error: failed to load spell list'
+      exit
```
ondreian, line 57: We should not do this, this is exit the Process

lib/infomon.rb
```diff
@@ -0,0 +1,515 @@
...
+                      %w[Feat Armor Weapon Shield CMan]
+                    end
+    need_psm = []
+    if $infomon_values['need_psm_update']
+      need_psm = psm_abilities.dup
+      $infomon_values['need_psm_update'] = false
+    else
+      psm_abilities.each do |ability|
+        if !defined?(Kernel.const_get(ability.to_s))
+          pp 'Encountered fatal error'
```
ondreian, line 76: I think instead of the myriad of `_respond` or `pp` we should write a `Logger` module that handles this stuff, I just ran into an issue with the `HMR` pr where I can't run rspec tests because a lot of the scripting globals aren't defined in CI

lib/infomon.rb
```diff
@@ -0,0 +1,515 @@
...
+        respond 'Now LOADING Stats, Skills, Spells, Society, Citizenship' if $infomon_debug
+        Stats.load_serialized   = $infomon_values['Stats']
+        Skills.load_serialized  = $infomon_values['Skills']
+        Spells.load_serialized  = $infomon_values['Spells']
+        Society.load_serialized = $infomon_values['Society']
+        Char.citizenship        = $infomon_values['citizenship']
+      rescue StandardError
+        echo $ERROR_INFO
+        echo $ERROR_INFO.backtrace[0..1]
+        exit
```
ondreian, line 118: This will log you out of the game with no info

lib/infomon.rb
```diff
@@ -0,0 +1,515 @@
...
+    $infomon_sleeping = false
+    $infomon_bound = false
+    $infomon_silenced = false
+    $infomon_calmed = true
+    $infomon_cutthroat = false
+    #
+    # Save function
+    #
+    def self.save_proc
+      @save_value = $infomon_values
```
ondreian, line 244: I don't understand why we are copying a global var to an instance var like this

lib/infomon.rb
```diff
@@ -0,0 +1,515 @@
...
+        Infomon::Monitor.save_proc
+      end
+    rescue StandardError
+      echo $ERROR_INFO
+      echo $ERROR_INFO.backtrace[0..1]
+    end
+    #
+    # Save current status on exit
+    #
+    before_dying do
```
ondreian, line 273: how does this even work at the top-level?

ondreian (changes requested):

lib/infomon.rb
mrhoribu: Removed this section completely in latest commits.

lib/infomon.rb
mrhoribu: Scoped the library back to being required to be loaded from a script, `infomon.lic` for now. Which allows for safe `exit` on this failure.

lib/infomon.rb
mrhoribu: Scoped the library back to being required to be loaded from a script, `infomon.lic` for now. Which allows for safe `exit` on this failure.

lib/infomon.rb
OSXLich-Doug: May be a bit too conservative, but I was capturing the current global var state in case it tried to change on me mid write.  Same thinking as behind the $1.dup convention.

What I'd really like to do is have an instance variable throughout - but I clearly have yet to conquer namespace and scope.  I'd rather not have $infomon_values, instead @infomon_values.  Perhaps, in time.

OSXLich-Doug:
Closed no merge - PR superseded by PR #325

PR #319, opened 2023-03-07
[lich.rbw] minor enhancements

mrhoribu:


MahtraDR:
I understand what 'l' is here, but does it hurt anything if we make it 'look' instead? It'll make the whole function easier to read and understand.

mrhoribu:
> I understand what 'l' is here, but does it hurt anything if we make it 'look' instead? It'll make the whole function easier to read and understand.

I don't see any reason for it not to be the full `look` instead of just `l`.

lich.rbw
```diff
@@ -1042,6 +1042,18 @@ module Games
...
+        # Add check to determine if the game server hung at initial response
+        
+        @@wrap_thread = Thread.new {
+          @last_recv = Time.now
+          while !@@autostarted && (Time.now - @last_recv < 6)
+            break if @@autostarted
+            sleep 0.2
+          end
+
+          puts 'l' if !@@autostarted
```
MahtraDR, line 1054:
```suggestion
          puts 'look' if !@@autostarted
```

PR #320, opened 2023-03-07
chore: script class to library conversion

OSXLich-Doug:
Moving script class out to lib


mrhoribu (approved):
Tested working with no issue so far.

PR #321, opened 2023-03-07
feat: add hmr module;

ondreian:


mrhoribu:
Shouldn't we require this somewhere in lich.rbw so we can then make use of it natively? Thinking perhaps building an upstream hook to just pass the argument? Something like `;reload <NAME>`

ondreian:
We can require it somewhere in there for sure, I just wanted to write the library for you to continue to work with so you weren't blocked

PR #330, opened 2023-03-15
chore: Adding infomon boolean status methods;

OSXLich-Doug:
Adds API for `Status.effect` and carries forward calmed? sleeping? etc.


spec/infomon_spec.rb
mrhoribu, line 1: Change the top `require "stats/stats_gs"` to `require "stats/stats"` and the tests should then be able to run properly.

PR #332, opened 2023-03-15
Chore/add active spell to infomon module

OSXLich-Doug:
This may ultimately be moved into `class Spell` but is placed here for the time being.


OSXLich-Doug:
Um, no.  Closing request (wrong merge target.)

PR #337, opened 2023-03-17
chore:add spells to infomon module

OSXLich-Doug:
Adds spells back into the mix for infomon


lib/infomon/parser.rb
```diff
@@ -94,7 +94,7 @@ def self.parse(line)
         when Pattern::Spell
           # todo: capture SK item spells here?
           match = Regexp.last_match
-          Infomon.set("spell.%s" % match[:name], match[:rank].to_i)
+          Infomon.set("spell.%s" % match[:name].gsub(' ', '_'), match[:rank].to_i)
```
ondreian, line 97: I think this should be normalized in the `_key` method

lib/stats/spells.rb
```diff
@@ -0,0 +1,61 @@
+require "ostruct"
+
+module Spells
+  def self.get_circle_name(num)
```
ondreian, line 4: Yes! well crafted @OSXLich-Doug

lich.rbw
```diff
@@ -2135,6 +1996,7 @@ module Games
     require_relative("./lib/weapon.rb")
     require_relative("./lib/infomon/infomon.rb")
     require_relative("./lib/stats/stats.rb")
+    require_relative("./lib/stats/spells.rb")
```
ondreian, line 1999: this should probably exist at `lib/spells.rb` since they aren't exactly a character stat, but I don't know, could go either way.

ondreian (changes requested):

lich.rbw
OSXLich-Doug: I can go either way as well - you probably saw the discussion with Tysong on it.  After we're done with the code, let me know and I'll put it in final place.  I have no real concern either way.

lib/stats/spells.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  def self.known
+    known_spells = Array.new
+    Spell.list.each { |spell| known_spells.push(spell) if spell.known? }
+    return known_spells
+  end
+
+  def self.serialize
+    [self.minor_elemental, self.major_elemental, self.minor_spiritual, self.major_spiritual, self.wizard, self.sorcerer, self.ranger, self.paladin, self.empath, self.cleric, self.bard, self.minormental]
```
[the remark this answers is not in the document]
ondreian: This should only be for backwards compat, and be slated for removal in the future

lib/stats/spells.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+
+  @@spell_lists = %i(major_elemental major_spiritual minor_elemental minor_mental minor_spiritual bard cleric empath paladin ranger sorcerer wizard)
```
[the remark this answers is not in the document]
ondreian: This should only be for backwards compat, and be slated for removal in the future

OSXLich-Doug:
agree - probably in `spell.rb` rewrite is where I'll tackle this, likely
not deprecating but just removing.

PR #339, opened 2023-03-18
chore: adding cli for infomon module

OSXLich-Doug:
Basic three function CLI.

Gussying up the output for `Infomon.show` is an exercise left to the reader.


ondreian (approved):
lgtm

PR #340, opened 2023-03-19
bugfix:init.rb - detecting 20 files vs detecting 20 lich debug files;

OSXLich-Doug:
Current Lich will sometimes fail to run for a few attempts and then mysteriously start with no other changes.  The issue is current Lich detects more than 20 files in TEMP_DIR and tries to delete `debug<numbers.log` files over 20.  But many activities (like `;repository`) also save files to TEMP_DIR, leading to a `TRUE` result for more than 20 files, but potentially a `FALSE` result for 20 debug files. 

This change detects only debug files, and actions only debug files to prevent the intermittent failure.


mrhoribu (dismissed):
Looks good.

mrhoribu (approved):
Looks good for merger.

PR #344, opened 2023-03-21
chore:Infomon additions (CLI, Update, Status)

mrhoribu:
Added an upstream hook in `global_defs.rb` for infomon CLI. Fixed `global_defs.rb` checks for various Status module checks.


mrhoribu:
Incorporated #342 changes into PR.

PR #345, opened 2023-03-22
[spell.rb] - v1.2.4 multiple updates

mrhoribu:
Fix for Spell.cast method to repeat casting if generated spell preparation time due to a debuff in effect.


mrhoribu:
Fix for affordable? to check Overexerted on stamina checks added as well

PR #347, opened 2023-03-29
chore: core lich5 global update stormfront -> wrayth

OSXLich-Doug:
Requires testing on all platforms.

Expect scripts to break due to expected value of 'stormfront' rather than 'wrayth' (new value).  Several of these are EoL scripts. (84 scripts in total in EOL / scripts alone, unknown on repo)


OSXLich-Doug:
Reqiures substantive rubocop cleanup - I'll wok on this throughout the week, but some of these failures are poor coding that need to be addressed.

OSXLich-Doug:
Redo with current lich.rbw file

PR #351, opened 2023-04-12
[lich.rbw] update entry for test servers

gs4-Xanlin:
GS uses the same login information as prime (assuming the same for DR).  Instead of duplicating login information under game GST, uses the login info from GS3 and applies the appropriate game code.


ondreian (approved):
lgtm, thanks!

PR #352, opened 2023-04-13
Bounty API: Adds River's Rest example for guard regex

sandersch:
Moved some of my leveling dudes to RR and noticed that the completed task regex wasn't handling things correctly there. This adds an example bounty to the tests and updates the parser to recognize RR info.


sandersch:
First commit is the relevant changes. I also updated these files to pass the rubocop checks in a separate commit since it was minimal work.

PR #354, opened 2023-05-17
correct infomon sync arguments passed

OSXLich-Doug:
Explicitly declare all arguments in `cli.rb` call to `util.rb` method `issue_command` to ensure proper argument to method parameter alignment.


mrhoribu (approved):
Verified syntax of  `issue_command` matching full method parameters. Looks good.

PR #358, opened 2023-05-22
Adding society and char to infomon.

OSXLich-Doug:
Adds `Society` and `Char` classes to infomon rewrite.

Note: Also loads `activespell` after XMLData is instantiated to resolve the `ActiveSpell::XMLData` issue @MahtraDR reported.


MahtraDR:
Looks good here. That error is gone now for me.

PR #370, opened 2023-06-05
lib/bounty/parser.rb update

gs4-Xanlin:
adds town to assignment tasks, adds capture for npc_name, and assist (pc name)


sandersch (approved):
LGTM

PR #383, opened 2023-06-10
chore: Infomon - upsert_batch speed improvement

OSXLich-Doug:
Now uses blob_set for INFO, EXP, SKILL, and PSM

### DIFF
--- a/lib/infomon/infomon.rb
+++ b/lib/infomon/infomon.rb
@@ -33,6 +33,7 @@ module Infomon
 
   def self.context!
     return unless XMLData.name.empty? or XMLData.name.nil?
+
     puts Exception.new.backtrace
     fail "cannot access Infomon before XMLData.name is loaded"
   end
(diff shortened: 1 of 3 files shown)


OSXLich-Doug:
@mrhoribu @ondreian - I finally got the specs to run.  Environment hasn't been stable, and the method of collecting lines means we should test :noop returns and :ok returns, but we can't just blanket accept :noop.  So I opted to print the status out during the runs so they can be validated and confirmed.

However, Rubocop doesn't love me, and for some reason, `heredoc` is not playing well.  So I am going to suggest we accept the recommendations from Rubocop, thank it, and continue.  Please review.

lib/infomon/infomon.rb
```diff
@@ -94,6 +98,26 @@ def self.set(key, value)
     self.upsert(key: self._key(key), value: self._validate!(key, value))
   end
 
+  def self.blob_set(*blob)
```
ondreian, line 101: Why did we add another method that does the same thing? Also we should not be wrapping this in `Thread.new` like that because we have now created a race conditon

ondreian (changes requested):
Please remove `Thread.new` logic and de-duplicate methods

mrhoribu (approved):
Looks good after the changes. We'll add the Threading and Mutex in a future PR.

PR #387, opened 2023-06-12
[skills.rb] bugfix for lore naming

mrhoribu:
Lore names in the @@skills variable had an extra `_` in their names when they're not being saved with it.


mrhoribu:
Closing as fixing key names to match other naming conventions so all uniform.

PR #391, opened 2023-06-12
feat: infomon/parser Mutex and threading

mrhoribu:
Add a mutex and thread pool to speed up writes and ensure data consistency between reads/writes


mrhoribu:
Migration over to #395 and this PR no longer needed. Closing.

PR #392, opened 2023-06-13
[infomon.rb] context! change to use underscore instead of decimal

gs4-Xanlin:
SQLite naming conventions use alphanumeric and underscore for naming, using a dot notation requires quoting the name.

Dots are for identifying objects, usually in the database.schema.table.column pattern.

Would be better to do this before shipping.

### DIFF
--- a/lib/infomon/infomon.rb
+++ b/lib/infomon/infomon.rb
@@ -40,7 +40,7 @@ module Infomon
 
   def self.table_name
     self.context!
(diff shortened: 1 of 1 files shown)


mrhoribu (approved):
In agreement, better to use good conventions before it's too late. Thanks for the catch on this.

OSXLich-Doug:
Baked into subsequent PR

PR #393, opened 2023-06-14
[parser.rb] additional regex anchoring

gs4-Xanlin:
redo of #380 since it got lost somewhere, and adds an anchor for SkillStart


OSXLich-Doug (approved):
Reviewed, looks strong.

PR #404, opened 2023-06-22
[parser.rb] bugfix for SKILL BASE output start msg

mrhoribu:
Update SkillStart to match for SKILL and SKILL BASE

### DIFF
--- a/lib/infomon/parser.rb
+++ b/lib/infomon/parser.rb
@@ -15,7 +15,7 @@ module Infomon
       TotalExp = /^\s+Total Exp: (?<total_experience>[\d,]+)\s+Death's Sting: \w+$/.freeze
       LTE = /^\s+Long-Term Exp: (?<long_term_experience>[\d,]+)\s+Deeds: (?<deeds>\d+)$/.freeze
       ExprEnd = /^\s+Exp (?:until lvl|to next TP): [\d,]+/.freeze
-      SkillStart = /^\s\w+\s\(at level \d+\), your current skill bonuses and ranks/.freeze
+      SkillStart = /^\s\w+\s\(at level \d+\), your (?:current|base) skill bonuses(?: and ranks|, ranks and goals)/.freeze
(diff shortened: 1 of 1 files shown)


OSXLich-Doug:
I'm not sure of the value to this change PR

mrhoribu:
> I'm not sure of the value to this change PR

Without this change, doing a SKILL BASE doesn't trigger the batch start processing for the skill group as it doesn't match the regex as the current regex only matches SKILLS or SKILLS FULL. Not SKILLS BASE or SKILLS FULL BASE. So you end up getting a massive error due to the mutex not getting locked and the initialization of the skills array to push values into.

mrhoribu:
SKILLS BASE is what is shown after updating your skills from a GOALS change.

OSXLich-Doug:
I see.  For goals (SKILL BASE) you get a clean run.  For full skill base, you get a ton of untrained skills.  Is there a way to isolate it in regex (rather than logic) so we don't get a ton of zero values in the DB?

OSXLich-Doug:
Sorry, wrong button.  Reopened

mrhoribu:
> I see. For goals (SKILL BASE) you get a clean run. For full skill base, you get a ton of untrained skills. Is there a way to isolate it in regex (rather than logic) so we don't get a ton of zero values in the DB?

There isn't any other logic to clean up. The only issue was triggering the START capture to lock the mutex. The rest still works fine. The output from `SKILL` vs `SKILL BASE` matches the other regex used to fill out various skills and end the capture. The only time you'd get 0 values is if you do a FULL for either, and this PR wasn't opened to address that.

OSXLich-Doug:
Agree that it wasn't opened to address it, but it did cause zero entry populations.  A lingering problem we have from the PSMs landing with 0 / 0 in the db as well.  But that's ok.  I'll address it in another PR.

OSXLich-Doug (approved):
LGTM for what it was intended to do.

PR #407, opened 2023-06-23
Society, Infomon.show, .delete! and better testing for infomon

OSXLich-Doug:
This update returns the `||=` method and adds `Infomon.delete!(key)` method to remove a row from infomon.db.

This update includes webbed, stunned and dead to module `Status`.

This update changes `Infomon.show` to display non-zero values, unless `Infomon.show(true)`.

This update includes parsing and db logic for `Society` joins, promotions, and resignations.

This update includes improved testing for new features and for existing features that did not fully test.

### DIFF
--- a/lib/infomon/cli.rb
+++ b/lib/infomon/cli.rb
@@ -34,13 +34,20 @@ module Infomon
     respond 'Infomon reset is now complete.'
   end
 
-  def self.show
+  def self.show(full = false)
     response = []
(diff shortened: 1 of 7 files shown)


mrhoribu (approved):
No glaring issues on my end so far with changes. Likely still needs more testing with actual society changes, but not sure best way to do that without actually doing society changes ingame.

PR #411, opened 2023-06-26
[parser.rb] update PSMStart to trigger on <PSM> INFO

mrhoribu:
Currently only triggers on <PSM> LIST. This allows <PSM> INFO to also allow capture of PSM info.


mrhoribu:
Unfortunately will not work due to issues with no end capture. Closing.

PR #412, opened 2023-06-26
[util.rb] fix to not squelch end_pattern

mrhoribu:
Bug fix to not squelch the end_pattern if not set to quiet.


OSXLich-Doug:
LGTM 👍

PR #415, opened 2023-07-02
[ascension.rb] creation of Ascension module and infomon parsing

mrhoribu:
Creates an Ascension module as part of PSMS due to relying on similar architecture and system backend.

### DIFF
--- a/lib/infomon/infomon.rb
+++ b/lib/infomon/infomon.rb
@@ -60,7 +60,12 @@ module Infomon
   end
 
   def self.reset!
-    self.mutex.lock
+    begin
+      self.mutex.lock
+    rescue StandardError
+      respond "--- Lich: error: self.reset!: #{$!}"
+      Lich.log "error: self.reset!: #{$!}\n\t#{$!.backtrace.join("\n\t")}"
+    end
     Infomon.db.drop_table?(self.table_name)
     self.cache.clear
     @cache_loaded = false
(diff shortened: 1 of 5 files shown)


OSXLich-Doug (commented):
in `parser.rb`

```          if category =~ /Ascesion/
            @psm_cat = 'ascension'
          else
            @psm_cat = 'psm'
          end
```

Suspect that should be `if category =~ /Ascension/

Three places.

mrhoribu:
> in `parser.rb`
> 
> ```
>           if category =~ /Ascesion/
>             @psm_cat = 'ascension'
>           else
>             @psm_cat = 'psm'
>           end
> ```
> 
> Suspect that should be `if category =~ /Ascension/`
> 
> Three places.

Good catch. Thanks.

mrhoribu:
Updated PR with additional debug output for Mutex lock/unlock/synchronize incase of error. Also updated SkillEnd pattern match to ignore if not currently in a mutex lock state as TPs being earned messaging was triggering match.
```
You've gained 1 physical and mental training point.
Unconverting 1 Mnt points to return 2 Phy points. (See HELP TPS)
--- Lich: error: Pattern::SkillEnd: Attempt to unlock a mutex which is not locked
Training Points: 67 Phy 0 Mnt (4326 Phy converted to Mnt)
```

OSXLich-Doug (approved):
LGTM

PR #416, opened 2023-07-03
[update.rb]:Adding beta support

OSXLich-Doug:
Adding beta support for update.rb


OSXLich-Doug:
Close, no merge (need to correct .DS_Store issue)

PR #418, opened 2023-07-04
[status.rb] typo for webbed using wrong indicator

mrhoribu:
Had `self.webbed?` using `XMLData.indicator['IconDEAD']` instead of `XMLData.indicator['IconWEBBED']`


OSXLich-Doug (approved):
GTG

PR #425, opened 2023-07-11
chore: refactor Lich::Util.issue_command

mrhoribu:
Refactor method to use named parameters instead of mandatory order of passed variables. Doing now before live so as easier to maintain going forward.


OSXLich-Doug (approved):
GTG

PR #426, opened 2023-07-11
feat: Resource module

mrhoribu:
Track GS4 profession service resource and suffusion. Relies on new `issue_command` from #425


OSXLich-Doug (approved):
LGTM

PR #427, opened 2023-07-12
[lich.rbw] fix #304 for constants in ARGV after loading constants.rb

### LINKED ISSUE #304: Some commandline arguments no longer work due to freezing of their values in [lib/constants.rb]
There are command line options to use custom directories for most (?all?) directories.
https://github.com/elanthia-online/lich-5/blob/40d87670b4a159c4b3e65c0151f83b09d58e39b7/lich.rbw#L4186-L4199

Due to the order these command line arguments are processed compared to the execution of lib/constants.rb, which freezes most of those cannot be set.
https://github.com/elanthia-online/lich-5/blob/40d87670b4a159c4b3e65c0151f83b09d58e39b7/lib/constants.rb#L2-L8

I'm not sure if the correct fix here is reordering the require of that constants.rb, or simply not freezing those.

gs4-Xanlin:
fix #304 for constants in ARGV after loading constants.rb

### DIFF
--- a/lich.rbw
+++ b/lich.rbw
@@ -36,6 +36,25 @@
 # Lich is maintained by Matt Lowe (tillmen@lichproject.org)
 # Lich version 5 and higher maintained by Elanthia Online and only supports GTK3 Ruby
 
+# process ARGV for constants before loading constants.rb: issue #304
+for arg in ARGV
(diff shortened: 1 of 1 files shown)


mrhoribu (approved):
looks good

PR #428, opened 2023-07-13
[lich.rbw][xmlparser.rb] Remove NAV tag rewrites, fix in xmlparser instead

MahtraDR:
Allows xmlparser to parse nav tags when nested in the same line as other attributes.

### DIFF
--- a/lib/xmlparser.rb
+++ b/lib/xmlparser.rb
@@ -236,17 +236,6 @@ class XMLParser
         end
       end
 
-      if name == 'compass' and $nav_seen
-        $nav_seen = false
-        @second_compass = true
-      end
-
-      if name == 'compass' and @second_compass
-        @second_compass = false
-        @room_count += 1
-        $room_count += 1
-      end
-
       if name =~ /^(?:a|right|left)$/
         @obj_exist = attributes['exist']
         @obj_noun = attributes['noun']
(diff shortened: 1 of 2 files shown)


lib/xmlparser.rb
```diff
@@ -713,6 +718,18 @@ def text(text_string)
 
   def tag_end(name)
     begin
+
+      if name == 'compass' and $nav_seen
+        $nav_seen = false
+        @second_compass = true
+      end
+
+      if name == 'compass' and @second_compass
```
MahtraDR, line 727:
```suggestion
      elsif name == 'compass' and @second_compass
```

lib/xmlparser.rb
```diff
@@ -713,6 +718,16 @@ def text(text_string)
...
     begin
+
+      if name == 'compass' and $nav_seen
+        $nav_seen = false
+        @second_compass = true
+      elsif name == 'compass' and @second_compass
+        @second_compass = false
+        @room_count += 1
+        $room_count += 1
+      end
```
MahtraDR, line 729:
```suggestion
      if @game =~ /^DR/
        if name == 'compass' and $nav_seen
          $nav_seen = false
          @second_compass = true
        end
        if name == 'compass' and @second_compass
          @second_compass = false
          @room_count += 1
          $room_count += 1
        end
      end
```

PR #435, opened 2023-07-20
[update.rb] bugfix for backing non-existent data files

mrhoribu:
Resolves trying to backup a file that does not exist, like the new effect-list.xml

```
--- Lich: error: client_thread: No such file or directory @ rb_sysopen - D:/Lich5/data/effect-list.xml
D:/Lich5/lib/update.rb:234:in `initialize'
```


mrhoribu:
Tested and confirmed working.

PR #438, opened 2023-07-22
[parser.rb] account for negative silver or fame

mrhoribu:


OSXLich-Doug (commented):
@mrhoribu 

```      
      StatEnd = /^Mana:\s+-?\d+\s+Silver:\s(?<silver>[-\d,]+)$/.freeze
      Fame = /^\s+Level: \d+\s+Fame: (?<fame>[-\d,]+)$/.freeze # serves as ExprStart
```
      
      Doesn't that need to have a ? (optional) for <fame> and for <silver>?  It already has one for Mana.

mrhoribu:
> @mrhoribu
> 
> ```
>       StatEnd = /^Mana:\s+-?\d+\s+Silver:\s(?<silver>[-\d,]+)$/.freeze
>       Fame = /^\s+Level: \d+\s+Fame: (?<fame>[-\d,]+)$/.freeze # serves as ExprStart
> ```
> 
> ```
>   Doesn't that need to have a ? (optional) for <fame> and for <silver>?  It already has one for Mana.
> ```

it shouldn't need the optional since we're capturing the whole thing with or without it, it's a character set and doesn't have to include them all. We could technically update the Mana one to be `[-\d]+` if we wanted. But we aren't capturing mana for anything so it's doesn't really matter. For Fame and Silver we do care about it and it being in a character set means it should capture with or without it.

PR #439, opened 2023-07-23
feat: Currency module

mrhoribu:
Add Lich::Currency module into the system using Infomon::Parser to populate data when seeing output from WEALTH, INFO, or TICKET BALANCE


mrhoribu:
Holding for post-Lich 5.7.0

PR #440, opened 2023-07-24
[task.rb] fix to :heirloom_found for guard?

mrhoribu:
Need to add `:heirloom_found` match for `guard?` method as you need to return to the guard first before going to the taskmaster.


sandersch (approved):
I have this change pending locally with a few others but have been too lazy to put a PR together. Rubocop's changes also make sense to me especially the safe navigation fix.

PR #451, opened 2023-07-28
[Bounty API] Updates heirloom-related methods for easier use by scripts

sandersch:
Some tweaks to the API mostly related to heirloom tasks based on consumption with client scripts.

### DIFF
--- a/lib/bounty/task.rb
+++ b/lib/bounty/task.rb
@@ -121,7 +121,7 @@ class Bounty
 
     def ready?
       [
-        :escort_assignment,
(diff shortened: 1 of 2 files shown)


mrhoribu (commented):
Overall looks good. Only thing I've found so far is when tasks are shared and before you get details of the task, those that are added to the task show `:none` for current task. Uncertain if should address now or do via another update later.

sandersch:
Gonna merge this one and then we can work on getting shared bounties to parse correctly in a followup.

PR #455, opened 2023-07-28
[update.rb] - change prod update path to zipball

OSXLich-Doug:
Correct `update.rb` default prod source from tarball to zipball to ensure proper files are downloaded at request.

### DIFF
--- a/lib/update.rb
+++ b/lib/update.rb
@@ -172,7 +172,7 @@ module Lich
         JSON::parse(update_info).each { |entry|
           if entry.include? 'tag_name'
             @update_to = entry[1].sub('v', '')
-          elsif entry.include? 'tarball_url'
+          elsif entry.include? 'zipball_url'
             @zipfile = entry[1]
           elsif entry.include? 'body'
             @new_features = entry[1].gsub(/\#\# What's Changed.+$/m, '')


lib/update.rb
```diff
@@ -172,7 +172,7 @@ def self.prep_update
         JSON::parse(update_info).each { |entry|
           if entry.include? 'tag_name'
             @update_to = entry[1].sub('v', '')
-          elsif entry.include? 'tarball_url'
+          elsif entry.include? 'zipball_url'
```
mrhoribu, line 175: This won't work, need to reference the ASSET not the source which this does. Should be able to mimic the beta code and change this to be as follows:

From:
```ruby
        JSON::parse(update_info).each { |entry|
          if entry.include? 'tag_name'
            @update_to = entry[1].sub('v', '')
          elsif entry.include? 'zipball_url'
            @zipfile = entry[1]
          elsif entry.include? 'body'
            @new_features = entry[1].gsub(/\#\# What's Changed.+$/m, '')
          end
        }
```
to
```ruby
        JSON::parse(update_info).each { |entry|
          if entry.include? 'tag_name'
            @update_to = entry[1].sub('v', '')
          elsif entry.include? 'assets'
            @holder = value
          elsif entry.include? 'body'
            @new_features = entry[1].gsub(/\#\# What's Changed.+$/m, '')
          end
        }
        beta_asset = @holder.find { |x| x['name'] =~ /lich-5.tar.gz/ }
        @zipfile = beta_asset.fetch('browser_download_url')
```

mrhoribu (changes requested):

OSXLich-Doug:
Closing due to inappropriate code submit.

PR #456, opened 2023-07-28
[update.rb] change method to pull asset tar file

OSXLich-Doug:
Changed prod tar retrieval based on suggestion to match beta file fetch method.  Change supports workflow change to delete data files from Lich repo and build EO scripts repo data files into build.

### DIFF
--- a/lib/update.rb
+++ b/lib/update.rb
@@ -171,7 +171,7 @@ module Lich
 
         JSON::parse(update_info).each { |entry, value|
           if entry.include? 'tag_name'
-            @update_to = entry[1].sub('v', '')
+            @update_to = value.sub('v', '')
           elsif entry.include? 'assets'
             @holder = value
           elsif entry.include? 'body'
--- a/lib/update.rb
+++ b/lib/update.rb
@@ -172,12 +172,14 @@ module Lich
         JSON::parse(update_info).each { |entry|
           if entry.include? 'tag_name'
             @update_to = entry[1].sub('v', '')
-          elsif entry.include? 'tarball_url'
-            @zipfile = entry[1]
+          elsif entry.include? 'assets'
+            @holder = value
           elsif entry.include? 'body'
-            @new_features = entry[1].gsub(/\#\# What's Changed.+$/m, '')
+            @new_features = value.gsub(/\#\# What's Changed.+$/m, '')
           end
         }
+        beta_asset = @holder.find { |x| x['name'] =~ /lich-5.tar.gz/ }
+        @zipfile = beta_asset.fetch('browser_download_url')
       end
 
       def self.download_update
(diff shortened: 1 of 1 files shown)


lib/update.rb
```diff
@@ -172,12 +172,14 @@ def self.prep_update
...
-            @zipfile = entry[1]
+          elsif entry.include? 'assets'
+            @holder = value
           elsif entry.include? 'body'
-            @new_features = entry[1].gsub(/\#\# What's Changed.+$/m, '')
+            @new_features = value.gsub(/\#\# What's Changed.+$/m, '')
           end
         }
+        beta_asset = @holder.find { |x| x['name'] =~ /lich-5.tar.gz/ }
+        @zipfile = beta_asset.fetch('browser_download_url')
```
mrhoribu, line 182: should probably change this from `beta_asset` to `release_asset` since it's not for the beta section

lib/update.rb
```diff
@@ -172,12 +172,14 @@ def self.prep_update
         JSON::parse(update_info).each { |entry|
           if entry.include? 'tag_name'
             @update_to = entry[1].sub('v', '')
-          elsif entry.include? 'tarball_url'
-            @zipfile = entry[1]
+          elsif entry.include? 'assets'
+            @holder = value
           elsif entry.include? 'body'
-            @new_features = entry[1].gsub(/\#\# What's Changed.+$/m, '')
+            @new_features = value.gsub(/\#\# What's Changed.+$/m, '')
```
mrhoribu, line 178: in order to use `value` we need to change the `.each` to have two keys so it's:
```ruby
        JSON::parse(update_info).each { |entry, value|
          if entry.include? 'tag_name'
            @update_to = value.sub('v', '')
          elsif entry.include? 'assets'
            @holder = value
          elsif entry.include? 'body'
            @new_features = value.gsub(/\#\# What's Changed.+$/m, '')
```

mrhoribu (changes requested):

lib/update.rb
```diff
@@ -169,15 +169,17 @@ def self.prep_update
         filename = "https://api.github.com/repos/elanthia-online/lich-5/releases/latest"
         update_info = URI.parse(filename).open.read
 
-        JSON::parse(update_info).each { |entry|
+        JSON::parse(update_info).each { |entry, value|
           if entry.include? 'tag_name'
             @update_to = entry[1].sub('v', '')
```
mrhoribu, line 174: This still needs updating to use `value` so it should end up looking like:
```ruby
@update_to = value.sub('v', '')
```

mrhoribu (changes requested):

lib/update.rb
OSXLich-Doug: 👍

mrhoribu (approved):
Lgtm

PR #457, opened 2023-07-29
[cli.rb] save last Infomon.sync to db

mrhoribu:
Save the last time `Infomon.sync` was ran to the db to have a valid check for first run sync and for other debugging needs by setting `Infomon.set('infomon.last_sync', Time.now.to_i)`

### DIFF
--- a/lib/infomon/cli.rb
+++ b/lib/infomon/cli.rb
@@ -26,6 +26,7 @@ module Infomon
       respond "Did #{command}." if $infomon_debug
     end
     respond 'Requested Infomon sync complete.'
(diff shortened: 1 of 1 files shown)


mrhoribu:
Tester had their Lich/Wrayth client crash on initial launch which upon logging in a second time, showed an incomplete sync causing a lot of values to not be populated and various scripts acting funky because of it. Adding this value to show that the full Infomon.sync completed that can be checked against in autostart.lic will prevent that edge case as well as be able to give us something to troubleshoot against by asking for the current value of it to know that it was done properly if troubleshooting with someone.

PR #459, opened 2023-07-30
[infomon] Refactor ActiveSpell watcher to use queue

sandersch:
When testing recent updates to lich, I noticed that my CPU usages was significantly higher (ie. redlining multiple CPUs on a modern machine).

After some investigation, it looks like the ActiveSpell watcher spawns a thread that repeatedly checks a new boolean value in the XMLData class.

I refactored this to try a couple of other approaches:

1) Invoke the ActiveSpell method directly instead of setting a boolean. This has greatly reduced CPU usage to less than 1% almost all the time. I understand the motivation for moving this to its own thread though, so I'm testing this especially with lots of scroll and spell updates to make sure it doesn't block processing in the main lich XML parsing routine. 

2) Continue to use the background thread but to switch to a [Queue](https://ruby-doc.org/core-2.5.0/Queue.html) instead of a busy loop to determine when there is work to pick up

To make testing easier, I also created a feature flag to swap between these behaviors without restarting Lich: `ActiveSpell.update_in_background = true / false`

Ran into a minor hiccup repeatedly spawning the watcher thread, but was able to solve this by memoizing the thread creation so that it only happens one time no matter how many times the code is loaded or the `watch!` method is called.

### Decision

After testing, the background thread version using a queue works well and won't block the main lich XML parsing thread. Removed the inline version and feature flag.

### DIFF
--- a/lib/game-loader.rb
+++ b/lib/game-loader.rb
@@ -13,7 +13,6 @@ module GameLoader
     require 'lib/infomon/status'
     require 'lib/experience'
     require 'lib/infomon/activespell'
-    ActiveSpell.watch!
     # PSMS (armor, cman, feat, shield, weapon) have moved
     # to ./lib/psms and are now called by psms.rb
     require 'lib/psms'
--- a/lib/game-loader.rb
+++ b/lib/game-loader.rb
@@ -13,6 +13,7 @@ module GameLoader
     require 'lib/infomon/status'
     require 'lib/experience'
     require 'lib/infomon/activespell'
+    ActiveSpell.watch!
     # PSMS (armor, cman, feat, shield, weapon) have moved
     # to ./lib/psms and are now called by psms.rb
     require 'lib/psms'
--- a/lib/infomon/activespell.rb
+++ b/lib/infomon/activespell.rb
@@ -117,10 +117,10 @@ module ActiveSpell
   def self.update_in_background?
     @update_in_background
   end
- 
+
   def self.request_update
     if update_in_background?
-      enqueue_update 
+      enqueue_update
     else
       update_spell_durations
     end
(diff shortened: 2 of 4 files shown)


mrhoribu:
To prevent it blocking processing on massive spell scroll, would it be a good idea to do something similar to how we broke out Infomon's main db writing into a worker thread that processes a queue? Perhaps doing something similar here where we push a value to a queue that ActiveSpell worker thread sees and then does it's evaluation. That way the main XML parser can continue without it being blocked out by ActiveSpell doing it's processing? Granted, depending on how long ActiveSpell takes to process, not sure if it would even amount to any visible difference/improvement in-game.

sandersch:
> Granted, depending on how long ActiveSpell takes to process, not sure if it would even amount to any visible difference/improvement in-game.

This is what I'd like to test / get feedback on. My intuition is that updating the spell timers is a relatively low compute task that happens entirely in memory so it shouldn't significantly slow down the main loop, but would like to verify that.

> would it be a good idea to do something similar to how we broke out Infomon's main db writing into a worker thread that processes a queue?

I think this is worth investigating and comparing to the inline update version. I was already considering using [Queue#pop](https://ruby-doc.org/core-2.5.0/Queue.html) so I have a pretty good idea of how to do this. Will try to put together a version with that approach tonight.

sandersch:
Pretty happy with the queue+thread approach here. I have run it through a couple Reim nights now without issue which is a pretty meaningful stress test. Only concern I have at this point is that I am on Linux and haven't done any testing on other platforms. If anyone is able to kick the times on this on Windows or other platforms, that would make me feel confident about merging this. Being a little more careful here than normal because the problematic behavior wasn't noticeable on Windows previously, so I don't want to regress the primary player platform experience.

mrhoribu:
Refactor looks good to me. Did a few rounds of bandits with 3 clients running Wrayth and no cpu spikes.

sandersch:
Ok gonna rip out the inline version and feature flag before one more test in Reim tonight and will merge if everything looks good.

PR #468, opened 2023-08-08
[parser.rb] Infomon::Parser regex update for redsteel marks

mrhoribu:
Fix for redsteel marks as no longer visible via `WEALTH` but `BEAST STATUS` instead.

```
Spirit Beast Information:

    Redsteel Marks:            3441
```


OSXLich-Doug:
Confirmed - merging

PR #469, opened 2023-08-10
[parser.rb] update Bounty::Parser guard regex

mrhoribu:
Add Hinterwilds guard completion regex
```
You have located a pair of bronze bangle earrings and should bring it back to the tavernkeeper at Rawknuckle's Common House.
```


OSXLich-Doug:
Confirmed - merging

PR #474, opened 2023-08-20
[parser.rb] logic fix Infomon::Parser skills current/base

mrhoribu:
Update parser to NOT parse SKILLS BASE output, it should be set to only grab SKILLS CURRENT so that it carries enhancive/ascension values.

### DIFF
--- a/lib/infomon/parser.rb
+++ b/lib/infomon/parser.rb
@@ -15,7 +15,7 @@ module Infomon
       TotalExp = /^\s+Total Exp: (?<total_experience>[\d,]+)\s+Death's Sting: \w+$/.freeze
       LTE = /^\s+Long-Term Exp: (?<long_term_experience>[\d,]+)\s+Deeds: (?<deeds>\d+)$/.freeze
       ExprEnd = /^\s+Exp (?:until lvl|to next TP): [\d,]+/.freeze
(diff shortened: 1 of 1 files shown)


OSXLich-Doug (approved):

mrhoribu:
Found issue with GOALS being set and then coming back it will show a SKILLS BASE. Which will not be parsed and system will think it's still on older values. A few options:

- Revert change and let the skills be reset to base or current depending on which one was last used.
- Capture both base and current and do proper logic to update current if base is more than what current is. This allows for tracking of non-asc/enhancive value SKILLS
- match against `/^Skill goals updated!$/` and send a SKILLS command to the client to grab current skills whenever a GOALS is completed.

lib/infomon/parser.rb
```diff
@@ -165,6 +175,21 @@ def self.parse(line)
...
           end
+        when Pattern::GoalsDetected
+          @goals_detected = true
+          :ok
+        when Pattern::GoalsEnded
+          if @goals_detected
+            @goals_detected = false
+            respond; respond monsterbold_start
+            respond "You just trained your character.  Lich will gather your updated skills."
+            respond monsterbold_end; respond
```
mrhoribu, line 186: This can be cleaned up a bit using native monsterbold messaging library:
```ruby
respond
respond Lich::Messaging.monsterbold('You just trained your character.  Lich will gather your updated skills.')
respond
```

PR #478, opened 2023-09-04
[lich.rbw] Gating some detached-mode init lines

MahtraDR:
Just gating some detached-mode initialization entries to GS.


mrhoribu (approved):
lgtm

PR #480, opened 2023-09-05
[stash.rb] fix emptyhands bug

OSXLich-Doug:
adding to regex to prevent false positive results detection


mrhoribu (approved):
Should work fine. May need eventual future updates to capture additional responses but should be good 99% of the time.

PR #481, opened 2023-09-06
[messaging.rb] mono API fix for \n AVALON bug

mrhoribu:
Fix Lich::Messaging.mono API to correct for Avalon's \n handling being bugged.


OSXLich-Doug:
Looks good.  Merging for 5.7.0

PR #485, opened 2023-09-09
[activespell.rb] ResistNature/620 Fix

mrhoribu:
Fix duration for Resist Nature 620 due to percentages in the name. Strip the percentage so that Lich can track properly

### DIFF
--- a/lib/infomon/activespell.rb
+++ b/lib/infomon/activespell.rb
@@ -42,6 +42,10 @@ module ActiveSpell
         makeychange << k
         update_spell_names.push('Next Bounty')
         next
+      when /(?:Resist Nature|620) \(\d{1,2}%\)/
+        makeychange << k
+        update_spell_names.push('Resist Nature')
+        next
       end
       update_spell_names << k
     end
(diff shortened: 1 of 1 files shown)


emcguinness:
As a Ranger, among others, fix works for me (with and without spell names).

lib/infomon/activespell.rb
```diff
@@ -42,6 +42,10 @@ def self.get_spell_info(spell_check = XMLData.active_spells)
         makeychange << k
         update_spell_names.push('Next Bounty')
         next
+      when /(?:Resist Nature|620) (?:- (?:Heat|Cold) \(\d\d%|- Steam \(\d\d|- Lightning|\(\d\d%\))/
```
FarFigNewGut, line 45: Could the regex be simplified a bit to:
`(?:Resist Nature|620)(?: - (?:Heat|Cold|Steam|Lightning))? \(\d{2}%\)`

https://rubular.com/r/uuwlYBJsybDyF6

It was kinda weirding me out heat/cold got lumped together but then the rest had to be separate. 😨

mrhoribu: It's because of truncing in the XML due to character limits by Simu as to why they're separated out like that.

PR #486, opened 2023-09-10
[cli.rb] extra regex match for 1212 stopping

mrhoribu:


FarFigNewGut:
LGTM - i know doug's looking now too or i'd hit the button :D

PR #488, opened 2023-09-11
[stash.rb] extra shield regex

mrhoribu:
Add extra shield regex for WEAR
```
You toss the shield casually up in the air, turning with apparent nonchalance to let it land precisely on the center of your back.
```
Noun is always generic shield for this type of scripted shield.


FarFigNewGut:
Should probably still change this to just check if the item is now in INV or whatever instead of the regex of doom but looks good for the existing setup. P.S. hi doug

PR #498, opened 2023-10-08
bugfix: Spellsong with Infomon

mrhoribu:
During rewrite of infomon, missed forcing Spellsong.renewed to be done when a song is sang or renewed. Also cleaned up spec-test to not throw as many internal errors on the spec run.


OSXLich-Doug (approved):
Looks solid, tests well

PR #500, opened 2023-10-18
chore: add individual PSM skill methods for short/long names

mrhoribu:
Allows for checking current skill rank by doing: `CMan.berserk` vs the current `CMan['berserk']` that currently existed.


ondreian (approved):
lgtm

PR #501, opened 2023-10-27
chore:isolate spellsong and fix duration

OSXLich-Doug:
Need to add a line to `autostart` as well, to fully capitalize on the bugfix in this chore (`Spellsong.sync`)


mrhoribu:
Not sure if this should live in `lib\attributes` instead of just `lib`. But overall looks good, need to test it a little. Also need to see if it causes any issue being loaded behind GS4 gametype. Theoritically it shouldn't, but I know there's some odd things with certain stuff getting loaded on the DR side that shouldn't be even when not used. So hoping Spellsong isn't referenced somewhere weird and moving it behind GS4 only causing issues. But those should likely be corrected anyways if so.

PR #508, opened 2023-11-19
[lich.rbw] fix: don't use WINE to launch custom frontends

sproctor:


sproctor:
This fixes #509

sproctor:
There's a deeper issue here. There's a bit of an overgeneralization between parsing the .sal files and creating the data in the GUI to launch the client. There's not intermediate representation, so there's no differentiation between parsing a .sal file and launching a custom command.

The problem: on Linux/Mac there should be a way to specify whether or not to use WINE. If the entry was saved, there's no way to know if it's a Windows binary or native.

The solution: There could be a WINE checkbox on non-Windows platforms, when unchecked would add a "WINE=disabled" argument or something like that. Or add a platform parameter to the frontend and on linux/mac add the WINE command as a prefix to Windows frontends.

sproctor:
The amount of nesting in the launch process makes it pretty hard to follow the logic. `custom_launcher` is set in almost every circumstance. It seems the only time it isn't is maybe running Avalon? Adding the wine binary where `custom_launcher` is set for the Windows executables fixes my issue and preserves the functionality when trying to run wrayth/wizard.

OSXLich-Doug:
> The amount of nesting in the launch process makes it pretty hard to follow the logic. `custom_launcher` is set in almost every circumstance. It seems the only time it isn't is maybe running Avalon? Adding the wine binary where `custom_launcher` is set for the Windows executables fixes my issue and preserves the functionality when trying to run wrayth/wizard.

This is an offshoot of retiring the old Windows methods where there were six different ways to launch Lich with an FE.  I'll see about streamlining it a bit more.  

Mac doesn't use WINE with the advent of the Mac-only-runs-64-bit-code and the FEs not being 64 bit code.  So for our purposes, WINE is a Linux only problem presently.

PR #511, opened 2023-11-20
[claim.rb] feat: add new claim module and associated changes to xmlparser

FarFigNewGut:
the `Group` and `Clustered` callouts are kinda holdovers from O's original script not sure if that should just get taken out. but believe this is working. I'm still not running 5.7 as a daily driver but was fine in some impromptu testing and was a port of code i've been running in slightly older lich for months without issue. but there were some changes in xmlparser so might need extra verification.

This would expose `Claim.mine?` that could be use in bigshot et. all checks as well as letting other scripts block on `Claim.last_room == <target room>` to minimize race condition issues parsing room objects, etc.

### DIFF
--- a/lib/claim.rb
+++ b/lib/claim.rb
@@ -84,7 +84,11 @@ module Claim
       @mine = true
       self.claim_room nav_rm unless nav_rm.nil?
     rescue StandardError => e
-      Log.out(e)
+      if defined?(Log)
+        Log.out(e)
+      else
+        respond("Claim Parser Error: #{e}")
+      end
     ensure
       Lock.unlock if Lock.owned?
     end
--- a/lib/claim.rb
+++ b/lib/claim.rb
@@ -49,7 +49,7 @@ module Claim
     info_table = Terminal::Table.new :headings => ['Property', 'Value', 'Description'],
                                      :rows => rows,
                                      :style => {:all_separators => true}
-    _respond "<output class=\"mono\"/>\n" + info_table.to_s + "\n<output class=\"\"/>")
+    Lich::Messaging.mono(info_table.to_s)
   end
 
   def self.mine?
--- a/lib/claim.rb
+++ b/lib/claim.rb
@@ -11,7 +11,7 @@ module Claim
   def self.claim_room(id)
     @claimed_room = id.to_i
     @timestamp    = Time.now
-    Log.out("claimed #{@claimed_room}", label: %i(claim room)) if defined? Log
+    Log.out("claimed #{@claimed_room}", label: %i(claim room)) if defined?(Log)
     Lock.unlock
   end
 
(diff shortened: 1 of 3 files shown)


lib/claim.rb
```diff
@@ -0,0 +1,93 @@
...
+      @others = pcs - self.clustered - self.members
+      @last_room = nav_rm
+      unless @others.empty?
+        @mine = false
+        return
+      end
+      @mine = true
+      self.claim_room nav_rm unless nav_rm.nil?
+    rescue StandardError => e
+      Log.out(e)
```
mrhoribu, line 87: This needs a `if defined? Log` added to the end of it otherwise will cause issues. Could do something like:
```ruby
if defined? Log
  Log.out(e)
else
  respond("Claim Parser Error: #{e}")
end

**Review by mrhoribu — CHANGES_REQUESTED** [review]
Looks good. Just missed one reference to `Log`. I feel it's likely safe to leave in the Group and Cluster for now as we try to target migrating more of Olib over to core Lich5. Group being one of the ones I want to get added asap.

File: lib/claim.rb [hunk 2]
```diff
@@ -0,0 +1,97 @@
...
+  end
+
+  def self.info
+    info = {'Current Room' => XMLData.room_id,
+            'Mine' => Claim.mine?,
+            'Claimed Room' => Claim.claimed_room,
+            'Checked' => Claim.checked?,
+            'Last Room' => Claim.last_room,
+            'Others' => Claim.others}
+    respond JSON.pretty_generate(info)
```
ondreian, line 49: we can/should terminal table this output

lib/xmlparser.rb
```diff
@@ -123,6 +125,11 @@ def initialize
     # real id updates
     @room_id = nil
     @previous_nav_rm = nil
+
+    # claim update
+    @arrival_pcs = []
```
ondreian, line 130: This seems like something we should make an attr_reader so it's able to be read from outside this module probably?

FarFigNewGut: attr_accessored arrival_pcs and the obvious_hidden_player values

FarFigNewGut:
- added the arrival_pcs and hidden_player values as attr_accessors in xmlparser
- tableized the Claim.info call (added a description field, might be too much now)
- also added exception handling for the Grouped and Clustered calls. I mean Log is still in there should probably chage that to Lich.log i guess but i figured Grouped especially might conflict with people's personal stuff so until we have an actual module to  CLAIM (GET IT?!?!) that namespace formally i figured i'd try to account for that.

lib/claim.rb
mrhoribu: This is used for primarily debugging/info purposes, so I'm ok with adding description column.

mrhoribu:
The only other consideration I would have is should we name scope this to something other than just `Claim`? Should it be something like `Lich::Claim` incase of namespace conflicts with any other code that may use a module with similar name?

mrhoribu (approved):
lgtm. Eventually we'll look at adding callbacks, but for now, this works.

PR #514, opened 2023-11-28
[parser.rb] negative TNL exp value

mrhoribu:
bugfix for negative Experience before leveling up


OSXLich-Doug:
Looks good, tests clean.

PR #516, opened 2023-12-04
[hooks] adds .sources for up/downstream hooks

FarFigNewGut:
```
;e UpstreamHook.sources
+---------------+------------+
| Hook          | Source     |
+---------------+------------+
| lnet          | lnet       |
+---------------+------------+
| alias-service | alias      |
+---------------+------------+
| group_ajar    | group_ajar |
+---------------+------------+
| infomon       | infomon    |
+---------------+------------+
```

Adds tracking what script has implemented an up/down stream hook into `UpstreamHook.sources` and `DownstreamHook.sources` that outputs a terminal table output. Useful for tracking down an errant hook from a random source potentially. Done in such a way as to not change any existing behavior the calls (in list, remove, etc.)

### DIFF
--- a/lich.rbw
+++ b/lich.rbw
@@ -266,6 +266,10 @@ class UpstreamHook
     Lich::Messaging.mono(info_table.to_s)
   end
 
+  def UpstreamHook.hook_sources
+    @@upstream_hook_sources
+  end
+
 end
 
 class DownstreamHook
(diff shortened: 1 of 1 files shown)


mrhoribu:
I like it, perhaps an option though to add `attr_accessor upstream_hook_sources` and `attr_accessor downstream_hook_sources` incase someone wants to utilize the sources for something other than just to see them printed in the FE?

mrhoribu (approved):
lgtm

PR #517, opened 2023-12-19
[psms.rb] stop clobbering with PSMS.name_normal, account for apostrophes

FarFigNewGut:
- apostrophes were not accounted for in `PSMS.name_normal` (Predator's Eye)
- stop clobbering the passed argument in name_normal by operating on a different object. the `gsubs!` were changing the passed name all the way up the stack causing unintended side effects

### DIFF
--- a/lib/psms.rb
+++ b/lib/psms.rb
@@ -12,10 +12,13 @@ module PSMS
   def self.name_normal(name)
     # there are five cases to normalize
     # "vault_kick", "vault kick", "vault-kick", :vault_kick, :vaultkick
+    # "predator's eye"
     # if present, convert spaces to underscore; convert all to downcase string
-    name.gsub!(' ', '_') if name =~ (/\s/)
-    name.gsub!('-', '_') if name =~ (/-/)
-    name.to_s.downcase
+    normal_name = name.to_s.downcase
+    normal_name.gsub!(' ', '_') if name =~ (/\s/)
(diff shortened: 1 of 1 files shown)


mrhoribu (commented):
Would we want to modify the `Lich::Util.normalize_lookup` to also handle an apostrophe as well?
```ruby
    def self.normalize_lookup(effect, val)
      caller_type = "Effects::#{effect}"
      case val
      when String
        (eval caller_type).to_h.transform_keys(&:to_s).transform_keys(&:downcase).include?(val.downcase.gsub('_', ' '))
      when Integer
        #      seek = mappings.fetch(val, nil)
        (eval caller_type).active?(val)
      when Symbol
        (eval caller_type).to_h.transform_keys(&:to_s).transform_keys(&:downcase).include?(val.to_s.downcase.gsub('_', ' '))
      else
        fail "invalid lookup case #{val.class.name}"
      end
    end
```
Adding a `.gsub("'", '')` to the end of the string/symbol eval?

FarFigNewGut:
pretty sure we only needed the _ gsub in there in there in the first place because the PSMS change was transforming input :D

Normalize lookup is doing a direct check on your effects, which has the apostrophe, so you wouldn't want to delete it. either way you WANT the apostrophe in the effect lookup and there's no way to know really where to put it so.

mrhoribu (approved):
lgtm

PR #520, opened 2023-12-29
[char.rb]-Adding Char.stance

OSXLich-Doug:
Putting `Char.stance` back into play.  May need to consider if any of the stance percent options need to return.

### DIFF
--- a/lib/attributes/char.rb
+++ b/lib/attributes/char.rb
@@ -9,8 +9,8 @@ class Char
     XMLData.name
   end
 
-  def Char.stance
-    XMLData.stance_text
+  def Char.stance(*args)
+    checkstance(*args)
   end
 
   def Char.health(*args)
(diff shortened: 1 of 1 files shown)


lib/attributes/char.rb
```diff
@@ -9,6 +9,10 @@ def Char.name
     XMLData.name
   end
 
+  def Char.stance
+    XMLData.stance_text
+  end
```
mrhoribu, line 14: This should likely be:
```
  def Char.stance(*args)
    checkstance(*args)
  end
```

mrhoribu: I couldn't find the original overload for `Char.stance` in 5.6.2 to compare to, but this would make a bit more sense to support the duality of using either `Char.stance` or `checkstance`

mrhoribu (approved):
lgtm

PR #523, opened 2024-01-06
[char.rb] migrate global functions to Char module

mrhoribu:
Migrate global namespace lookups from `global_defs.rb` that relate to Char module and change global methods to reference them.

Created a new `Lich.depreciated` method to log usage of depreciated methods and to be able to point out references to functions that should no longer be used.

Used new depreciated method for global methods that are now usable from Char module. May need to comment out if too spammy as not sure just how much the global methods are used within core Lich functionality and don't want to overly spam the debug file.

### DIFF
--- a/lib/global_defs.rb
+++ b/lib/global_defs.rb
@@ -919,32 +919,8 @@ def percentconcentration(num = nil)
   end
 end
 
-def checkstance(num = nil)
-  if num.nil?
-    XMLData.stance_text
-  elsif (num.class == String) and (num.to_i == 0)
-    if num =~ /off/i
-      XMLData.stance_value == 0
-    elsif num =~ /adv/i
-      XMLData.stance_value.between?(01, 20)
-    elsif num =~ /for/i
-      XMLData.stance_value.between?(21, 40)
-    elsif num =~ /neu/i
-      XMLData.stance_value.between?(41, 60)
-    elsif num =~ /gua/i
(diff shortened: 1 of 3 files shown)


mrhoribu:
Also not sure if these functions are utilized by DragonRealms at all. Will need to check that prior to migrating them to calls placed behind GS4 only libraries.

mrhoribu:
From what I can see, `checkhealth` is used by DragonRealms. The others are not. So likely would need some sort of coding done to detect game. Downside is `char.rb` currently isn't loaded unless GS4.

lib/global_defs.rb
```diff
@@ -812,94 +812,64 @@ def checksaturated
...
 end
 
-def checkmana(num = nil)
-  if num.nil?
-    XMLData.mana
-  else
-    XMLData.mana >= num.to_i
-  end
+def checkmana(*args)
+  Lich.depreciated('checkmana', 'Char.mana')
```
ondreian, line 816: `deprecated` not `depreciated`

lib/lich.rb
```diff
@@ -58,6 +58,15 @@ def Lich.log(msg)
     $stderr.puts "#{Time.now.strftime("%Y-%m-%d %H:%M:%S")}: #{msg}"
   end
 
+  def Lich.depreciated(old_object = '', new_object = '')
```
ondreian, line 61: This should be `deprecated` not `depreciated`

PR #524, opened 2024-01-08
[skills.rb] removal of combatleadership

mrhoribu:
Remove erroneous skill `combatleadership` from skills list as hasn't been implemented in a very long time.


OSXLich-Doug:
LGTM - merging.

PR #532, opened 2024-01-15
feat: Group & Disk class

mrhoribu:
Add the `Group` and `Disk` class for GS4.


lib/character/disk.rb
```diff
@@ -0,0 +1,48 @@
+module Games
+  module Gemstone
+    class Disk
+      NOUNS = %w{disk coffin}
+
+      def self.is_disk?(thing)
+        NOUNS.include?(thing.noun)
```
FarFigNewGut, line 7: does this need to match on something more like `[A-Z][a-z]+ (?:disk|coffin)` to avoid some false positives with random other things? I guess bigshot already doesn't do this it's probably fine they aren't exactly common. Also lol Waldo2, etc. Poachin them like whoa.

PR #533, opened 2024-01-22
Update lich.rbw for drb

OSXLich-Doug:
Edit require drb line to proper current init


abirmingham (approved):
Tested locally with no apparent ill effect.

mrhoribu:
In my searching of core Lich5 and it's scripts, I haven't found any usage of drb at all. Is it even needed to have Lich5 require it to begin with? Why not just let scripts that use it require it instead of having Lich5 load it and never use it's code unless you happen to run one of the scripts that uses it, which those scripts all also have a require statement in them to ensure loading of the gem.

mrhoribu:
There isn't really any reversal needed. REQUIRE only loads a file the first time it's called for a ruby process. The next attempt will just not load it as it's already been loaded once. So the benefit of not having it required in Lich5 core if it's not being used is less time (milliseconds though, so likely unnoticeable) and letting the scripts themselves require the file. And if we ever start using the gem/files code in Lich5 and require it there, it would have no impact on the scripts that already try to require the same file/gem. Granted, it does require everyone trying to reference the gem/file the same way. IE `drb\drb` != `drb`

OSXLich-Doug:
Leaving it here means we don't need to ask script owners to update every time something in the interface changes.  I'm also curious, in a kind of offhand way, how many gems are loaded by scripts that aren't loaded by core.  I'm going to guess that prior to 2018, there are a few, but those scripts aren't well maintained anymore.

The time-savings goes both ways.  Milliseconds in core, means milliseconds saved in scripts.  And we have it when we gemify Lich.  No significant reason to move it.  I am curious about gems used by others though - never know what we're missing in terms of cool functionality!

PR #535, opened 2024-02-26
[all][lich.rbw] debug logfile Game:Character

mrhoribu:
Add a Lich.log entry for logged in Game:Char to the debug file to assist in determining what character/game the debug file belongs to


abirmingham:
Verified functioning as expected - prints character name in debug log near the beginning of the log.

PR #536, opened 2024-02-26
[all][init.rb] expand debug log files include milliseconds

mrhoribu:
Add `%L` to debug filenames due to rapid lich launcher launch within a single second causing issues with debug files not being created.

### DIFF
--- a/lib/init.rb
+++ b/lib/init.rb
@@ -680,7 +680,7 @@ Lich.init_db
 # only keep the last 20 debug files
 #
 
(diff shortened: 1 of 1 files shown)


abirmingham:
Looks good to me, though `%r[^debug(?:-\d+)+\.log$]` might be slightly easier to read.

abirmingham:
Verified log filenames include milliseconds and that delete still occurs (did not test the case of deleting logs without microseconds).

OSXLich-Doug:
Good update

PR #553, opened 2024-04-23
[all] chore: Sqlite3 gem syntax code cleanup

mrhoribu:
With the release of sqlite3 gem v2+ it depreciated doing db.execute of passing parameters not in an array. This cleans up all sqlite commands I could find to pass parameters in an array instead of as a comma delimited parameters. See https://github.com/sparklemotion/sqlite3-ruby/blob/main/CHANGELOG.md for additional details. 

> Removed support for non-Array bind parameters to methods Database#execute, #execute_batch, and #query, which has been deprecated since v1.3.0. 
> Deprecated code looks like this:
```
@db.query("select * from foo where a = ? and b = ? and c = ?", 1, 2, 3)
```
> For these cases, pass the bind parameters as an array:
```
@db.query("select * from foo where a = ? and b = ? and c = ?", [1, 2, 3])
```

### DIFF
--- a/lich.rbw
+++ b/lich.rbw
@@ -346,7 +346,7 @@ module Setting
     values = Array.new
     for setting in args
       begin
-        v = Lich.db.get_first_value('SELECT value FROM script_setting WHERE script=? AND name=?;', script.name.encode('UTF-8'), setting.encode('UTF-8'))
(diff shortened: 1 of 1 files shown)


MahtraDR:
What is the minimal version well support after this? Is it ruby or the gem?

PR #571, opened 2024-05-06
[all][lich.rbw] feat: recommended Ruby version support

mrhoribu:
Add support for showing a warning to end-users for a recommended version of Ruby to help push people into updating their version of Ruby before it becomes to late.


MahtraDR:
Works for me. This is the one we have in dependency.lic https://github.com/rpherbig/dr-scripts/blob/28d7920409eae106e08071fd075c743568246b5f/dependency.lic#L1817 I'd remove the one in dependency if we merge this here, and keep it all in one place.

PR #579, opened 2024-05-16
[all][update.rb] feat: Update core scripts from source

mrhoribu:
To be used for first Lich5 run as well as when doing in-game updates. Uses new lich.db3 variable & method to remember last run based on LICH_VERSION stored in the variable. Only issue right now is `gameobj-data.xml` is not stored on github and only way we currently distribute it is via `;repository`, Lich5 releases, and from GitHub directly.

### DIFF
--- a/lib/update.rb
+++ b/lib/update.rb
@@ -9,8 +9,8 @@ module Lich
       require 'zlib'
 
       @current = LICH_VERSION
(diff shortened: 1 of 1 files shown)


OSXLich-Doug:
👍

MahtraDR:
We only need dependency. I really need to bring dependency into lich proper.

PR #581, opened 2024-05-17
[all][global_defs.rb] update monsterbold methods

mrhoribu:
Remove standalone `<b>`/`</b>` for profanity as Profanity supports pushBold/popBold without issue.

### DIFF
--- a/lib/global_defs.rb
+++ b/lib/global_defs.rb
@@ -2020,10 +2020,8 @@ end
 def monsterbold_start
   if $frontend =~ /^(?:wizard|avalon)$/
     "\034GSL\r\n"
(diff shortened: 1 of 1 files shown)


MahtraDR:
Tested and works well.

vtcifer:
should maybe include genie in there as well?

MahtraDR:
> should maybe include genie in there as well?

Can you have a go testing that? If it works as is, just adding `genie` to that line, that's easy enough to add.

vtcifer:
Yeah, it's been in my local copy for a while, I was noticing this when going through and finally comitting and doing PRs against the repo here.

PR #591, opened 2024-05-31
feat: proposal for sequel based Settings;

ondreian:


OSXLich-Doug:
@ondreian - Regret that I stomped on this on the breakout branch.  Just FYI - I pulled your `method SettingsNew` and pasted over the `method Settings` that is currently live.  Also had to bring in the deprecated calls, particularly `Settings.load` because yes - seems every script calls it.

But this works.  It just works.  Of course.  I'll keep playtesting it, but I think this is solid at this point.

OSXLich-Doug:
@ondreian - Ready for your review.

OSXLich-Doug:
One of the things I like about this, at its core, is that it does away
with the bulk 'save' requests.  There isn't an equivalent 'bulk upcert'
function here like there is in infomon.  If a setting is changed, it's
saved, right then.  If it's not changed, it's not saved.  No unnecessary
saving entire settings strings every 5 minutes.

I am sure we can tighten it up, but I'd like to see some numbers before
going too hog wild.  We aren't chasing the same problem, at this point.

PR #592, opened 2024-06-05
[dr][global_defs.rb] add missing climb failure message

vtcifer:


lib/global_defs.rb
OSXLich-Doug, line 1: I think our current standard is to make the regex as discreet as possible.  This already exists in the regex

```^You make your way up the .*\.\s*Partway up, you make the mistake of looking down\. Struck by vertigo, you cling to the .* for a few moments, then slowly climb back down\.$```

@vtcifer - do you see any problem with just modifying that existing and complete regex with the required `(?:up|down)` component?

vtcifer: Nah, seems like a better idea.  

I think I just missed it in my scan of what conditions we already trapped.

PR #621, opened 2024-06-26
[gs][stash.rb] bugfix find_container case sensitivity

mrhoribu:
Remove case sensitivity on param match on `find_container` method


OSXLich-Doug:
RTM

PR #622, opened 2024-06-27
[all][lich.rbw] Extend class Integer for sleep times

OSXLich-Doug:
I would like to put my calculators and spreadsheets down.

Usage

`sleep 20.minutes`
`sleep 2.hours`
`sleep 5.days`  <- Hey, in Shattered you just never know

### DIFF
--- a/lib/class_exts/integer.rb
+++ b/lib/class_exts/integer.rb
@@ -1,0 +1,20 @@
+# Add Integer extension to Lich
+# extension to class Integer 2024-06-27
+
+class Integer
+  def seconds
+    return self
+  end
+
+  def minutes
+    return self * 60
+  end
+
+  def hours
+    return self * 3600
+  end
+
+  def days
+    return self * 86400
+  end
+end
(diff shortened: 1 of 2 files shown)


Kiriawen:
This could be useful in DR too.  With how our F2P goes there is a lot of downtime or times people want to be inactive for extended periods.

OSXLich-Doug:
> While it does seem nice, at this point, why recreate the wheel and instead just push to include ActiveSupport gem so you could then just do:

Um, because Rails adds a lot of needless cruft we won't value?  And as you point out - no gem.

OSXLich-Doug:
> Definitely wouldn't want to include all of it, but there are some nice QOL things that Rails does do, hence the suggestion. Wouldn't hurt to include it in the gems for future readiness, doesn't need to be required right now. But would give us the option for the future if we wanted to utilize some of the additional extensions it provides.

While I was aware Rails did this, I can't speak from a perspective of Rails Expert so I don't know all the QOL things it may have.  I'm not opposed to future discussions about ease of implementation as needs are identified and brought in.  

But for the proposed 15 lines of code, I am on the YAGNI side of Rails presently - hence this submit.

OSXLich-Doug:
I'm not sure about the `as_time` discussion, it seems that's working as it always has.  I do like the suggestion to avoid creating a new file.  I'm closing this PR and opening one with `numeric.rb` modified, instead.

PR #625, opened 2024-07-01
[gs][critranks.rb] Add critrank resolution and tables

OSXLich-Doug:
Several contributors - if you were one, please let me know so I can attribute!

This PR includes crit tables from the wiki (largely from Mark's research) cross referenced with Gib's old crit script.  It includes all existing tables, plus one 'generic' crit table for creatures that often do not display a crit message (looking at siphons in the Rift, and many others).

The basic query mechanism in `critranks.rb` is 

`ruby
CritRanks.match(string)
`
A common mechanism, if one simply wanted to see all the crits, might include:

`ruby
    crit_response = CritRanks.match(combat_line)
    unless crit_response.empty?
      crit_response.each_value do |v|
        @crit_results = v
      end
    end
`
Happy slaying!


ondreian:
The uppercase words like `HEAD` and `PLASMA` should not be uppercase if we are going to move this into core lich

snperry85:
Would it be possible to make a minor change to the regex that will make it usable with xml as well. 
Just replacing `the .*?` with `.*?` because the xml has `the` after the `<pushBold/>`. I don't foresee any issues with this change but maybe someone else will.

`                                   :regex           => /Deep slash to the .*?'s neck severs an artery./ },`
`Deep slash to <pushBold/>the <a exist="1769462" noun="warlock">triton warlock</a><popBold/>'s neck severs an artery!`

OSXLich-Doug:
> The uppercase words like `HEAD` and `PLASMA` should not be uppercase if we are going to move this into core lich. I also think a significant amount of the strings should probably be symbols like `:left_eye` etc
> 
> There is the possibility we should even add a dictionary like:
> 
> ```ruby
> module Body
>   module Locations
>     LeftEye = :left_eye
>     RightEye = :right_eye
>     ... and all the other normal areas so that we cannot have a typo accidentally
> ```

Done with two exceptions - first, I opted to leave the string for position in all caps.  And second, I have yet to do the help dictionary - that belongs in critranks.rb.  I'll work that in today or tomorrow.

OSXLich-Doug:
> Would it be possible to make a minor change to the regex that will make it usable with xml as well. Just replacing `the .*?` with `.*?` because the xml has `the` after the `<pushBold/>`. I don't foresee any issues with this change but maybe someone else will.
> 
> ` :regex => /Deep slash to the .*?'s neck severs an artery./ },` `Deep slash to <pushBold/>the <a exist="1769462" noun="warlock">triton warlock</a><popBold/>'s neck severs an artery!`

This is done.

I'm interested in ensuring the fuctionality still works as intended.  In particular watch for the `'s` to properly register in XML, and I'll watch in standard text returns.

lib/crit_tables/slash_critical_table.rb
```diff
@@ -0,0 +1,2505 @@
...
+                                  :crippled        => false,
+                                  :sleeping        => false,
+                                  :dazed           => false,
+                                  :limb_favored    => false,
+                                  :roundtime       => 0,
+                                  :silenced        => false,
+                                  :slowed          => false,
+                                  :wound_rank      => 1,
+                                  :secondary_wound => nil,
+                                  :regex           => /Slash (?:to the|along) .*?'s lower back./ },
```
snperry85, line 1048: this one is having trouble with the xml since `the` was in the group.

`/Slash (?:to|along) .*?'s lower back./` works with both the xml and nonxml

OSXLich-Doug:
Ah, yep.  Might be a few more of that type as well.  I'll scan for more and
update here in a couple hours.

PR #633, opened 2024-08-01
[gs][DB_Store.rb] Update library to support vars

OSXLich-Doug:
Update to read / save uservars to module.  Several uses, copying profiles between characters / game instances and supporting renaming characters.


OSXLich-Doug:
> The only thing I would think to do is to incorporate the new mutex for Lich.db that we implemented. That way it doesn't cause conflict in the rare race condition that a save/load happens the same time as a normal Settings/Vars save and load happens.

Done for writes, not needed for reads.  Would like to recommend we convene a group to power through our desired DB, config, etc.  We have plenty of unresolved points that we need to no longer cherry pick, but solve for.  I'll put something up in the discord.

PR #638, opened 2024-08-09
[gs4][infomon::parser.rb] Update for 640 Wall of Thorns

OSXLich-Doug:
Update Infomon's parser to detect / report 640 poisoning effects

### DIFF
--- a/lib/infomon/parser.rb
+++ b/lib/infomon/parser.rb
@@ -67,6 +67,10 @@ module Infomon
       CalmNoActive = /^You are enraged by .*? attack!|^The feeling of calm leaves you\./.freeze
       CutthroatActive = /slices deep into your vocal cords!$|^All you manage to do is cough up some blood\.$/.freeze
       CutthroatNoActive = /^\s*The horrible pain in your vocal cords subsides as you spit out the last of the blood clogging your throat\.$|^That tingles, but there are no head injuries to repair\.$/.freeze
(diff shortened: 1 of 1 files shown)


OSXLich-Doug:
> This tracks the game messaging, but is there any way to know for 100% certaintity that it's not currently active that we could also watch for, otherwise say you get 640 thorned on one installation, move to a different installation and it wears off, then you go back to the prior installation, it's now still set. Not sure best way to account for that if no way to check/clear it based on ingame health status.

Yes, this worried me too.  The only potential way I could think of was the poison progression / poison deprogression messages - that's the only reason I included those.  Start / End should be enough except we can't see beyond game sessions / new installs with 100% accuracy.  I put them in place, but `:noop` out the capture, just for this discussion.  Interested in thoughts on best way to handle.

ondreian (approved):
lgtm

OSXLich-Doug:
> My concern with this is there is no way to "reset" this status without seeing the clear messaging or doing a full `;infomon reset`. Case in point, using two installs and migrating between them while the effect is active. Propose instead of using infomon db value, just sitting a runtime variable to FALSE and then setting it to TRUE/FALSE based on infomon messaging seen. Less likely to have issues but also could run into the reverse where you log out and back in and then Lich doesn't see that you're infected still. The overall status of this spell effect is just very wonky.

Yep, as I said above, agreed.  After our discussion, we agreed that if we could tie together infomon db and Effects, as we did in other cases in `Status` then this risk is greatly reduced.  I've modified the PR to match our discussion.  I have generated some more information about the spell effect as well.

PR #639, opened 2024-08-09
[all][games.rb]Bugfix - correct for missing room info in some cases

OSXLich-Doug:
This fixes a bug wherein an edgecase could cause a CLI user not to be able to use the `--start-scripts` feature.

### DIFF
--- a/lib/games.rb
+++ b/lib/games.rb
@@ -204,7 +204,7 @@ module Games
                   @@infomon_loaded = true
                 end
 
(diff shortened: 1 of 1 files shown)


OSXLich-Doug:
No, you're right too, Tysong.  I spoke to the symptom, but you spoke to the contributing issue.  I think that's important to retain.

ISSUE:  expected code execution does not occur in edge case where room descriptions are off at startup.

RESULT:  CLI users who declare `--start-scripts` rather than or in addition to using `;autostart' features and who have room descriptions OFF will have the `--start-scripts` request ignored - not executed.

PR #646, opened 2024-08-19
[gs][xmlparser.rb] bugfix for Claim module

mrhoribu:
Fix issue where a `<component>` XML tag for `room players` is not sent from server causing Claim to return false. This should allow for parsing of `Also here: ` line sent from the server and check names against what's already in `XMLData.arrival_pcs` and add to it if they don't exist.


ondreian (approved):
lgtm

PR #649, opened 2024-08-27
[bug][init.rb] Fix WINE entries for Linux / MacOS

OSXLich-Doug:
Version 9 WINE and LINUX have diverged.  Unknown how many use this combination since it is so very slow.  This however addresses the divergence.

### DIFF
--- a/lib/init.rb
+++ b/lib/init.rb
@@ -63,11 +63,22 @@ if (RUBY_PLATFORM =~ /mingw|win/i) && (RUBY_PLATFORM !~ /darwin/i)
     end
   end
 elsif defined?(Wine)
-  ## Needs improvement - iteration and such.  Quick slam test.
-  $sf_fe_loc = Wine.registry_gets('HKEY_LOCAL_MACHINE\\SOFTWARE\\WOW6432Node\\Simutronics\\STORM32\\Directory') || ''
-  $wiz_fe_loc_temp = Wine.registry_gets('HKEY_LOCAL_MACHINE\\Software\\Wow6432Node\\Simutronics\\WIZ32\\Directory')
-  $sf_fe_loc_temp = Wine.registry_gets('HKEY_LOCAL_MACHINE\\Software\\Wow6432Node\\Simutronics\\STORM32\\Directory')
-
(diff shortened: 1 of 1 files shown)


OSXLich-Doug:
> I don't currently have a simple way to test this, and I'm not super well versed in WINE, but wouldn't it be better to do some sort of detection of what version of WINE is installed (9+ or whatever) and adapt accordingly vs what platform?

That depends on a couple of things.  From my perspective, WINE hasn't been useful for quite some time and I support retaining it mostly for those who have no other choice and aren't heavy combat scripters.  The current iterations of WINE are non performant.  The old versions of WINE aren't supported any longer on any modern OS and lack support to be installed.

Net / Net - any WINE user needs to be on the latest to even have a chance to run.  It's going to be a very small user base.  And if push comes to shove, I'm very happy in this instance saying 'just update your WINE'.

MahtraDR:
I also don't have a simple way to test this... I don't use WINE at all. I think we're better off removing WINE support, tbh.  I don't picture linux users wanting to use Wrayth/Wizard, personally.

Personally, I'd prefer we keep things clean, and suggest folks use native clients for their operating systems. Linux users can use Frostbite, or Profanity, or any MUD client? Just my opinion here.

OSXLich-Doug:
> I also don't have a simple way to test this... I don't use WINE at all. I think we're better off removing WINE support, tbh. I don't picture linux users wanting to use Wrayth/Wizard, personally.
> 
> Personally, I'd prefer we keep things clean, and suggest folks use native clients for their operating systems. Linux users can use Frostbite, or Profanity, or any MUD client? Just my opinion here.

Yeah, my bad.  I remembered from last year you were in a discussion on WINE, and didn't check - it was actually @sproctor who drove the PR, and I remembered your activity by association.  No need to review. 

I think I'm going to push it through to prod - I know of at least two who still use it, but I'm very interested to see how many break (we don't know about).

sproctor:
I don't use wine either. I'm the developer of Warlock. It was impossible to run Warlock through an older version of Lich because it pretended the wine command to any custom command on Linux.

PR #668, opened 2024-09-05
[dr] [lib/main/main.rb] Push XML details to genie when it first connects in detached FE mode

vtcifer:
Similar to how it pushes xml info to profanity on first connect.

Still WIP, ironing out some issues with room (weirdness in how room titles come across for DR vs GS) and inventory/stow containers (not currently tracked AFAIK, may need to be done later).


MahtraDR:
This change was made by Django early on, so there was a good reason, which alas, I don't recall now... I miss him.

PR #670, opened 2024-09-11
[all][xmlparser.rb] Split out if elif block to account for DR having multiple…

MahtraDR:
… tags in one line

I have not tested all of these (and I cannot test all of these from DR alone...) 

This is a bit brute and may need tweaking.

### DIFF
--- a/lib/xmlparser.rb
+++ b/lib/xmlparser.rb
@@ -6,9 +6,11 @@ xmlparser.rb: Core lich file that defines the data extracted from SIMU's XML.
     game: Gemstone
     tags: CORE, spells
     required: Lich > 5.7
-    version: 1.3.0
+    version: 1.3.1
 
   changelog:
+    v1.3.1 (2024-09-11)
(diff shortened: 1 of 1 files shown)


OSXLich-Doug:
Testing this one, now - I'm interested in potential impacts (favorable and non-favorable).  This one probably needs to bake a bit, though, unless we get rspec tests designed or unless more folks help test.

MahtraDR:
> Testing this one, now - I'm interested in potential impacts (favorable and non-favorable). This one probably needs to bake a bit, though, unless we get rspec tests designed or unless more folks help test.

It can only break on instances where either of a given tag should be parsed from a single line, which is what the large if/elseif block did. I don't think it'll break anything in the end. In DR, we often have more than one tag in a line, and as the case for the `nav` tag and now `component`, for example the existing structure prevented two tags from being parsed from the same line.

OSXLich-Doug (approved):
No oddities noted.  LGTM

PR #671, opened 2024-09-11
[gs][crit_tables] fix: CritRanks regex fixes

snperry85:
Got a few of the crit tables pretty well tested. All of my edits are additive, as I have no way of knowing if it's alternate messaging or if the crit tables on the wiki are so old that some of the messaging has since been updated. I'm pretty happy with the following tables.

`Grapple, Steam, Vacuum, Impact, Slash, Disruption, Non-Corporeal`

I'll expand to the other tables as I can but they may not been on the menu immediately.

I also undid the edits done to the non-corporeal table to include the entire regex. Having all the lines causes the regex to not make a partial match with the split up lines. Sorry I asked about why they were not included without investigating Doug,


lib/crit_tables/slash_critical_table.rb
```diff
@@ -87,7 +87,7 @@ module CritRanks
                                   :slowed          => false,
                                   :wound_rank      => 2,
                                   :secondary_wound => nil,
-                                  :regex           => /(?!.*removes skull.)Blow to head\./ },
+                                  :regex           => /(?!.*removes skull.)Blow to head(?:\.|\!)/ },
```
OSXLich-Doug, line 90: Preferred to have 'any one character' regex (`.`) over explicitly declaring punctuation.  This may be a problem with the initial submit - an actual period rather than a 1-char regex.  Can this be reverted to that format as the correction?

`/(?!.*removes skull.)Blow to head./ },`

lib/crit_tables/steam_critical_table.rb
```diff
@@ -281,7 +281,7 @@ module CritRanks
                                   :slowed          => false,
                                   :wound_rank      => 1,
                                   :secondary_wound => nil,
-                                  :regex           => /Hot burst to neck leaves .*? a li'l steamed./ },
+                                  :regex           => /Hot burst (?:to to|to) neck leaves .*? a li'l steamed./ },
```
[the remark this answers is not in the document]
snperry85: It is indeed an in game typo.

```
** Your wyrwood arrow erupts with a plume of steam! **
    ... 15 points of damage!
    Hot burst to to neck leaves the triton fanatic a li'l steamed!

***   ERROR no match for crit:    Hot burst to to neck leaves the triton fanatic a li'l steamed!   dam: 15 flare: Steam***
```

OSXLich-Doug (approved):
Looking good.

OSXLich-Doug (approved):

OSXLich-Doug (approved):

lib/crit_tables/cold_critical_table.rb
```diff
@@ -341,7 +341,7 @@ module CritRanks
                                   :wound_rank      => 3,
                                   :secondary_wound => nil,
                                   :regex           =>
-                                                      /Icy blast to neck freezes .*?'s words in mid\-speech and leaves it speechless./ },
+                                                      /Icy blast to neck freezes .*? words in mid\-speech and leaves (?:her|him|it) speechless\.\.\.permanently./ },
```
OSXLich-Doug, line 344: Why strike the `'s` from the regex?  It exists in 550 plus other locations, and works fine that way.  Seems to be needed here, too, to get just 'the triton warlock'.

snperry85: Icy blast to neck freezes <pushBold/>the <a exist="862580855" noun="warlock">triton warlock</a><popBold/>'s words in mid-speech and leaves <pushBold/><a exist="862580855" noun="warlock">her</a><popBold/> speechless...permanently!

mistake.

OSXLich-Doug (approved):
👍

OSXLich-Doug (commented):
Should we prefer a standard `.*?'s` or a standard `.*?` where possessive pronouns are used?  I think the vast majority is `.*?'s` but would prefer a stated standard and use that standard.  Using slash_critical_table update as example.

OSXLich-Doug:
Ah, got it.

OSXLich-Doug (approved):
👍

PR #675, opened 2024-09-25
Delete lib/infomon directory

OSXLich-Doug:


OSXLich-Doug:
Mistake!  Do not process!

PR #682, opened 2024-10-18
[gs][games.rb] bugfix: RealID update

mrhoribu:
Update to show Lich's UID if enabled instead of the game feed real ID when flag is turned on as well.

Changes:
```
[The Garden Shack - 31978] (u8086350) (8086350)
```
Back to:
```
[The Garden Shack - 31978] (u8086350)
```
When both `SET ShowRoomID ON` and `;display uid true`


OSXLich-Doug (approved):
LGTM

PR #685, opened 2024-10-24
[all][xmldata.rb] Add XMLData.room_name attribute

MahtraDR:
Adds `XMLData.room_name as a readable attribute.

```
>;eq echo XMLData.room_name
[exec1: [Willow Walk, Garden Path]] 
```

### DIFF
--- a/lib/xmlparser.rb
+++ b/lib/xmlparser.rb
@@ -86,6 +86,7 @@ class XMLParser
 
     @room_count = 0
     @room_title = String.new
+    @room_name = String.new
     @room_description = String.new
     @room_exits = Array.new
     @room_exits_string = String.new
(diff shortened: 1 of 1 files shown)


MahtraDR:
> Only issue is this xml tag will get updated in GS4 whenever a locate spell or wizard familiar "look" is used as that tag gets sent whenever you use any of those abilities. So we wouldn't be able to use this tag for mapdb current room logic as it would be unreliable. In contrast, the current mapdb tag that's used for room name/title `XMLData.room_title` tag only gets sent upon entrance to a room as it's part of the streamWindow main story window for setting the title of that window which is only done upon movement, not ability/spell usage.

That's exactly how it works in DR too. The point of this tag isn't to use it for map db stuff, it's precisely to simplify the check in your other PR. 😀

PR #698, opened 2024-11-16
[all][games.rb] bugfix: correct $_SERVERSTRING_ prior to FE

mrhoribu:
Previously we only corrected `$_SERVERSTRING_` for bad XML after it was sent to the FE for situations of nested single quotes or nested double quotes. This PR corrects this so that it's properly escaped prior to sending to FE.

### DIFF
--- a/.github/workflows/rubocop.yml
+++ b/.github/workflows/rubocop.yml
@@ -22,16 +22,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
-
-      - name: Get changed files
-        id: changed-files
-        uses: tj-actions/changed-files@v44
-        with:
-          files: |
-            **/*.lic
-            **/*.rb
-#            lich.rbw
-        
+      
       - uses: ruby/setup-ruby@v1
         with:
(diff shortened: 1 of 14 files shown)


lib/games.rb
```diff
@@ -204,6 +204,38 @@ def Game.open(host, port)
...
+                      Lich.log "Invalid nested double quotes XML tags detected: #{$_SERVERSTRING_.inspect}"
+                      $_SERVERSTRING_.gsub!(data[1], data[1].gsub!(/"/, '&quot;'))
+                      Lich.log "Invalid nested double quotes XML tags fixed to: #{$_SERVERSTRING_.inspect}"
+                      retry
+                    end
+                    $stdout.puts "error: server_thread: #{$!}\n\t#{$!.backtrace.join("\n\t")}"
+                    Lich.log "Invalid XML detected - please report this: #{$_SERVERSTRING_.inspect}"
+                    Lich.log "error: server_thread: #{$!}\n\t#{$!.backtrace.join("\n\t")}"
+                  end
+                  XMLData.reset
```
MahtraDR, line 236:
```suggestion
```

PR #699, opened 2024-11-17
[all][games.rb] fix: make show Room Exits of non-StringProc/Obvious Exits toggable for all games

mrhoribu:
Allows for showing of Lich's mapdb room exits after a current room's description. Off by default.

```
>go lodge
[The Wayside Lodge] (3003)
This ramshackle hostel bears the signs of rough use by hard-bitten travellers and headstrong adventurers.  Several blade-scarred tables and low stools are lined up against the walls, while the central area is dominated by a large firepit around which is a low platform covered with a few ratty bearskins.  The clatter of clay tankards and the smell of humble victuals provide some small comfort to the steady stream of wayfarers who pause here to meet and greet one another.  You also see some inviting tables, a steep stair and a doorway leading south.
Obvious exits: south, out
Room Exits: go steep stair
>
```


OSXLich-Doug:
Why would this not be a feature?

OSXLich-Doug:
👍

lib/games.rb
```diff
@@ -206,38 +206,45 @@ def Game.open(host, port)
...
+                      room_number = ""
+                      room_number += "#{Map.current.id}" if Lich.display_lichid
+                      room_number += " - " if Lich.display_lichid && Lich.display_uid
+                      room_number += "#{XMLData.room_id}" if Lich.display_uid
+                      unless room_number.empty?
+                        alt_string = "Room Number: #{room_number}\r\n#{alt_string}"
+                        unless ['genie', 'frostbite'].include?($frontend)
+                          alt_string = "<streamWindow id='main' title='Story' subtitle=\" - [#{XMLData.room_title[2..-3]} - #{room_number}]\" location='center' target='drop'/>\r\n#{alt_string}"
+                          alt_string = "<streamWindow id='room' title='Room' subtitle=\" - [#{XMLData.room_title[2..-3]} - #{room_number}]\" location='center' target='drop' ifClosed='' resident='true'/>#{alt_string}"
+                        end
```
MahtraDR, line 236: This doesn't work for DR at all. To get our room number, the room title has to have been determined already. This isn't rewriting the room title to append the room number, it's just adding additional lines to the stream that aren't doing anything. It isn't possible at present to rewrite the room title window to append room number because room number cannot be determined before the whole room is parsed. It's why we have `Room Exits:` appended to the room description.

MahtraDR: Fair enough. Wrayth only makes sense.

PR #715, opened 2024-12-28
[all][util.rb] bugfix: allow issue_command to use fput or put

mrhoribu:
Sometimes you want to use an `fput` and sometimes you want to use a `put` command depending on the situation and to prevent race conditions of spammy additional scripts running causing a false RT detection. This allows to switch the command being sent to use `put` but defaults to usage of `fput`.

### DIFF
--- a/lib/util.rb
+++ b/lib/util.rb
@@ -93,7 +93,7 @@ module Lich
               line
             end
           })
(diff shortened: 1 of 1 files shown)


OSXLich-Doug (approved):
Good change

PR #719, opened 2025-01-05
[dr][map_dr.rb] cleanup and support for UIDs

MahtraDR:
It is not pretty, but it seems to work. :)


MahtraDR:
Been using this the last few days and it looks good. I'll make another pass later cleaning it up further, removing defs we don't use, etc...

PR #721, opened 2025-01-06
[all][gui-saved-login.rb] fix: update Stormfront label

mrhoribu:
Change label for Stormfront to show as Wrayth. Also append "custom" to the FE label if a custom launch cmd is given


OSXLich-Doug (approved):
Good fix

PR #728, opened 2025-01-09
[dr][games.rb] bugfix: Genie FE real ID#s not supported, hide by default

mrhoribu:
Due to real ID#s being an option and now being required for future Lich5 release on DragonRealms, but GenieFE not supporting them for mapping function, hide real ID#s until client updated.


MahtraDR (approved):
Tests well. Let's roll it.

PR #729, opened 2025-01-10
[all][update.rb] purge old LIB folder prior to update release

mrhoribu:
Prevent old stale files from being left behind in LIB folder by purging prior to copying files into LIB folder


OSXLich-Doug (approved):
Great add

PR #733, opened 2025-01-24
[all][main.rb] fix: create session file in regular connect method

mrhoribu:
Currently we only create a session file in `File.join(Dir.tmpdir, "simutronics", "sessions")` if Lich is started with CLI arg of `--detachable-client=PORT`. This PR corrects a logic mistake in that detection as well as creates the same session file for launching from GUI that FEs can utilize if not able to take passed launch parameters to connect to a host/port. Namely Mudlet.

Also adds new `Account.character` api to allow for session file to be written properly.

### DIFF
--- a/lib/main/main.rb
+++ b/lib/main/main.rb
@@ -285,7 +285,7 @@ reconnect_if_wanted = proc {
       end
       accept_thread = Thread.new { $_CLIENT_ = SynchronizedSocket.new(listener.accept) }
       localport = listener.addr[1]
-      Frontend.create_session_file(Account.character, listener.addr[2], listener.addr[1])
+      Frontend.create_session_file(Account.character, listener.addr[2], listener.addr[1], display_session: false)
       if custom_launch
         sal_filename = nil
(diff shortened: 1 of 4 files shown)


lib/main/main.rb
```diff
@@ -285,6 +285,7 @@
       end
       accept_thread = Thread.new { $_CLIENT_ = SynchronizedSocket.new(listener.accept) }
       localport = listener.addr[1]
+      Frontend.create_session_file(Account.character, listener.addr[2], listener.addr[1])
```
OSXLich-Doug, line 288: This creates a session file for official FE - I don't think it's useful there.  And this exists in the detachable client, where I think it needs to live if the client can utilize it.  Suggest dropping this one, and updating the one in detachable client to use `Account.character`.

OSXLich-Doug: Ok, to resolve I'm going to add a `front-end.rb` to permit apps / FEs / scripts to find the session file location for the current character.  And I'm going to modify this one to remove the now duplicative session file creation under detachable client.  That way they'll be ready for review.

OSXLich-Doug (approved):
Good here - nice updates

PR #734, opened 2025-01-24
first update to update.rb for 5.11

OSXLich-Doug:
This is a fundamental change to our update methodology.  New feature is 

`;lich5-update --install --version=<tag> --lib=<file.rb>`

--install is required to ensure this is really what the user wants
--version is required and must match an existing release
--lib is optional and is intended to allow point update.

This needs thorough testing, but - I would strongly recommend you not test it against your existing cloned Lich.  It does delete files / folders, making the commit status an absolute mess.

And remember during your testing, if you see something you didn't like, `;lich5-update --revert` is your best friend.

### DIFF
--- a/lib/util/update.rb
+++ b/lib/util/update.rb
@@ -9,46 +9,67 @@ module Lich
       require 'zlib'
 
       @current = LICH_VERSION
-      @snapshot_core_script = ["alias.lic", "autostart.lic", "dependency.lic", "ewaggle.lic", "foreach.lic", "go2.lic", "infomon.lic",
+      @back_rev = false
+      @snapshot_core_script = ["alias.lic", "autostart.lic", "dependency.lic",
+                               "ewaggle.lic", "foreach.lic", "go2.lic", "infomon.lic",
                                "jinx.lic", "lnet.lic", "log.lic", "logxml.lic", "map.lic", "repository.lic", "vars.lic", "version.lic"]
+
+      INSTALL_REGEX = %r[--install(?:\s+--version=(?<tag>(?:v)?[\d\.]+(?:-[\w\d\.]+)?))?(?:\s+--(?<type>script|lib(?:rary)?|data)=(?<file>.*\.(?:lic|rb|xml|json|yaml|ui)))?]
+      BETA_REGEX = %r[--(?:beta|test)((?:\s+--(?<type>script|lib(?:rary)?|data))=(?<file>.*\.(?:lic|rb|xml|json|yaml|ui)))?]
+      VERSION_REGEX = %r[(?:v)?[\d\.]+(?:-[\w\d\.]+)?]
 
       def self.request(type = '--announce')
         case type
         when /--announce|-a/
           self.announce
-        when /--(?:beta|test)(?: --(?:(script|library|data))=(.*))?/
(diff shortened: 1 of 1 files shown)


lib/util/update.rb
```diff
@@ -118,74 +159,50 @@ def self.snapshot
...
+          filename = "https://api.github.com/repos/elanthia-online/lich-5/releases"
+          update_info = URI.parse(filename).open.read
+          record = JSON::parse(update_info).first # assumption: Latest beta release always first record in API
+        when VERSION_REGEX
+          requested_version = version.gsub(/v/, '') # if present, remove 'v' to standardize on version numbers
+          filename = "https://api.github.com/repos/elanthia-online/lich-5/releases"
+          update_info = URI.parse(filename).open.read
+          temp_record = JSON::parse(update_info).select { |h1| h1['tag_name'] == "v#{requested_version}" }
+          unless temp_record.nil? or temp_record.empty?
+            record = temp_record.first
```
OSXLich-Doug, line 178: Which is what is processed in the next three lines.  If the requested version is not in the releases api, it fails out, with an error message.

lib/util/update.rb
```diff
@@ -291,22 +324,26 @@ def self.update_file(type, rf, version = 'production')
...
           case version
           when "production"
             remote_repo = "https://raw.githubusercontent.com/elanthia-online/lich-5/master/lib"
-          when "beta"
+          when "beta" # FIXME: need to correct to releases?
             remote_repo = "https://raw.githubusercontent.com/elanthia-online/lich-5/staging/lib"
+          when VERSION_REGEX # user asked for a file from a particular release
+            version.gsub!(/^v/, '') # if present, remove 'v' to standardize on version numbers
+            remote_repo = "https://raw.githubusercontent.com/elanthia-online/lich-5/refs/tags/v#{version}/lib/#{requested_file}"
+            remote_repo = nil unless validate_url_request(remote_repo)
```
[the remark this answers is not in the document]
OSXLich-Doug: Like the preceding feedback point, this also will simply error out if the specific URL doesn't exist - whether it's by version or by file, with an error message.  However, in re-reviewing to ensure I had my fact correct, I believe there is a problem here to address.  Specifically, with all of our subdirs, the above does not account for namespace / folder location nor is there any captured input from the user.  While this fails out appropriately for a file misnamed in the lib folder, if fails out inappropriately for a file in a subfolder in lib.

I'll have to consider that solution.  I'm not a huge fan of `--lib=commons/gtk.rb` and all that might entail.   I'm also not sure I'm interested in walking a specific version tree hunting for a file location either.  Choices, choices.   It's current location is intended to ensure no file activities (inluding zero byte file creation) occur before the determination is made.  I'll look over it again, while solving the subfolder debacle.

lib/util/update.rb
OSXLich-Doug: This goes to the first review comment - I mistyped something, I am sure.

OSXLich-Doug:
This requires a significant restructure after launch of 5.11 to capitalize on new branch / release methodology.  We'll launch 5.11 first with the new namespace information, then probably pare the branches down to one, then update the `update.rb` in time to catch 5.11.1 to implement the new `tag` features.

PR #736, opened 2025-01-25
[all][main.rb] fix: Mudlet support

mrhoribu:
Mudlet doesn't support real Simutronics integration, so force it if launcher_cmd contains it


OSXLich-Doug (approved):
LGTM

PR #739, opened 2025-01-26
[dr][drdefs.rb] Simplify npc detection and add ordinals to them

MahtraDR:
We now parse room_objects once and detect live and dead NPCs in one place, adding ordinals to all NPCs.

Works well for me, but should wait till the next release is out, so don't merge it yet, please.


OSXLich-Doug:
Mahtra ok'd, merging.

PR #740, opened 2025-01-26
[all][games.rb]bug:fix namespace constants

OSXLich-Doug:
This permits symbols for namespace constants to be defined.


OSXLich-Doug:
Mahtra Ok'd, merging.

PR #744, opened 2025-01-27
[all][gui-saved-login.rb] fix: Wrayth display non-tabbed entries

mrhoribu:
This fixes showing "Wrayth" instead of "Stormfront" on the non-tabbed interface, as well as adds "custom" to the entry if custom_launch for it


OSXLich-Doug:
👍

PR #747, opened 2025-02-01
[gs][stash.rb] bugfix: UserVars String class check

mrhoribu:
Prevent UserVars used for stash module from being used unless they're Strings.


OSXLich-Doug (approved):
👍

PR #751, opened 2025-02-03
[all][games.rb] feat: toggleable display of StringProcs with clickable links

mrhoribu:
Add an additional `;display stringprocs` option to show StringProcs for a room. Also makes Room Exits and Stringprocs clickable for FrontEnds that support clickable links.


OSXLich-Doug (approved):
Tests clean, no issues discovered.  👍

PR #757, opened 2025-02-10
[all][update.rb] fix: safe escape lich character

mrhoribu:
Use `$clean_lich_char` in place of `;` to support FEs that use a different character.


OSXLich-Doug (approved):
👍

PR #760, opened 2025-02-11
[dr][equipmanager.rb] fix: safe escape lich character

mrhoribu:


OSXLich-Doug (commented):
👍 Another @MahtraDR but also standard solid Tysong stuff.

MahtraDR (approved):

PR #761, opened 2025-02-11
[dr][common-crafting.rb] fix: safe escape lich character

mrhoribu:


OSXLich-Doug (commented):
👍 Probably want @MahtraDR to look this one over too, but it is straightforwrard.

MahtraDR (approved):

PR #762, opened 2025-02-11
[gs][infomon/parser.rb] bugfix: non-lich auth Account module

mrhoribu:
Allows for population of AccountName and Subscription into `Account` module if they're nil due to not using Lich to authenticate


OSXLich-Doug (approved):
👍

PR #763, opened 2025-02-11
[dr][drparser.rb] bugfix: non-lich auth Account module

mrhoribu:
Allows for population of Account.name and Account.subscription into Account module if they're nil due to not using Lich to authenticate


OSXLich-Doug:
👍  - Probably need to have @MahtraDR weigh in on this one, but looks solid.

PR #768, opened 2025-02-17
[all][games.rb] bugfix: Wrayth <settings> bad XML

mrhoribu:
Move the ignore for Wrayth `<settings>` being sent to before trying to REXML parse since it's not anything Lich5 needs and is not true XML and can cause issues parsing


OSXLich-Doug:
👍 

Tested, functionally no change.

Customer reported issue requires addressing prior to launch.  Customer sideload testing success reported.

PR #777, opened 2025-02-22
[gs][skills.rb] bugfix: proper name of Skill bonus lookup

mrhoribu:
Incorrectly used `Infomon.get("skill.%s.bonus" % ranks)` instead of how it's actually stored as `Infomon.get("skill.%s_bonus" % ranks)`


OSXLich-Doug (approved):
👍 

Merging. . .

PR #778, opened 2025-02-23
feat(gs): add olib Container;

ondreian:
I'm going to get rid of `Containers` because `Container` vs `Containers` is confusing.  When I originally wrote this I was trying to follow Lich naming conventions ala (`Settings` vs `Setting` vs `GameSettings` versus `CharSettings`), but we have all sense wisely soured on those in favor of Ruby standard conventions.

However, I just wanted to open this draft so we can start seeing some WIP and maybe Tysong can give me feedback/contribute as well.

My eventual goal is to have `Container` feel familiar to an `ActiveRecord` model, with common patterns for querying/accessing/etc, and a few smart enhancements.  This will not be a "kitchen sink" implementation, I think initially we just ship a thin API.


mrhoribu:
Closing as will need to refactor a bit and change name scope location. Will readdress in new PR in the future.

PR #785, opened 2025-02-27
chore(dr): common-arcana - predicate to return backfire status and spell regex patterns update

mdr55:
backfired? provides a boolean status on whether the last spell cast using DRCA.cast? backfired.
  - increase selectivity for auto_mana to pick out a backfire vs the (super-)set of all spell failures which shouldn't be penalised.
 
Updated spell failure patterns to include sorcery backlash, more specific regex for backfire, and missing target (dead before spell finished prepping)

### DIFF
--- a/lib/dragonrealms/commons/common-arcana.rb
+++ b/lib/dragonrealms/commons/common-arcana.rb
@@ -293,7 +293,6 @@ module Lich
         Flags.add('cyclic-too-recent', 'The mental strain of initiating a cyclic spell so recently prevents you from formulating the spell pattern')
         Flags.add('spell-full-prep', /^This pattern may only be cast with full preparation/)
         Flags.add('spell-backfired', /^Your spell .*backfires/)
-        @@backfired_status = false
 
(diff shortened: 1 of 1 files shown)


lib/dragonrealms/commons/common-arcana.rb
```diff
@@ -318,6 +324,8 @@ def cast?(cast_command = 'cast', symbiosis = false, before = [], after = [])
           DRC.bput('release mana', 'You release all', "You aren't harnessing any mana")
         end
 
+        @@backfired_status = true if Flags['spell-backfired']
```
MahtraDR, line 327: What does `Flags['spell-backfired']` return when it isn't set? Could `@@backfired_status` just be equal to that?

mdr55: Updated, could just be set to equal.  Removed the earlier = false as well, not needed now.

MahtraDR:
Thank you!

PR #790, opened 2025-03-04
fix(gs): migrate PSMS.name_normal to Lich::Util.name_normal

mrhoribu:
Allows for usage of `name_normal` method to be utilized by other functions within Lich without the reference of PSMS to make it more generic usage.


OSXLich-Doug (approved):
LGTM 👍

PR #799, opened 2025-03-13
fix(gs): FEAT Covert Arts colon normalization and stamina costs

mrhoribu:
Bug in conversion as `Covert Art: Escape Artist` would get translated into `covert_art__escape_artist` with current substitution


OSXLich-Doug (approved):
👍

PR #835, opened 2025-03-27
fix(dr): drparser.rb Clean up exp tracking code

MahtraDR:
Moving this to exp-tracking.lic as per https://github.com/elanthia-online/dr-scripts/pull/7146


MahtraDR:
> Appears ok, only issue is it's got all the code from #834 in it as well.

Thanks. Fixed.

PR #836, opened 2025-03-27
fix(all): games.rb proper fix for single/double nested quotes

mrhoribu:


lib/games.rb
```diff
@@ -887,7 +887,7 @@ def Game.open(host, port)
                       end
                       # Fixes invalid XML with nested double quotes in it such as:
                       # <subtitle=" - [Avlea's Bows, "The Straight and Arrow"]">
-                      if (matches = $_SERVERSTRING_.scan(/"([^=]*"[^=]*)"/))
+                      unless (matches = $_SERVERSTRING_.scan(/"([^=]*"[^=]*)"/)).empty
```
MahtraDR, line 890:
```suggestion
                      unless (matches = $_SERVERSTRING_.scan(/"([^=]*"[^=]*)"/)).empty?
```

?

PR #841, opened 2025-03-27
fix(gs): infomon parsing for CHE membership

mrhoribu:
Update infomon parsing to match on PROFILE house membership for only self profile. Also allow for RESIGN output to match and update status as well.

----

> [!IMPORTANT]
> Update infomon parser to handle self-profile CHE membership and resignation parsing with new regex patterns and state management.
> 
>   - **Behavior**:
>     - Update `ProfileHouseCHE` regex to match house membership only for self-profile.
>     - Add `ResignCHE` regex to handle resignation output and update status.
>   - **State Management**:
>     - Introduce `Profile` state in `State` module to manage parsing state transitions.
>     - Use `State.set()` and `State.get()` in `parse()` to handle state transitions for profile and goals.
>   - **Patterns**:
>     - Add `ProfileStart` and `ProfileName` patterns to identify self-profile parsing.
>     - Update `All` regex union to include new patterns.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=elanthia-online%2Flich-5&utm_source=github&utm_medium=referral)<sup> for b96c0842a6d47495a313b42e7a638970c5f3fa04. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/lib/gemstone/infomon/parser.rb
+++ b/lib/gemstone/infomon/parser.rb
@@ -111,8 +111,8 @@ module Lich
             case state
             when Goals, Profile
               unless @state.eql?(Ready)
-                respond "--- Lich: error: Infomon::Parser::State is in invalid state(#{@state} - caller: #{caller[0]}"
-                Lich.log "error: Infomon::Parser::State is in invalid state(#{@state} - caller: #{caller[0]}"
+                Lich.log "error: Infomon::Parser::State is in invalid state(#{@state}) - caller: #{caller[0]}"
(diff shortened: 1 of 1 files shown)


lib/gemstone/infomon/parser.rb
```diff
@@ -412,7 +415,27 @@ def self.parse(line)
               else
                 :noop
               end
-            when Pattern::HouseCHE
+            when Pattern::ProfileStart
+              @profile_self_output = true
```
ondreian, line 419: are we doing this in a lot of places? We should probably rewrite this to be `@operation = <some enum>`

PR #894, opened 2025-05-19
fix(dr): Update regex and parsing for active spells

MahtraDR:
Occasionally the old regex would match multiple lines as one and get erroneous spell durations. This addresses that.

Old: <https://rubular.com/r/3pZ3mvyMODuWpy>
image

New: <https://rubular.com/r/I0UCfcftNoRbcF>

### DIFF
--- a/lib/common/xmlparser.rb
+++ b/lib/common/xmlparser.rb
@@ -662,18 +662,25 @@ module Lich
             spell = nil
             duration = nil
             case text_string
-            when /(?<spell>[^<>]+?)\s+\((?:\D*)(?<duration>\d+)\s*(?:%|roisae?n)\)/i
+            when /(?<spell>^[^\(]+)\((?<duration>\d+|Indefinite|OM|Fading)\s*(?:%|roisae?n)?\)/i
               # Spell with known duration remaining
-              spell = Regexp.last_match[:spell]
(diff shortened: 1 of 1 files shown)


lib/common/xmlparser.rb
```diff
@@ -684,10 +691,10 @@ def text(text_string)
               @dr_active_spells_stellar_percentage = Regexp.last_match[:percentage].to_i
               unit = Regexp.last_match[:unit]
               duration = unit == 'anlaen' ? duration * 30 : duration
-            when /(?<spell>[^<>]+?)\s+\(.+\)/i
+            when /(?<spell>^[^\(]+)\(.+\)/i
               # Spells with inexact duration verbiage, such as with
               # Barbarians without knowledge of Power Monger mastery
-              spell = Regexp.last_match[:spell]
+              spell = Regexp.last_match[:spell].strip
```
KatoakDR, line 697: Good catch, should line 689 also do a strip?

KatoakDR (approved):

lib/common/xmlparser.rb
MahtraDR: That one shouldn't need a strip because in that instance, spell is always `Stellar Collector` We could even clean that up a bit, but it's not hurting anything either.

PR #1079, opened 2025-12-02
fix(all): [lib][global-defs] - move update

BinuDR:
When trying to use `def move` while lying down, it doesn't have a match for `You must be standing.`, so I added it.

----

> [!IMPORTANT]
> Update `move()` in `global_defs.rb` to handle lying down by matching `You must be standing.`
> 
>   - **Behavior**:
>     - Update `move()` in `global_defs.rb` to handle the case where the user is lying down by matching the message `You must be standing.`
>   - **Misc**:
>     - Added regex pattern `^You must be standing` to existing conditions in `move()` to trigger standing up when lying down.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=elanthia-online%2Flich-5&utm_source=github&utm_medium=referral)<sup> for dff1d2d17f4309c43f2ed2917cc3253c577a4210. You can [customize](https://app.ellipsis.dev/elanthia-online/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/lib/global_defs.rb
+++ b/lib/global_defs.rb
@@ -640,7 +640,7 @@ def move(dir = 'none', giveup_seconds = 10, giveup_lines = 30)
         sleep 0.3
       end
       put_dir.call
-    elsif line =~ /will have to stand up first|must be standing first|^You'll have to get up first|^But you're already sitting!|^Shouldn't you be standing first|^That would be quite a trick from that position\.  Try standing up\.|^Perhaps you should stand up|^Standing up might help|^You should really stand up first|You can't do that while sitting|You must be standing to do that|You can't do that while lying down/
+    elsif line =~ /will have to stand up first|must be standing first|^You'll have to get up first|^But you're already sitting!|^Shouldn't you be standing first|^That would be quite a trick from that position\.  Try standing up\.|^Perhaps you should stand up|^Standing up might help|^You should really stand up first|You can't do that while sitting|You must be standing to do that|You can't do that while lying down|^You must be standing/
       fput 'stand'
       waitrt?
       put_dir.call


lib/global_defs.rb
```diff
@@ -640,7 +640,7 @@ def move(dir = 'none', giveup_seconds = 10, giveup_lines = 30)
         sleep 0.3
       end
       put_dir.call
-    elsif line =~ /will have to stand up first|must be standing first|^You'll have to get up first|^But you're already sitting!|^Shouldn't you be standing first|^That would be quite a trick from that position\.  Try standing up\.|^Perhaps you should stand up|^Standing up might help|^You should really stand up first|You can't do that while sitting|You must be standing to do that|You can't do that while lying down/
+    elsif line =~ /will have to stand up first|must be standing first|^You'll have to get up first|^But you're already sitting!|^Shouldn't you be standing first|^That would be quite a trick from that position\.  Try standing up\.|^Perhaps you should stand up|^Standing up might help|^You should really stand up first|You can't do that while sitting|You must be standing to do that|You can't do that while lying down|^You must be standing/
```
MahtraDR:
```suggestion
    elsif line =~ /(?:stand(?:ing)? up|standing|get up) first|(?:already|while) (?:sitting|lying down)|must be standing|(?:try )?standing up|should stand up/i
```
https://rubular.com/r/6fisvpjhOwzSJz

mrhoribu: While it may match, it loses the precision and requires a bit more processing to remove the extra words vs having them due to how regex works. 938 steps vs 792 steps.

PR #1092, opened 2025-12-14
fix(gs): cman.rb hamstring regex update when can't perform on target

ServerFailure:
Updated hamstring results

----

> [!IMPORTANT]
> Update regex for `hamstring` maneuver in `cman.rb` to handle cases where the action cannot be performed.
> 
>   - **Behavior**:
>     - Updated regex for `hamstring` maneuver in `cman.rb` to include `/You cannot hamstring .+\./` pattern.
>     - Handles cases where hamstring action cannot be performed.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=elanthia-online%2Flich-5&utm_source=github&utm_medium=referral)<sup> for 3e12fd890abe9fbf79fa30fb6378dfa8aab3c123. You can [customize](https://app.ellipsis.dev/elanthia-online/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/lib/gemstone/psms/cman.rb
+++ b/lib/gemstone/psms/cman.rb
@@ -289,7 +289,7 @@ module Lich
           :regex      => Regexp.union(/You lunge forward and try to hamstring .+ with your .+!/,
                                       /The .+ is too unwieldy for that\./,
                                       /You need to be holding a weapon capable of slashing to do that\./,
-                                      /You cannot hamstring .+\./),
+                                      /You cannot hamstring .+\./,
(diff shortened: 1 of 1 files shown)


lib/gemstone/psms/cman.rb
mrhoribu: Would it make more sense to do this generically in the results_regex built from the name?
```
/^You cannot #{name} .+\./,
```
This way it matches for more than just hamstring?

mrhoribu: Also if that's the case, probably wouldn't hurt to update all the other PSMS `.use` to have similar in their .use function.

PR #1114, opened 2026-01-13
feat(gs): add QStrike module for optimal stamina-based RT reduction

Nisugi:
## Summary

Adds `Lich::Gemstone::QStrike` module for calculating optimal QSTRIKE roundtime reduction based on current stamina, weapon setup, and attack costs.

## Key Features

- **Automatic equipment detection** - Reads weapon stats from both hands, applies speed multipliers (1.5x for two-handed/polearm, 2.5x for ranged)
- **Striking Asp support** - Automatically applies cost reduction when stance is active (rank 1: 33%, rank 2: 50%, rank 3: 67% discount)
- **Attack cost integration** - Looks up CMan/Weapon/Shield technique stamina costs
- **Per-character defaults** - Settings persisted via DB_Store

## API

```ruby
# Calculate optimal reduction that won't cause negative stamina
QStrike.calculate(reserve: 1, attack_name: :cripple)
# => { seconds: 2, stamina_cost: 30, qstrike_cmd: "qstrike -2", ... }

# Get just the command string
QStrike.command(reserve: 5)
# => "qstrike -3"

# Check if any reduction is affordable
QStrike.affordable?(attack_name: :tackle)
# => true

# Execute qstrike + attack in one call
QStrike.use(reduction: -3, attack: "cripple", target: "kobold")
# Sends: "qstrike -3" then CMan.use("cripple", "kobold")

# Target absolute RT (reduce to 2 second RT)
QStrike.use(reduction: 2, attack: "attack", target: "troll")

# Use adaptive mode - if requested reduction is too expensive, use max affordable
QStrike.use(reduction: -5, attack: "mstrike", target: "troll", adaptive: true)
# If -5 costs too much, automatically uses max affordable reduction instead

# Set per-character defaults
QStrike.set_default(:reserve, 5)
QStrike.set_default(:adaptive, true)

# Get cost per second of RT reduction for current weapon setup
QStrike.cost_per_second_reduction
# => 15

# Debug breakdown of cost calculation
QStrike.debug_calculation
```

## Formula

```
Cost per second = (10 + primary_equipment_speed + secondary_equipment_speed/2) * striking_asp_multiplier
```

Where equipment_speed = weapon_base_rt × category_multiplier

## Test Plan

- [x] 58 RSpec tests pass locally

### DIFF
--- a/lib/gemstone/psms/qstrike.rb
+++ b/lib/gemstone/psms/qstrike.rb
@@ -23,7 +23,7 @@ module Lich
     #
     # @example Set defaults
     #   QStrike.set_default(:reserve, 5)
-    #   QStrike.set_default(:force, true)
+    #   QStrike.set_default(:adaptive, true)
     #
     module QStrike
       # Speed multipliers by weapon category
@@ -60,7 +60,7 @@ module Lich
       # Default settings
       DEFAULT_SETTINGS = {
         reserve: 1,
-        force: false
+        adaptive: false
       }.freeze
 
       # === DEFAULTS MANAGEMENT (Persisted via DB_Store) ===
@@ -71,7 +71,7 @@ module Lich
         load_settings
         {
           reserve: @settings[:reserve],
-          force: @settings[:force]
+          adaptive: @settings[:adaptive]
         }
       end
 
@@ -84,7 +84,7 @@ module Lich
       end
 
       # Set a single default value (persists to DB_Store for per-character storage)
-      # @param key [Symbol] Setting name (:reserve, :force)
+      # @param key [Symbol] Setting name (:reserve, :adaptive)
       # @param value [Object] New value
       def self.set_default(key, value)
         load_settings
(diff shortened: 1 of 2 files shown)


lib/gemstone/psms/qstrike.rb
```diff
@@ -0,0 +1,813 @@
...
+    #   # Executes: put "qstrike -3" then CMan.use("cripple", "kobold")
+    #
+    # @example Absolute RT mode (reduce to 2 second RT)
+    #   QStrike.use(reduction: 2, attack: "attack", target: "troll")
+    #
+    # @example Set defaults
+    #   QStrike.set_default(:reserve, 5)
+    #   QStrike.set_default(:adaptive, true)
+    #
+    module QStrike
```
mrhoribu: If the module is located at `Lich::Gemstone::QStrike` it really should be located at `lib\gemstone\qstrike.rb` to follow proper folder structure for library files in Ruby ecosystem. Granted, there's a few other lib files that don't follow that, but they also need corrected :D

mrhoribu (approved):

PR #1301, opened 2026-03-31
fix(all): use :__unset for --without-frontend CLI request YAML matches

OSXLich-Doug:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved CLI login behavior to properly handle the `--without-frontend` flag when looking up saved login entries. The system now ensures that headless mode settings don't incorrectly constrain saved-entry lookup, providing more consistent behavior during authentication.

### DIFF
--- a/lib/common/authentication/login_helpers.rb
+++ b/lib/common/authentication/login_helpers.rb
@@ -507,6 +507,23 @@ module Lich
           [instance, frontend, custom_launch]
         end
 
+        # Resolves which frontend should be used when matching a saved entry for
+        # CLI login.
+        #
+        # Headless CLI launches still carry an intended frontend for downstream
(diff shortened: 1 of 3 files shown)


MahtraDR:
I don't understand this one. What's the issue exactly?

mrhoribu (approved):

mrhoribu:
> I don't understand this one. What's the issue exactly?

Fixes doing CLI launches like:
```
ruby -W0 lich5/lich.rbw --dragonrealms --without-frontend --detachable-client=8001 --frostbite --login Bob
```
To not fail if the saved entry doesn't have the `frontend: frostbite` which normally it does not. It now ignores trying to find a matching frontend when `--without-frontend` is used.

MahtraDR (approved):
Works for me! :)

PR #1303, opened 2026-04-02
fix(dr): Add top-level aliases for ArgParser and SetupFiles

MahtraDR:
## Summary

- Add `ArgParser = Lich::Common::ArgParser` alias at the bottom of `lib/common/arg_parser.rb`
- Add `SetupFiles = Lich::Common::SetupFiles` alias at the bottom of `lib/common/setup_files.rb`

Core lich defines these classes under `Lich::Common`, but DR scripts (via dependency.lic) reference them as bare top-level constants (`ArgParser.new`, `SetupFiles.new`). When the `CORE_ARGPARSER` / `CORE_SETUPFILES` sentinel gates in dependency.lic detect the core versions and skip their inline definitions, bare constant lookups fail with `NameError`.

This was introduced in #1273 and affects any user on lich >= 5.16.0 running the latest dependency.lic.

## Test plan
- [x] `ruby -c` passes on both files
- [ ] In-game: verify `ArgParser.new` resolves after dependency skips inline definition
- [ ] In-game: verify `SetupFiles.new` resolves after dependency skips inline definition
- [ ] In-game: verify `get_settings` / `parse_args` work end-to-end

## Summary by CodeRabbit

* **New Features**
  * Added top-level constant aliases `ArgParser` and `SetupFiles` for more convenient access to argument parsing and setup file management utilities without requiring full namespace references.

### DIFF
--- a/lib/common/arg_parser.rb
+++ b/lib/common/arg_parser.rb
@@ -139,3 +139,7 @@ module Lich
     end
   end
 end
+
+# Top-level alias so scripts can use bare `ArgParser.new` without
+# fully qualifying the Lich::Common namespace.
+ArgParser = Lich::Common::ArgParser unless defined?(::ArgParser)
--- a/lib/common/setup_files.rb
+++ b/lib/common/setup_files.rb
@@ -285,3 +285,7 @@ module Lich
     end
   end
 end
+
+# Top-level alias so scripts can use bare `SetupFiles.new` without
+# fully qualifying the Lich::Common namespace.
+SetupFiles = Lich::Common::SetupFiles unless defined?(::SetupFiles)


OSXLich-Doug:
`Lich::Common` in its entirety is brought to top level in `lich.rbw`.  

```
;e echo LICH_VERSION
--- Lich: exec1 active.
[exec1: 5.16.2]
--- Lich: exec1 has exited.
>;e echo defined?(ArgParser.new)
--- Lich: exec1 active.
[exec1: method]
--- Lich: exec1 has exited.
;e echo defined?(SetupFiles.new)
--- Lich: exec1 active.
[exec1: method]
--- Lich: exec1 has exited.
```

Since this already lives toplevel, this PR is probably not necessary and is likely masking a load order or context issue that may become a larger problem.

MahtraDR:
You're right -- closing this.

`include Lich::Common` at `lich.rbw:118` already makes `ArgParser` and `SetupFiles` resolvable as bare top-level constants through Object's ancestor chain. On top of that, all script bindings (`TRUSTED_SCRIPT_BINDING` and `Scripting.new.script`) are created inside `Lich::Common` in `lib/common/script.rb`, so `ArgParser` is found via lexical nesting before Object is even consulted.

The NameError claim in the PR description was assumed, not observed -- your in-game test confirms it resolves fine. Adding explicit aliases would mask any real context issue rather than surfacing it.

I'm also removing the redundant alias lines from the dependency.lic sentinel gates in dr-scripts#7372 (`ArgParser = Lich::Common::ArgParser unless defined?(::ArgParser)` is dead code since `defined?(::ArgParser)` is always truthy after the include).

Thanks for catching this.

MahtraDR:
Closing per discussion -- `include Lich::Common` already handles constant resolution.

PR #1306, opened 2026-04-02
feat(all): Add Map.apply_wayto_overrides for personal map overrides

MahtraDR:
## Summary

- Add `Map.apply_wayto_overrides` to `MapBase::ClassMethods` -- reads `base_wayto_overrides`, `personal_wayto_overrides`, and `personal_map_targets` from YAML settings and patches the in-memory map
- Add `CORE_MAP_OVERRIDES` sentinel in `Lich::Common`
- Game-agnostic -- available in both DR and GS via the shared MapBase module

Replaces dependency.lic's `make_map_edits` (which delegated to `ScriptManager.make_map_edits`). Logic is identical but now lives in core alongside the map code. Adds a nil guard for rooms that don't exist in the current map.

## Test plan
- [x] `ruby -c lib/common/map/map_base.rb` passes
- [x] All 163 existing map specs pass
- [ ] In-game: verify `Map.apply_wayto_overrides` applies wayto overrides from YAML
- [ ] In-game: verify personal_map_targets merge into GameSettings

## Summary by CodeRabbit

* **New Features**
  * Map customization: players can override navigation routes and specify custom travel times for specific connections.
  * Personal map targets: user-defined targets can be added and will persist in settings across sessions.
* **Bug Fixes / Improvements**
  * Personal overrides now take precedence over base overrides.
  * Travel-time values are parsed as numbers when possible; invalid or malformed entries are safely ignored.


MahtraDR:
@coderabbitai Both nitpicks resolved in 153adfc:

**1. nil.to_i targeting room 0 -- RESOLVED.** Added `next unless values.is_a?(Hash) && values['start_room']` before `.to_i`. This skips malformed entries early rather than relying on the downstream `next unless start_room` guard, which would pass silently since room 0 exists in the map.

**2. Map-loaded guard -- RESOLVED.** Added `self.load unless loaded?` at method entry, consistent with `get_free_id`, `estimate_time`, and `dijkstra` in the same ClassMethods module. This follows the established convention and prevents silent no-ops if `apply_wayto_overrides` is called before map initialization (e.g., early in the autostart sequence).

Both are cheap one-line guards that follow SOLID's defensive design principle -- fail fast on invalid input, and maintain consistent preconditions across the class interface.

OSXLich-Doug (approved):
Should be a good add for GS, hence placement in map_base.  Approved.

PR #1312, opened 2026-04-08
fix(all): GTK hardening via guards and timer protections

OSXLich-Doug:
## Summary by CodeRabbit

* **New Features**
  * Process-wide guards for the shared GTK main loop with an explicit core-only shutdown escape hatch; launcher idle handler kept active.
  * Registries that retain GTK/GLib callbacks and signal handlers to prevent premature GC.

* **Bug Fixes**
  * Safer, more consistent GUI shutdown across multiple early-exit flows.
  * Improved lifecycle handling to avoid dangling callbacks and related crashes.

* **Documentation**
  * Added a GTK hardening guide describing the safety measures.

* **Tests**
  * New and updated tests/mocks validating hardening, retention, and shutdown behavior.


OSXLich-Doug:
@coderabbitai - review, and during review update the now erroneous DocString observation as well

OSXLich-Doug:
@coderabbitai - I will address two of the items raised.  However, theGLib::Idle.add callback is intentionally perpetual, so one retained idle entry living for process lifetime is consistent with the current retention model. I will tweak it for clarity, though:

change it to sleep 0.01; true

add I will add a short comment that this keepalive is intentionally repeating
optionally note in GTK-HARDENING.md that long-lived callbacks remain retained by design

OSXLich-Doug:
For Mahtra recommends what changed:

 - gtk.rb now blocks Gtk.main_quit unconditionally unless the core-owned Gtk.lich_main_quit escape hatch is active.
 - gtk.rb also drops the unnecessary cleanup_installer proc indirection, so the destroy-hook retention path is easier to read.
 - spec_helper.rb now contains a shared mock GTK hardening environment context, and both gtk_spec.rb and gtk_edge_cases_spec.rb use it instead of duplicating the large fake GTK/GLib harness.
 - GTK-HARDENING.md now documents the current limits around destroy-based handler cleanup, timer/idle cleanup boundaries, and the intentional perpetual launcher idle keepalive.
 - gtk_edge_cases_spec.rb now also verifies that Gtk.main_quit is blocked even before the shared loop starts.

OSXLich-Doug:
Clean run from CR, so ready for reviews.

OSXLich-Doug:
Clean run from CR, ready for review.

OSXLich-Doug:
Close no commit.

OSXLich-Doug:
@coderabbitai review

OSXLich-Doug:
@coderabbitai let's discuss docstring coverage metrics.  I do not know if the current ~ 37% measurement comes from a much older review or if it was touched up with this recent review.  But I find this delta to be disquieting.  Please explain what you're measuring so I can compare it to below and improve.

```Measured with YARD, scoped to the changed production files and with .yardopts disabled so it does not silently scan lib/**/*.rb:

Files:      4
Modules:   10 (4 undocumented)
Methods:   51 (3 undocumented)
88.52% documented
Undocumented objects in that scoped YARD pass:

Lich
Lich::Common
Lich::Common::Authentication
Lich::Common::GUI
Lich::Common::GtkIdleRetention#add
Lich::Common::GtkSignalHandlerRetention#signal_connect
Lich::Common::GtkTimeoutRetention#add```

OSXLich-Doug:
@coderabbitai so just to confirm my understanding, you rely on the repo's `.yardopts` configuration to generate your docstring coverage analysis.  

If accurate, it then follows that the repo `.yardopts` should be configured in such a way as to prevent your analysis from the whole repository's code base.

Or alternatively, I could provide you instruction similar to what I use locally.  My runs generally are `--no-yardopts --no-document --no-cache --no-save`.

I feel I should update the coderabbit configuration with that as an instruction set.  Feasible?  Appropriate?

Edited to include no yardopts config in local run.

OSXLich-Doug:
@coderabbitai perhaps I didn't clarify well.  The measurement I'm referring to is in the initial coderabbit feedback run - it shows the status (running, rate limted, no actionable comments), the review stack, the preflight (which is where this figure is reported) and so on.

How specifically do I get the preflight to come in somewhat directionally aligned with what I can generate locally, and which apparently you could also validate independently.  I just want it (more?) correct in preflight reporting.

OSXLich-Doug:
Superseded for merge purposes by #1385; kept open for branch history/review context.

OSXLich-Doug:
Close no merge - retain for documentation on CR interaction.

PR #1313, opened 2026-04-09
refactor(dr): replace EquipmentManager verb_data hardcoded patterns with DRCI constants

MahtraDR:
## Summary

Follow-up to [#1256](https://github.com/elanthia-online/lich-5/pull/1256). The `verb_data` method in EquipmentManager had its own hardcoded match patterns for remove/untie/get verbs that duplicated and diverged from the DRCI constants consolidated in #1256. When new game messages were added to DRCI (e.g., cold-enchanted item removal), `verb_data` didn't pick them up -- users had to manually patch patterns in two places.

### What changed

- **verb_data :worn** now splats `DRCI::REMOVE_ITEM_SUCCESS/FAILURE_PATTERNS` instead of hardcoding `'you tug'`, `'Remove what'`, `'slide themselves off'`, etc.
- **verb_data :tied** now splats `DRCI::UNTIE_ITEM_SUCCESS/FAILURE_PATTERNS` instead of hardcoding `'^What were you referring'`, `'^Untie what'`, etc.
- **verb_data :stowed** now splats `DRCI::GET_ITEM_SUCCESS/FAILURE_PATTERNS` instead of hardcoding `'^What were you referring'`, `'You are already'`, etc.
- **New DRCI patterns**: `/^A brisk chill leaves you as you/` in `REMOVE_ITEM_SUCCESS_PATTERNS`, `"too busy"` patterns in `UNTIE_ITEM_FAILURE_PATTERNS`
- **New exhausted constants**: `UNTIE_EXHAUSTED_PATTERNS` and `GET_EXHAUSTED_PATTERNS` separate "item not found" from recoverable failures, preventing case/when ordering bugs
- **return_held_gear**: now passes `failure_patterns: DRCI::WEAR_ITEM_FAILURE_PATTERNS` to `stow_helper`, matching the pattern used in `stow_weapon`

### Bug fixes found during review

- **:tied `*matches` splat bug**: The failure_recovery proc used `*matches` (wrapping the response in an Array), then matched with `case matches when [/regex/]`. Since `Array === Array` uses `==` (not element-wise `===`), the "too busy" recovery branches never triggered -- every failure silently fell through to `stow_weapon` instead of `retreat`/`stop_playing`. Fixed by accepting `response` as a single String parameter.
- **:tied `/^You remove/` missing anchor**: Without the `$` end anchor, every successful untie response ("You remove a sword from your belt") matched the failures entry, routing it to `stow_weapon` recovery instead of the success branch. Fixed with `/^You remove$/`.
- **Plain String patterns in failures/exhausted**: Several entries like `'You get'`, `'What were you referring'` used String comparison in `case/when`, which does exact equality (`==`) -- never matching game text with trailing words. Converted to Regexp for proper `===` pattern matching.

## Test plan

- [x] `rspec spec/lib/dragonrealms/commons/equipmanager_spec.rb` -- 186 examples, 0 failures
- [x] `rspec spec/lib/dragonrealms/commons/common_items_spec.rb` -- 239 examples (new: cold-enchanted remove, too-busy untie)
- [x] Full suite: `bundle exec rspec` -- 3807 examples, 0 failures
- [x] `rubocop -A` clean on all 4 files
- [ ] In-game: verify remove/untie/get/transform operations with gear sets
- [ ] In-game: verify return_held_gear correctly detects wear failures

### New spec coverage

- **Constant sync guards**: every entry in DRCI success/failure arrays must appear in verb_data matches
- **Exhausted subset checks**: UNTIE_EXHAUSTED_PATTERNS and GET_EXHAUSTED_PATTERNS verified as subsets of their DRCI parents
- **No hardcoded string duplicates**: specs assert no plain strings remain for patterns now covered by DRCI regexes
- **Failure/exhausted ordering safety**: no pattern appears in both arrays (would skip recovery)
- **Structural invariants**: all 4 types have required keys, callable recovery procs, and exhausted entries present in matches
- **All Regexp enforcement**: failures/exhausted arrays contain only Regexp (no String case/when traps)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of cold-enchanted item removal messages.
  * Recognize additional “too busy” untie failure responses and classify non-recoverable untie failures correctly.
  * Ensure wear/stow/return flows forward correct failure handling and verify item retrieval reliably (including better hand-identity checks and transform handling).

* **Tests**
  * Expanded test coverage for item removal, untying, stowing/returning gear, retrieval behavior, and equipment verb logic.


MahtraDR:
All three CodeRabbit findings addressed in 8eedd3c:

**Comment 1 (return_held_gear)**: Removed the hard-coded `true` after both `stow_helper` and `stow_by_type` calls so their return values propagate. If wearing fails, `return_held_gear` now returns `false` and `empty_hands` falls back to `DRCI.stow_hands`. Added spec: `'propagates stow_helper failure (returns false, not hard-coded true)'`.

**Comment 2 (non-exhausted GET failures)**: Expanded exhausted to cover ALL non-recoverable DRCI failure patterns so nothing falls through to the 5s hand-change timeout:
- `:worn` already used `DRCI::REMOVE_ITEM_FAILURE_PATTERNS` (no change needed)
- `:tied` `UNTIE_EXHAUSTED_PATTERNS` now includes `/^You don't seem to be able to move/` and `/^You fumble with the ties/` (only "too busy" excluded -- it has recovery in failures)
- `:stowed` and `:transform` now use `DRCI::GET_ITEM_FAILURE_PATTERNS` directly (removed `GET_EXHAUSTED_PATTERNS` constant entirely)
- Added a **coverage invariant spec**: for every verb type, every DRCI failure pattern in `matches` must be in either `exhausted` or `failures`. If someone adds a new pattern to DRCI later, this spec fails until they categorize it -- preventing silent fallthrough to the timeout branch.

**Comment 3 (transform failures)**: Added `/You'll need a free hand to do that!/` and `/You don't seem to be holding/` to transform's `matches` array so `bput` returns on them and the recovery proc can fire. Previously these would cause a bput timeout since they weren't in `matches`. Added spec: `'includes failure patterns in matches so bput returns on them'`.

Full suite: 3810 examples, 0 failures. rubocop clean.

MahtraDR:
**Comment 4 (transform recovery bput nil patterns)**: Fixed in 39c8e00.

`verb_data(item)[:matches]` is `nil` because the top-level hash is keyed by `:worn`/`:tied`/`:stowed`/`:transform` -- not `:matches`. The transform retry `bput` call was receiving `nil` patterns, causing a timeout instead of matching. Pre-existing bug.

Fixed to `*verb_data(item)[:transform][:matches]` (with splat for bput's variadic signature).

MahtraDR:
Re: [review 4082151025](https://github.com/elanthia-online/lich-5/pull/1313#pullrequestreview-4082151025) -- transform recovery `in_hands?` checks the pre-transform form:

**Acknowledged but not fixing here.** This is a pre-existing behavioral issue in untouched code, well outside the diff scope.

The problem is real: after the transform recovery proc re-runs the transform verb, `get_item_helper` checks `DRCI.in_hands?(transform_item)` -- but `transform_item` is the *source* form (e.g., orb), which no longer exists after a successful transform. The *target* form (e.g., armor) is in hand, so `in_hands?(orb)` returns false and `get_item_helper` reports failure despite success.

The correct fix requires `get_item_helper` to understand transform semantics -- either accepting an alternate item to check, or having the recovery proc return the transformed item. That's a design-level change to `get_item_helper`'s contract that belongs in a dedicated transform-handling PR, not a pattern consolidation refactor.

MahtraDR:
**Comment 5 (transform post-recovery in_hands? checks wrong item)**: Fixed at the source in 73ff82d.

Replaced `DRCI.in_hands?(item)` with `snapshot != [DRC.left_hand, DRC.right_hand]` in the post-recovery check. This is consistent with the success (`else`) branch which already uses the same snapshot-changed logic for all types. The previous `in_hands?` check failed for `:transform` because the item changes identity (orb -> armor), but snapshot-changed correctly detects success regardless of what the item became.

MahtraDR:
@coderabbitai review the last two commits and see last two comments

OSXLich-Doug (approved):
LGTM 👍

PR #1316, opened 2026-04-10
fix(all): use send(:respond) in StatusReporter fallback path

MahtraDR:
## Summary
- `StatusReporter.respond_mono` has a fallback branch for when `Lich::Messaging` is unavailable. The bare `respond` call can't resolve the top-level private method from a module class method context -- it raises `NoMethodError` instead.
- Changed to `send(:respond, text)` to correctly invoke the private top-level method.

## Context
The fallback path is rarely hit in practice (Lich::Messaging is loaded before any update code runs), but if it ever fires -- e.g. during early boot errors or test isolation -- it would crash instead of displaying the message.

## Test plan
- [ ] Full spec suite passes (3753 examples, 0 failures)
- [ ] Manually verify: if `Lich::Messaging` were undefined, `respond_mono("test")` would call top-level `respond` instead of raising

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fallback for status reporting so messages are correctly delivered when the primary messaging path is unavailable.
  * Prevents silent failures by ensuring the alternate response path is invoked reliably, resulting in more consistent and dependable status updates for users.


MahtraDR:
@coderabbitai please review in full again

OSXLich-Doug (approved):

PR #1317, opened 2026-04-10
fix(all): ensure SessionLifecycle.stop runs on abnormal exit

MahtraDR:
## Summary
- `SessionLifecycle.stop` was only called on the clean shutdown path in `main.rb`. If an exception occurred after `.start` (e.g. propagated via `abort_on_exception` on a child thread), the process would be left marked as running/stale in session reports.
- Wrapped the main event loop in `begin..ensure` so `.stop` always runs on both normal and abnormal exits.
- `SessionLifecycle.stop` is idempotent (guarded by `@started` flag), so calling it from the ensure block is safe even when the normal shutdown path already ran it.

## Test plan
- [ ] Full spec suite passes (3753 examples, 0 failures)
- [ ] Normal shutdown: verify session is marked as cleanly exited (same as before)
- [ ] Abnormal shutdown: kill a thread with `abort_on_exception = true` and verify session is still cleaned up

## Summary by CodeRabbit

* **Chores**
  * Improved shutdown reliability so the app cleans up consistently on both normal and abnormal exits.
  * Reduces hangs and ensures session/state persistence and resource release (sockets, saved settings) during exit.
  * Makes reconnection/quit behavior more predictable for end users.


MahtraDR:
@coderabbitai Good catch — resolved in 5fd212a.

`ActiveSessions::Lifecycle.stop` is now in the `ensure` block alongside `SessionLifecycle.stop`. Both are idempotent (guarded by `@started` under mutex), so the double-call on normal shutdown is a safe no-op.

OSXLich-Doug (approved):
Two C's (Claude and Codex) agree!

PR #1320, opened 2026-04-13
fix(all): Add in-memory TTL cache to FeatureFlags

MahtraDR:
## Summary

- Adds a mutex-protected in-memory cache with 60-second TTL to `FeatureFlags.enabled?`, eliminating repeated SQLite reads for values that rarely change
- On `.set`, the resolved value is written through to the cache (not just invalidated), closing a TOCTOU race where a concurrent read could re-cache a stale value
- DB errors and nil DB handles are intentionally not cached so the next call retries immediately
- A public `.clear_cache!` method is provided for operational/testing use

## Motivation

`FeatureFlags.enabled?` hits the SQLite database on every call. With the `ActiveSessions` heartbeat running every 5 seconds per process and N characters running, this creates unnecessary DB load for values that are set once and read thousands of times. A short TTL cache is a natural fit -- flags propagate within 60 seconds while DB reads drop by an order of magnitude.

## Implementation details

- `CACHE_TTL_SECONDS = 60` -- configurable constant
- `@cache_mutex` protects all cache reads/writes for thread safety
- `cache_read` returns `nil` for missing, expired, or negative-age (backward clock) entries
- `cache_write` stores `{ value: <bool>, at: <float> }` keyed by flag name
- On successful `.set`, the resolved boolean is written through to the cache
- Failed `.set` (DB error) does **not** update the cache
- When `fetch_db` returns nil (runtime not yet initialized), the default is returned without caching
- Each Lich process maintains its own independent cache; cross-process convergence is bounded by the TTL

## Test plan

- [x] First call hits DB, second call within TTL serves from cache
- [x] After TTL expires, next call re-reads from DB
- [x] Different flag names have independent cache entries
- [x] Cache stores resolved boolean, not raw DB string
- [x] nil DB result (unpersisted flag) is cached as default value
- [x] DB unavailable (nil handle) does NOT cache -- next call retries
- [x] DB error does NOT cache fallback (next call retries)
- [x] `.set` writes the resolved value through to cache
- [x] Failed `.set` does NOT update cache
- [x] `.clear_cache!` forces DB reads; safe when empty; safe under concurrent calls
- [x] Thread safety: concurrent `enabled?` returns valid results
- [x] Thread safety: concurrent `enabled?` + `set` -- no deadlocks
- [x] Thread safety: no starvation under mutex contention (timeout guard)
- [x] TTL boundary: entry at exactly TTL is still valid (strict >)
- [x] TTL boundary: entry just past TTL is expired
- [x] Negative TTL: every read goes to DB, no crash
- [x] Clock moving backward: negative age treated as expired
- [x] Stale value served until TTL, then fresh value returned
- [x] Rapid set/enabled? interleaving: write-through serves correct value
- [x] DB goes down after cache warm: cached values served until TTL, then fallback

All 41 specs pass (18 existing + 23 new).

## Summary by CodeRabbit

* **New Features**
  * Feature flags now use an in-memory cache with expiration for faster checks and reduced backend reads.
  * Cache updates when flags are changed, and a manual cache-clear capability was added.

* **Reliability**
  * Improved behavior when backend is unavailable and safer handling around cache expiry and clock edge cases.

* **Tests**
  * Extensive tests added for caching semantics, TTL edge cases, concurrency, and error scenarios.

### DIFF
--- a/lib/common/feature_flags.rb
+++ b/lib/common/feature_flags.rb
@@ -24,6 +24,16 @@ module Lich
       # persisted value in `lich_settings` always overrides the default.
       DEFAULTS = {}.freeze
 
+      # Cached flag values are considered fresh for this many seconds.
+      # Feature flags change infrequently; this avoids repeated DB reads
+      # on every heartbeat cycle.
+      #
+      # @return [Integer]
+      CACHE_TTL_SECONDS = 60
+
+      @cache = {}
+      @cache_mutex = Mutex.new
+
       # Returns whether a feature flag is enabled.
       #
       # @param name [String, Symbol] the feature flag name
@@ -33,11 +43,14 @@ module Lich
       #   characters
       def self.enabled?(name)
         flag_name = validate_flag_name!(normalize_name(name))
+        cached = cache_read(flag_name)
+        return cached unless cached.nil?
+
         begin
           stored = read_flag(flag_name)
-          return default_for(flag_name) if stored.nil?
-
-          truthy?(stored)
+          resolved = stored.nil? ? default_for(flag_name) : truthy?(stored)
+          cache_write(flag_name, resolved)
+          resolved
         rescue StandardError => e
           log_failure('read', flag_name, e)
           default_for(flag_name)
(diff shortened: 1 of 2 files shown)


MahtraDR:
Re: @coderabbitai nitpick on the `Concurrent` detection pattern (lines 286-287) -- applied in ee51b7e. Good catch, the ternary is cleaner.

MahtraDR:
Closing as recommended by Claude. Once #1325 is merged this is moot.

PR #1322, opened 2026-04-15
feat(dr): Support matching multiple times in Flags and keep count of the number of matches

vtcifer:
## Summary by CodeRabbit

* **New Features**
  * Added per-flag match counting so you can monitor how often each flag is triggered.
  * You can now retrieve counts for individual flags or a snapshot of all flag counts.
  * Flag resets no longer clear these counts — counts persist independently of flag reset operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/elanthia-online/lich-5/pull/1322?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/lib/dragonrealms/drinfomon/drparser.rb
+++ b/lib/dragonrealms/drinfomon/drparser.rb
@@ -76,6 +76,7 @@ module Lich
           regexes.each do |regex|
             if (matches = server_string.match(regex))
               Flags.flags[key] = matches
+              Flags.counts[key] = Flags.count(key) + 1
               break
             end
           end
--- a/lib/dragonrealms/drinfomon/drparser.rb
+++ b/lib/dragonrealms/drinfomon/drparser.rb
@@ -76,7 +76,7 @@ module Lich
           regexes.each do |regex|
             if (matches = server_string.match(regex))
               Flags.flags[key] = matches
-              Flags.counts[key] = Flags.count(key) + 1
+              Flags.increment_count(key)
               break
             end
           end
--- a/lib/dragonrealms/drinfomon/events.rb
+++ b/lib/dragonrealms/drinfomon/events.rb
@@ -5,6 +5,7 @@ module Lich
     class Flags
       @@flags = {}
       @@matchers = {}
+      @@counts = {}
 
       def self.[](key)
         @@flags[key]
@@ -16,16 +17,19 @@ module Lich
 
       def self.add(key, *matchers)
         @@flags[key] = false
+        @@counts[key] = 0
         @@matchers[key] = matchers.map { |item| item.is_a?(Regexp) ? item : /#{item}/i }
       end
 
       def self.reset(key)
+        @@counts[key] = 0
         @@flags[key] = false
       end
 
       def self.delete(key)
         @@matchers.delete key
-        @@flags.delete key
+        @@counts.delete key
+        @@flags.delete key    
       end
 
       def self.flags
@@ -34,6 +38,14 @@ module Lich
 
       def self.matchers
         @@matchers
+      end
+
+      def self.count(key)
+        @@counts[key] || 0
+      end
+
+      def self.counts
+        @@counts
       end
     end
   end
--- a/lib/dragonrealms/drinfomon/events.rb
+++ b/lib/dragonrealms/drinfomon/events.rb
@@ -29,7 +29,7 @@ module Lich
       def self.delete(key)
         @@matchers.delete key
         @@counts.delete key
-        @@flags.delete key    
+        @@flags.delete key
       end
 
       def self.flags
--- a/lib/dragonrealms/drinfomon/events.rb
+++ b/lib/dragonrealms/drinfomon/events.rb
@@ -22,7 +22,6 @@ module Lich
       end
 
       def self.reset(key)
-        @@counts[key] = 0
         @@flags[key] = false
       end
 
@@ -45,7 +44,11 @@ module Lich
       end
 
       def self.counts
-        @@counts
+        @@counts.dup.freeze
+      end
+
+      def self.increment_count(key)
+        @@counts[key] = (@@counts[key] || 0) + 1
       end
     end
   end


MahtraDR:
Claude and I had a chat about this PR just now, and I have two big concerns (Claude has some smaller ones in addition)

1. Race condition: read-modify-write on @@counts is not atomic
[drparser.rb:79](vscode-webview://0jhipd6ugodasrh7h6j9919vp7h4cqq4fqt247lgk434v24vff3f/lib/dragonrealms/drinfomon/drparser.rb#L79):


Flags.counts[key] = Flags.count(key) + 1
This is a classic TOCTOU race. Flags.count(key) reads @@counts[key], then the caller adds 1, then writes back. check_events is called from process_game_specific_data in the game connection thread, but scripts running in separate threads also call Flags.reset (which zeros the count). Between the read and the write, a script thread could reset the count to 0, and this line would silently overwrite it with the stale pre-reset value + 1. The existing @@flags[key] = matches assignment on line 78 doesn't have this problem because it's a simple assignment, not a read-modify-write.

Fix: At minimum, use @@counts[key] = (@@counts[key] || 0) + 1 inside a method on Flags itself (e.g., increment_count) so the read and write are at least in the same method and could be wrapped in a mutex later.

2. reset zeroes the count -- semantic mismatch with how reset is used
Every consumer of Flags.reset in the codebase uses it to mean "I've consumed this flag, re-arm it for the next match." For example, [common-items.rb:2018](vscode-webview://0jhipd6ugodasrh7h6j9919vp7h4cqq4fqt247lgk434v24vff3f/lib/dragonrealms/commons/common-items.rb#L2018):


Flags.reset("pouch-full")
This is called in a loop -- match, consume, reset, wait for next match. The PR makes reset also zero the count. This means the count is only useful between add and the first reset, which in most real usage patterns is a single iteration of a loop. The count will perpetually be 0 or 1 for any flag that follows the standard consume-reset pattern. This makes the feature nearly useless for its stated purpose of "tracking match frequency."

If the intent is to count total matches over the flag's lifetime, reset should NOT zero the count. If the intent is to count matches since the last reset, it should be documented, because that's a different (and much less useful) semantic.

```markdown
## Race condition in `Flags.counts` increment

### Threading model

Lich runs two categories of threads that interact with `Flags`:

1. **Game server thread** (`games.rb:337`) -- priority 4, reads lines from the socket, calls `DRParser.check_events` on every line. This is the **writer**.
2. **Script threads** -- each user script runs in its own thread at default priority. These **consume** flags via `Flags[key]` and re-arm them via `Flags.reset(key)`.

There is no mutex or synchronization anywhere in `Flags`. CRuby's GIL prevents truly parallel execution but does NOT prevent interleaving -- the GIL releases between Ruby method calls, and the scheduler checks for yield points every ~10ms.

### The vulnerable code

Game thread (`drparser.rb:78-79`):
```ruby
Flags.flags[key] = matches                  # simple assignment -- safe
Flags.counts[key] = Flags.count(key) + 1     # read-modify-write -- NOT safe
```

Script thread (`events.rb`, called from e.g. `common-items.rb:2018`):
```ruby
def self.reset(key)
  @@counts[key] = 0
  @@flags[key] = false
end
```

### Interleaving 1: reset is silently overwritten

Assume `@@counts["pouch-full"]` is `5`. A script thread calls `reset` while the game thread processes a new match.

| Step | Thread | Operation | `@@counts` value |
|------|--------|-----------|-----------------|
| 1 | Game | `Flags.count(key)` reads `5` | 5 |
| 2 | Game | Computes `5 + 1 = 6`, holds result | 5 |
| -- | *GIL yields* | | |
| 3 | Script | `reset` sets `@@counts[key] = 0` | **0** |
| 4 | Script | Sets `@@flags[key] = false` | 0 |
| -- | *GIL yields back* | | |
| 5 | Game | `Flags.counts[key] = 6` writes stale value | **6** |

**Result:** The reset is silently clobbered. Count jumps to 6 instead of being 1.

### Interleaving 2: a match is lost from the count

| Step | Thread | Operation | `@@counts` value |
|------|--------|-----------|-----------------|
| 1 | Script | Enters `reset`, about to write | 5 |
| -- | *GIL yields* | | |
| 2 | Game | `Flags.count(key)` reads `5` | 5 |
| 3 | Game | `Flags.counts[key] = 6` | **6** |
| -- | *GIL yields back* | | |
| 4 | Script | `@@counts[key] = 0` overwrites the new match | **0** |

**Result:** The match that just happened is lost. `@@flags[key]` holds MatchData but count says 0 -- flag and count are out of sync.

### Why the existing `Flags.flags[key] = matches` doesn't have this problem

Line 78 is a simple assignment -- it doesn't read the old value. Last-writer-wins is correct for a "did this match?" flag. The count increment is fundamentally different: `new = old + 1` requires `old` to still be valid when `new` is written. Without a mutex or atomic increment, that invariant can be violated.

### Why this will happen in practice

The standard Flags usage pattern is a tight consume-reset loop:

```ruby
# common-items.rb:2017-2018
if Flags["pouch-full"]
  Flags.reset("pouch-full")
  # ...
end
```

The game thread pumps every line of server output through `check_events`. Script threads poll and reset in loops. The window between `Flags.count(key)` and `Flags.counts[key] = ...` is exactly where the script thread's `reset` lands when both are active on the same flag.
```

MahtraDR:
I did an independent review of this PR and I think we need a design conversation before we iterate on the code.

The core tension is that `reset` zeroes the count, but nearly every consumer of Flags in dr-scripts follows a tight consume-reset loop -- `if Flags["thing"]` then `Flags.reset("thing")`. That means the count will almost always be 0 or 1 in practice, which makes the feature a no-op for its stated purpose of tracking match frequency. If the intent is lifetime counts, `reset` shouldn't touch them. If the intent is counts-since-last-reset, that's a different (and much narrower) tool, and it should be documented as such.

On top of that, the increment in drparser is a read-modify-write (`Flags.counts[key] = Flags.count(key) + 1`) that's vulnerable to interleaving with script threads calling `reset`. It's not a simple assignment like the existing `Flags.flags[key] = matches` -- it reads the old value, adds one, and writes back, so a `reset` landing between the read and write gets silently clobbered. The `counts` accessor also hands back the live mutable hash, which makes it impossible to add synchronization later without hunting down every direct write.

But stepping back further -- I wonder if this is a good time to think about what Flags should actually be. Right now it's a DR-specific class living under `Lich::DragonRealms`, but it's really just a pattern-matching event system -- there's nothing inherently DR-specific about "register a regex, fire when it matches server output, let scripts poll the result." If we're going to extend the API with counting, should we first pull Flags up into a game-agnostic layer so GS scripts can use the same infrastructure? And if we're rethinking the API anyway, maybe we should address the known issue where `Flags[key]` stores MatchData but scripts compare it with `== String` (which is always false) -- that's been a quiet source of bugs for a while.

@vtcifer, what's the concrete scenario you're trying to enable with match counting? Knowing the use case would help us figure out whether this is the right shape for the API or whether we should design something broader.

vtcifer:
Discussed with @MahtraDR on discord, proposed changes put in.

MahtraDR:
@coderabbitai review this pr in full now

PR #1324, opened 2026-04-18
fix(gs): Claim corrections to handle mounts in room_players stream

mrhoribu:
## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate detection of arriving players by ignoring non-arrival identifiers and avoiding duplicates.
  * Prevented creation of spurious player records; only genuine arrivals are recorded.
  * More reliable clearing and updating of player title/status fields and guarded handling of status text to avoid incorrect merges.


OSXLich-Doug (approved):
Tests clean in integration, ready for prime time.

PR #1325, opened 2026-04-19
fix(all): address ActiveSessions lifecycle hot path feature-flag polling

OSXLich-Doug:
## Why
- The original Active Sessions API implementation introduced repeated nested checks of `FeatureFlags.enabled?(:active_sessions_api)` on a heartbeat-driven path.
- In multi-session usage, that design amplified SQLite-backed feature-flag reads until lock pressure became visible as `SQLite3::BusyException` warnings.
- This change addresses the root cause in the Active Sessions lifecycle path rather than only masking the resulting symptoms.

## What Changed
- Gate `active_sessions_api` once at `Lifecycle.start`.
- Latch that admitted state in the lifecycle instance for the rest of the session.
- Add an internal `assume_enabled:` path in `ActiveSessions` so lifecycle hot paths do not re-read the same feature flag on heartbeat, listener update, or teardown.
- Preserve local listener state independent of transient feature-flag failures.
- Keep the normal public feature-gate behavior intact for callers that do not use the internal admitted path.

## Scope
- `lib/internal_api/active_sessions.rb`
- `lib/internal_api/active_sessions/lifecycle.rb`
- `spec/lib/internal_api/active_sessions_spec.rb`
- `spec/lib/internal_api/active_sessions/lifecycle_spec.rb`

## Validation
- `rspec` clean
- `rubocop` clean

## Notes
- This PR is intentionally narrow.
- It is meant to separate root-cause cleanup from follow-on symptom mitigation such as SQLite timeout/WAL tuning, caching, and listener-state hardening.

## Summary by CodeRabbit

* **New Features**
  * Added admission-specific session entry points to manage already-admitted sessions without triggering service bootstrap.

* **Improvements**
  * Lifecycle captures the feature-flag state at start to avoid repeated checks during runtime; stop/cleanup use that latched state.
  * Listener and heartbeat flows are gated by lifecycle start state.
  * Cleanup now queries live service state and avoids attempting to bootstrap a service.

* **Tests**
  * Expanded tests covering admission paths, bootstrap restrictions, and cleanup behavior under feature-flag changes.


OSXLich-Doug:
@coderabbitai - review

OSXLich-Doug:
@coderabbitai - review the latest submission for closure on feedback

MahtraDR (approved):

PR #1340, opened 2026-05-01
fix(all): enable heartbeat-driven failover for ActiveSessions service owner

MahtraDR:
## Summary

- When the service-owning Lich process exits while other sessions remain, the TCP service dies with no failover. Surviving sessions' heartbeats use `allow_bootstrap: false`, so they cannot elect a new owner.
- External consumers (e.g. the launcher) see `:no_service` and bypass all session safety checks (same-account conflict detection, existing session reuse), potentially starting duplicate logins.
- Changes `register_session_admitted` to `allow_bootstrap: true` so surviving sessions can race to become the new owner on their next heartbeat tick (~5s).
- The kill-switch is preserved: `ensure_service_internal!` independently re-checks `enabled?` before bootstrapping a new owner (line 98), so disabling the feature flag still prevents new owners from being created.
- Leaves `unregister_session_admitted` as `allow_bootstrap: false` -- there is no value in bootstrapping a server just to unregister during shutdown.

### Port-binding race safety

When multiple sessions race to become owner, exactly one wins the `TCPServer.new` bind on port 42857. Losers get `Errno::EADDRINUSE`, caught by `Server#start`'s rescue, silently fail that heartbeat tick, and discover the winner's service via the discovery file on their next tick.

## Test plan

- [ ] Start multiple Lich sessions with `--detachable-client=0`
- [ ] Identify which PID owns the service (`cat temp/lich-active-sessions.json`)
- [ ] Kill the owner process
- [ ] Verify a surviving session bootstraps a new service within ~5s (check discovery file for new PID)
- [ ] Verify the launcher correctly detects existing sessions after failover
- [ ] Verify disabling the feature flag still prevents new owner creation (kill-switch preserved)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience of the active sessions system: when no healthy owner is available, the system can now bootstrap a replacement server while still honoring the kill-switch/feature flag safeguards.

* **Tests**
  * Added specs covering the bootstrapping behavior and the negative case when the kill-switch/feature flag is disabled.

### DIFF
--- a/lib/internal_api/active_sessions.rb
+++ b/lib/internal_api/active_sessions.rb
@@ -141,10 +141,15 @@ module Lich
       # Registers or updates a session record from a lifecycle path that
       # already admitted the feature gate at startup.
       #
+      # When the current service owner has exited, this allows a surviving
+      # session to bootstrap a replacement owner so session visibility is
+      # maintained. The kill-switch is still enforced: ensure_service_internal!
+      # re-checks enabled? before creating a new owner.
+      #
       # @param payload [Hash] normalized session metadata
       # @return [Boolean] true when the service accepted the update
       def self.register_session_admitted(payload)
-        return false unless ensure_service_internal!(allow_bootstrap: false)
+        return false unless ensure_service_internal!(allow_bootstrap: true)
 
         service_client&.upsert(payload)&.fetch(:ok, false) || false
       end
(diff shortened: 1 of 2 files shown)


OSXLich-Doug (approved):
This submission closes concerns raised and is ready to deploy.  Approving without final CR review under break/fix.

OSXLich-Doug (approved):
All concerns addressed.  Looks good.

PR #1341, opened 2026-05-02
fix(gs): add support for RESERVE xml in GameObj.reserve

mrhoribu:
## Summary by CodeRabbit

* **New Features**
  * Introduced a new reserve registry system for game objects, enabling creation, access, and management of reserved items.
  * Reserve objects are now fully integrated into game object lookups and remain protected from index cleanup.
  * Game stream parser now automatically handles reserve item population and clearing during gameplay.


OSXLich-Doug (approved):
Looks good, one small note for future dev:

Add a parser-level regression that feeds a reserve stream and asserts `GameObj.reserve` receives the parsed object.

PR #1343, opened 2026-05-02
fix(dr): DRC text2num correction for thousand

BinuDR:
Added option for 'thousand' to text2num method because you can stack blank scrolls up to 1000 and it was throwing an error.

## Summary by CodeRabbit

* **New Features**
  * Text-to-number conversion now recognizes the word "thousand" in input text, enabling accurate numerical conversion for larger values expressed in text format.


BinuDR:
This could have also been fixed by adding 100 and 1000 to the NUM_MAP in drvariables.rb, but since hundred was already done here I just added thousand here as well.

MahtraDR (approved):

mrhoribu (approved):

PR #1345, opened 2026-05-04
fix(all): copy Gemfile and LICENSE during self-update

mrhoribu:
The self-updater only copied `lib/` and `lich.rbw` from release archives, silently skipping `Gemfile` and `LICENSE`. This meant those files could go stale or be missing entirely after an update.

**Changes:**
- Add `TOP_LEVEL_FILES` constant listing non-lib, non-lich.rbw release files (`Gemfile`, `LICENSE`)
- Iterate and copy those files to `LICH_DIR` during `perform_update`
- Include `TOP_LEVEL_FILES` in `validate_lich_structure` so releases missing them are rejected
- Adding future top-level release files is now a single-line change to the constant

### DIFF
--- a/lib/common/update/release_installer.rb
+++ b/lib/common/update/release_installer.rb
@@ -12,6 +12,10 @@ module Lich
   module Util
     module Update
       class ReleaseInstaller
+        # Top-level files (besides lib/ and lich.rbw) to copy from release archive.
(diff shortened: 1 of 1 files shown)


OSXLich-Doug (approved):
Good add.

PR #1355, opened 2026-05-17
fix(all): recover from cross-process zombie in ActiveSessions service

MahtraDR:
## Summary

When the Active Sessions server owner's accept thread dies, the owner should self-heal via its heartbeat thread, which calls `ensure_service_internal!` every 2 seconds to detect and restart the dead thread. But the heartbeat's `rescue StandardError` was **outside** the loop body -- a single exception from `upsert_current_session` killed the heartbeat thread permanently. Without the heartbeat, the owner lost its only self-healing path, the accept thread stayed dead, and the process held port 42857 as a permanent zombie that no other process could reclaim (`EADDRINUSE`).

### Changes

**Primary fix -- heartbeat resilience (lifecycle.rb):**
- Move `rescue StandardError` from outside the heartbeat loop to inside the loop body. Individual tick failures are now logged (with backtrace) and retried on the next interval instead of killing the thread.
- Reduce heartbeat interval from 5s to 2s. The zombie detection window (time between accept thread death and owner self-recovery) is now 2 seconds, reducing the window where peers see an unresponsive service.

**Defense-in-depth -- cross-process zombie detection (active_sessions.rb):**
- When `ensure_service_internal!` finds the service unavailable, check the discovery file's `owner_pid` liveness via `Process.kill(0, pid)`:
  - Owner dead: delete stale discovery, bootstrap normally.
  - Owner alive but service unresponsive: delete stale discovery, return `false` immediately -- no bind attempt, no EADDRINUSE, no log spam.
- This is **noise reduction and stale-state cleanup**, not recovery. Only the owner process can recover its own bound port via its heartbeat. Peers cannot reclaim a port held by a living process.
- `service_available?` recheck after loading discovery prevents a race where a newly-bootstrapped owner could be mistakenly flagged as a zombie.
- Discovery deletion uses compare-before-delete (`delete_discovery_if_owner`) to avoid racing with another process that wrote a valid discovery between our read and our delete.

**Noise reduction (active_sessions.rb):**
- Clear `@server` after failed `start` so the next heartbeat tick doesn't misidentify a never-started Server as an in-process zombie, eliminating the misleading "server thread died" message when the real issue is a port conflict.

**Diagnostic instrumentation (server.rb):**
- Log accept thread start, every exit path in `accept_loop` (nil server, IOError/EBADF shutdown, fatal exception with backtrace), and thread exit via `ensure`. Permanent -- captures the root cause if the accept thread dies again.

## Test plan

- [x] All 53 specs pass (active_sessions_spec, server_spec, server_client_spec, lifecycle_spec)
- [x] Rubocop clean on all modified files
- [ ] Start multiple Lich sessions, kill the owner process, verify failover within ~2s
- [ ] Verify heartbeat survives transient errors (spec covers this)
- [ ] Verify owner self-recovery when accept thread dies (spec covers this)
- [ ] Verify cross-process zombie detection logs warning and avoids EADDRINUSE spam (spec covers this)
- [ ] Verify peers do not thrash with bind attempts while zombie owner is alive (spec covers this)
- [ ] Verify the launcher can query the service after failover


MahtraDR:
@coderabbitai Good catch on the race window -- addressed in 924f2386. Added `return true if service_available?` recheck after loading discovery, before the liveness probe. Also fixed the `Lint/RescueException` rubocop offense in the same commit.

MahtraDR:
@coderabbitai review this PR one more time to account for https://github.com/elanthia-online/lich-5/pull/1355/changes/7e5a5fdc633783ac522b274f335af51795e66c06

OSXLich-Doug:
@coderabbitai - update review on this pr to include all commits.

OSXLich-Doug (approved):

PR #1359, opened 2026-05-22
fix(all): Improve shutdown session state handling

OSXLich-Doug:
This change requires PR #1354 

## Summary
- preserve ActiveSessions registry presence while marking sessions disconnected at shutdown start
- keep unregister late, with guarded shutdown phases and buffered slow/failure debug tracing
- clean up expected ActiveSessions accept-loop stop logging so intentional shutdown does not report IOError/warnings

## Tests
- ruby -c lib/main/main.rb
- ruby -c lib/internal_api/active_sessions/lifecycle.rb
- ruby -c lib/internal_api/active_sessions/server.rb
- rspec spec/lib/internal_api/active_sessions/lifecycle_spec.rb spec/lib/internal_api/active_sessions/server_spec.rb
- git diff --check HEAD~1..HEAD

## Summary by CodeRabbit

* **Chores**
  * Improved session lifecycle management with explicit connection state tracking
  * Enhanced shutdown sequence with better state management during client disconnection
  * Refined offline mode behavior handling

[Review Change Stack](https://app.coderabbit.ai/change-stack/elanthia-online/lich-5/pull/1359?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


OSXLich-Doug:
@coderabbitai - review

OSXLich-Doug:
Item | Valid? | Action? | Reason
-- | -- | -- | --
stopping? reads @stopping without mutex | Mostly valid | Accept | The next commit will address this low-cost consistency fix. The practical risk is tiny on MRI, but @stopping is written under @mutex, so reading it under the same mutex is cleaner and easy.
Reconnect hook comment is misleading | Valid | Accept | The next commit will address this.  The call is active. The comment says “taking this out of play,” which is false/confusing. Remove or replace it.
next nil semantics | Partly valid, overstated | Reject action | It is an intentional fix. The old nil did not fall through into elsif, but it did allow the thread to continue after the conditional. next nil clearly exits the thread block. A comment is optional, not necessary.
SessionLifecycle lacks update_connected(false) | Not valid for this PR | Reject | Lich::Common::SessionLifecycle does not expose a connected-state contract. It tracks/registers/stops session summary state. The connected API belongs to Lich::InternalAPI::ActiveSessions::Lifecycle, which #1359 updates intentionally.
Duplicate session_name / session_role vars | Mild readability concern | Reject | True that names are rebound, but the two lifecycle systems have distinct resolvers and start points. Renaming locals would be harmless, but it is cleanup, not needed for this shutdown fix.

mrhoribu (approved):
good stuff

PR #1368, opened 2026-05-29
fix(all): Script#kill remove GC.start and add optional instrumentation

OSXLich-Doug:
`Script#kill` has explicitly triggered `GC.start` throughout Lich 5. Recent runtime and lifecycle work made this path more timing-sensitive, and that exposed problems around script teardown and shutdown behavior.

We tried moving cleanup into the GTK queue and briefly routed it through `MemoryReleaser`, but that proved too heavy for rapid script exits. We then returned to a direct `GC.start` call while the immediate blocker was addressed separately with SQLite busy timeouts.

After revisiting the inherited design, we agreed that script teardown should not be responsible for forcing process-wide garbage collection. The cost of a major GC scales with the whole Ruby heap, not just the script being stopped, so doing that work inside `Script#kill` couples script lifecycle to global memory policy.

This PR removes the forced GC from `Script#kill` and adds opt-in aggregate instrumentation for non-shutdown script stops. The instrumentation is disabled by default, but gives us a way to measure script-stop volume and teardown duration if this area becomes suspicious again.

## Summary by CodeRabbit

* **New Features**
  * Added opt-in, per-minute aggregated runtime diagnostics for non-shutdown script kills (avg/max duration and failure counts).
* **Bug Fixes**
  * Consolidated cleanup into a single path, always run cleanup and log errors; fall back to inline cleanup with a warning if background thread creation fails.
* **Tests**
  * Added deterministic tests for metric aggregation, minute-rollover logging, failure reporting, and feature-flag behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/elanthia-online/lich-5/pull/1368?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/lib/common/script.rb
+++ b/lib/common/script.rb
@@ -300,9 +300,29 @@ module Lich
         end
       }
       @@running = Array.new
+      @@kill_metrics_mutex = Mutex.new
+      @@kill_metrics = {
+        :minute            => nil,
+        :runtime_stops     => 0,
+        :duration_total_ms => 0.0,
+        :duration_max_ms   => 0.0,
+        :failures          => 0
+      }
 
       attr_reader :name, :vars, :safe, :file_name, :label_order, :at_exit_procs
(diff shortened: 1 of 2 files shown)


mrhoribu:
@coderabbitai remove all previous comments and review in full

MahtraDR:
@coderabbitai where is that full review of this PR?

OSXLich-Doug:
@coderabbitai remove all previous comments and review in full

MahtraDR:
@coderabbitai remove all previous comments and review in full

mrhoribu (approved):

PR #1381, opened 2026-06-02
ci(all): implement custom/ascii only cop rubocop 

OSXLich-Doug:
Adds a project-local RuboCop cop, Custom/AsciiOnlySource, to detect non-ASCII source characters and escape sequences that produce non-ASCII runtime text.

The cop is enabled with temporary exclusions for the current known baseline offenses so CI can remain green while a follow-up cleanup PR classifies and resolves each occurrence. It supports explicit autocorrection for a small set of known display glyphs, but normal RuboCop runs only report offenses.

## Summary by CodeRabbit

* **Chores**
  * Added a new code quality check to enforce ASCII-only source code. The check detects non-ASCII characters and escape sequence issues in source files, with selected paths excluded to accommodate dependencies and configuration requirements.


mrhoribu:
@coderabbitai review

PR #1382, opened 2026-06-02
fix(all): rubocop custom cop ascii only cleanup

OSXLich-Doug:
## Summary
Classifies and resolves the current Custom/AsciiOnlySource baseline so the cop can run without exclusions.

This keeps intentional Unicode where it is purposeful and supported:

GTK display glyphs are retained with scoped cop disables.
Unicode password/credential fixtures are retained with scoped cop disables.
CLI status output is converted to plain ASCII text.
Also updates the 0x92 apostrophe cleanup in lib/games.rb to use byte-level detection/replacement and adds a focused spec for that repair path.

Validation:

rbenv exec bundle exec rubocop --cache false --only Custom/AsciiOnlySource
rbenv exec bundle exec rspec

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of poorly encoded characters in data processing.

* **Chores**
  * Updated CLI message formatting for encryption mode transitions (replaced checkmarks with "ok:" prefix and standardized warning text format).
  * Enhanced code quality tooling to enforce ASCII-only source code with exceptions for UI elements and tests requiring Unicode support.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -80,14 +80,3 @@ Style/AsciiComments:
 
 Custom/AsciiOnlySource:
   Enabled: true
-  Exclude:
-    - 'lib/common/cli/cli_encryption_mode_change.rb'
-    - 'lib/common/gui/account_manager_ui.rb'
-    - 'lib/common/gui/manual_login_tab.rb'
-    - 'lib/common/gui/master_password_prompt_ui.rb'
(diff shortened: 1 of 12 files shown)


OSXLich-Doug:
@coderabbitai review

mrhoribu (approved):

PR #1385, opened 2026-06-03
feat(all): Add modern script command shutdown method

OSXLich-Doug:
## Shutdown Train Bonus Round

This PR adds a script-safe way to request orderly Lich teardown without sending exit through the game/front-end I/O path which generates EOF teardown.

Core shape:
- Adds lich_shutdown as the approved script-facing helper.
- Routes script shutdown through Lich::Common::OrderlyShutdown.request_user_exit.
- Reuses the same orderly user-exit path as frontend shutdown.
- Excludes the calling script from the script drain so it does not kill itself mid-request.
- Preserves the shutdown modernization behavior: drain scripts, save Vars, close Game, record summary logging.
- Refactors frontend user-exit handling to call the shared helper rather than duplicating setup.
- Adds focused spec coverage, including cleanup for the shutdown summary buffer so specs remain order-independent.

## Summary by CodeRabbit

* **New Features**
  * Unified shutdown flow for clean user exits and connection-loss handling
  * GUI-safe GTK shutdown helpers and hardened GTK lifecycle behavior
  * Detection and safe abort on irrecoverable game stream desync

* **Bug Fixes**
  * More robust socket read/timeout and EOF handling
  * Improved detection/reporting of fatal client write failures and connection disruptions

* **Documentation**
  * GTK hardening guidance added/updated

* **Tests**
  * Extensive test coverage for shutdown, GTK hardening, and game stream behaviors


OSXLich-Doug:
@coderabbitai review

OSXLich-Doug:
@coderabbitai Thanks, this is a fair catch on the duplicated predicate methods. I agree those can be extracted into a tiny shared helper/mixin so completed?, failed?, scripts_drained?, and vars_saved? have one implementation.

I would not merge the two Result structs, though. OrderlyShutdown::Result intentionally includes game_closed, while BestEffortShutdownCleanup::Result intentionally does not because that path avoids managing the game socket directly.

I would also avoid extracting the larger runner/helper flow. The two runners share some mechanics, but their semantics, coordinator entry points, completion criteria, and log language differ in important ways. Keeping those flows explicit is preferable here; only the identical predicate boilerplate should be shared.

mrhoribu (approved):
working great for me, only suggestion is changing from a feat to refactor perhaps?

OSXLich-Doug:
@coderabbitai review

PR #1389, opened 2026-06-10
feat(main): add --pipe mode using stdin/stdout as client transport

sproctor:
## What

Adds a `--pipe` launch flag that uses **stdin/stdout** as the front-end client transport instead of a TCP socket, so Lich's parse/hook pipeline can be driven over pipes.

- **stdin** carries what a front-end would send: the login key, version string, then commands (→ upstream hooks → server)
- processed downstream server output is written to **stdout**
- **EOF on stdin** marks the client gone and triggers the normal shutdown path

## Usage

```
ruby lich.rbw -g HOST:PORT --pipe --no-gtk
```

`-g HOST:PORT` connects straight to the game server (no SGE/eaccess). Pair with `--no-gtk` for headless use. It also works via the normal launch-data / `--login` / `--sal` path.

## How it works

- **`PipeIO`** (`lib/common/pipe_io.rb`): a duplex stdin/stdout adapter wrapped in the existing `SynchronizedSocket`. It reports `closed?` once stdin hits EOF, so `SynchronizedSocket#alive?` becomes EOF-based with **no change to that class** — the server read loop and client thread stop cleanly on EOF.
- **`main.rb`**: a `--pipe` branch in the `$_CLIENT_` setup (`Frontend.client = 'unknown'`, `$_CLIENT_ = SynchronizedSocket.new(PipeIO.new)`) on both the launch-data path and the `-g HOST:PORT` path, plus a guard so the front-end listener step is skipped when no listener was opened.
- `Frontend.client = 'unknown'` gives raw passthrough; the client thread reads the login key + version from stdin first, then enters the normal read loop.

## Tests

`spec/lib/common/pipe_io_spec.rb` covers gets/write/puts, EOF-based `closed?`, and the EOF → `alive?` transition when wrapped in `SynchronizedSocket`. Full suite green; rubocop clean.

Verified end-to-end with a local loopback harness (fake server + `-g 127.0.0.1:PORT --pipe`): the upstream login key reaches the server and downstream server XML flows through to stdout.

> Note: this branch also includes an unrelated one-line `.gitignore` addition for `.idea`.


sproctor:
My motivation is to be able to use pipes to connect to a client, but it also simplifies my upcoming benchmarking work. It's relatively low impact. If you don't like this, I can rework my other PRs.

PR #1392, opened 2026-06-10
perf(all): parse the server stream with Ox instead of REXML

sproctor:
## What

Replaces REXML with [Ox](https://github.com/ohler55/ox) for parsing the **GemStone server stream**. `REXML::Document.parse_stream` was ~41% of server-thread CPU; Ox's SAX parser is ~16x faster at tokenizing. `XMLData` implements the `Ox::Sax` interface directly, so Ox parses straight into it and the large existing `tag_start`/`text`/`tag_end` logic is untouched.

Commits, in order:
- **parse with Ox instead of REXML** — Ox is permissive, so it handles Simu's not-quite-XML (nested attribute quotes, missing `</d>`, etc.) without REXML's clean/retry dance; that error-handling is dropped (a rescue remains as a crash safety net). REXML's `XMLCleaner` is kept (still unit-tested).
- **drop the `<root>` wrapper** — only REXML needed a single root; Ox handles multiple top-level elements and bare text directly, saving a per-line allocation and the synthetic root callbacks.
- **tag parsed text as Windows-1252** — the stream is Windows-1252; Ox returns ASCII-8BIT, so tag it with its true encoding rather than UTF-8 (which would be invalid for high bytes).
- **inline the Ox bridge into XMLData** — drop the separate adapter object.
- **strip the server string once per line** — `strip_xml` ran twice (downstream feed + Infomon); compute once and share, dropping an unnecessary `.dup`.
- **drop `strip_xml`'s `type:` arg / simplify the multiline buffer** — only ever `'main'` now.

## Impact

`REXML::Document.parse_stream` (~41% of server-thread CPU) drops out of the profile; ~+22% pipeline throughput from the parser swap, plus `strip_xml` from ~11% to ~7%. (Measured with a separate `--pipe` benchmark; this change is independent of it.)

## Scope / dependencies

This swaps **only the GemStone server-stream parse**. REXML is still used elsewhere (DR parser, `gameobj`/`spell` data files, `textstripper`) and **remains a dependency** — so this *adds* Ox for the hot path rather than removing REXML. Full REXML removal can be a focused follow-up.

## Heads-up: new native dependency

Ox is a C extension. It compiles from source (lockfile entry is platform-agnostic, no `bundle lock --add-platform` churn), but it's worth confirming it builds cleanly across the Windows/Mac/Linux targets Lich ships to.

## Tests

`spec/lib/common/xmlparser_sax_spec.rb` covers the SAX→listener translation, multi-element parsing without a wrapper, and Windows-1252 tagging. Full suite passes; rubocop clean.


sproctor:
Moving all of this to my branch wip/ox-hard that is stacked with and dependent on the existing work. So I don't want to have 2 PRs with the same work open and this one will never be ready.

PR #1397, opened 2026-06-11
feat(all): enumerate script subdirs with file glob

FarFigNewGut:
After updating and hitting this code path i noticed *extreme* slowdown in script instantiation. Running the following:
```
;e t = Time.now;Script.start('test');echo Time.now - t
```

where test is just "echo test", to get start time from the function (note it is non-blocking script.start, not script.run, so only measuring time to bring up the new Script object) I was seeing 1.5 - 2.5 second script startup times.

Most likely do not notice this as they probably have a combination of fewer custom scripts + are playing on local SSD drives. I am currently playing from a spinning disk network share, and while i expect some performance loss this was obvious enough to warrant investigation.

The numerous File.directory? calls stacked up to i in testing ~1500 MS. Post change to .glob i'm seeing the startup time from the exec above at more like 0.04 seconds instead of 1.5, which is what i would expect. Formatting of the return from the glob call is mutated to match the output from the Dir.children call so everything else works as expected.

This is likely all just a good reminder i should delete a bunch of random things i never run out of my custom directory :D

## Performance
* Improved script loading speed on slower filesystems by optimizing directory detection.

### DIFF
--- a/lib/common/script.rb
+++ b/lib/common/script.rb
@@ -67,10 +67,10 @@ module Lich
         custom_dirs = []
         if File.directory?(custom_base)
           custom_dirs << custom_base
-          Dir.children(custom_base).sort.each do |child|
-            child_path = File.join(custom_base, child)
-            custom_dirs << child_path if File.directory?(child_path)
-          end
+          # Dir.glob with trailing '/' matches directories only via libc readdir d_type,
+          # avoiding a File.directory? stat() per file. Critical on slower filesystems
+          # where each stat() costs ~5-10ms.
+          custom_dirs.concat(Dir.glob(File.join(custom_base, "*/")).map { |p| p.chomp("/") }.sort)
         end
         file_list = custom_dirs.flat_map { |dir|
           prefix = dir.sub(SCRIPT_DIR, '')


MahtraDR:
@coderabbitai review

MahtraDR (approved):
## Approve

I reviewed this change for correctness, behavioral equivalence, and the performance claim. It is a sound optimization. Approving.

### Performance claim: verified

I reimplemented both code paths and benchmarked them at varying directory sizes (local SSD, 200 iterations):

| files in custom/ | subdirs | OLD | NEW | ratio |
|---|---|---|---|---|
| 50 | 5 | 0.126ms | 0.051ms | 2.5x |
| 500 | 5 | 0.971ms | 0.298ms | 3.3x |
| 2000 | 5 | 3.850ms | 1.166ms | 3.3x |
| 2000 | 50 | 4.083ms | 1.262ms | 3.2x |

- OLD time scales linearly with the number of **files**, not directories (rows 3 vs 4 are nearly identical despite 10x the subdirs). That linear scaling is the signature of one `File.directory?` stat() per entry, which is exactly the waste this PR removes.
- The reported 1500ms is self-consistent: the old loop does N stats where N = entries in custom/. At 5-10ms/stat on a network/spinning share, ~150-300 files lands right at ~1500ms. On local SSD the same stats cost microseconds, so I only see ~3x, which is why SSD users never noticed it.
- The persistent 3x even on fast storage confirms glob is taking the cheaper `readdir` + `d_type` path rather than stat-per-file. A C-vs-Ruby overhead difference alone could not produce the author's 37x; only eliminating per-file stat() round-trips can.

### Correctness: behavior preserved, with one intentional delta

Tested old vs new on a fixture tree (files, subdirs, symlinked dir, hidden dir, mixed-case names):

- Sort order is byte-identical for non-hidden entries (the common prefix means full-path sort == basename sort).
- Regular files excluded, symlinks-to-directories included, empty dirs handled: all equivalent.
- One real difference: `Dir.glob` omits `File::FNM_DOTMATCH`, so **hidden directories** (e.g. `.git`) are no longer enumerated, whereas `Dir.children` included them. This is almost certainly an improvement (you do not want to scan `custom/.git/` for runnable scripts), but it is a silent behavior change worth acknowledging.

### Non-blocking suggestions

1. The code comment states the d_type behavior as universal. It is filesystem-dependent: when `readdir` returns `DT_UNKNOWN` (some NFS configs), Ruby's glob falls back to stat() per candidate and the gain shrinks toward 1x. Consider softening to "uses d_type when the filesystem provides it, falling back to stat()". Your measurement confirms your filesystem supplies it.
2. A one-line note (here or in the comment) that hidden subdirectories are now intentionally excluded would prevent a future "where did my dir go" surprise.

Neither blocks merge. The optimization is real, the mechanism is correct, and equivalence is preserved for all practical cases.

PR #1405, opened 2026-06-16
feat(all): add --bind-address option to set listen socket address

sproctor:
## What

Adds a `--bind-address=IP` CLI option that controls the local address Lich binds its listen sockets to.

## Why

Previously the frontend listener and the `--game` proxy listener were hardcoded to bind `127.0.0.1`. Only the detachable-client listener could target another local address (via `--detachable-client=IP:PORT`). This made it impossible to, e.g., bind the frontend listener to a LAN interface so a remote front-end could connect.

## Changes

- `lib/main/argv_options.rb`
  - Parse `--bind-address=IP` into `@argv_options[:bind_address]`.
  - The detachable-client listener now defaults its host to `bind_address` (an explicit `--detachable-client=IP:PORT` still wins).
- `lib/main/main.rb`
  - Frontend listener (`TCPServer.new`) and `--game` proxy listener (`ReusableTCPServer.create`) now bind `bind_address`, defaulting to `127.0.0.1`.
  - The `GAMEHOST=` value written into the generated `.sal` launch file follows `bind_address` so a spawned front-end connects to the chosen address.
- `lib/main/help_text.rb`
  - Documents `--bind-address=IP` under the `advanced` help topic.

## Behavior

- Default is unchanged: `127.0.0.1`.
- `--bind-address=0.0.0.0` makes all three listeners accept connections from other hosts; any specific IP binds that interface.

Usage: `lich --login Mychar --bind-address=0.0.0.0`

## Notes

- This only changes the bind address; it adds no authentication. Exposing a listener on `0.0.0.0` means anything that can reach the port can drive the session, so it's only appropriate on a trusted network.
- The option name was chosen as `--bind-address` (over `--interface`/`--listen-host`) because "interface" collides with the front-end client concept and "host" implies a resolvable DNS name, whereas this is specifically a local bind address.

## Testing

- `bundle exec rspec spec/lib/main/` passes.
- `bundle exec rubocop` clean on the changed files.
- Syntax-checked all three files.
- No new spec added: this parsing layer (`ArgvOptions`) has no existing spec coverage, and requiring it in isolation pulls in the full CLI/auth/GUI require chain. Happy to add coverage if maintainers prefer.

## Summary by CodeRabbit

* **New Features**
  * Added `--bind-address=IP` CLI option to control the local address Lich binds its listening sockets to (frontend, game proxy, and detachable-client listeners). Defaults to `127.0.0.1`; use `0.0.0.0` to bind to all interfaces.
* **Documentation**
  * Expanded advanced help text to describe `--bind-address` behavior and default.


MahtraDR:
@coderabbitai review

MahtraDR:
That outside-diff comment from CodeRabbitAI looks valid and should be resolved. This would need to be tested, or gated, for that use case.

sproctor:
Addressed the CodeRabbit finding in f89f2d5.

Wildcard binds (`0.0.0.0` / `::`) are valid listen addresses but not valid connect targets, so mirroring them into `GAMEHOST` would leave the spawned frontend unable to connect. The launch file now mirrors only a *specific* `--bind-address` into `GAMEHOST` (the listener binds that one address, so the frontend must reach it there) and falls back to loopback (`127.0.0.1` / `localhost`) for wildcard binds.

MahtraDR:
@coderabbitai review this PR again taking all commits into account

OSXLich-Doug (approved):

PR #1406, opened 2026-06-16
feat(all): Refactor Util::issue_command to better handle possible single/multi-line matches

vtcifer:
Support known single line and known multiple line matches AND where you don't know if the output will be single OR multiple line match before you issue the command. Preserves the end_pattern = `:ignore` symbol behavior by then setting end_pattern = start_pattern

## Summary by CodeRabbit

* **Refactor**
  * Improved command output capture behavior for edge cases involving start/end pattern matching.
  * More reliably stops capturing and decides whether to include or suppress the matching boundary line, including when the start and end occur on the same line.
  * Enhances consistency with the existing capture options, reducing unexpected omissions or extra lines in captured output.

### DIFF
--- a/lib/util/util.rb
+++ b/lib/util/util.rb
@@ -102,10 +102,10 @@ module Lich
         Timeout::timeout(timeout, Interrupt) {
           DownstreamHook.add(name, proc { |line|
             if filter
-              if ignore_end || line =~ end_pattern
+              if line =~ end_pattern
                 DownstreamHook.remove(name)
                 filter = false
-                if quiet && !ignore_end
+                if quiet
                   next(nil)
                 else
                   line
--- a/lib/util/util.rb
+++ b/lib/util/util.rb
@@ -87,7 +87,8 @@ module Lich
       result = []
       name = self.anon_hook
       filter = false
-      ignore_end = end_pattern.eql?(:ignore)
+      end_pattern = start_pattern if end_pattern.eql?(:ignore)
+      matched_start_and_end = false
 
       save_script_silent = Script.current.silent
       save_want_downstream = Script.current.want_downstream
@@ -116,6 +117,14 @@ module Lich
                   line
                 end
               end
+            elsif line =~ start_pattern && line =~ end_pattern
+              matched_start_and_end = true
+              DownstreamHook.remove(name)
+              if quiet
+                next(nil)
+              else
+                line
+              end
             elsif line =~ start_pattern
               filter = true
               if quiet
@@ -131,16 +140,10 @@ module Lich
 
           until (line = get) =~ start_pattern; end
           result << line.rstrip
-          unless ignore_end
-            until (line = get) =~ end_pattern
-              result << line.rstrip
-            end
+          until matched_start_and_end || (line = get) =~ end_pattern
+            result << line.rstrip
           end
-          unless ignore_end
-            if include_end
-              result << line.rstrip
-            end
-          end
+          result << line.rstrip if include_end && !matched_start_and_end 
         }
       rescue Interrupt
         nil
--- a/lib/util/util.rb
+++ b/lib/util/util.rb
@@ -143,7 +143,7 @@ module Lich
           until matched_start_and_end || (line = get) =~ end_pattern
             result << line.rstrip
           end
-          result << line.rstrip if include_end && !matched_start_and_end 
+          result << line.rstrip if include_end && !matched_start_and_end
         }
       rescue Interrupt
         nil


mrhoribu:
@coderabbitai review

lib/util/util.rb
```diff
@@ -87,7 +87,8 @@ def self.issue_command(command, start_pattern, end_pattern = /<prompt/, include_
       result = []
       name = self.anon_hook
       filter = false
-      ignore_end = end_pattern.eql?(:ignore)
```
mrhoribu, line 90: You removed this variable which is used in the actual method further down.

mrhoribu: Findings from CodeRabbitAI found this.

vtcifer: Good catch.

mrhoribu:
@coderabbitai remove all previous comments and review in full again

sproctor:
## Code review

The core idea is sound and fixes a real pre-existing over-capture bug — the old code, on a line matching both patterns, would `get` *past* it looking for a separate end line. The common path is correct. The main concern is that the single-line signal is plumbed through a **cross-thread shared flag** when it could be computed locally. Findings ranked below.

### 1. (Medium) Single-line detection couples the capture loop to a hook flag set on another thread — a real, if low-probability, race

`lib/util/util.rb:121` (write) → `:143` / `:146` (read)

`matched_start_and_end` is assigned inside the `DownstreamHook` proc, which runs on the **server thread** (`games.rb:494` → `DownstreamHook.run`), and is read by the capture loop `until matched_start_and_end || (line = get) =~ end_pattern`, which runs on the **script thread**. The script's downstream buffer is populated *first* (`games.rb:600`, inside `process_xml_data` at line 491) and the hook runs *after* (`games.rb:494`). So the script thread can `get` the start line and evaluate the `until` condition **before** the hook sets the flag.

- **Failure scenario:** for a single-line match (including every `:ignore` call, where `end_pattern == start_pattern`), if the server thread is preempted in the micro-window between the buffer push and the hook run, the loop sees `matched_start_and_end == false`, calls `get` again, and appends the *next* line to `result` (which doesn't match `end_pattern`), over-capturing one extra line; worst case it blocks until `timeout`. It's timing-dependent and rare in practice (the server thread normally runs push→hook synchronously before the script's next 0.05 s poll), but it makes the captured result non-deterministic.
- **Deeper point:** the script thread already holds the captured start line in `line`. Single-line-ness is a *local* property — `line =~ end_pattern` — that needs no coordination with the hook. Routing it through a server-thread-written flag is unnecessary coupling between two independently-fed streams.
- **Suggested fix:** keep the hook's combined branch (it's needed so the hook removes itself / suppresses the line for `quiet`), but drop the cross-thread read and decide locally:

```ruby
until (line = get) =~ start_pattern; end
result << line.rstrip
unless line =~ end_pattern              # start line also matches end => single line
  until (line = get) =~ end_pattern
    result << line.rstrip
  end
  result << line.rstrip if include_end
end
```

This is race-free and behaves identically for `:ignore` (where the start line always matches `end_pattern`), genuine single-line, and multi-line.

### 2. (Low) Duplicated suppress/emit decision in the hook proc

`lib/util/util.rb:103-128`

`if quiet then next(nil) else line end` is repeated in four branches, and the `if filter` block's `then`/`else` branches are identical except that `then` also does `DownstreamHook.remove(name); filter = false`. The line-vs-suppress decision could be computed once (`quiet ? nil : line`) after the branches set their side effects — removing the hand-synchronized copies, which a future tweak to `quiet` could otherwise desync across the start/mid/end cases.

### 3. (Low) No direct test coverage for the new behavior

`spec/lib/util/`

There's no `issue_command` spec (the dir has only `memoryreleaser`, `opts`, `textstripper`), and callers stub it out. This is public API consumed by external `.lic` scripts, and CLAUDE.md states "`spec/` mirrors `lib/`". The new single-line path, the `:ignore` rewrite, and `quiet`/`include_end` interactions would all regress with CI green. A spec driving the hook proc + capture loop over canned single-line and multi-line inputs would lock the behavior in.

### 4. (Nit) YARD doc not updated

`lib/util/util.rb:78`

`@param end_pattern` still reads "Use `:ignore` for single-line capture." The PR generalizes this so *any* caller's output is captured as a single line when one line matches both patterns — worth a sentence so callers passing a real `end_pattern` know a coincidental start+end match short-circuits capture.

---

### Investigated and cleared (not bugs)

A natural worry is **multi-line truncation** for `usexml: false` callers (`inv combat`, `flag`, `exp mods`, …) — the theory being the hook matches start+end against a whole multi-line chunk while the script reads stripped lines, dropping the body. **This does not occur:** `server_string = @socket.gets` is newline-delimited (`games.rb:394`, comment at `games.rb:203`), and `clean_serverstring`'s only combiners (`buffer_room_objs`, `fix_xml_tags`) reassemble a *single* split logical line, never two distinct output lines. So the hook receives one line per invocation; for a multi-line command the start line matches start-only (sets `filter`, not `matched_start_and_end`) and the end line arrives later, so `matched_start_and_end` stays false and the full body is captured. Also fine: the `quiet` + `:ignore` front-end visibility (net suppression is identical to the old code), and `rstrip`-on-non-String (this path always sets exactly one of `want_downstream*`, so `get` returns a String).

Net: ship-able after addressing **#1** (cross-thread flag → local check); #2–#4 are polish.

<sub>🤖 Review generated with [Claude Code](https://claude.com/claude-code)</sub>

PR #1408, opened 2026-06-17
fix(all): surface a failed Game.open in the connect timeout

sproctor:
## Summary
A failed `Game.open` during startup is silently swallowed: the bounded-connect logic polls `Thread#status`, which is `nil` when the connect thread died with an exception and `false` when it finished normally — but the post-loop `if connect_thread.status` only handles the still-running (timeout) case. So a `Game.open` that **raises** (e.g. connection refused, DNS failure, unreachable host) is treated as success, and startup proceeds with a dead socket.

On the main connect path this also defeats the `break_game_host_port` **retry** — it lives in the `rescue` that never runs — so Lich neither retries the alternate host nor exits cleanly.

This is a **long-standing bug**, present since the original connect loop in "Base Lich 5" (verified by reproducing it on `main`), not a recent regression.

## The fix
Extract the connect-with-timeout logic into `Lich::GameBase::Game.open_with_timeout(host, port, timeout = 30)` and use `connect_thread.join(timeout)`:

- returns `nil` on timeout → `kill` + `raise "timed out"`,
- returns the thread on success → proceed,
- **re-raises** the thread's exception on failure → reaches the caller's `rescue` (retry on the first attempt, clean exit on the second).

`report_on_exception = false` in the thread so the error is surfaced once (by the `rescue`) rather than also auto-printed. `main.rb`'s two inline connect loops are wired to the new method (also de-duplicating them).

`join` is interruptible (`Thread#raise` / signals are delivered at the join) and event-driven (wakes the instant the connect finishes), so responsiveness is preserved/improved versus the 0.1s poll loop.

## Tests
TDD: a failing test was added first (`f8d8d79`), then the fix (`51e5172`):
- ✅ `Game.open` succeeds → returns
- ✅ connect hangs past the timeout → raises `/timed out/`
- ✅ `Game.open` raises → propagates (was the failing case)

Full suite green; rubocop clean. (`lib/main/main.rb`'s runtime path has no spec coverage; the connect logic is now covered via the extracted method.)

## Notes
- The `--pipe` connect path has the same pattern; it's addressed on the Ox-migration branch separately. This PR fixes the two main (non-pipe) connect sites.
- Possible future cleanup (not in this PR): refactor `Game.open` to use `Socket.tcp(connect_timeout:)` and drop the wrapper thread entirely.

## Summary by CodeRabbit

* **New Features**
  * Added connection timeout mechanism (default 30 seconds) to prevent indefinite hangs when connecting to game servers.
  * Improved connection failure handling with automatic retry attempts and enhanced error logging.


MahtraDR:
@coderabbitai review

OSXLich-Doug (approved):

PR #1234, opened 2026-02-26
fix(all): GameObj object dedupe & GC

mrhoribu:
<html><body>
<html><head></head><body><h1>Refactor <code>GameObj</code> — deduplication, persistent O(1) index with TTL GC, DRY cleanup, idioms &amp; YARD docs</h1>
<h2>Summary</h2>
<p>This PR refactors <code>Lich::Common::GameObj</code> to eliminate duplicate object allocation,
replace a linear O(n) deduplication scan with a single persistent O(1) hash-index,
introduce time-based garbage collection for the index that is safe for long sessions,
reduce code repetition across factory and accessor methods, align with Ruby best
practices, and add full YARD documentation. The existing public API is fully preserved
— no breaking changes.</p>
<hr>
<h2>Motivation</h2>
<p>The class had several long-standing issues:</p>
<ul>
<li><strong>Duplicate <code>GameObj</code> instances</strong> — every <code>new_npc</code>, <code>new_loot</code>, <code>new_inv</code>, etc. call
unconditionally created and pushed a new object, even when an identical one already
existed in the registry. In active game sessions this caused unnecessary memory growth
with no automatic eviction — the GC cannot collect anything still referenced by a class
variable array, so duplicates lived until the next explicit <code>clear_*</code> call.</li>
<li><strong>No memory management for long sessions</strong> — even with deduplication, a persistent
index holding every unique entity seen across a session could accumulate thousands of
references in marathon or automated runs with no way to reclaim them.</li>
<li><strong>Repeated boilerplate</strong> — factory methods, registry accessors, XML parsing, and
type-matching logic all contained near-identical code with no shared abstraction.</li>
<li><strong>Style drift</strong> — mixed use of <code>Array.new</code>/<code>[]</code>, bare <code>rescue</code>/<code>$!</code>, <code>for</code> loops,
manual array building where <code>filter_map</code> fits, and <code>def GameObj.method</code> instead of
<code>def self.method</code>.</li>
<li><strong>No documentation</strong> — no YARD coverage on any method, making the class difficult to
work with for contributors unfamiliar with the codebase.</li>
</ul>
<hr>
<h2>Changes</h2>
<h3>Deduplication — <code>find_or_create</code> with persistent O(1) <code>@@index</code></h3>
<p>All registry factory methods now route through a single private class method. Duplicate
detection uses a <strong>single shared hash index</strong> (<code>@@index</code>) keyed on the composite string
[...]

PR #12, opened 2021-09-01
Effects to downcase for compare

OSXLich-Doug:
Addresses Issue #11 

summary of changes

bumped version to 5.0.18
creaeted begin / rescue clause to gracefully fail out with 'false' in .active?
structured .active? to resolve request to keys in downcase when not Integer
structured .active? to resolve Symbol requests to keys in downcase

Issue #11, opened 2021-09-01
Effects Module Errors


OSXLich-Doug:
The Effects module code as modified to support IDs from SIMU has three general error states that need to be corrected

1) Attempts to search by name when the name is not present in the :name keys generates an exception and does not fail gracefully

2) Attempts to search by name with improper capitalization as sent by Simu generates an exception and does not fail gracefully

3)  Symbols are not supported

PR #1410, opened 2026-06-18
fix(all): bound long-session memory growth (index/hook/watchfor/buffer leaks)

sproctor:
## What & why

Lich runs for hours or days as a proxy, so unbounded in-memory growth shows up
as steadily climbing RSS over a session. This PR adds a memory-profiling
harness, uses it to characterize the footprint, and fixes the four
unbounded-growth leaks it surfaced. All fixes are backstops/auto-cleanup that
are no-ops for already-well-behaved paths.

## Changes

**Profiling harness (`bench/`, dev-only, not loaded at runtime)**
- `memory_sampler.rb` — RSS + GC/heap/ObjectSpace stats at named checkpoints.
- `lich_env.rb` — headless bootstrap of the real subsystems (skips `main.rb`'s
  live login/socket thread); data/log/script dirs sandboxed to a temp dir.
- `memory_profile.rb` — synthesizes a "normal run" (boot -> GameLoader ->
  synthetic GS stream -> GameObj index growth -> real Script start/kill churn),
  sampling per phase. Offline-reproducible via the repo's `effect-list.xml`.

**`fix(common)`: GameObj identity-index auto-prune** — `@@index` grew unbounded
(`prune_index!` existed but was never called automatically). Now pruned on a
throttle from the insert path, reusing the existing live-registry guard so live
entries are never evicted. Tunable via `GameObj.auto_prune*`.
_Long-session sim (room changes -> stale entries): 20,000 -> ~600 entries._

**`fix(common)`: hook lifetime on script death + watchfor cleanup** — an
orphaned hook proc pins the dead script's binding, keeping the whole script
alive. But some scripts deliberately register a hook and exit (`;alias`
registers an UpstreamHook that runs off the in-memory module API; a few do the
same with downstream hooks), so blindly removing a dead script's hooks would
break them. Hooks now declare intent via `add(..., persist:)`: `true` keeps it
past death, `false` scopes it to the script and is removed on death, and
**undeclared (the default for every existing script) is kept, with a one-time
advisory warning** to nudge the author. So `cleanup_on_death` only removes
`persist: false` hooks — existing register-and-exit scripts are unaffected.
Cleanup keys on the owning script's `object_id` (a `force: true` sibling that
shares a name keeps its own hooks) and is dispatched through a small
`ScriptDeath` callback registry each subsystem subscribes to, rather than the
kill path naming subsystems by hand. `@watchfor` is always cleared (those procs
can never fire once the script leaves `@@running`). The engine's own in-tree
hook callers declare their disposition explicitly.
_A/B, 100 `persist: false` start/exit cycles: 100 -> 0 leaked hooks; +100 -> 0
retained procs._

**`fix(common)`: prune dead-thread entries from Buffer/SharedBuffer indices** —
`@@index`/`@@streams`/`@buffer_index` are keyed by `Thread#object_id` and were
never pruned (cleanup methods existed but were never called). Now invoked on a
throttle from the thread-registration path; cleanup is also made mutex-safe and
snapshots live thread ids once instead of rescanning `Thread.list` per entry.
_50 short-lived reader threads: 50 -> 1 index entry._

## Not included

The static `GameLoader` footprint (~34 MiB at first GS login) was profiled: it
is ~22 MiB of retained data spread thin across core subsystems plus ~12 MiB of
native sqlite3/Sequel (infomon). `GC.compact` reclaims nothing, iseq is
negligible, and there is no single high-ROI reducible target, so no static
change is included here.

## Testing

- `bundle exec rspec` — full suite green (incl. new specs for GameObj
  auto-prune, hook persist/cleanup-on-death via the `ScriptDeath` registry, and
  Buffer/SharedBuffer thread cleanup).
- `bundle exec rubocop` — clean (ASCII-only cop included).
- Behavior verified end-to-end with `bench/memory_profile.rb`.

## Summary by CodeRabbit

* **New Features**
  * Added memory profiling and benchmarking tools for performance analysis.
  * Hook persistence control: explicitly mark hooks as persistent or scoped to script lifecycle.

* **Bug Fixes**
  * Improved thread cleanup to prevent unbounded resource growth in long-running sessions.
  * Enhanced script lifecycle management with centralized death-handling callbacks.
  * Added concurrency protection for shared registries to prevent data races.

* **Tests**
  * Added comprehensive coverage for memory profiling, thread cleanup, hook persistence, and script cleanup behavior.


sproctor:
## Code review — bounding long-session memory growth

Reviewed at max effort. The intent (bounding the Buffer/SharedBuffer thread-id indices, orphaned hooks, watchfor, and the GameObj identity index) is sound and well-tested; ASCII/conventions/spec-isolation all check out. Findings ranked most-severe first, correctness before cleanup.

### Correctness

**1. `lib/common/gameobj.rb:1125` — `@@index` is mutated concurrently from multiple threads with no lock; auto-prune adds a second `delete_if` racer** `[HIGH]`
`@@index` has no mutex anywhere. `MemoryReleaser` auto-starts by default (`auto_start: true`) and its background worker calls `prune_index!` -> `sweep_stale!` -> `@@index.delete_if` every 900s. The parser thread inserts new keys via `find_or_create`/`index_or_create`, and this PR now adds `maybe_auto_prune` -> `sweep_stale!` -> `@@index.delete_if` on that same insert path. `Hash#delete_if` yields to a Ruby block per entry, so the GVL can switch threads mid-iteration. When MemoryReleaser's sweep overlaps a parser insert (key add) or the parser's own `maybe_auto_prune` sweep, Ruby raises `RuntimeError: can't add a new key into hash during iteration` on the game-parse thread, aborting parsing of that line. The lock-free `@@last_auto_prune_at` throttle also lets two threads pass the gate and sweep simultaneously. Suggest guarding all `@@index` access with a mutex shared by `maybe_auto_prune`/`prune_index!`.

**2. `lib/common/script.rb:829` (+ `downstreamhook.rb:47`, `upstreamhook.rb:47`) — `remove_by_source(@name)` tears out a same-named sibling script's live hooks** `[MED]`
Hooks are tracked by script *name*, and `Script.start(..., force: true)` bypasses the already-running guard (`script.rb:89`), so two concurrently-running scripts can share a `@name`. When instance A exits, `remove_by_source(@name)` removes every hook whose source equals that name — including instance B's still-active hooks — silently stopping B's stream processing with no error.

**3. `lib/common/script.rb:836` — kill nils `@downstream_buffer`/`@upstream_buffer` while the script is still in `@@running`, exposing a `nil.push` race** `[MED]`
`Script.new_downstream`/`new_upstream` run on the parser thread, take no shared lock with kill cleanup, and call `script.downstream_buffer.push(...) if script.want_downstream` with no nil-guard and no rescue. In the window between `@downstream_buffer = nil` (836) and `@@running.delete(self)` (837), a downstream line raises `NoMethodError` on nil. Pre-existing, but it sits in the function this PR heavily edits, and the PR fixed the exactly-analogous `@watchfor` read (835, set to `{}` to avoid `nil.empty?`) while leaving the sibling buffers nil. Same fix: assign `[]`, or delete from `@@running` first.

**4. `lib/common/buffer.rb:81` / `:161` (and `sharedbuffer.rb:123`) — auto-run `cleanup` (mutex-held `delete_if`) races unsynchronized writers** `[MED]`
The PR makes `cleanup` run automatically via `maybe_cleanup`, holding `@@mutex` while `delete_if`-ing `@@index`/`@@streams`. But `Buffer.rewind` writes both hashes outside `@@mutex`, and for a thread whose first Buffer call is `rewind` that *adds* keys — a key-add during a concurrent `cleanup` iteration raises `can't add a new key into hash during iteration`. The `@last_cleanup_at` throttles are also read-modified-written lock-free. (Evicting the live caller's own entry is *not* a risk — `Thread.list` always includes it.)

**5. `lib/common/gameobj.rb:1126` — `sweep_stale!` live-guard keys on `obj.id`, but the index keys on `id|noun|name`, so stale name-variants are pinned forever** `[MED]`
`live_ids.include?(obj.id)` keeps any entry whose id is live in a registry. When the same exist-id appears under different name strings (PC title changes, NPC descriptor changes -> distinct keys), one live variant pins every stale same-id sibling regardless of age, so auto-prune can never evict them — a residual unbounded-growth path. (Pre-existing in `prune_index!`, but the PR's goal is to bound this index.)

**6. `bench/lich_env.rb:51` — sandbox dir constants are never set; profiling writes into the real repo tree** `[MED, non-production]`
`Object.const_set(c, dir) unless defined?(c)` checks the *block parameter* `c`, and `defined?(local_var)` is always `"local-variable"` (truthy), so `const_set` never runs for `TEMP_DIR/DATA_DIR/LOG_DIR/BACKUP_DIR/MAP_DIR`. `constants.rb` then `||=`s them to `LICH_DIR/{temp,data,logs,...}` (the repo root), so a profiling run writes `infomon.db`, logs, and `effect-list.xml` into the repo tree — contradicting the file's header guarantee that it "never touches a real character's data/logs/db." Fix: `unless Object.const_defined?(c)` — the correct form already used for `SCRIPT_DIR` two lines below.

### Cleanup / Altitude / Efficiency

**7. `lib/common/gameobj.rb:98` — auto-prune duplicates the existing MemoryReleaser scheduler** `[altitude]`
`MemoryReleaser` is already the designated periodic GameObj-index pruner (auto-started, `prune_index!(ttl:)` every 900s). This PR bolts a second scheduler onto the data structure with its own TTL/threshold/interval/`GC.start` — two pruners with two cadences to reconcile, and the root of finding #1. Pushing the size-threshold trigger into MemoryReleaser instead of sweeping from the hot insert path would be deeper.

**8. `buffer.rb:161` / `sharedbuffer.rb:123` / `gameobj.rb:1097` — the throttle dance is hand-rolled three times** `[reuse]`
All three copy "monotonic now; return if `(now - last) < interval`; stamp; sweep." No shared helper, so a fix to one silently skips the others. Extract one throttle helper.

**9. `lib/common/gameobj.rb:629` — 4 tuning class vars + 9 accessors with no production callers** `[simplification]`
`auto_prune`/`_threshold`/`_interval`/`_ttl` (+ `auto_prune?`) add a config surface only the spec exercises. The sibling Buffer/SharedBuffer fix uses one `CLEANUP_INTERVAL` constant and no accessors. Collapse to private constants (keep at most an on/off flag).

**10. `lib/common/script.rb:829-835` — kill hardcodes DownstreamHook/UpstreamHook/Watchfor internals** `[altitude]`
No `on_script_death`/observer mechanism exists, so kill reaches into each subsystem directly (and pokes the Watchfor-owned `@watchfor`, duplicating `Watchfor.clear`). Every future per-script global must remember to add another line here. A death-callback registry each subsystem subscribes to would invert the coupling.

**11. `lib/common/gameobj.rb:1101` — redundant `clock_gettime` on the hottest insert path** `[efficiency]`
`find_or_create`/`index_or_create` already compute `now` (1068/421) immediately before calling `maybe_auto_prune`, which calls `Process.clock_gettime` again. Pass `now` in.

**12. `lib/common/gameobj.rb:1147` / `:1106` — the sweep allocates a splat mega-array and runs `GC.start` synchronously on the parse thread** `[efficiency]`
`live_registry_ids` splats ~11 registry arrays + `@@contents.values.flatten` then `.compact.map(&:id)` -> `Set`, and the prune calls `GC.start` — all inline on the latency-sensitive parse thread. Build the id Set by iterating registries without the intermediates, and let MemoryReleaser own the GC hint off-thread.

**13. `lib/common/upstreamhook.rb:47` / `downstreamhook.rb:47` — `remove_by_source` is a byte-identical twin** `[reuse]`
The method plus its 14-line YARD block is copy-pasted across the two near-identical hook classes (only the cvar prefix differs). Fixes to source-matching (e.g. #2) must be made twice. Factor a shared module/base.

---

**Verified clean:** ASCII-only source, no hand-edited `LICH_VERSION`, no Ruby-2.6-breaking syntax, no test-only methods in production, full YARD on new public methods, and the new/modified specs reset shared class state and pass under `--order random`. The `@watchfor = {}` choice (vs `nil`) is correct and race-safe; `remove_by_source` snapshots before mutating and clears both hook maps; Buffer cleanup keeps `@@index`/`@@streams` in sync and never evicts the live caller.

The headline issue is **#1** — auto-prune adds unsynchronized concurrent mutation to a hash a background thread already prunes; that one warrants a mutex before merge.

<sub>Generated with [Claude Code](https://claude.com/claude-code) `/code-review`</sub>

lib/common/script.rb
```diff
@@ -820,6 +820,19 @@ def __run_kill_cleanup(source:, context:, record_metrics:)
...
+              # dead script so it can never be garbage collected. Removal is
+              # idempotent, so this is a no-op for scripts that cleaned up in a
+              # before_dying block.
+              DownstreamHook.remove_by_source(@name)
+              UpstreamHook.remove_by_source(@name)
+              # Release per-script watchfor procs (and the bindings they
+              # capture). The script is removed from @@running below, so they can
+              # never fire again; cleared to {} rather than nil so a concurrent
+              # new_downstream sweep still sees a safe, empty collection.
+              @watchfor = {}
```
mrhoribu: This is going to cause issues for a handful of current scripts that create hooks that run without their primary script staying running in the background (as the hook just uses the now in memory module/class API). Key example is ;alias that creates an upstream hook. But there are a few scripts that do downstream hooks as well in the same manner.

sproctor:
## Review follow-up — all findings addressed

Pushed 9 commits (`fc64f0e..a467af0`) resolving every item from the review above. Full suite is green (4158 examples, 0 failures) and rubocop is clean across all changed files.

### Correctness
| Finding | Commit | Resolution |
|---|---|---|
| **#1** GameObj `@@index` data race | `7a8fb37` | New `@@index_mutex` guards inserts and the `delete_if` sweeps, so the parser thread and MemoryReleaser's background worker can no longer hit `can't add a new key into hash during iteration`. The live-object set is computed outside the lock (no reentrancy). |
| **#5** stale same-id name variants pinned | `7a8fb37` | The live guard now keys on object **identity** (`live_registry_objects`) instead of `obj.id`, so a stale noun/name variant sharing an id with a live object is evicted. Regression test added. |
| **#3** `nil.push` race on dead-script buffers | `ebd3728` | Kill resets `@downstream_buffer`/`@upstream_buffer` to fresh empty arrays (not nil), matching the existing `@watchfor = {}` fix. |
| **#4** `rewind` adds keys outside the mutex | `078b6c3` | `Buffer.rewind` / `SharedBuffer#rewind` now write under the buffer mutex. |
| **#2** `remove_by_source` over-removes a same-named `force:true` sibling's hooks | `507ef0f` | Hooks now record their owning script's `object_id`; the kill path calls `remove_by_owner`, so a same-named sibling keeps its own hooks. |
| **#6** bench sandbox dirs never set | `3a420ca` | `Object.const_defined?(c)` instead of `defined?(c)` (which is always truthy for a block local). |

### Altitude / reuse / simplification / efficiency
| Finding | Commit | Resolution |
|---|---|---|
| **#7** duplicate pruner + **#9** over-built config + **#12** sweep/GC on parser thread | `bb4455a` | Removed the insert-path auto-prune, the four `auto_prune*` knobs, and their nine accessors. Pruning now happens only off-thread in `MemoryReleaser` (which has called `prune_index!` since #1234), so there's no longer an O(n) sweep or `GC.start` on the parse path. |
| **#8** 3-way throttle duplication | `4b20fda` | Extracted `Lich::Common::Throttle`; both buffers use it. |
| **#13** Down/UpstreamHook twins | `f1909ba` | Factored the shared registration/bookkeeping into `Lich::Common::HookRegistry`; the genuinely different `run` methods stay per-class. |
| **#10** kill hardcodes subsystem internals | `a467af0` | Added `Lich::Common::ScriptDeath`, a small on-death callback registry; the hook registries self-register their cleanup at load and kill just calls `ScriptDeath.run(self)`. |

### Notes
- **#11** (redundant `clock_gettime`) was made moot by **#7** — there is no insert-path prune left to optimize.
- The headline race (**#1**) is the load-bearing fix: `@@index` is now the only piece that's mutex-guarded, since `MemoryReleaser` prunes it from a background thread.

<sub>Generated with [Claude Code](https://claude.com/claude-code)</sub>

lib/common/script.rb
sproctor: Good catch -- you're right, and this would have broken `;alias` and the other register-and-exit hook scripts. Fixed in 983c4f8: hooks now **persist past script death by default**.

`HookRegistry.add` takes a `persist:` keyword:
- `persist: true` -- kept when the owning script dies (the `;alias` case)
- `persist: false` -- scoped to the script's lifetime, removed on death
- **undeclared (the default for every existing script)** -- kept, with a one-time advisory warning nudging the author to declare intent

So `cleanup_on_death` only removes `persist: false` hooks. Existing scripts that register a hook and exit are unaffected -- nothing is torn out unless it opts in. The engine's own in-tree callers (map_gs peer, xmlparser nerve tracker, combat tracker, inventory-box toggles, util command hooks) now declare their disposition explicitly so they don't trip the warning.

OSXLich-Doug:
@coderabbitai review

lib/common/script.rb
mrhoribu: Nice work, I'll take a look when I get some time back in front of a PC to test this a bit

OSXLich-Doug (approved):

PR #1413, opened 2026-06-24
feat(dr): xmlparser: add structured parser for the assess (combat situation) stream

MahtraDR:
## Summary

Adds structured capture of GemStone `ASSESS` output to `Lich::Common::XMLParser`, exposed as `XMLData.assess` (an array of per-entity hashes) plus an `XMLData.assess_creatures` convenience filter (positive-id mobs only).

Previously the `assess` stream was unparsed -- scripts wanting combat positioning had to scrape raw lines (and `logxml` drops the stream entirely unless run with `streams`). This surfaces the data as first-class `XMLData` state, alongside the existing `bounty`/`society`/`familiar` stream handling.

## How it works

Each assess line arrives in its own `pushStream`/`popStream`; the subject and target `GameObj` ids live in `<d cmd='look #id'>` attributes (subject first, then target), and the trailing `  | F` is a separate `<d cmd='face #id'>` hint. The parser:

- inits a per-line text buffer + look-id list on `pushStream id="assess"`
- collects look ids (ignoring `face` hints) from `<d>` tags in the stream
- reassembles the full line text via the existing `text()` pipeline
- parses name, id, assess number, status, relation, target (+ id/number), and range (`:melee` / `:pole` / `:missile`) on `popStream`
- resets the list on `clearStream id="assess"` (start of a fresh `ASSESS`)

Self lines (no subject id) and PCs (negative ids) are distinguished from creatures, matching the existing negative-id PC convention already used in this file.

### Each entry

```ruby
{ name:, id:, number:, status:, relation:, target:, target_id:,
  target_number:, range:, self:, pc: }
```

`XMLData.assess_creatures` returns the subset that are creatures (positive ids; excludes self and PCs) -- an **array of these hashes**, e.g.:
```ruby
[{ name: "A jeol moradu", id: "89511379", number: 1, status: "cursed and nimbly balanced",
   relation: "behind", target: "Tenuk", target_id: "-10592168", target_number: nil,
   range: :melee, self: false, pc: false },
 # ... through number: 7, range: :pole
]
```

A script that wants a compact projection can map over it, e.g. `assess_creatures.map { |c| [c[:number], c[:name], c[:id], c[:range]] }`.

## Tests

Adds `spec/lib/common/xmlparser_spec.rb` -- the first spec for this class. Two layers:

- **`#parse_assess_line`** unit coverage (header rejection, self/creature/PC lines, `you` targets, `moving to flank` -> `flanking` normalization, all three ranges)
- **stream integration** driving the real `REXML::StreamListener` pipeline with verbatim game output, confirming the `tag_start`/`text`/`popStream`/`clearStream` wiring reassembles lines and ids correctly

`15 examples, 0 failures`. Rubocop clean on both files.

## Compatibility

Purely additive -- new attr_reader, new instance vars, new stream branch. No change to existing parsing paths or public API.


MahtraDR:
@coderabbitai review in full

MahtraDR:
@coderabbitai Thanks for the thorough review. Addressed in caea6dfa:

- **1. `is_self` double-negation** -- done. Switched to `name.casecmp?('you')` (also exact-match rather than line-anchored).
- **2. Pipe-strip too greedy** -- done. Tightened to `/\s+\|\s+\S+\s*$/`, anchored to the pipe + trailing token so a stray `|` mid-line can't eat text. (Worth noting: the strip is cosmetic anyway -- the main match ends at `range\b` with no end-anchor, so the `| F` hint never reaches a captured field -- but the tighter form is clearer.)
- **3. Redundant `||=`** -- done, removed. It was dead: `@assess_ids` is set in `initialize` and reset on every `pushStream id="assess"`, and the `<d>` branch only fires once `@current_stream == 'assess'`, which only happens after that pushStream.
- **6. PR description return type** -- fixed. Description now shows the array-of-hashes return and notes the compact form is a `.map` projection.
- **7. Top-level spec constant** -- done. Moved the corpus into a `let(:assess_corpus)` inside the describe block.
- **8. Untested `number: nil` on self line** -- done, added the assertion.

Two I did **not** change, with reasoning:

- **4. `@assess = []` vs `.clear`** -- keeping reassignment. The parser runs on the REXML parse thread while scripts read `XMLData.assess` from their own threads. Reassignment hands each caller a stable, complete snapshot; the next `ASSESS` rebinds `@assess` to a fresh array and the caller's reference is untouched. `.clear` + `<<` would instead empty and refill the *same* array a script may be mid-iteration over, producing torn/empty reads. So the "callers holding a reference see the reset" behavior is the hazard here, not the goal. It's also consistent with the sibling `@bounty_task = String.new` in the same handler (which likewise allocates a new object rather than mutating in place).
- **5. `assess_creatures` delegation** -- no change needed. `XMLData = Lich::Common::XMLParser.new` (lich.rbw:120), so `XMLData` *is* the parser instance; `assess_creatures` resolves by ordinary method dispatch with no delegation/`method_missing` layer to update.

Spec is green (15/15) and rubocop is clean on both files.

OSXLich-Doug (approved):

PR #1414, opened 2026-06-25
fix(dr): equipmanager.rb - Fix unload weapon detection when AimTimerDialog tag is prepended

MahtraDR:
## Summary

The game introduced an aim/firing timer tag that can be prepended to the unload line:

```
<dialogData id='AimTimerDialog'><timer id='firingTimer' value='0' /> </dialogData>You unload the shortbow.
```

The `^`-anchored patterns in `UNLOAD_WEAPON_SUCCESS_PATTERNS` (`lib/dragonrealms/commons/common-items.rb`) no longer matched when this tag was present, so unloads went undetected by `EquipmentManager#unload_weapon`. The line can still arrive plain (no tag) as well.

## Change

Allow an optional `(?:<dialogData.*?</dialogData>)?` prefix on the two `You unload` / `unloading` success patterns. The non-greedy `.*?` keeps it lenient about the tag's inner attributes/values (e.g. `value='0'` may vary) while preserving the `^` anchor so it won't match mid-sentence.

## Testing

Added two tagged-line cases to the `UNLOAD_WEAPON_SUCCESS_PATTERNS` spec (a plain `You unload` and the stealth `unloading` variant). Full unload spec passes:

```
12 examples, 0 failures
```

Rubocop clean on both touched files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved unload success detection so messages with an optional timer-tagged dialog wrapper are recognized correctly.
  * Unload actions now match successfully even when extra XML dialog data appears before the confirmation text.
  * Added coverage for unload messages that include the new wrapped format.


MahtraDR:
@coderabbitai review

OSXLich-Doug (approved):

PR #1417, opened 2026-06-30
feat(all): add central method to determine lich ver

OSXLich-Doug:
To remove the copy / pasta of warning messages from scripts - as desired.

## Summary by CodeRabbit

* **New Features**
  * Added script version checks based on each script’s declared minimum Lich version.
  * Improved script startup so scripts are found more reliably in standard and custom locations.

* **Bug Fixes**
  * Fixed version lookup for scripts in custom subfolders.
  * Prevented scripts from running when the installed Lich version is too old, with a clearer warning message.
  * Made script version parsing more tolerant of comments, formatting, and version suffixes.


OSXLich-Doug:
@coderabbitai review PR

MahtraDR (approved):
# Review: PR [#1417](https://github.com/elanthia-online/lich-5/pull/1417)

**Title:** feat(all): and central method to determine lich ver
**Author:** OSXLich-Doug
**Base:** main | **Head:** `d07b26d` | **+506 / -33** across 2 files
**Reviewed:** 2026-06-30

---

## 1. What the PR does

Centralizes script-file discovery and adds a per-script "minimum Lich version"
gate, replacing the copy-pasted version-check + warning idiom that scripts have
each open-coded (the stated motivation: "remove the copy / pasta of warning
messages from scripts").

1. **Shared resolver `Script.__find_script_file(script_name)`**
   ([script.rb:323-345](https://github.com/elanthia-online/lich-5/blob/d07b26d/lib/common/script.rb#L323)).
   Extracts the file-discovery logic that previously lived inline in
   `Script.start`. Searches `custom/` root, then each `custom/<subdir>/`, then
   `SCRIPT_DIR`, matching exact -> case-sensitive-prefix -> case-insensitive
   prefix. Now the single resolver shared by `Script.start`, `Script.version`,
   and `Script.required_lich_version` so they can never diverge on what they see.
2. **Shared header extractor `Script.__extract_header_comments(script_data)`**
   ([script.rb:356-376](https://github.com/elanthia-online/lich-5/blob/d07b26d/lib/common/script.rb#L356)).
   Pulls the `=begin/=end` block (preferred) or the leading run of `#` comments.
   One source of truth for "the script header."
3. **`Script.version` refactor** -- rewritten to delegate to the two helpers;
   behavior preserved (returns `Gem::Version`, or boolean against a required
   arg, or nil if missing).
4. **New public gate API:**
   - `Script.required_lich_version(script_name = nil)` -- reads the
     `required: Lich <op> X.Y.Z` header; nil-arg path reads the running script's
     own `file_name` directly.
   - `Script.lich_version_satisfied?(minimum = required_lich_version)` -- compares
     `LICH_VERSION` numerically; missing/blank/unparseable minimum passes.
   - `Script.require_lich_version!(minimum = required_lich_version)` -- the
     one-liner: warn via `Lich::Messaging` and `current.exit` when too old.
   - Private `__to_gem_version` (fail-soft parse) and `__warn_lich_too_old`
     (frontend-aware notice).
5. **New spec** `spec/lib/common/script_version_gate_spec.rb` -- 325 lines, full
   coverage of the new API plus characterization of the refactored
   `Script.version`, using an in-memory virtual filesystem.

## 2. Headline assessment

Solid, well-documented, genuinely DRY PR. **Recommend approve / merge.**

- **Correct:** the resolver consolidation is behavior-preserving for
  `Script.start` (verified the regex collapse is an exact equivalence, 3.1), and
  the gate API's fail-open-on-malformed design is deliberate and consistent.
- **DRY:** a real de-duplication -- one resolver replaces two near-identical
  copies, one header extractor replaces two, and one canonical reader replaces
  the per-script copy-paste (`dependency.lic` open-codes the entire gate with raw
  `<pushBold/>` xml; this PR routes through `Lich::Messaging` so it renders on
  every frontend -- a correctness improvement, not just tidiness).
- **Documentation:** exemplary YARD. Every helper explains the *why*
  (fail-open avoidance, the lossy `@name`->file round-trip, frontend routing),
  not just the what.
- **Tests:** 33 examples, 0 failures; rubocop clean (verified locally, section 8).

Findings below are all **minor / non-blocking**. The only one with any
correctness teeth is the gz edge case (3.4) -- a compressed script *crashes* its
own header read (confirmed empirically); pre-existing for `Script.version` and a
very rare configuration, but newly reachable via `require_lich_version!`.

---

## 3. Correctness

### 3.1 Verified -- the `Script.start` resolver extraction is behavior-preserving

The old inline tier-1 match was
`find { |val| val =~ /...\.ext.../ || val =~ /...\.ext.../i }`. The new code is
`find { |val| val =~ /...\.ext.../i }`. These are **equivalent**: the `||`
short-circuits *within a single element*, and the case-insensitive pattern is a
superset of the case-sensitive one, so the block was already true exactly when
the element matched case-insensitively. Tiers 2 and 3 (case-sensitive prefix,
then case-insensitive prefix) are carried over verbatim. The `custom/` subdir
enumeration, the `Dir.glob(".../*/")` directory-only optimization, and the
extensionless sort key are all moved unchanged. No regression.

### 3.2 Verified -- `Script.version`'s search scope widens intentionally (alignment, not bug)

The old `Script.version` searched only `custom/` root + `SCRIPT_DIR`; it did
**not** look in `custom/<subdir>/`. Routing it through `__find_script_file` means
it now also searches subdirs -- which is the entire point ("so they can never
diverge in which files they can see"). The risk would be a name that resolves to
a *different* file than before; but because the resolver mirrors `Script.start`'s
order, `version()` now agrees with what actually *starts*, which is strictly more
correct. `Script.version` has no in-repo callers outside `script.rb`; it is a
public API consumed by scripts doing dependency checks. Low risk, net
improvement.

### 3.3 Verified -- the nil-arg path correctly avoids a fail-open bypass

`@file_name` is the full absolute path; `@name` is only the basename
([script.rb:834-835](https://github.com/elanthia-online/lich-5/blob/d07b26d/lib/common/script.rb#L834):
`@name = /.*[\/\\]+([^\.]+)\./.match(@file_name)...`). So a running
`custom/<subdir>/foo.lic` knows itself only as `foo`, and a `foo`->file
round-trip could resolve to the wrong `foo` (or none). Reading
`Script.current.file_name` directly, as the nil-arg path does, sidesteps this
entirely. The doc block calls this out and the spec pins it
("enforces a custom/<subdir>/ script's own declared floor (regression: no
fail-open bypass)"). Correct and well-reasoned.

### 3.4 Note -- gz-compressed scripts *crash* the header read (pre-existing, minor)

`required_lich_version` and `Script.version` both read with plain `File.read`.
For a `.lic.gz` script that returns the raw deflate bytes tagged UTF-8 but
invalid, and the very first header regex (`script_data =~ /^=begin.../m`) raises
**`ArgumentError: invalid byte sequence in UTF-8`** -- the check *crashes* rather
than gating. I confirmed this empirically: `File.read` on a gz file yields a
string with `valid_encoding? == false`, and `=~` against it raises. (So this is
*not* a quiet fail-open -- it is a hard raise.) This is **pre-existing** for
`Script.version` (the old `File.open(...).read` + `=~` did the same; only the
instance `@data` loader at
[script.rb:882](https://github.com/elanthia-online/lich-5/blob/d07b26d/lib/common/script.rb#L882)
handles gz), and is newly reachable via `require_lich_version!`'s nil-arg path
when a *running* compressed script reads its own `file_name`. gz scripts are
rare, so this is not a blocker -- but note CodeRabbit's suggested
`rescue SystemCallError` (section 7) would **not** catch this, since it is an
`ArgumentError` from the regex, not an I/O error. A guard around the read+extract
that also handles `ArgumentError` (or reading via the gz-aware loader) closes
both gaps at once.

### 3.5 Verified -- fail-soft parsing is deliberate and consistent

`__to_gem_version` rescues `ArgumentError` and returns nil; blank -> nil.
`lich_version_satisfied?` treats nil minimum as "no requirement" -> true. So a
malformed `required:` header can never *block* a script (only a well-formed one
can). The regex captures only the dotted-numeric run (`([\d.]+)`), so `5.0x`
yields `5.0` and `Lich 4.3.12, Olib` yields `4.3.12`. All three operator forms
(`>=`, bare `>`, no-operator) are normalized to a floor -- and the bare-`>` form
is not hypothetical: core's `infomon.rb` header is `required: Lich > 5.6.2`.

### 3.6 Verified -- frontend-correct, null-safe output

`Lich::Messaging.msg(type, msg)` matches the
[messaging.rb:127](https://github.com/elanthia-online/lich-5/blob/d07b26d/lib/messaging.rb#L127)
signature and resolves xml/gsl/plain rendering centrally (vs the raw `<pushBold/>`
tags scripts copy-paste, which only render on xml frontends). `current&.exit` and
`current&.name || 'script'` are null-safe -- exercised by the "no current script
to exit" example.

---

## 4. SOLID / DRY (production code)

**Strong.**

- **SRP:** clean single responsibilities -- `__find_script_file` (resolution),
  `__extract_header_comments` (header isolation), `__to_gem_version` (safe parse),
  `__warn_lich_too_old` (notice). Each does one thing.
- **DRY -- the core win:** one resolver replaces two near-identical copies
  (`start` + `version`), one extractor replaces two, and one canonical
  `required:` reader replaces the per-script copy-paste idiom. This is a genuine
  de-duplication, not a cosmetic move.
- **OCP / composition:** the three public methods layer cleanly
  (`required_lich_version` -> `lich_version_satisfied?` -> `require_lich_version!`),
  each usable independently, with sensible defaulting that reads the header at
  most once per call.
- **Encapsulation:** `__`-prefix + `private_class_method` convention applied
  consistently to every internal helper.
- **Documentation:** above and beyond. The YARD blocks explain rationale
  (fail-open avoidance, the lossy name->file round-trip, frontend routing), and
  the inline comment preserves the `Dir.glob` stat-avoidance note.

Minor points (none blocking):

- **4.1 Small residual DRY seam.** The "resolve name -> `File.read` ->
  `__extract_header_comments`" sequence appears in both `Script.version`
  ([390](https://github.com/elanthia-online/lich-5/blob/d07b26d/lib/common/script.rb#L390))
  and the named-arg branch of `required_lich_version`
  ([428-431](https://github.com/elanthia-online/lich-5/blob/d07b26d/lib/common/script.rb#L428)).
  A tiny `__header_lines_for(script_name)` would fold the last seam. The two
  consumers then differ only in which header line they grep. Optional.
- **4.2 Double-slash path.** `"#{SCRIPT_DIR}/#{file_name}"` yields
  `.../scripts//custom/...` when `file_name` carries the `/custom/` marker.
  Harmless on POSIX (and the spec normalizes `//` to match), but `File.join`
  would be tidier and intent-revealing.
- **4.3 Warning shows basename only.** `__warn_lich_too_old` receives
  `current&.name`, so a gated `custom/<subdir>/foo` is reported as `foo`. Cosmetic.

---

## 5. SOLID / DAMP (specs)

**Strong.** 33 examples, all descriptive and self-documenting.

- **DAMP:** example names read as specifications ("treats a bare '>' the same as
  a minimum floor", "compares numerically, not lexically (5.9.0 < 5.10.0)",
  "captures only the dotted-numeric run, dropping a stray suffix (5.0x -> 5.0)").
  Non-obvious examples carry comments explaining the production condition they
  stand in for.
- **Isolation done right:** a per-example in-memory virtual filesystem stubs
  `Dir.glob`, `Dir.children`, and `File.read`; no real script is ever touched.
  The `add_script(..., subdir:)` helper makes each example's setup read locally
  without hiding what it does (DAMP over DRY in the right place). The lib is
  loaded once and its constants removed in `after(:context)` so sibling specs
  still see the stub -- the same disciplined pattern as `script_kill_metrics_spec`.
- **Characterization safety net:** the `.version (characterization)` block pins
  the refactored method's behavior (Gem::Version return, parenthetical
  stripping, 0.0.0 default, boolean-against-required, nil-on-missing) so the
  helper extraction is provably behavior-preserving.
- **Adversarial / boundary coverage:** unparseable minimum, blank/nil minimum,
  exact-equal boundary, lexical-vs-numeric, bare `>`, no-operator, suffix strip,
  dependency-list truncation, two-segment version, custom-root vs subdir
  resolution, missing file, and nil-current safety are all exercised.
- **Self-checking harness:** the "does not double the /custom/ prefix" example
  asserts `Dir.children(custom_base)` returns *bare* names -- it guards the
  *stub's own fidelity* against silently re-introducing the `/custom/custom/`
  bug. Nice touch.

Minor:

- **5.1 One characterization gap.** The `.version` characterization block only
  uses root-level scripts; it does not assert that `version()` now resolves a
  `custom/<subdir>/` script (the scope widening of 3.2). The gating methods do
  cover subdir resolution, so the resolver is tested -- but a one-line
  `version()`-through-a-subdir example would lock in the alignment explicitly.
- **5.2** No gz example (the in-memory VFS only ever stores plain source, so the
  crash-on-compressed behavior of 3.4 is neither exercised nor pinned). If 6.2 is
  taken, a spec asserting a gz script's header read degrades gracefully (rather
  than raising) would lock the fix in.

---

## 6. Suggestions (non-blocking)

- **6.1** Apply CodeRabbit's `File.read` guard (section 7) for parity with the
  fail-soft philosophy `__to_gem_version` already follows.
- **6.2** Make the header read gz-aware (or rescue `ArgumentError` around the
  read+extract) so a compressed script does not crash its own version check
  (3.4). This is a superset of CodeRabbit's I/O guard, not the same fix.
- **6.3** Optional `__header_lines_for` extraction to close the last DRY seam
  (4.1); `File.join` for the read path (4.2).

---

## 7. CodeRabbit findings

One comment, profile CHILL, severity **nitpick / trivial** (the only finding):

> **Guard `File.read` so a transient read failure can't crash a gated script at
> startup.** `__to_gem_version` is fail-soft, but the `File.read` in
> `required_lich_version` (lines 437-439) and `Script.version` (line 390) is
> unguarded -- a delete/permission race between resolve and read would raise
> instead of degrading to "no requirement."

**Assessment: valid, low priority, currently unaddressed** (PR head `d07b26d`
matches the reviewed commit). The window is small (resolve-then-read race) and
the failure is a raise rather than a wrong answer, but wrapping the read in
`rescue SystemCallError => nil` would make the gate consistently fail-soft, which
is the module's stated design goal. Worth taking. Not a merge blocker.

---

## 8. Verification performed

- Fetched PR head `d07b26d` (`upstream pull/1417/head`) and checked it out.
- `rspec spec/lib/common/script_version_gate_spec.rb` (Ruby 4.0.1):
  **33 examples, 0 failures.**
- `rubocop --force-exclusion` on both touched files (incl. the custom
  `AsciiOnlySource` cop): **no offenses.**
- Confirmed the `Script.start` regex collapse is an exact equivalence (3.1).
- Confirmed `@file_name` = full path, `@name` = basename, validating the nil-arg
  design (3.3).
- Confirmed `Lich::Messaging.msg` signature (3.6) and that the bare-`>` header
  form exists in core (`infomon.rb`).
- Confirmed the copy-paste idiom this replaces (`dependency.lic` raw-xml gate).

**Bottom line:** correct, genuinely DRY, exemplarily documented, and thoroughly
+ descriptively tested. **Approve.** Take CodeRabbit's `File.read` guard (6.1)
and add a gz note (6.2); both are polish, not blockers.

PR #1420, opened 2026-07-01
fix(all): use bundler to check for required Ruby gems

mrhoribu:
## Summary by CodeRabbit

* **New Features**
  * Added a boot-time check to verify required Ruby gems before startup.
  * When dependencies are missing, the app shows a platform-specific alert and records a detailed failure report for troubleshooting.
  * On Windows, users can proceed to the latest release from the alert; on macOS/Linux, the alert does not launch the release link.

* **Bug Fixes**
  * Improved behavior when Bundler reports missing gems or out-of-scope dependency issues, including safe logging and graceful fallback messaging.

* **Tests**
  * Added automated coverage for gem verification, alerting, logging, and platform dialog selection.


mrhoribu:
@coderabbitai review

mrhoribu:
@coderabbitai review

mrhoribu:
# Review: PR #1420 (updated)

**Title:** fix(all): use bundler to check for required Ruby gems
**Author:** mrhoribu
**Base:** main | **Head:** `357a31b` | **+805 / -8** across 3 files
**Reviewed:** 2026-07-01 (re-review; supersedes the review at `e227436`)

---

## 0. What changed since the last review

Two corrective commits landed on top of the previously reviewed `e227436`:

- **`02bb652`** "Refactor error handling and ensure TEMP_DIR exists" -- broadens
  the `verify!` rescue and creates the temp dir before logging.
- **`357a31b`** "Add test for BundlerError subclass handling" -- adds the
  regression spec for the broadened rescue.

These resolve the two findings with any teeth from the prior review (3.1 and
3.2) and close the one open CodeRabbit item. The `:default`-only scope question
(3.5) is the only substantive item still open, and it is a design question for
the author rather than a defect.

---

## 1. What the PR does

Adds a boot-time `Lich::GemCheck.verify!` gate that runs immediately after
constants/version load and before the rest of Lich boots, so a user missing
required gems gets a native OS dialog (with a log-file fallback) instead of a
raw stacktrace deep in startup.

1. **New module `lib/gemcheck.rb`** (263 lines). `verify!(*groups)`
   ([[gemcheck.rb:26-52](https://github.com/elanthia-online/lich-5/blob/357a31b/lib/gemcheck.rb#L26)](https://github.com/elanthia-online/lich-5/blob/357a31b/lib/gemcheck.rb#L26))
   runs a two-stage check: first a direct detector `missing_gems`
   ([[60-68](https://github.com/elanthia-online/lich-5/blob/357a31b/lib/gemcheck.rb#L60)](https://github.com/elanthia-online/lich-5/blob/357a31b/lib/gemcheck.rb#L60))
   using `Gem::Specification.find_all_by_name(dep.name, dep.requirement)`; if
   that finds anything it alerts and `exit 1` before ever calling Bundler. Only
   if the detector is clean does it run `Bundler.setup(*groups)` inside a
   `Bundler.settings.temporary(without: excluded)` block with a forced
   `Bundler.definition(true)` re-resolve
   ([[35-40](https://github.com/elanthia-online/lich-5/blob/357a31b/lib/gemcheck.rb#L35)](https://github.com/elanthia-online/lich-5/blob/357a31b/lib/gemcheck.rb#L35)).
2. **Bundler-error scope filter.** `bundler_error_out_of_scope?`
   ([[80-88](https://github.com/elanthia-online/lich-5/blob/357a31b/lib/gemcheck.rb#L80)](https://github.com/elanthia-online/lich-5/blob/357a31b/lib/gemcheck.rb#L80))
   + `extract_gem_name` parse the gem name out of a Bundler error and swallow the
   failure only when it concerns a group that was not requested. In-scope (and
   unattributable) failures alert + exit.
3. **Diagnostics.** `write_log`
   ([[143-189](https://github.com/elanthia-online/lich-5/blob/357a31b/lib/gemcheck.rb#L143)](https://github.com/elanthia-online/lich-5/blob/357a31b/lib/gemcheck.rb#L143))
   ensures the temp dir exists, then appends a timestamped report
   (Ruby/Bundler/Gemfile/lockfile, groups, a per-dependency `dependency_report`,
   release URL on Windows). Every external lookup is wrapped in `safe_call`, and
   the whole write is `rescue StandardError`.
4. **Platform dialogs.** `alert_windows` (WScript.Shell Popup with an
   OK-to-open-download-page branch), `alert_macos` (osascript display dialog),
   `alert_linux` (zenity -> kdialog -> xmessage -> `warn` fallback).
5. **Startup wiring in `lich.rbw`** -- constants/version requires move up, and
   `gemcheck.rb` + `verify!` are inserted at
   [[lich.rbw:36-37](https://github.com/elanthia-online/lich-5/blob/357a31b/lich.rbw#L36)](https://github.com/elanthia-online/lich-5/blob/357a31b/lich.rbw#L36),
   ahead of the stdlib require block; the old later require site is removed.
6. **Spec** `spec/lib/gemcheck_spec.rb` (532 lines) -- 64 examples covering the
   verify flow, detector, scope filter, alert composition, logging, dependency
   report, platform dialog dispatch, command detection, and (new) the broadened
   rescue.

## 2. Headline assessment

Correct, defensively coded, cleanly scoped, thoroughly tested. The two prior
findings that had correctness weight are resolved, and the fixes are minimal and
verified. **Recommend approve / merge**, with one one-line confirmation from the
author on the `:default`-only scope (3.5).

- **Fail-soft everywhere it matters:** the detector short-circuits the ordinary
  missing-gem case before Bundler is touched (3.4), and `Bundler.setup` failures
  now uniformly route to the alert path (3.1, fixed).
- **Well-scoped diff:** three files, no unrelated refactoring, no new runtime
  dependency (Bundler is already required). The corrective commits stayed
  surgical -- rescue breadth, a mkdir guard, one doc line, one spec.
- **Documentation:** near-complete YARD (CodeRabbit measured 93.75%); rationale
  comments on the non-obvious bits.
- **Tests:** 64 examples, 0 failures; rubocop clean incl. the custom ASCII cop
  (section 8).

---

## 3. Correctness

### 3.1 RESOLVED -- `verify!` now rescues all Bundler setup failures

Line 41 is now `rescue Bundler::BundlerError => e`
([[gemcheck.rb:41](https://github.com/elanthia-online/lich-5/blob/357a31b/lib/gemcheck.rb#L41)](https://github.com/elanthia-online/lich-5/blob/357a31b/lib/gemcheck.rb#L41)).
A `Bundler.setup` failure that is not `GemNotFound`/`GitError` (e.g.
`GemRequireError` for an installed-but-unloadable gem, `PathError`,
`SolveFailure`) now reaches the alert + exit path instead of propagating as an
unhandled crash. The broadening is safe: `bundler_error_out_of_scope?` only
swallows when it can positively attribute the error to an unrequested group, so
any error it cannot attribute (which is all the newly-covered subclasses) falls
through to alert + exit. **Verified:** I temporarily reverted the rescue to the
old two-class form and the new spec (3.4 / 5.2) fails with the error propagating
out of `verify!`; with the broadened rescue it passes. The finding is closed and
guarded.

### 3.2 RESOLVED -- the fresh-install log the alert cites now gets written

`write_log` now runs `Dir.mkdir(TEMP_DIR) unless File.exist?(TEMP_DIR)` before
opening the file
([[gemcheck.rb:145](https://github.com/elanthia-online/lich-5/blob/357a31b/lib/gemcheck.rb#L145)](https://github.com/elanthia-online/lich-5/blob/357a31b/lib/gemcheck.rb#L145)),
mirroring `init.rb`'s own guard exactly and needing no new require. Because
`TEMP_DIR`'s parent (`LICH_DIR`, the running-script directory) always exists,
the mkdir succeeds in the normal case; a rare failure (custom `--home` at a
missing path, permission denial) is still absorbed by the method's
`rescue StandardError`. **Verified empirically:** pointing `TEMP_DIR` at a
not-yet-existing `temp/` and calling `write_log` creates the directory, writes
the log, and the log contains the detected gem -- so the dialog's
"See ... for details" line is now accurate on a first run. Finding closed.

### 3.3 Verified -- the out-of-scope workaround is correctly gated (unchanged)

`bundler_error_out_of_scope?` returns true only when a gem name is extractable,
is a known dependency, and its groups do not intersect the requested groups
([[80-88](https://github.com/elanthia-online/lich-5/blob/357a31b/lib/gemcheck.rb#L80)](https://github.com/elanthia-online/lich-5/blob/357a31b/lib/gemcheck.rb#L80));
any failure of those returns false -> alert + exit. All four branches remain
pinned by the spec. Correct.

### 3.4 Verified -- detector runs first and short-circuits before Bundler (unchanged)

The spec asserts that an in-scope detected gem triggers `alert` and never calls
`Bundler.setup`, and that the post-setup alert re-runs the detector for its list
([[spec:68-105](https://github.com/elanthia-online/lich-5/blob/357a31b/spec/lib/gemcheck_spec.rb#L68)](https://github.com/elanthia-online/lich-5/blob/357a31b/spec/lib/gemcheck_spec.rb#L68)).
This is the property that kept the (now-fixed) rescue gap low-severity in the
first place.

### 3.5 Note (still open) -- the call site checks only `:default`

`lich.rbw` still invokes `verify!` with no args
([[lich.rbw:37](https://github.com/elanthia-online/lich-5/blob/357a31b/lich.rbw#L37)](https://github.com/elanthia-online/lich-5/blob/357a31b/lich.rbw#L37)),
so only the `:default` group is gated and `gtk`, `development`, `profanity`,
`vscode` are excluded. A GUI user missing `gtk3` passes this gate and fails
later where `gtk3` is required. This is very likely **intentional** (headless /
ProfanityFE users have no GTK), and it was left unchanged deliberately -- altering
the boot-scope is a cross-platform behavior decision that belongs to the author,
not to a review-driven patch. Worth a one-line confirmation, e.g. whether GUI
front ends should pass `:gtk`. Behavior question, not a defect.

### 3.6 Verified -- the `lich.rbw` require reordering is behavior-preserving (unchanged)

`constants.rb` (uses only `File`), `version.rb` (pure constant assignments), and
`gemcheck.rb` (requires `bundler` itself) have no dependency on the stdlib block
that now loads after them, so moving them up cannot regress. Unchanged by the
corrective commits.

### 3.7 Verified -- macOS AppleScript escaping is correct for realistic inputs (unchanged)

`alert_macos` uses `body.split("\n").map(&:inspect)`; Ruby's `String#inspect`
emits `\"` and `\\`, matching AppleScript's in-string escaping. Correct for the
ASCII input that flows here.

### 3.8 RESOLVED (documented) -- `missing_gems` rubygems-source assumption

`missing_gems` now carries a YARD note that it relies on installed
`Gem::Specification`s and does not detect git/path-sourced gems
([[gemcheck.rb:56-57](https://github.com/elanthia-online/lich-5/blob/357a31b/lib/gemcheck.rb#L56)](https://github.com/elanthia-online/lich-5/blob/357a31b/lib/gemcheck.rb#L56)).
Still moot for the current all-rubygems Gemfile, but the caveat is now recorded
against a future git-sourced dependency. Closed.

---

## 4. SOLID / DRY (production code)

**Strong**, and unchanged in character by the corrective commits.

- **SRP / fail-soft policy** as before: single-purpose helpers, `safe_call`
  wrapping every fragile lookup, and rescues that degrade rather than raise.
- The two production edits are additive and local; they did not disturb the
  module's structure.

Minor (non-blocking):

- **4.1 TEMP_DIR guard still cosmetically inconsistent.** `build_alert_body`
  guards with `if defined?(TEMP_DIR)` while `write_log` references `TEMP_DIR`
  unguarded (now on two lines: the `File.join` and the new `Dir.mkdir`). The
  outcome remains consistent -- an undefined `TEMP_DIR` raises `NameError`, which
  `write_log`'s `rescue StandardError` absorbs, and `build_alert_body` then omits
  the citation -- so this is harmless. Left as-is is defensible; unifying on one
  convention would be pure polish.
- **4.2 `.sort.uniq`** ([[68](https://github.com/elanthia-online/lich-5/blob/357a31b/lib/gemcheck.rb#L68)](https://github.com/elanthia-online/lich-5/blob/357a31b/lib/gemcheck.rb#L68))
  works; `.uniq.sort` is marginally more idiomatic. Correctly left unchanged --
  no functional difference, not worth the churn.

## 5. SOLID / DAMP (specs)

**Strong.** 64 examples (was 63), all descriptive, all green.

- **5.2 RESOLVED -- the broadened rescue is now covered.** New context
  "when Bundler.setup raises another BundlerError subclass"
  ([[spec:107-124](https://github.com/elanthia-online/lich-5/blob/357a31b/spec/lib/gemcheck_spec.rb#L107)](https://github.com/elanthia-online/lich-5/blob/357a31b/spec/lib/gemcheck_spec.rb#L107))
  raises `Bundler::PathError` -- deliberately neither of the originally-rescued
  classes -- and asserts `alert(missing: [], ...)` + `exit 1`. I confirmed this
  is a genuine regression guard: it fails against the old narrow rescue and
  passes against the broadened one (3.1). Good addition; the setup matches the
  sibling in-scope/out-of-scope contexts so it reads consistently.
- **DAMP / isolation / adversarial coverage** as praised previously: behavior-as-
  name examples, doubled Bundler/Gem/system/platform constant, data-driven
  platform matrix, filesystem-error-swallowed case, append-not-overwrite, and the
  Windows-only release URL.

Minor (unchanged from prior review):

- **5.1 Shared-constant load-order fragility.** The spec defines top-level
  `TEMP_DIR` only `unless defined?` and registers `after(:suite)` cleanup inside
  that guard ([[spec:8-13](https://github.com/elanthia-online/lich-5/blob/357a31b/spec/lib/gemcheck_spec.rb#L8)](https://github.com/elanthia-online/lich-5/blob/357a31b/spec/lib/gemcheck_spec.rb#L8));
  `update_spec_helper.rb:16` uses the identical idiom, so in a full-suite run the
  cleanup registration depends on which file loads first. This matches existing
  repo convention, so it is consistent rather than wrong. Noting for awareness;
  not worth restructuring.

---

## 6. Suggestions (status)

- **6.1 (was: broaden the rescue)** -- DONE (3.1).
- **6.2 (was: ensure temp dir before logging)** -- DONE (3.2).
- **6.3 (confirm `:default`-only scope, or pass `:gtk` for GUI)** -- STILL OPEN;
  author's call (3.5).
- **6.4 (doc the rubygems assumption; `.sort.uniq`)** -- doc note DONE (3.8);
  `.sort.uniq` correctly left alone.

Nothing remaining is a merge blocker.

---

## 7. CodeRabbit findings (status)

All three CodeRabbit items (raised against `140652e`) are now closed:

1. **Rescue clause misses other Bundler errors** -- ADDRESSED at `357a31b`
   (broadened to `Bundler::BundlerError`, 3.1) and covered by a new spec.
2. **TEMP_DIR never cleaned up** -- ADDRESSED (the `after(:suite)` cleanup, still
   present).
3. **kdialog test does not assert `xdg-open`** -- OBSOLETE (the Linux alert is a
   plain `--msgbox`; there is no `xdg-open` call to assert, and the kdialog
   example correctly checks the msgbox invocation).

---

## 8. Verification performed

- Re-fetched `pull/1420/head`, confirmed the new head `357a31b` sits on the
  previously reviewed `e227436` via commits `02bb652` and `357a31b`, and diffed
  both files to confirm the changes are exactly the two fixes + one spec.
- `rspec spec/lib/gemcheck_spec.rb`: **64 examples, 0 failures.**
- `rubocop` with the repo config (including the custom `AsciiOnlySource` cop) on
  `lib/gemcheck.rb` and `spec/lib/gemcheck_spec.rb`: **no offenses.** **0
  non-ASCII bytes** across all three changed files (ASCII-only contract rule).
- **3.1 regression guard proven:** reverting the rescue to the old
  `GemNotFound, GitError` form makes the new example fail (the `PathError`
  propagates out of `verify!`); restoring `BundlerError` makes it pass.
- **3.2 fix proven:** with `TEMP_DIR` pointed at a not-yet-existing directory,
  `write_log` creates the directory, writes the log, and the log cites the
  detected gem -- so the alert's "See ... for details" line is now accurate on a
  fresh install.
- Re-confirmed the boot ordering, the `version.rb` reorder safety, and the
  all-rubygems Gemfile from the prior review remain valid.

**Environment caveat (unchanged):** run under Ruby 3.2.3 (the repo targets 4.0;
the pinned rubocop for 4.0 needs rubygems.org, unreachable in my sandbox, so
rubocop was run at `TargetRubyVersion: 3.2` with the repo's cop set otherwise
intact). The code uses no 4.0-only features, so this does not affect the
findings; a maintainer on the pinned 4.0 toolchain should see the same clean
result.

**Bottom line:** the corrective commits resolve both substantive findings with
minimal, verified, well-guarded changes, and close the open CodeRabbit item.
**Approve.** The only remaining item is a one-line confirmation that gating just
`:default` (not `:gtk`) is intended (3.5).

OSXLich-Doug (approved):

PR #1426, opened 2026-07-02
fix(all): increase read timeout on idle socket to 300 seconds

OSXLich-Doug:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved socket read waiting behavior so game connections are less likely to time out too quickly.
  * Updated related timeout expectations to match the new, longer wait period.

### DIFF
--- a/lib/games.rb
+++ b/lib/games.rb
@@ -238,7 +238,7 @@ module Lich
     # Base Game class with common functionality
     class Game
       # Seconds to wait for readable game socket data before one read timeout.
-      READ_TIMEOUT_SECONDS = 30
+      READ_TIMEOUT_SECONDS = 100
 
       # Consecutive read timeouts allowed before treating the game link as dead.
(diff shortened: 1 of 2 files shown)


OSXLich-Doug:
@coderabbitai - review

lib/games.rb
```diff
@@ -238,7 +238,7 @@ class GameStreamDesyncError < StandardError; end
     # Base Game class with common functionality
     class Game
       # Seconds to wait for readable game socket data before one read timeout.
-      READ_TIMEOUT_SECONDS = 30
+      READ_TIMEOUT_SECONDS = 100
```
MahtraDR, line 241: The PR title says 300 seconds.

mrhoribu: Yes, but it retries 3 times, total of 300 seconds I believe is where that comes in to play

MahtraDR: Ah. Fair.

mrhoribu (approved):

PR #1427, opened 2026-07-03
feat(gs): add creature template library

OSXLich-Doug:


OSXLich-Doug:
@coderabbitai review

Notes: This is largely a data library, most files  are under the creatures/ path and can be checked as template entities.  Logic is in lib/gemstone/creature.rb - I am tracking 4 minor findings that will not likely prevent release.

1) Schema fields accepted but silently dropped (in this PR, the plumbing is in another PR).

2) Data in instance variables may have no reader (the core consumer of this data will be scripts not evidenced in this repo - this is acceptable at this time because no consumer knows of the existance yet).

3) field: nil type entries are converted to 0, yes.  The only problematic one might be height, but again - see 2 above.

4) Lookup key collisions silently overwrite at runtime, yes.  The data is created via a process that detects this condition and any files involved are not released until corrected.  We had one in this run - spectre.rb, which now contains a note concerning the duplicative key issue for its comrade the shadowy spectre.

Anything outside these four findings is welcome.

OSXLich-Doug:
@coderabbitai - review --lib/gemstone && specs/lib/gemstone diff only

OSXLich-Doug:
@coderabbitai - correction for review to spec/lib/gemstone

PR #1429, opened 2026-07-03
fix(all): monkeypatch gi for memoryreleaser / asyncprocessor until patched upstream

OSXLich-Doug:
Patches to mitigate real crash: GC.compact could relocate a Proc gobject-introspection caches outside Ruby's GC visibility for boxed/object type conversion (Gdk::Event, Pango::Attribute), leaving a dangling reference that SIGBUS/SIGSEGVs a live GTK session sometime after periodic compaction runs.

Lich::Util::GtkCompaction pins those converters via Fiddle (rb_gc_register_mark_object) before gtk3 loads, so GC.compact stays safe until a natively-patched gem is detected, or by disabling compaction, if pinning couldn't install in time (not preferred). MemoryReleaser and Combat::AsyncProcessor now route through it, with adversarial spec coverage for the safety branches.

## Summary by CodeRabbit

* **New Features**
  * Added a GTK-safe heap compaction helper and routed compaction through it during shutdown and periodic cleanup.
  * Added early initialization to install the compaction safety behavior before GTK loads.

* **Bug Fixes**
  * Prevented direct heap compaction calls that could interfere with GTK/GObject introspection behavior.
  * Added safety gating and fallback behavior to skip compaction when conditions aren’t safe, with warn-once messaging.

* **Tests**
  * Added/updated RSpec coverage to verify delegation to the safe helper and correct cleanup timing and gating behavior.

### DIFF
--- a/lib/gemstone/combat/async_processor.rb
+++ b/lib/gemstone/combat/async_processor.rb
@@ -5,6 +5,7 @@
 #
 
 require 'concurrent'
+require_relative '../../util/gtk_compaction'
 
 module Lich
   module Gemstone
@@ -62,9 +63,11 @@ module Lich
           @thread_pool.each(&:join)
           @thread_pool.clear
 
-          # Force GC after shutdown to help with memory fragmentation
+          # Force GC after shutdown to help with memory fragmentation.
+          # Compaction is routed through Lich::Util::GtkCompaction, which
+          # keeps it safe to use alongside gtk3.
           GC.start
-          GC.compact if GC.respond_to?(:compact) # Ruby 2.7+
+          Lich::Util::GtkCompaction.safe_compact!
         end
 
         def stats
@@ -100,7 +103,7 @@ module Lich
           # Periodic heap compaction to reduce fragmentation (every hour)
           if GC.respond_to?(:compact) && (Time.now - @last_compact) > 3600
             GC.start
-            GC.compact
+            Lich::Util::GtkCompaction.safe_compact!
             @last_compact = Time.now
             respond "[Combat] Triggered hourly GC compaction" if Tracker.debug?
           end
--- a/lib/util/gtk_compaction.rb
+++ b/lib/util/gtk_compaction.rb
@@ -1,0 +1,191 @@
+require 'fiddle'
+
+module Lich
+  module Util
+    # Keeps Lich's periodic GC.compact calls (MemoryReleaser,
+    # Combat::AsyncProcessor) safe to use alongside gtk3/glib2/gdk3/pango --
+    # on a stock, unpatched ruby-gnome install, with GC.compact still
+    # actually running.
+    #
+    # ruby-gnome's native gems cache a Ruby Proc for each "boxed"/"object"
+    # GType conversion (Gdk::Event, Pango::Attribute, ...) inside a plain C
+    # struct that Ruby's GC can't see or update (BoxedInstance2RObjData /
+    # ObjectInstance2RObjData in gobject-introspection's rb-gi-loader.c).
+    # GC.compact can relocate that Proc without anything updating this one
+    # untracked reference to it, so the next GTK signal marshaled through it
+    # dereferences a dangling pointer and crashes with SIGBUS/SIGSEGV.
+    #
+    # The clean fix is a one-line C patch: pin each converter Proc as a
+    # permanent GC root (rb_gc_register_mark_object) instead of caching a
+    # raw, GC-invisible reference to it. That function is a public,
+    # documented libruby API -- exported from an already-loaded shared
+    # library -- which means we can call it ourselves, from pure Ruby, via
+    # Fiddle, without patching or recompiling anything:
+    #
+    #   1. GObjectIntrospection::Loader.register_boxed_class_converter and
+    #      .register_object_class_converter are the same *public* methods
+    #      gdk3/pango/etc. call to install these converters in the first
+    #      place. We wrap them so our wrapper captures the block as `&block`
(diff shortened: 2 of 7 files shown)


OSXLich-Doug:
@coderabbit review PR

MahtraDR:
Really nice work on this one -- the core mechanism is clever and I came away more convinced of it after digging in. Pinning the converter Procs from pure Ruby via a public libruby API, gating on a native-patch fallback, and writing an explicit RETIREMENT section with a concrete removal condition is exactly the right shape for a stopgap. I did an adversarial pass (grounded in code + a few live experiments against a real GTK stack) and wanted to share what held up and the one thing that needs fixing before merge.

## The blocker: the fallback's tests are green in CI only because the GTK gems aren't there

`spec/lib/util/gtk_compaction_spec.rb:121` and `:141` -- the only two examples that exercise the *positive* path of `native_gem_patched?` -- fail deterministically on any machine that actually has the `gtk`/`gobject-introspection` gems installed:

```
77 examples, 2 failures
rspec ./spec/lib/util/gtk_compaction_spec.rb:121 # returns true when the loaded gem source carries the patch marker
rspec ./spec/lib/util/gtk_compaction_spec.rb:141 # matches a .so extension the same as .bundle
```

Root cause, confirmed step by step:

1. `install!` runs as a side effect of requiring the file, and calls `require 'gobject-introspection'`. Merely requiring the module for the spec injects the *real* gem's bundle into `$LOADED_FEATURES` (verified: empty before the require, real path after).
2. `native_gem_patched?` does `$LOADED_FEATURES.find { ... }` -- first match wins. The real (stock, unpatched) gem precedes the fake path the test appends, so `find` returns the real one, reads its unpatched source, and returns `false`.

CI is green because `rspec_tests.yaml` sets `BUNDLE_WITHOUT: 'gtk:vscode:profanity'`. With the gem absent, `require 'gobject-introspection'` raises `LoadError`, `$LOADED_FEATURES` stays clean, the fake is the only match, and the tests pass. I verified both directions: fake-only -> `true`, real-present -> `false`. So the branch that decides whether we compact-or-crash on an unpinned stock gem is effectively unvalidated in CI and broken on our own dev boxes.

Fix is small: give `native_gem_patched?` a seam that takes the candidate path (or stub `$LOADED_FEATURES` to contain only the fake) instead of relying on array order, and don't let `install!` fire as a require-time side effect inside specs. Happy to push these spec fixes to a branch if useful.

## Smaller things worth a look

- **Eager `require` at load time.** `install!` running at file-require means you can't load the module (in a test, a tool, a script) without pulling GTK's introspection layer into the process. `init.rb`'s `require 'gtk3'` is gated on `--no-gtk`/no-DISPLAY, so on a headless box that happens to have the gem, this now loads gi at boot when it previously wouldn't -- and `gtk_loaded?` then reports `true` in a `--no-gtk` session. Compaction still works, so it's minor, but it's also the mechanism behind the test pollution above.
- **Silent permanent no-op.** In the one case that matters (gtk loaded, not pinned, not natively patched), `safe_compact!` returns `nil` with no signal, so compaction is disabled for the whole session and nothing shows in the logs. A one-time warn there would save a future debugging session. (Contrast the hourly path, which logs the happy case.)
- **`native_gem_patched?` greps the `.c` source, not the compiled binary.** Source on disk can diverge from the loaded `.bundle` (patched source, stale binary, or vice versa). Fine as a last-resort heuristic -- worth a comment noting the source != binary assumption.
- **DRY:** the two `define_method` wrappers are identical modulo boxed/object; a small loop over the two method names drops the duplication (and the paired alias lines).
- **Minor:** the only callers of these converters in the current ruby-gnome stack are gdk3 (rectangle, Event) and pango (Attribute) -- all *boxed*. Nothing calls `register_object_class_converter`, and both named examples in the module doc (`Gdk::Event`, `Pango::Attribute`) are boxed too, so the object-converter half is defensive/untested against real gems. Not a bug, just noting the doc reads as if there's an object example when there isn't.

## What checked out (credit where due)

- The wrapper is on the correct receiver and routes subclass calls: both `Gdk::Loader` and `Pango::Loader` resolve `register_boxed_class_converter` to the Ruby wrapper via singleton-class inheritance, and I watched `Pango::Attribute` actually get intercepted.
- Ordering is sound: no converters register during the base `gobject-introspection` require; they all register during `require 'gtk3'`, which runs after `install!` in `lich.rbw`. (I'd initially worried about a pre-wrapper gap here -- there isn't one.)
- Fiddle usage is correct: `Fiddle.dlwrap(block)` yields the object's `VALUE`, passed as `UINTPTR_T` to `rb_gc_register_mark_object(VALUE)`; `rb-gi-loader.c` confirms these are singleton methods taking `(gtype)` + block, matching the wrapper. The blanket rescue for JRuby/TruffleRuby/renamed-symbol is the right call.
- No call sites missed (`GC.compact` lived only in `memoryreleaser.rb` and `async_processor.rb`, both routed through `safe_compact!`), no public methods removed, `ruby -c` clean on all 5 files, rubocop 0 offenses.

Net: the design is good and I'd be glad to see it land -- just want the fallback's tests to actually exercise the fallback (and go red when it regresses) before we rely on it. Let me know if you want me to take the spec fixes.

OSXLich-Doug:
Excellent feedback, actioned.  Thank you.

@coderabbitai - pull all context of offered findings and observations, ensure your prior comments were addressed and closed.  Then, review diff in response to @MahtraDR feedback.

MahtraDR (approved):

mrhoribu (approved):

PR #1431, opened 2026-07-04
feat(all): setup_files: support scripts/data/custom for user data overrides

MahtraDR:
## Summary

Lets users drop their own data files in `scripts/data/custom/` that take precedence over the shipped `scripts/data/` files, mirroring how `scripts/custom/` already takes precedence over `scripts/` when loading scripts.

Today, a user who wants to tweak a shipped `base-{type}.yaml` (e.g. `base-items.yaml`) has to edit it in place, which conflicts on every update. With this change they can place `scripts/data/custom/base-items.yaml` and it wins without touching the tracked file.

## How it works

`SetupFiles#get_data_glob_patterns` now scans `scripts/data/custom` before `scripts/data`. Because `load_files` keys its cache by basename with first-match-wins (`||=`), the custom file overrides the base one. Custom-only data files with no base equivalent are also picked up.

The custom directory is entirely optional: `Dir.glob` on a missing path returns nothing, so behavior is unchanged when no custom data exists. The cache also self-heals back to the base file if a custom file is later removed and `reload` runs.

## Tests

Added specs to `setup_files_spec.rb` covering:
- custom file overrides base
- custom-only file with no base equivalent
- fallback to base when no custom file exists
- fallback to base after a custom file is deleted

All existing specs still pass; rubocop clean on both touched files.

## Summary by CodeRabbit

* **Bug Fixes**
  * User-added data files now take priority over bundled data when both exist.
  * The app now correctly falls back to bundled data if no custom file is present.
  * Reloading data updates the active file selection after custom files are removed.

* **Tests**
  * Added coverage for data loading with a custom data directory and priority-based file selection.

### DIFF
--- a/lib/common/setup_files.rb
+++ b/lib/common/setup_files.rb
@@ -133,6 +133,13 @@ module Lich
         @scripts_data_path ||= File.join(SCRIPT_DIR, 'data')
       end
 
+      # Lazy memoized path -- user-supplied data files that take precedence
+      # over the shipped base data files, mirroring how scripts/custom takes
+      # precedence over scripts/ when loading scripts.
+      def scripts_data_custom_path
+        @scripts_data_custom_path ||= File.join(scripts_data_path, 'custom')
+      end
+
       # Returns the character name for filename construction.
       # Prefers Account.character (available from authentication, before XML stream)
       # with fallback to checkname (XMLData.name, available after XML stream starts).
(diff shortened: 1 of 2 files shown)


MahtraDR:
@BinuDR Have a test of this one?

MahtraDR:
@coderabbitai review in full

mrhoribu:
@coderabbitai review in full

mrhoribu (approved):

PR #1423, opened 2026-07-02
fix(all): pin lich directory for gemfile check

OSXLich-Doug:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved app startup to reliably use the correct Gemfile even when launched from a different working directory.
  * Preserves any existing Bundler Gemfile setting, avoiding unexpected environment changes.

PR #1432, opened 2026-07-06
fix(all): keep Gemfile.lock in sync on self-update and tolerate stale locks at boot

MahtraDR:
## Problem

A self-update (`lich5-update --update` / `--branch=...`) copies `Gemfile` into the install but **not** `Gemfile.lock` (`TOP_LEVEL_FILES = %w[Gemfile LICENSE]`). Whenever the `Gemfile` changed, the leftover lock is now stale relative to it.

At boot, `GemCheck.verify!` forces a Bundler re-resolution (`Bundler.definition(true)`). Against a stale, multi-platform `Gemfile.lock`, that resolve runs **offline** (`Bundler.setup` only sees locally installed gems) and fails even though every required gem is installed for the current platform:

```
Bundler::GemNotFound
Could not find gems matching 'ffi' valid for all resolution platforms
(aarch64-linux, arm-linux, arm64-darwin, x64-mingw-ucrt, x86-linux,
 x86_64-darwin, x86_64-linux) in locally installed gems.
  The source contains the following gems matching 'ffi':
    * ffi-1.17.4-x86_64-linux-gnu
```

Because `ffi` is in the `:default` group, the error is treated as fatal and Lich `exit 1`s. The GemCheck log even shows `Missing gems (detected): none` and every dependency `OK` -- the gems are present; only the lockfile can't be re-resolved. The user is left unable to log in, with no in-app path to recover (the fix requires a shell `bundle install`).

## Fix (defense in depth)

**1. `ReleaseInstaller` keeps the lock in sync.** `Gemfile.lock` is copied alongside `Gemfile` so the resolved dependency set stays consistent after an update. Validation moves to its own `REQUIRED_ARCHIVE_ITEMS` constant so the optional lock never *gates* an update (older release tarballs may omit it), and the copy loop is extracted into a testable `copy_top_level_files`.

**2. `GemCheck.verify!` degrades gracefully.** When a Bundler resolution error occurs but the platform-aware `missing_gems` detector confirms every required gem is installed, GemCheck now logs the failure, warns the user to run `bundle install`, and **continues boot** on plain RubyGems activation (how Lich loaded before GemCheck adopted `Bundler.setup`) instead of exiting. Genuinely missing gems still alert and exit as before.

Either fix alone resolves the reported failure; together they also cover bundler-version skew and dead-platform locks from any source.

## Testing

- `rubocop -A` clean on all changed files (0 offenses).
- New `spec/lib/common/update/release_installer_spec.rb`: lock copy + structure validation (incl. optional-lock tolerance).
- Updated `spec/lib/gemcheck_spec.rb`: tolerate-and-continue behavior + `warn_stale_lock`.
- Full update suite + gemcheck: **100 + n examples, 0 failures**.

## Summary by CodeRabbit

* **New Features**
  * Updates now handle partial release archives more gracefully, copying available top-level files while continuing even if optional files are missing.
  * Bundler setup failures are now treated more leniently when dependencies are otherwise intact, allowing startup to continue with a warning.

* **Bug Fixes**
  * Prevents updates from failing when optional archive files like `Gemfile.lock` are absent.
  * Adds a clearer warning path for stale or mismatched lockfile situations instead of stopping the app.

### DIFF
--- a/lib/common/update/release_installer.rb
+++ b/lib/common/update/release_installer.rb
@@ -12,9 +12,23 @@ module Lich
   module Util
     module Update
       class ReleaseInstaller
-        # Top-level files (besides lib/ and lich.rbw) to copy from release archive.
-        # lich.rbw is handled separately due to its dynamic target name.
-        TOP_LEVEL_FILES = %w[Gemfile LICENSE].freeze
+        # Top-level files (besides lib/ and lich.rbw) copied verbatim from the
+        # release/branch archive into LICH_DIR during a self-update. lich.rbw is
+        # handled separately due to its dynamic target name.
+        #
+        # Gemfile.lock travels alongside Gemfile so the resolved dependency set
+        # stays consistent with the Gemfile after an update. When only Gemfile is
+        # replaced, the leftover lock is stale relative to the new Gemfile, which
+        # forces Bundler to re-resolve at boot; against a multi-platform lock that
+        # re-resolution can fail even though every required gem is installed.
+        # Entries absent from the archive are skipped (see #copy_top_level_files).
+        TOP_LEVEL_FILES = %w[Gemfile Gemfile.lock LICENSE].freeze
+
+        # Archive entries that must exist for an extracted download to count as a
+        # structurally valid Lich installation. Deliberately distinct from
+        # TOP_LEVEL_FILES so that optional payload (e.g. Gemfile.lock, which older
+        # release tarballs may omit) never gates an update.
+        REQUIRED_ARCHIVE_ITEMS = %w[lib lich.rbw Gemfile LICENSE].freeze
 
         # @param client [GitHubClient] GitHub API client instance
         # @param resolver [ChannelResolver] channel resolver instance
@@ -255,14 +269,7 @@ module Lich
           respond "All Lich lib files have been updated."
           respond
 
-          # Copy top-level release files to LICH_DIR
-          TOP_LEVEL_FILES.each do |filename|
-            src = File.join(source_dir, filename)
-            if File.exist?(src)
-              FileUtils.cp(src, File.join(LICH_DIR, filename))
-              respond "Updated #{filename}."
-            end
-          end
+          copy_top_level_files(source_dir)
 
           file_updater = FileUpdater.new(@client, @resolver)
           file_updater.update_core_data_and_scripts(version)
@@ -273,13 +280,30 @@ module Lich
           true
         end
 
-        # Validates extracted directory contains required Lich files.
+        # Copies each top-level file present in the extracted archive into
+        # LICH_DIR. Files missing from the archive are skipped rather than
+        # treated as errors, so a partial archive (e.g. one without Gemfile.lock)
+        # updates cleanly instead of aborting.
+        #
+        # @param source_dir [String] extracted tarball directory
+        # @return [void]
+        def copy_top_level_files(source_dir)
+          TOP_LEVEL_FILES.each do |filename|
+            src = File.join(source_dir, filename)
+            next unless File.exist?(src)
+
+            FileUtils.cp(src, File.join(LICH_DIR, filename))
+            respond "Updated #{filename}."
+          end
+        end
+
+        # Validates the extracted directory contains every archive entry Lich
+        # requires to install safely.
         #
         # @param dir [String] directory to check
-        # @return [Boolean] true if valid
+        # @return [Boolean] true if all required entries are present
         def validate_lich_structure(dir)
-          required_items = ['lib', 'lich.rbw'] + TOP_LEVEL_FILES
-          required_items.all? { |item| File.exist?(File.join(dir, item)) }
+          REQUIRED_ARCHIVE_ITEMS.all? { |item| File.exist?(File.join(dir, item)) }
         end
 
         # Checks if current Ruby meets minimum version requirement.
--- a/lib/gemcheck.rb
+++ b/lib/gemcheck.rb
@@ -14,13 +14,25 @@ module Lich
     TITLE           = 'Lich5: Missing Ruby Gems'
     RELEASE_URL     = 'https://github.com/elanthia-online/lich-5/releases/latest'
     LOG_FILENAME    = 'lich5-missing-gems.log'
+    STALE_LOCK_MESSAGE = 'Lich: Gemfile.lock could not be resolved, but every required gem ' \
+                         'is installed; continuing boot. Run \'bundle install\' from your ' \
+                         'Lich5 folder to refresh the lockfile.'
 
     module_function
 
     # Runs Bundler.setup for the given groups; on failure, alerts the
-    # user and exits. If Bundler complains about a gem outside the
-    # requested groups (which some Bundler versions do despite `without`
-    # settings), the complaint is silently ignored and boot continues.
+    # user and exits. Two classes of Bundler failure are tolerated instead of
+    # being fatal, because in both the gems Lich actually needs are present:
+    #
+    # 1. Bundler complains about a gem outside the requested groups (which some
+    #    Bundler versions do despite `without` settings).
+    # 2. Bundler cannot resolve the lockfile, yet our platform-aware detector
+    #    confirms every required gem is installed -- typically a stale,
+    #    multi-platform Gemfile.lock re-resolved offline after a self-update
+    #    replaced Gemfile but not Gemfile.lock. Boot proceeds on plain RubyGems
+    #    activation (how Lich loaded before GemCheck adopted Bundler.setup),
+    #    with a non-fatal notice pointing at `bundle install`.
+    #
     # @param groups [Array<Symbol>] Bundler groups to verify
     # @return [void]
     def verify!(*groups)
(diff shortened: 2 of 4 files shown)


MahtraDR:
@coderabbitai review in full

MahtraDR:
Thanks @coderabbitai — good catch on the diagnostic. Fixed in 48e96bfa: `warn_stale_lock` now takes `groups` and forwards it to `write_log`, so the log's "Groups checked" line matches the actual `verify!` invocation, consistent with the fatal `alert` path. Behavior is unchanged; added a spec asserting non-default groups reach the log.

mrhoribu:
This will need us to update the release builder Github action as well as our `.gitignore` to include the Gemfile.lock file to work properly won't it? Currently we don't ship the lock file with Lich releases, we'd have to include that in the .zip release package and remove from the .gitignore. Originally thinking is let the client machine use `bundle install` and create their own lock file based on whatever sources their Ruby gets. We can definitely change it up. Developers or those that want bleeding edge are likely going to be smart enough to delete their lock file to do bundle install to get the latest version of the Gems instead of older versions specified from lock file.

OSXLich-Doug:
Good catch on release packaging: we need to update stable, prerelease (maybe?), and rebuild release asset workflows to include Gemfile.lock in both .tar.gz and .zip. 

But .gitignore does not need changing because Gemfile.lock is already tracked and not ignored. Shipping the lockfile is consistent with treating Lich releases as a reproducible app/runtime package; advanced users can still opt out locally.

I say 'maybe?' about prerelease because given the update.rb capability of loading a full branch, all the activities in prerelease may actually be cruft at ths point.  Something I'm weighing anyway.

If this PR comes to fruition, happy to take on updating the CI to match expectations.  Or it can be included directly here as part of the change.  However, I'm getting ready to put up a pretty hefty CI change anyway - to cover the gem / Installer factory and reconsumption to save time.  A commit to cover will not be a burden.

OSXLich-Doug:
However, there is this. . . 

### Findings

Major: [lib/gemcheck.rb (line 60)](/Users/doug/dev/test/lich-5/lib/gemcheck.rb:60) now treats missing_gems(groups).empty? as proof that boot can safely continue after any Bundler::BundlerError, but [missing_gems (line 102)](/Users/doug/dev/test/lich-5/lib/gemcheck.rb:102) only checks direct Gemfile dependencies and returns [] on its own Bundler/introspection failures at [line 108 (line 108)](/Users/doug/dev/test/lich-5/lib/gemcheck.rb:108). That means this PR can now tolerate real Bundler failures that are not stale-lock recovery cases, such as a missing transitive dependency, malformed Gemfile/lockfile, Ruby/platform incompatibility, or another Bundler failure where direct declared gems happen to be installed. Before this PR, those errors alerted and exited; now they log a “every required gem is installed” warning and continue boot.

The spec at [spec/lib/gemcheck_spec.rb (line 132)](/Users/doug/dev/test/lich-5/spec/lib/gemcheck_spec.rb:132) actually locks in the broad behavior by asserting that other BundlerError subclasses are tolerated whenever the mocked detector returns empty. I’d narrow the continue path so “detector found no missing gems” is not the only gate. At minimum, distinguish detector success from detector failure, and only tolerate the specific stale multi-platform lock resolution shape this PR is targeting. A regression test should cover Bundler.setup raising for an unknown/transitive gem while missing_gems returns []; that should remain fatal.

### No Other Blocking Findings
ReleaseInstaller looks sound: splitting TOP_LEVEL_FILES from REQUIRED_ARCHIVE_ITEMS preserves older archive compatibility while allowing Gemfile.lock to be copied when present. The method extraction is small and in-family with the existing code.

Docs/comments are mostly useful, but the GemCheck comments and STALE_LOCK_MESSAGE inherit the correctness problem above: they state “every required gem is installed,” while the implementation only proves direct declared gems are installed, and sometimes proves nothing if missing_gems rescues.
Tests are otherwise good around lockfile copying, optional lock validation, group forwarding, and alert preservation for the explicitly stubbed missing-gem path.

Verification

Ran with rbenv exec:
bundle exec rspec spec/lib/gemcheck_spec.rb spec/lib/common/update/release_installer_spec.rb: 81 examples, 0 failures
bundle exec rspec spec/lib/common/update spec/lib/gemcheck_spec.rb: 173 examples, 0 failures
bundle exec ruby -c lib/gemcheck.rb: Syntax OK
bundle exec ruby -c lib/common/update/release_installer.rb: Syntax OK
git diff --check origin/main...HEAD: clean

No files changed.

mrhoribu:
Err meant `.gitattributes` not `.gitignore`. Error on my part.

mrhoribu:
Actually, the entire usage of `Bundler.setup` is causing a bit of an issue for my on Linux. It is preventing usage of gtk/discordrb/etc any gem not in core Ruby or part of the gemfile is now not part of the $LOAD_PATH. Working with Claude to come up with a resolution for it.

MahtraDR:
> This will need us to update the release builder Github action as well as our `.gitignore` to include the Gemfile.lock file to work properly won't it? Currently we don't ship the lock file with Lich releases, we'd have to include that in the .zip release package and remove from the .gitignore. Originally thinking is let the client machine use `bundle install` and create their own lock file based on whatever sources their Ruby gets. We can definitely change it up. Developers or those that want bleeding edge are likely going to be smart enough to delete their lock file to do bundle install to get the latest version of the Gems instead of older versions specified from lock file.

I did figure out what the issue was eventually, but the way the problem manifested was extremely non obvious. As it was, as a user, I ran my lich update, I ran the bundle install for the gems I didn't have, and still couldn't log in. Gem management isn't my strong suit and I think I already did more than regular users would have. I think this issue will generate queries on Discord.

mrhoribu:
Good — that's the right thing to interrogate. Before I propose the shape of a verify-without-locking boot, I want to know *why* `Bundler.setup` was adopted in the first place, so we don't quietly regress whatever it was meant to fix. Let me trace the history and the current engine load path.The history makes the design intent clear. Let me pull the introducing commit's version and then validate the core safety property empirically.Good call, and the history backs it. GemCheck came in through your own #1420, and its stated job is narrow — "a boot-time check to verify required Ruby gems before startup" plus a clear alert when something's missing. `Bundler.setup` was just the mechanism you reached for; locking the load path was never the goal, it's a side effect. And it's the side effect doing all the damage: it breaks runtime-gem scripts (discordrb) *and* produces the stale-lock false-fatal that #1432 is trying to paper over. #1432 is really you backing away from setup's strictness one rescue arm at a time — (c) removes the root cause instead.

The structural fact that makes this clean: verification is already independent of locking. `missing_gems` (gemcheck.rb:72) reads declared deps from `Bundler.definition.current_dependencies` and checks each against `Gem::Specification.find_all_by_name` — Gemfile-as-source-of-truth plus RubyGems, no `Bundler.setup` involved. The only thing `Bundler.setup(*groups)` adds at line 52 is activating/pinning the bundle, i.e. the load-path takeover. I confirmed setup appears nowhere else in the tree and nothing downstream depends on the bundle being active — the engine loads its own gems with plain `require 'ox'`, `require 'sequel'`, etc. via RubyGems, which works with or without setup.

I validated the safety property empirically: with a Gemfile listing one installed gem and one genuinely missing gem, and no `Bundler.setup`, the `missing_gems` logic still reports only the missing one, and a plain `require` of the installed gem succeeds. So dropping setup keeps "tell the user what's missing and exit" fully intact while restoring runtime-gem loading.

Concretely, `verify!` collapses to detection plus alert:

```ruby
def verify!(*groups)
  groups = [:default] if groups.empty?
  configure_gemfile!

  missing = missing_gems(groups)
  return if missing.empty?

  alert(missing: missing, groups: groups)
  exit 1
end
```

That keeps `configure_gemfile!` (still needed — `missing_gems` calls `Bundler.definition`, which needs `BUNDLE_GEMFILE` pinned per your #1423 fix), `missing_gems`, and the entire alert/log stack unchanged. What falls out as dead code once setup is gone is the `Bundler::BundlerError` rescue and everything that exists only to interpret it: `bundler_error_out_of_scope?`, `extract_gem_name`, `warn_stale_lock`, and `STALE_LOCK_MESSAGE`.

The honest tradeoffs of not locking:

- We give up Bundler's runtime version pinning. RubyGems will auto-activate the newest installed version of each gem, which only diverges from the lock when multiple versions of a gem coexist; on a normal Ruby4Lich5 / `bundle install` machine there's one of each, so it's a no-op — and it's exactly how Lich loaded for years before #1420.
- We stop pre-flighting the full transitive graph. `missing_gems` checks declared deps only, so a broken transitive dependency would surface as a LoadError at its `require` site rather than at the GemCheck alert. It still fails loudly, and this was never the case GemCheck targeted.
- Scripts can load gems outside the Gemfile — which is the whole point, but it does mean less isolation.

Given Lich is an engine hosting arbitrary scripts that self-install gems, that flexibility is effectively a requirement, and the isolation was never really compatible with the model.

Three calls I want from you before I write anything:

1. **#1432.** (c) supersedes its GemCheck half — no setup means no forced re-resolution, so the stale-lock tolerance becomes moot. Its `ReleaseInstaller` change (copying `Gemfile.lock` on self-update) is still good, independent hygiene. Trim #1432 to just the ReleaseInstaller fix and land (c) separately, or fold both into one PR?
2. **Dead-code removal.** Remove the now-unused BundlerError helpers (larger but clean diff) or leave them (smaller diff)? I'd lean remove.
3. **Version pinning.** Confirm you're fine giving up runtime pinning — I think it's the right trade here, per above.

Answer those and I'll draft the work unit plus the gemcheck change, including the `gemcheck_spec.rb` updates (the existing spec asserts the tolerate-and-continue path, so those examples change with it).

PR #1433, opened 2026-07-06
fix(all): Mapdb sort tags by downcase and original case

mrhoribu:
Resolves issue where two tags that are identical letters (ex: Pawnshop, pawnshop) could end up being swapped back and forth when saving, causing issues with DIFF showing changes for just array placement of tags.


mrhoribu:
@coderabbitai review in full

OSXLich-Doug (approved):

PR #1434, opened 2026-07-06
fix(all): stop locking the load path at boot; keep Gemfile.lock in sync on self-update

mrhoribu:
Supersedes #1432. This keeps that PR's `Gemfile.lock`-in-sync fix but replaces its
"tolerate a failed `Bundler.setup`" approach with removing the boot-time `Bundler.setup`
altogether. GemCheck goes back to being a presence check, and boot no longer locks the
load path. Recommend closing #1432 in favor of this.

## Problem

Two failures, one root cause.

Since #1420, `GemCheck.verify!` runs `Bundler.setup(:default)` at boot. `Bundler.setup`
does not merely confirm gems are present -- it **locks the process to the bundle**: it
prunes `$LOAD_PATH` (and the in-memory `Gem::Specification` set) down to the gems declared
in Lich's `Gemfile`. Any installed gem that is not in the `Gemfile` becomes unrequireable.

**1. Runtime-gem scripts break.** Community scripts install and `require` their own gems at
runtime (e.g. `zdiscord` / `discordhook` -> `discordrb`). `discordrb` is not in Lich's
`Gemfile`, so once boot clears GemCheck the `require` fails even though the gem is installed:

    --- Lich: error: cannot load such file -- discordrb
            .../ruby/4.0.0/bundled_gems.rb:60:in 'Kernel.require'
            .../ruby/4.0.0/bundled_gems.rb:60:in 'block (2 levels) in Kernel#replace_require'
    --- Lich: zdiscord has exited.

`gem list` shows the gem, and the scripts' own installed-gem guard
(`Gem::Specification.map(&:name)`) sees it too -- both ignore Bundler -- so the scripts skip
reinstalling and go straight to the failing `require`. The `bundled_gems.rb` frames are just
Ruby 4.0's `require` wrapper; the actual cause is the pruned load path.

**2. Stale lockfiles false-fatal at boot** (the original #1432 report). A self-update replaced
`Gemfile` but not `Gemfile.lock`; `verify!`'s `Bundler.definition(true)` re-resolution then
failed offline against the multi-platform lock and `exit 1`ed, even though every required gem
was installed for the current platform:

    Bundler::GemNotFound
    Could not find gems matching 'ffi' valid for all resolution platforms
    (aarch64-linux, arm-linux, ...) in locally installed gems.

#1432 addressed (2) by catching the `Bundler.setup` failure and continuing on plain RubyGems.
But the deeper issue is that `Bundler.setup` -- locking the load path -- is fundamentally
incompatible with an engine that hosts scripts installing gems at runtime. It is the source of
both failures, and tolerating it one exception at a time (as #1432 did) leaves (1) unsolved.
No post-`setup` reactivation from a script fixes it either: `gem "discordrb"` raises
`not part of the bundle`, and `Gem::Specification.find_by_name(...).activate` raises
`MissingSpecError` even after `Gem::Specification.reset`, because Bundler has replaced the
registry.

## Fix

**1. `GemCheck.verify!` stops locking the load path.** It stays a presence check: read the
declared dependencies for the requested groups, confirm each is installed (via
`Gem::Specification`, platform-correct for the current machine), and alert + `exit 1` if any
are missing. It no longer calls `Bundler.setup`, `Bundler.definition(true)`, or
`Bundler.settings.temporary`. Boot then proceeds on plain RubyGems activation -- how Lich
loaded before #1420 -- so scripts can again `require` gems they install at runtime. Genuinely
missing gems still alert and exit exactly as before.

This also makes the #1432 failure **impossible** rather than merely tolerated: with no forced
re-resolution and no `setup`, a stale lock can no longer fail boot. The tolerance code that is
now dead is removed: `warn_stale_lock`, `STALE_LOCK_MESSAGE`, `bundler_error_out_of_scope?`,
and `extract_gem_name`.

**2. `ReleaseInstaller` keeps `Gemfile.lock` in sync on self-update** (retained from #1432).
`Gemfile.lock` is copied alongside `Gemfile` so the on-disk resolved set stays consistent for a
later `bundle install` / dev use. Structural validation lives in its own
`REQUIRED_ARCHIVE_ITEMS` constant so the optional lock never gates an update (older release
tarballs may omit it), and the copy loop is extracted into the testable `copy_top_level_files`.
Its comment is updated to drop the now-inaccurate "forces re-resolution at boot" rationale.

Detection itself is unaffected: `missing_gems` reads `Bundler.definition.current_dependencies`
(Gemfile as source of truth) and checks `Gem::Specification`, needing no `setup`; it fails open
if Bundler cannot parse the Gemfile, so a broken Gemfile never blocks login.

### Trade-off

Not locking means RubyGems auto-activates the newest installed version of each gem rather than
the exact lockfile pin. This only diverges when multiple versions of a gem coexist; on a normal
Ruby4Lich5 / `bundle install` machine there is one of each, and this is how Lich loaded for
years before #1420. Transitive-only breakage is no longer pre-flighted by GemCheck, but it still
surfaces as a clear `LoadError` at the offending `require`.

## Testing

- New/updated specs pass: **65 examples, 0 failures** across `spec/lib/gemcheck_spec.rb` and
  `spec/lib/common/update/release_installer_spec.rb`, verified on a clean `main` checkout with
  all changed files applied.
- `spec/lib/gemcheck_spec.rb` rewritten for the presence-check contract: missing -> alert +
  `exit 1`; nothing missing -> returns and never calls `Bundler.setup`; group forwarding;
  `configure_gemfile!` invoked.
- `spec/lib/common/update/release_installer_spec.rb` unchanged (lock copy + optional-lock
  tolerance).
- `ruby -c` clean; ASCII-only clean (`Custom/AsciiOnlySource`).
- `rubocop -A` clean on all changed files.

## Summary by CodeRabbit

* **New Features**
  * Update installs now copy `Gemfile.lock` along with other top-level files when available.

* **Bug Fixes**
  * Installation checks are less strict, allowing partial update archives to complete when optional files are missing.
  * Gem dependency verification now focuses on missing gems for the selected groups and exits cleanly when issues are found.

* **Tests**
  * Updated coverage to match the new update and gem-check behavior.

### DIFF
--- a/lib/common/update/release_installer.rb
+++ b/lib/common/update/release_installer.rb
@@ -12,9 +12,22 @@ module Lich
   module Util
     module Update
       class ReleaseInstaller
-        # Top-level files (besides lib/ and lich.rbw) to copy from release archive.
-        # lich.rbw is handled separately due to its dynamic target name.
-        TOP_LEVEL_FILES = %w[Gemfile LICENSE].freeze
+        # Top-level files (besides lib/ and lich.rbw) copied verbatim from the
+        # release/branch archive into LICH_DIR during a self-update. lich.rbw is
+        # handled separately due to its dynamic target name.
+        #
+        # Gemfile.lock travels alongside Gemfile so the resolved dependency set
+        # on disk stays consistent with the Gemfile after an update. Shipping a
+        # new Gemfile without its lock would leave a stale lock that a later
+        # `bundle install` (or any Bundler invocation) has to re-resolve.
+        # Entries absent from the archive are skipped (see #copy_top_level_files).
+        TOP_LEVEL_FILES = %w[Gemfile Gemfile.lock LICENSE].freeze
+
+        # Archive entries that must exist for an extracted download to count as a
+        # structurally valid Lich installation. Deliberately distinct from
+        # TOP_LEVEL_FILES so that optional payload (e.g. Gemfile.lock, which older
+        # release tarballs may omit) never gates an update.
+        REQUIRED_ARCHIVE_ITEMS = %w[lib lich.rbw Gemfile LICENSE].freeze
 
         # @param client [GitHubClient] GitHub API client instance
         # @param resolver [ChannelResolver] channel resolver instance
@@ -255,14 +268,7 @@ module Lich
           respond "All Lich lib files have been updated."
           respond
 
-          # Copy top-level release files to LICH_DIR
-          TOP_LEVEL_FILES.each do |filename|
-            src = File.join(source_dir, filename)
-            if File.exist?(src)
-              FileUtils.cp(src, File.join(LICH_DIR, filename))
-              respond "Updated #{filename}."
-            end
-          end
+          copy_top_level_files(source_dir)
 
           file_updater = FileUpdater.new(@client, @resolver)
           file_updater.update_core_data_and_scripts(version)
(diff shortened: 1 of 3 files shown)


mrhoribu:
@coderabbitai review

MahtraDR (approved):

PR #1436, opened 2026-07-07
feat(all): Modify running? method to support optional exact matching of script names

vtcifer:
Updated running? method to include exact_match parameter for stricter matching of script names.
Preserves default behavior of partial (starts with) matching.

## Summary by CodeRabbit

* **New Features**
  * Added an option for more precise script-name checks, allowing exact matching when needed.
  * Existing matching behavior remains unchanged by default, continuing to support partial name matching.

### DIFF
--- a/lib/global_defs.rb
+++ b/lib/global_defs.rb
@@ -131,8 +131,11 @@ def stop_script(*target_names)
   end
 end
 
-def running?(*snames)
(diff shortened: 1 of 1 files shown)


mrhoribu:
@coderabbitai review

mrhoribu (approved):

PR #1438, opened 2026-07-08
feat(all): add roomlinks/roommono toggles for room display

MahtraDR:
## Context

Core Lich now renders room numbers and exits natively, replacing the retired dr-scripts `roomnumbers.lic` ([dr-scripts#7458](https://github.com/elanthia-online/dr-scripts/pull/7458)). Core's output differs from the old script in two visible ways:

- **Font** - the old script emitted via `respond`, which wraps output in `<output class="mono"/>`, so it rendered fixed-width. Core injects text straight into the stream, so it renders in the proportional game font.
- **Links** - core wraps exits in clickable `<d cmd='...'>` tags; the old script emitted plain text.

## What this adds

Two independent, persisted toggles:

| Command | Controls |
|---|---|
| `;display roomlinks` | exits as clickable command links vs plain text |
| `;display roommono` | Lich room lines in fixed-width mono vs the game font |

**Game-aware defaults so nothing regresses:** DragonRealms defaults to the classic look (mono + plain text, matching `roomnumbers.lic`); GemStone and any other game keep the current core look (proportional + clickable links).

The two toggles are independent - mono lines can still contain live links.

## Refactor (DRY/SOLID)

The near-duplicate StringProc/exit rendering that lived in both the GemStone and DragonRealms `#process_room_display` methods is extracted into a shared, game-agnostic `GameBase::RoomFormatter` mixin - one source of truth for the toggle behavior, the obvious-exit pattern, and the room-window frontend list. Each game keeps only its own tail (GS room-window echo, DR room-number line). No methods removed; no existing signatures changed.

## Hardening

A dangling `wayto` reference (destination room missing from the mapdb) previously raised `NoMethodError` inside the downstream hook; it now skips that exit gracefully.

## Tests

Adds adversarial, DAMP specs: module units on a throwaway host class (styling, link/plain rendering, routable-only filtering, the `is_a?(StringProc)` quirk, dangling-ref guard, mono+links coexistence) plus GS/DR parity and integration through `process_room_display`. Full suite green across seeds (`4531 examples, 0 failures`); rubocop clean on all touched files.

Docs: the [Room Number & Exit Display Toggles](https://github.com/elanthia-online/lich-5/wiki/Room%E2%80%90Number-&-Exit-Display-Toggles) wiki page is updated with both settings.

## Summary by CodeRabbit

* **New Features**
  * Added persistent display settings to render room exits as clickable links and to toggle monospace room text.
  * Added `display roomlinks` and `display roommono` commands, and updated the `display` help output.

* **Bug Fixes**
  * Standardized how room exits and related annotations are generated and injected across game modes.
  * Improved room-window rendering so exit/link behavior and mono styling respect the active frontend.

* **Tests**
  * Expanded unit tests for room rendering toggles and added Gemstone/DragonRealms parity checks with expected per-game differences.

### DIFF
--- a/lib/games.rb
+++ b/lib/games.rb
@@ -90,9 +90,10 @@ module Lich
           next unless timeto.is_a?(Numeric) || (timeto.is_a?(StringProc) && timeto.call.is_a?(Numeric))
           # Guard against a dangling wayto reference (destination room missing
           # from the mapdb) rather than crashing the whole downstream hook.
-          next if Map[key].nil?
+          dest = Map[key]
+          next if dest.nil?
 
-          label = "#{Map[key].title.first.gsub(/\[|\]/, '')}#{Lich.display_lichid ? "(#{Map[key].id})" : ''}"
+          label = "#{dest.title.first.gsub(/\[|\]/, '')}#{Lich.display_lichid ? "(#{dest.id})" : ''}"
           entries << (room_links? ? "<d cmd=';go2 #{key}'>#{label}</d>" : label)
         end
         entries
@@ -120,17 +121,17 @@ module Lich
       # Prepends the shared "StringProcs:" and "Room Exits:" lines (in that
       # order, so exits end up above StringProcs and any later room-number line
       # ends up above both) to alt_string, each only when it has entries. This is
-      # the composition both games share.
+      # the composition both games share. Entries are passed in (built once by the
+      # caller) so a game that also reuses them - e.g. the GemStone room-window
+      # mirror - does not iterate Map.current.wayto twice.
       # @param alt_string [String] the server string being rewritten
+      # @param stringproc_entries [Array<String>] pre-built StringProc entries
+      # @param exit_entries [Array<String>] pre-built obvious-exit entries
       # @return [String] alt_string with the room lines prepended
       # @api private
-      def prepend_room_exit_lines(alt_string)
-        stringprocs = room_stringproc_entries
-        alt_string = "#{room_styled("StringProcs: #{stringprocs.join(', ')}")}\r\n#{alt_string}" unless stringprocs.empty?
-
-        exits = room_exit_entries
-        alt_string = "#{room_styled("Room Exits: #{exits.join(', ')}")}\r\n#{alt_string}" unless exits.empty?
-
+      def prepend_room_lines(alt_string, stringproc_entries, exit_entries)
+        alt_string = "#{room_styled("StringProcs: #{stringproc_entries.join(', ')}")}\r\n#{alt_string}" unless stringproc_entries.empty?
+        alt_string = "#{room_styled("Room Exits: #{exit_entries.join(', ')}")}\r\n#{alt_string}" unless exit_entries.empty?
         alt_string
       end
     end
@@ -1129,9 +1130,9 @@ module Lich
       # @param alt_string [String] the server string being rewritten
       # @return [String] the rewritten server string
       def process_room_display(alt_string)
-        alt_string = prepend_room_exit_lines(alt_string)
-
         exits = room_exit_entries
+        alt_string = prepend_room_lines(alt_string, room_stringproc_entries, exits)
+
         if !exits.empty? && ROOM_WINDOW_FRONTENDS.include?(Frontend.client) && Map.current.id != Game.instance_variable_get(:@last_id_shown_room_window)
           alt_string = "#{alt_string}<pushStream id='room' ifClosedStyle='watching'/>Room Exits: #{exits.join(', ')}\r\n<popStream/>\r\n"
           Game.instance_variable_set(:@last_id_shown_room_window, Map.current.id)
@@ -1252,7 +1253,7 @@ module Lich
       # @param alt_string [String] the server string being rewritten
       # @return [String] the rewritten server string
       def process_room_display(alt_string)
-        alt_string = prepend_room_exit_lines(alt_string)
+        alt_string = prepend_room_lines(alt_string, room_stringproc_entries, room_exit_entries)
 
         # DR-specific room number display
         room_number = ""
--- a/lib/games.rb
+++ b/lib/games.rb
@@ -21,6 +21,120 @@ module Lich
   end
 
   module GameBase
+    # Game-agnostic formatting for the Lich-injected room annotations
+    # (room number, obvious exits, and StringProc exits). Both the GemStone
+    # and DragonRealms game instances mix this in so the font and clickable-link
+    # toggles behave identically regardless of game: the module decides *how* an
+    # annotation looks, while each game instance decides *which* lines/streams to
+    # emit. Extracting it removes the near-duplicate exit/StringProc rendering
+    # that previously lived in both #process_room_display implementations.
+    #
+    # Two independent settings drive it:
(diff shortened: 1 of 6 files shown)


MahtraDR:
@coderabbitai I don't see your review. Post it?

MahtraDR:
@coderabbitai review

MahtraDR:
@coderabbitai Both nitpicks resolved in 71553ff:

1. **`room_stringproc_entries` repeated `Map[key]` lookups** - now cached in a local `dest` right after the nil guard and reused for the title and id (exactly as suggested).

2. **`room_exit_entries` computed twice per room display** - fixed by making the prepend helper a pure formatter that takes pre-built entries: `prepend_room_exit_lines(alt_string)` -> `prepend_room_lines(alt_string, stringproc_entries, exit_entries)`. Each game now builds the entry lists once; the GemStone room-window mirror reuses the same `exits` array instead of re-iterating `Map.current.wayto`. This also tightens the module's single responsibility (the helper only formats; the game instances own the data-gathering).

Behavior is unchanged, specs updated for the new signature, and the full suite is green (4531 examples, 0 failures) with rubocop clean on the touched files.

MahtraDR:
@coderabbitai review this PR in full taking all commits into account

MahtraDR:
@coderabbitai The non-String `wayto` defensive-fix nitpick is resolved in 970cc72d.

`room_exit_entries` now derives the command with `value.to_s.dump[1..-2]`, matching the defensive `to_s` used in the obvious-exit check one line above. As you noted, `wayto` values are always `String` or `StringProc` in practice (the map-JSON loader and `apply_wayto_overrides` are the only writers, and StringProcs are filtered out immediately above), so this is defensive rather than a reachable bug - and it is byte-identical for `String` input since `String#to_s` returns self. The `room_links?` branch and all other logic are unchanged.

Added a spec driving a non-`String` (Integer) `wayto` value through the helper to lock in the behavior. Full suite green (4532 examples, 0 failures) and rubocop clean on the touched files.

Please take another look.

OSXLich-Doug:
@MahtraDR - regret that I am late to the party.


## Audit result: not ready to merge.
P1: The two new persisted display flags have no required flag record or release-note/default-flip entry. roommono defaults on for DR, changing the current default output. This violates the local feature-flag policy. [implementation (line 987)](/Users/doug/dev/test/lich-5/lib/lich.rb:987) · [policy (line 1)](/Users/doug/dev/AI_CONTEXT/architecture/lich5-feature-flag-semver-foundation.md:1)

P2: Room-window behavior duplicates frontend names instead of using the existing capability API. A newly registered room_window frontend will not receive the room output. Replace ROOM_WINDOW_FRONTENDS.include?(Frontend.client) with Frontend.supports_room_window?. [constant (line 55)](/Users/doug/dev/test/lich-5/lib/games.rb:55) · [use (line 1156)](/Users/doug/dev/test/lich-5/lib/games.rb:1156)

P2: Specs mock the new settings as plain accessors, so they do not test the actual defaulting, persistence, or “game not identified” behavior in lib/lich.rb. Add direct settings specs. [mock (line 684)](/Users/doug/dev/test/lich-5/spec/spec_helper.rb:684) · [production getter (line 941)](/Users/doug/dev/test/lich-5/lib/lich.rb:941)

YARD and inline documentation are otherwise strong; the shared formatter is a good SOLID/DRY extraction, and the focused specs are appropriately DAMP.
Validation: focused spec 130/0; RuboCop 5 files/0 offenses. Full suite reached 4,532 examples.

MahtraDR:
@OSXLich-Doug thanks for the thorough look - much appreciated. Addressed both P2s in b236fea; pushing back on the P1 with evidence below.

**P2 (capability API) - agreed, fixed.** You're right, and my original reasoning for hardcoding the list was wrong. I'd assumed `Frontend.client` and `$frontend` could differ, but `Frontend.client` literally returns `$frontend` ([front-end.rb:155-157](https://github.com/elanthia-online/lich-5/blob/main/lib/common/front-end.rb#L155-L157)), which is exactly what `supports_room_window?` reads - so the swap is behavior-identical today (room_window capability = wrayth+stormfront) and correctly picks up any future room_window frontend. Dropped `ROOM_WINDOW_FRONTENDS`; both the GS echo and DR streamWindow now use `Frontend.supports_room_window?`.

**P2 (settings specs) - agreed, added.** New `spec/lib/lich_room_display_settings_spec.rb` covers the real `Lich.display_room_links`/`display_room_mono` getters - game-aware defaults (DR mono-on/links-off, GS mono-off/links-on) and the "no game identified yet" -> nil state. The main suite mocks `Lich` with plain accessors, and requiring `lib/lich.rb` inline would clobber the global `Lich.db`/`Lich.log` mocks other specs depend on, so each case runs the real getter in an isolated subprocess against a no-row DB stub. Full suite green (4538 examples, 0 failures), rubocop clean.

**P1 (feature-flag policy / "changes DR default output") - respectfully disagree; I don't think this is actionable.** Two parts:

1. *"roommono defaults on for DR, changing the current default output"* - this doesn't hold up factually. For DR, every content flag that produces a Lich room line defaults **off**: `display_lichid` and `display_uid` are `GS ? true : false` (so off for DR), and `display_exits`/`display_stringprocs` default off for all games. With no lichid/uid the `Room Number:` string is empty and no line is emitted, and there are no exit/stringproc lines either - so a fresh DR install emits **zero** Lich room lines, and `roommono` has nothing to style. There is no out-of-box default output for it to change. (For a DR user who has *already* opted into those flags, `roommono` restores the mono look of the retired `roomnumbers.lic` - which is the intended, signed-off behavior, not a silent regression.)

2. *"violates the local feature-flag policy"* - the policy you linked lives at `/Users/doug/dev/AI_CONTEXT/architecture/...` on your machine, not in this repo. The repo's actual feature-flag system is `Lich::Common::FeatureFlags` (`feature_flag:`-prefixed keys, empty `DEFAULTS`), which these settings correctly do **not** use. `display_room_links`/`display_room_mono` follow the exact established pattern of the five existing `display_*` settings (`display_lichid`, `display_uid`, `display_exits`, `display_stringprocs`, `display_expgains`) - none of which register a flag record or ship a default-flip note, and two of which (`display_lichid`/`display_uid`) already carry game-aware defaults. If EO wants a formal flag-registration/semver policy for user display settings, I'm glad to help draft it, but it'd be a separate cross-cutting change rather than a blocker on this PR.

Happy to keep iterating if I've missed something on P1.

OSXLich-Doug:
@MahtraDR 

 Assessment:

Capability P2: closed. Both call sites now use Frontend.supports_room_window?.
Settings-spec P2: mostly closed. The new isolated real-getter specs verify DR/GS defaults and the pre-game nil state; 136 focused examples pass and RuboCop is clean.
Remaining narrow gap: the new test stub always returns no DB row, so it does not test persisted true/false values overriding game defaults, nor either setter’s write path. [spec (line 30)](/Users/doug/dev/test/lich-5/spec/lib/lich_room_display_settings_spec.rb:30)

P1 is withdrawn, compelling position that was stated upfront and overlooked.  Apologies for the needless churn.

OSXLich-Doug (approved):
Welcome to the club!

mrhoribu (approved):

PR #1440, opened 2026-07-14
fix(dr): fix_standing infinite loop (#3668) and prone/sitting detection under HidePostStrings (#4529)

MahtraDR:
Two related robustness fixes in the DR commons.

## #3668 - safe-room spams STAND while unconscious
`DRC.fix_standing` only broke its loop on `standing?`. Responses that are in the STAND match list but never make `standing?` true -- unconscious, plummeting, held, overburdened, no-room -- were matched (so no 15s dump) but never ended the loop, so the script spammed STAND forever. Added `CANNOT_STAND_PATTERN` and a break when the STAND response matches it.

## #4529 - prone/sitting detection breaks under HidePostStrings
`DRDefsPattern::LYING_DOWN` / `SITTING` only matched the "who is lying down" / "who is sitting" long forms. With the HidePostStrings option on, posture shows as "(prone)" / "(sitting)", so `DRRoom.pcs_prone` / `pcs_sitting` came back empty and safe-room's wait-for-others logic walked away early. Broadened both patterns to also match the short forms; `PARENTHETICAL` already strips the "(prone)" suffix before the name is extracted.

## Testing
Extended `common_spec.rb` and `drdefs_spec.rb` with adversarial cases: `fix_standing` issues STAND exactly once for each non-recoverable state (loop-prevention), `CANNOT_STAND_PATTERN` is frozen, and "(prone)"/"(sitting)" matching + extraction including mixed long/short forms and cross-posture negatives. All added examples pass; touched files are rubocop-clean.

Note: `drdefs_spec.rb`'s pre-existing `add_ordinals_to_duplicates with more than 20 duplicates` example fails on current main as well (unrelated to this change).

Generated with [Claude Code](https://claude.com/claude-code)


mrhoribu (approved):

MahtraDR:
Splitting into one PR per topic: #1441 (fix_standing #3668), #1442 (posture detection #4529). Closing this combined PR.

PR #1451, opened 2026-07-19
fix(all): keep ActiveSessions owner discovery stable under session churn

MahtraDR:
## Problem

During heavy session churn -- the weekly bulk-login run does mass premium-account swaps (killing/spawning owner-eligible sessions) then rapidly launches ~14 short-lived alt sessions per account -- the Active Sessions service becomes unreliable: sessions that are genuinely running are **absent from the snapshot for seconds-to-minutes**. Operator tooling reported "session not found" for 14/14 characters whose Lich logs show they ran to completion; a calm retry minutes later succeeded 14/14. The service was healthy before and after.

## Root cause

The cross-process "zombie" heuristic in `ensure_service_internal!` deleted the discovery pointer of a **still-healthy, still-listening owner** whenever:

1. a single 1s `Client#ping` timed out transiently (the host is saturated by bulk logins), or
2. the recorded `owner_pid` had been **recycled** by an unrelated churn process (so `Process.kill(0, pid)` reported it "alive").

Because the owner only wrote discovery at bootstrap and **never republished it**, and because a successor could not bind the still-held port (`EADDRINUSE`), the service went dark for every non-owner process until the owner process exited. In production this showed as bursts of 5-19 concurrent `clearing stale discovery` warnings at churn timestamps and multi-hour gaps between owner elections while the owner held the port throughout.

## Fix

Makes failover resilient without a persistence rewrite:

- **Owner self-heal** -- a process that owns a running server now re-asserts the shared discovery pointer on its heartbeat, rewriting it only when missing or stale. A pointer wrongly cleared by a peer reappears within one heartbeat instead of staying lost until the owner exits. (This alone would have prevented the incident.)
- **Tolerate transient probe failures** -- a discovered owner is re-probed with a short backoff before any takeover, so a churn-induced timeout doesn't trigger a disruptive election.
- **Bind-as-arbiter** -- the port bind is now the sole authority on whether a live owner remains. A successful bind takes over and overwrites a stale pointer (covers recycled pids and clean exits); a failed bind leaves the peer's pointer untouched and backs off. **The service never deletes a peer's discovery pointer**, so a healthy owner is never knocked offline by a transient false positive. This also removes the platform-fragile reliance on bare pid liveness.

The election is decomposed into small single-responsibility helpers (`serving_owner_reasserted?`, `service_responsive?`, `release_in_process_zombie!`, `bootstrap_owner!`), each fully YARD-documented.

## Tests

`spec/lib/internal_api/active_sessions_spec.rb` updated to assert the corrected contract and extended with adversarial/boundary coverage:

- owner self-heal after a peer **deletes** the pointer (recreated within one tick)
- owner self-heal after a peer **redirects** the pointer to another owner (overwritten)
- idempotence: no rewrite when the pointer already matches
- **retry tolerance**: an owner that answers on the 2nd probe is reused, never re-elected
- **recycled-pid takeover**: pid looks alive but the port is free -> bind succeeds and overwrites
- **contended bind back-off**: bind fails (real owner holds the port) -> return false and leave the live owner's pointer intact

All ActiveSessions specs pass (77 examples, 0 failures); `rubocop -A` reports no offenses on the touched files.

## Scope / follow-up

Focused on the owner/failover instability. A natural follow-up is to seed a new owner's in-memory registry from the existing persistent session store (`SessionsSettings` / `SessionDatabaseAdapter`) so a legitimate re-election repopulates instantly instead of over one heartbeat interval; filed separately to keep this change reviewable.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved service ownership handling during startup and recovery.
  * Prevented healthy but temporarily overloaded services from being unnecessarily taken over.
  * Added self-healing for missing or incorrect service discovery information.
  * Improved recovery from stale ownership records and interrupted server states.
  * Preserved existing service ownership during competing takeover attempts.

* **Tests**
  * Expanded coverage for service recovery, ownership contention, retry behavior, and discovery self-healing.

### DIFF
--- a/lib/internal_api/active_sessions.rb
+++ b/lib/internal_api/active_sessions.rb
@@ -45,6 +45,22 @@ module Lich
       # @return [String]
       DISCOVERY_FILENAME = 'lich-active-sessions.json'
 
+      # Number of times a discovered owner is probed before it is treated as
+      # unreachable during an election decision.
+      #
+      # A single probe can time out transiently while the host is saturated by
+      # bulk session churn (mass login/logout), even against a perfectly healthy
+      # owner. Re-probing before electing a replacement prevents a disruptive,
+      # unnecessary takeover.
+      #
+      # @return [Integer]
+      OWNER_PROBE_ATTEMPTS = 3
+
+      # Delay, in seconds, between owner responsiveness probes.
+      #
+      # @return [Float]
+      OWNER_PROBE_BACKOFF_SECONDS = 0.1
+
       @registry = nil
       @server = nil
       @service_client = nil
@@ -89,6 +105,12 @@ module Lich
       # @param allow_bootstrap [Boolean] when false, never start a new server
       # @return [Boolean]
       def self.ensure_service_internal!(allow_bootstrap:)
+        # This process already owns a healthy server: keep the shared discovery
+        # pointer authoritative and reuse it. This self-heals a pointer that was
+        # cleared or overwritten by a peer, without a disruptive re-election.
+        return true if serving_owner_reasserted?
+
+        # A healthy peer owner is already reachable on the fast path.
         return true if service_available?
         return false unless allow_bootstrap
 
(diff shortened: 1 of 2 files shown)


MahtraDR:
@coderabbitai review in full

MahtraDR:
Closing this after a live reproduction on the deployment showed the changes here do not address the actual production failure.

**What actually happened:** a game-dead but non-exited Lich session was left holding the hardcoded ActiveSessions port (`DEFAULT_PORT = 42_857`) -- socket still bound, accept loop dead (kernel accept-queue piling up, no protocol responses), and no discovery file present. Because the port is fixed and the stale process still owned it, every new session hit `EADDRINUSE` when trying to become the owner, so no owner was ever elected and nothing could register. `lich <char>` logged in fine but timed out waiting to register. Killing the stale process instantly restored the service.

The mechanisms this PR hardens (owner self-heal, retried probe, bind-as-arbiter, non-destructive discovery deletion) cannot help when a *live* process squats a *fixed* port -- a successor still cannot bind it.

**The real fixes** belong in a separate PR: (1) bind an ephemeral port and publish it in the discovery file so clients/CLI/launcher read the port instead of hardcoding 42857 (removes the single-point-of-failure squat), and (2) guarantee a shutting-down session actually exits and releases the listener (a shutdown watchdog / always-close-on-teardown), which is the deeper bug. Reworking accordingly.

PR #1453, opened 2026-07-19
feat(gs): gameobj add full_name matcher to type/sellable classification

mrhoribu:
## Summary
Adds an optional `<full_name>` field to `<type>` and `<sellable>` entries in `gameobj-data.xml`. When present, its regex is matched against an object's composed full name — `before_name` + `name` + `after_name` (the value returned by `GameObj#full_name`) — instead of only `name` or `noun`. This lets classification target items whose identity depends on prefix/suffix text (a "glowing" adjective, an "of power" suffix, etc.) that the bare `name` doesn't capture.

## Changes
All in `lib/common/gameobj.rb`:
- **Parser** (`parse_data_section`): reads the optional `<full_name>` child into the entry's `:full_name` pattern. Participates in base + custom-XML merging through the existing code path — no new merge logic.
- **Matcher** (`matching_data_keys`): adds `full_name` to the existing OR chain, so an entry matches on `name` **or** `noun` **or** `full_name`, still subject to the (name-based) `<exclude>`. `full_name` is computed once per lookup rather than once per candidate key.
- **Type cache** (`type`): keyed by a composite of `noun`, `name`, and `full_name` — see note below.
- YARD comments on the three methods updated to describe the new field and cache key.

`gameobj-data.xml` is intentionally **not** modified: this PR wires up the capability only and adds no classification entries.

## Behavioral note: type cache key
`#type` memoizes in `@@type_cache`. The key is now a `"|"`-delimited composite of the exact values the matcher inspects:

```ruby
cache_key = "#{@noun}|#{@name}|#{full_name}"
```

The format mirrors the composite keys already used by `@@index` in the same file (`"id|noun|name"`).

Previously the key was `name` alone. That was unsafe in two ways: (1) the new `<full_name>` matching depends on `before_name`/`after_name`, which `name` doesn't capture; and (2) — a **pre-existing** latent gap — the matcher already read `noun`, so two objects sharing a `name` but with different `noun` could collide even before this PR. Because this PR is what touches the cache key, it closes both. Concretely, with a `<type>` matching on `<noun>sword</noun>`, an object `(noun: sword, name: "a gleaming blade")` and `(noun: dagger, name: "a gleaming blade")` now classify independently instead of the second inheriting the first's result.

Classification **output** is unchanged — only the internal cache-key representation differs (visible via `type_cache`, which is used only in tests).

## Backward compatibility
Entries with no `<full_name>` tag are unaffected: the absent pattern is `nil`, and `String#=~ nil` is falsy — the same nil path the existing `name`/`noun` checks already rely on. All current `gameobj-data.xml` entries classify exactly as before.

## Example
```xml
<type name="magic">
  <full_name>glowing wand of .*</full_name>
</type>
```

## Testing
New specs in `spec/lib/common/gameobj_spec.rb` cover `full_name` matching for both `type` and `sellable`, the nil-safe absent-tag path, and cache non-collision along both matcher dimensions — same `name`/different `before_name`, and same `full_name`/different `noun`. The existing type-cache spec was updated to assert the composite key. Full `gameobj_spec.rb` suite passes with no regressions; the new `full_name` and noun-collision assertions fail against the pre-change code, confirming they exercise the fix.

- [x] Tests added and passing
- [x] Zero regression (existing suite green)
- [x] YARD updated

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved item type and sellable classification for objects with prefixes, suffixes, or other composed names.
  * Prevented classification errors when items share the same base name but differ in their full displayed name or noun.
  * Preserved existing name- and noun-based classification when no full-name rule is configured.

* **New Features**
  * Added support for matching classification rules against an item’s complete displayed name.


mrhoribu:
@coderabbitai review

mrhoribu:
@coderabbitai can you check the recent commits and review to make sure everything is addressed and that nothing else stands out at you that may have been missed?

OSXLich-Doug (approved):

PR #1437, opened 2026-07-08
feat(all): keyword bind hosts for the detachable client (tailscale/lan/any)

Nisugi:
## Summary

Adds keyword bind hosts to `--detachable-client` and `--headless` so users can expose the detachable listener on the right interface without ever discovering an IP address — the Lich-side groundwork for attaching mobile clients to a headless session over Tailscale/WireGuard/LAN.

- `tailscale:PORT` — binds this machine's Tailscale address (scans for a `100.64.0.0/10` CGNAT address; fails with guidance if Tailscale isn't running)
- `lan:PORT` — binds the private LAN address, preferring the default-route interface so Docker bridges / WSL adapters / VM host-only interfaces aren't picked by mistake
- `any:PORT` — binds `0.0.0.0`
- plain `HOST:PORT` now accepts hostnames and bracketed IPv6, not just dotted-quad IPv4

Since the detachable port is unauthenticated and can drive the game session, every non-loopback resolution surfaces a security warning at launch, and the help text steers users toward `tailscale` over `lan`/`any`.

## Changes

- **New** `Lich::Common::BindHostResolver` — keyword resolution + advisory warnings. The `lan` keyword learns the default-route source address via a UDP `connect()` probe (no packet sent) before falling back to an RFC1918 range scan ordered to deprioritize `172.16/12`.
- **New** `Lich::Main::DetachableClientTarget` — pure parser for `PORT` / `auto` / `HOST:PORT` / `HOST:auto` / `[IPv6]:PORT`, extracted so it's spec-able. Malformed `--detachable-client` values now fail loudly with usage text instead of being silently ignored.
- `--headless` alias accepts the same `HOST:PORT` forms and rewrites to `--detachable-client=...`.
- `ReusableTCPServer` derives the socket family from `Addrinfo` instead of hardcoding `:INET` (enables IPv6/hostname binds; no behavior change for IPv4 callers).
- The detachable listener echoes `--- Lich: detachable client listening on <addr>:<port>` to the console (previously log-file only), so `auto` ports and keyword-resolved addresses are visible to the user.
- Help text (`login` + `advanced` topics) documents the new forms.

## Testing

- New specs: `spec/lib/common/bind_host_resolver_spec.rb`, `spec/lib/main/detachable_client_target_spec.rb`; extended `spec/lib/main/arg_normalization_spec.rb` — 43 examples, 0 failures
- RuboCop clean on all touched files
[...]

PR #1461, opened 2026-07-21
feat(all): --bind-address speaks the same keywords as --detachable-client

Nisugi:
## What this fixes

PR #1437 gave detachable-client hosts a keyword vocabulary (`tailscale`/`lan`/`any`) but left `--bind-address` (#1405) taking only literal IPs. The split hides a trap on 5.19 today:

```
lich --bind-address=tailscale --detachable-client=8000
```

hands the literal string `"tailscale"` to `TCPServer.new` for the frontend and `--game` listeners -- a getaddrinfo failure at best, a real DNS lookup of a host named `tailscale` at worst -- while the detachable port-only form inherits the same raw string and `warning_for_explicit` treats it as a hostname it can't judge, so no warning either.

## What this does

`BindHostResolver` becomes the single bind-token vocabulary for every listener Lich opens:

- New `Lich::Main::BindAddressOption` (pure, spec-loadable, mirroring the `DetachableClientTarget` pattern) resolves `--bind-address` once in `SideEffects`, **before** the detachable handler, and stores the concrete address back. An unresolvable keyword (Tailscale not running) fails fast with the resolver's existing guidance.
- All four consumers converge on the resolved address: the frontend listener, the `--game` proxy, GAMEHOST mirroring (a spawned frontend is now told the actual tailnet/LAN IP), and a detachable client inheriting the bind address.
- The exposure warning prints exactly once; the detachable handler's duplicate inherit-warning is removed.
- Warning text no longer claims only the detachable port is at risk -- the frontend and `--game` sockets are exactly as unauthenticated -- so the resolver speaks of Lich's listen sockets generally.
- `DetachableClientTarget` is untouched: it was already a pure `HOST:PORT` parser, which is its whole job under this split.

## Testing

- New `spec/lib/main/bind_address_option_spec.rb`: keyword resolution, warning carry-through, error-instead-of-raise on resolver failure, real-resolver passthrough for explicit/`any` tokens.
- `bind_host_resolver_spec` updated for the reworded tailscale guidance (regex-level; all other assertions unchanged).
- Bind-family specs: 46 examples, 0 failures. (The full suite has 8 failures on my machine that reproduce identically on a clean 5.19.1 checkout -- Ruby 4.0.3/mingw environment issues unrelated to this change.)

## Follow-up commits

- `bbb5bfb3` -- rubocop compliance: ASCII-only comments (em-dashes replaced), and the spec's constant-in-block swapped for a helper method.
- `42082e55` -- finished the dedup this PR is named for: `handle_bind_address` and `handle_detachable_client` each carried their own stdout+log warning block and stdout+log+exit error block; a shared `announce`/`die` pair now serves both. End state: one parser (`DetachableClientTarget`), one resolver (`BindHostResolver`), one applier (`BindAddressOption`), and handlers that only orchestrate their flag-specific flow.


Nisugi:
Two follow-up pushes since the initial review request:

1. **Rubocop compliance** (`bbb5bfb3`): ASCII-only comments per `Custom/AsciiOnlySource`, and `Lint/ConstantDefinitionInBlock` in the new spec resolved with a helper method instead of a block-level constant.
2. **Completed the dedup** (`42082e55`): the two argv bind handlers each had their own warn-to-stdout+log and error-to-stdout+log+exit blocks -- collapsed into shared `announce`/`die` helpers in `SideEffects`. With that, the division of labor is fully single-purpose: `DetachableClientTarget` parses, `BindHostResolver` resolves, `BindAddressOption` applies, and the handlers only orchestrate.

Specs 46/46 green, rubocop clean across the touched files.

mrhoribu (approved):
# PR Review: #1461 -- `feat(all): --bind-address speaks the same keywords as --detachable-client`

| | |
|---|---|
| **Repo** | `elanthia-online/lich-5` |
| **PR** | [#1461](https://github.com/elanthia-online/lich-5/pull/1461) |
| **Author** | Nisugi |
| **Base** | `main` @ `44cb27e` |
| **Head** | `Nisugi/lich-5:feat/bind-address-keywords` @ `c23d545` |
| **Size** | 4 commits, +171 / -35, 6 files |
| **Mergeable** | clean |

## Verdict

**Approve -- no blocking issues.** Minimal, surgical, and correctly touches only the six files that needed changing. The bug being fixed is real: on 5.19 today, `lich --bind-address=tailscale --detachable-client=8000` hands the literal string `"tailscale"` to `TCPServer.new`, and the detachable port-only form inherits the same raw string while `warning_for_explicit` can't judge it -- so no warning either. The consolidation onto a single bind-token vocabulary is the right shape and the diff is clean.

Two low-severity items to eyeball before merge and three non-blocking nits (two with ready-to-apply spec drafts below) follow.

---

## Verification stack

| Gate | Result |
|---|---|
| `ruby -c` (all 6 files) | Syntax OK |
| Non-ASCII byte scan | Clean -- the em-dash fix in `bbb5bfb` holds |
| RuboCop (changed files) | 6 files, **0 offenses** |
| RSpec -- relevant set, PR head | **53 examples, 0 failures** |
| RSpec -- same set, merge base | 47 examples, 0 failures (= 53 - 6 new) |
| Regression | Clean for everything runnable |

**Relevant set** = `bind_address_option_spec`, `bind_host_resolver_spec`, `detachable_client_target_spec`, `arg_normalization_spec`, `help_text_spec`.

### Sandbox caveats (disclosed)

- Ran on **Ruby 3.2.3**, not the repo's target **4.0.3**. The touched files use only 3.2-compatible syntax (`frozen_string_literal`, `Struct.new(..., keyword_init: true)`), so this doesn't affect the result -- but the full 4.0 target was not exercised here.
- Ran **RuboCop 1.39**, not the pinned **1.86.1** (rubygems.org is unreachable in this environment; used the apt-packaged toolchain and a temp config overriding `TargetRubyVersion` to 3.2, since 1.39 predates Ruby 4.0). RuboCop version drift means a cop added between 1.39 and 1.86.1 could still fire in CI. **Re-run RuboCop 1.86.1 locally to be certain.**
- The **full** suite could not load here: missing native gems (`ox`, and downstream `sqlite3` / `gtk3` / `sequel`) that require rubygems.org. This is a harder sandbox limitation than the author's -- their 8 reported failures are a separate Ruby 4.0.3/mingw environment issue, unrelated to this change. The bind-family specs are pure and ran clean on **both** branches, which is what matters for this PR.

---

## Behavioral trace (evidence-grounded)

The core mechanism -- resolve `--bind-address` once, write it back, let every listener read the concrete value -- checks out. Each link verified against the code:

- **Single call site, correct order.** `SideEffects.execute` (`argv_options.rb:425`) calls `handle_bind_address` (`:162`) **before** `handle_detachable_client` (`:163`).
- **Store-back.** `handle_bind_address` resolves via `BindAddressOption.apply`, writes the concrete host back into `argv_options[:bind_address]` (`:190`), and announces the warning once.
- **Four consumers, all reading `argv_options[:bind_address]`:** frontend `TCPServer` (`main.rb:312`), GAMEHOST mirror (`main.rb:333-334`), `--game` proxy (`main.rb:462`), detachable inherit (`argv_options.rb:211`). **The three `main.rb` consumers are pre-existing and unchanged** -- the PR only fixes the *value* they receive (keyword -> concrete IP). Correct scoping; no `main.rb` edit is needed. (The PR body's "four consumers converge" slightly oversells this -- they already converged; the PR feeds them a resolvable value.)
- **Explicit public-IP warning is preserved.** Moving the warning off the detachable port-only branch does **not** drop it for `--bind-address=<public IP>`: `resolve`'s `else` branch routes explicit tokens through `warning_for_explicit` (`bind_host_resolver.rb:117`), which is directly tested (`bind_host_resolver_spec.rb:121-125`, `203.0.113.9` -> `/not a private address/`). No regression in the security-warning path.
- **Flagship dedup confirmed by replay.** Replaying the exact `handle_bind_address` + `handle_detachable_client` sequence through the real components: `--bind-address=any --detachable-client=8000` resolves to `0.0.0.0`, emits **exactly one** warning, and the port-only detachable inherits `0.0.0.0` with no second warning. The "prints exactly once" promise holds.
- **`--headless` keyword claim in the help text is accurate.** `arg_normalization.rb:52-53` rewrites `--headless` into the `--detachable-client=` form, so it inherits the vocabulary; the help text's grouping is not overclaiming.
- **No orphaned methods.** The removed inline warning/error blocks and the removed `else` branch don't leave dangling references -- `warning_for_explicit` is still called by `resolve` (verified). No public method was removed.
- The unconditional `detachable_client_host = bind_address || '127.0.0.1'` default (`argv_options.rb:211-212`) is **pre-existing** (`main:183`), not introduced here.

### Design review (DRY / SOLID / YARD)

- **DRY** -- net win. Two handlers previously each carried a duplicated stdout+log warning block and a stdout+log+exit error block; a shared `announce` / `die` pair now serves both. End state is one parser (`DetachableClientTarget`), one resolver (`BindHostResolver`), one applier (`BindAddressOption`).
- **SOLID** -- `BindAddressOption` is a single-responsibility pure module; `apply` takes an injectable `resolver:` (dependency inversion, and it's what the spec uses). Adding a keyword touches only `resolve`'s `case`.
- **YARD** -- `bind_address_option.rb` documents `@param` / `@return` / `@since` and the `Result` struct fields; `resolve` carries `@param` / `@return` / `@raise`. The new `SideEffects` methods use plain comments rather than YARD tags, but that's consistent with every sibling in that module.

---

## Findings

### Blocking

None.

### Confirm before merge (low risk, worth a glance)

1. **New hard-exit path.** An unresolvable `--bind-address` keyword now calls `die` -> `exit 1` at startup (`argv_options.rb:188`). `--bind-address` never resolved before, so this abort path is genuinely new. It's the intended fail-fast (far better than a cryptic `getaddrinfo` failure deep in `TCPServer`), but confirm you want a startup that aborts here -- e.g. `--bind-address=tailscale` with Tailscale down now stops immediately with guidance.
2. **`@since 5.19.2` doc tag** (`bind_address_option.rb:16`). Since versions aren't hand-edited (release-please derives the `feat` -> minor bump), make sure the hardcoded tag matches whatever the next release actually lands as, or drop it to avoid drift.

### Non-blocking nits

1. **No spec exercises the `SideEffects` wiring** -- the bind-before-detachable ordering, the store-back, the once-only warning, and a port-only detachable inheriting the *resolved* keyword. `bind_address_option_spec` tests the pure applier in isolation; the integration seam (exactly where the fixed bug lived) currently rests on manual reasoning, so a future reorder of `execute` could silently reintroduce the double-warn or unresolved-inherit bug with every unit spec still green. This is *consistent with repo convention* -- `SideEffects` has never had a spec because it's coupled to global `ARGV`, `Lich.log`, and `exit`; the pattern is "extract the pure unit and spec that." Defensible to merge without it. Draft below. **(higher value)**
2. **Coverage micro-gap:** `bind_address_option_spec` doesn't directly assert `apply` relays the explicit-public-IP warning. It's a thin passthrough, the resolver layer covers the public IP, and the relay is covered via the mocked `lan` and real `any` cases -- so this is symmetry-only. One-line draft below.
3. **Doc wording inconsistency (cosmetic):** the resolver's class comment (`bind_host_resolver.rb:13-15`) lists "the frontend socket, the `--game` proxy, and the detachable client" but omits `--headless`; `BindAddressOption`'s rdoc mentions `--headless`. Harmless.

---

## Nit 1 -- `SideEffects` integration spec (draft)

**New file:** `spec/lib/main/side_effects_bind_spec.rb`

```ruby
# frozen_string_literal: true

require 'rspec'

require_relative '../../../lib/main/argv_options'

# Exercises the argv wiring that #1461 relies on but that the pure
# BindAddressOption spec cannot see: that --bind-address is resolved *before*
# --detachable-client inherits it, that the resolved address is written back,
# that the exposure warning is announced exactly once, and that an
# unresolvable keyword aborts before any socket work.
#
# SideEffects reads the global ARGV directly and writes to both $stdout and
# Lich.log, and its failure path calls exit -- so all three are isolated here.
RSpec.describe Lich::Main::ArgvOptions::SideEffects do
  around do |example|
    saved_argv = ARGV.dup
    example.run
  ensure
    ARGV.replace(saved_argv)
  end

  before do
    allow(Lich).to receive(:log)
    @stdout = []
    allow($stdout).to receive(:puts) { |line| @stdout << line.to_s }
  end

  def warnings_matching(substring)
    @stdout.select { |line| line.include?(substring) }
  end

  it 'resolves the bind keyword and lets a port-only detachable inherit it' do
    ARGV.replace(['--detachable-client=8000'])
    options = { bind_address: 'any' }

    described_class.execute(options)

    expect(options[:bind_address]).to eq('0.0.0.0')
    expect(options[:detachable_client_host]).to eq('0.0.0.0')
    expect(options[:detachable_client_port]).to eq(8000)
  end

  it 'announces the exposure warning exactly once for the shared binding' do
    ARGV.replace(['--detachable-client=8000'])

    described_class.execute(bind_address: 'any')

    expect(warnings_matching('every network').size).to eq(1)
  end

  it 'aborts before touching sockets when the bind keyword is unresolvable' do
    allow(Lich::Common::BindHostResolver).to receive(:resolve)
      .with('tailscale')
      .and_raise(Lich::Common::BindHostResolver::Error, 'no tailscale here')

    expect { described_class.execute(bind_address: 'tailscale') }
      .to raise_error(SystemExit)
    expect(warnings_matching('no tailscale here')).not_to be_empty
  end

  it 'leaves a port-only detachable on loopback, unwarned, without --bind-address' do
    ARGV.replace(['--detachable-client=8000'])
    options = {}

    described_class.execute(options)

    expect(options[:detachable_client_host]).to eq('127.0.0.1')
    expect(options[:detachable_client_port]).to eq(8000)
    expect(warnings_matching('every network')).to be_empty
  end
end
```

**Validation status:** `ruby -c` OK, ASCII-clean, RuboCop clean (1.39 / temp 3.2 target). The spec's four assertions were **independently verified true** by replaying the exact `handle_bind_address` + `handle_detachable_client` logic through the real `BindAddressOption` / `BindHostResolver` / `DetachableClientTarget` components. The spec file itself could **not be executed** in the review sandbox because `argv_options.rb` transitively requires Windows FFI credential-manager code (`FFI::Struct` / `FFI::Library`) that isn't available here -- on a full checkout with the Gemfile.lock gems, the plain `require_relative` loads normally and the spec should run as-is. **Run it locally before relying on it.**

**Notes for the author:**
- The unresolvable-keyword test stubs `BindHostResolver.resolve` to raise so it's deterministic regardless of whether the CI host has Tailscale.
- Test 1 (outcome-based) already proves ordering: the detachable can only inherit the *resolved* value if `handle_bind_address` ran first. If you'd prefer it explicit, add an `.ordered` expectation on `handle_bind_address` / `handle_detachable_client` -- but that's belt-and-suspenders.

---

## Nit 2 -- public-IP passthrough assertion (draft)

**File:** `spec/lib/main/bind_address_option_spec.rb`, inside the existing `describe 'through the real resolver'` block.

```diff
@@ RSpec.describe Lich::Main::BindAddressOption do
       expect(result.host).to eq('0.0.0.0')
       expect(result.warning).to match(/every network/)
     end
+
+    it 'carries the not-a-private-address warning for an explicit public IP' do
+      result = described_class.apply('203.0.113.9')
+
+      expect(result.host).to eq('203.0.113.9')
+      expect(result.warning).to match(/not a private address/)
+      expect(result.error).to be_nil
+    end
   end
 end
```

**Validation status:** **Executed green** -- applied to the spec and ran in the sandbox: **7 examples, 0 failures** (was 6). `ruby -c` OK, ASCII-clean.

---

## Recommendation

Merge after glancing at the two "confirm" items. Of the nits, only **Nit 1** carries real value (it locks the PR's two headline guarantees against a future refactor); Nit 2 is symmetry and Nit 3 is a comment. All three are optional and can land in this PR or a follow-up, author's call.

PR #1464, opened 2026-07-23
fix(all): no-op Gtk.lich_main_quit when no main loop is nested

simtel12:
## Summary
- Guard `Gtk.lich_main_quit` so it no-ops when `main_level <= 0`, avoiding Gtk-CRITICAL `main_loops != NULL` after the loop has already unwound.
- Apply the same `gtk_main_loop_running?` check in `shutdown_gtk!` for headless / `--without-frontend` exits (e.g. Genie disconnect).
- Add specs covering the no-op path and direct teardown failure cleanup.

## Test plan
- [x] `rspec spec/lib/common/gtk_spec.rb`
- [x] Headless / `--without-frontend` exit after FE disconnect does not emit Gtk-CRITICAL about `main_loops`
- [x] Normal GUI shutdown still quits the nested GTK main loop

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GTK shutdown safety by preventing quit requests when the main loop is inactive or its state cannot be verified.
  * Preserved expected cleanup and shutdown behavior while the GTK main loop is active.

* **Tests**
  * Expanded coverage for inactive, unavailable, and active GTK main-loop states.


OSXLich-Doug:
@simtel12 - would like to discuss, some code duplicative.  DM me on Discord `lich_doug` pleasse?

OSXLich-Doug:
@coderabbitai review

OSXLich-Doug (approved):

PR #1467, opened 2026-07-24
fix(dr): resolve no-UID rooms carrying a stored UID instead of returning nil

MahtraDR:
## Problem

DragonRealms room matching returns `nil` for any room that carries a UID when the game exposes **no** live UID (`XMLData.room_id == 0`). Both `match_current` and `match_fuzzy` resolve a matched room with:

```ruby
room.uid.include?(XMLData.room_id) ? room.id : nil
```

Stored UIDs are never `0`, so `include?(0)` is always false. The result: any room the server does not surface a UID for -- or any room that has had a stray/provisional UID stamped onto it -- becomes **permanently unresolvable**. `Map.current` goes `nil`, which breaks `go2`/automap for that room. In the DR mapdb this is the documented "Ancient Tower 9375" class of bug (a stray UID on a genuinely no-UID room makes `find` bail to `nil`).

This also blocks a broader effort: pre-seeding UIDs onto rooms from an external source is unsafe today, because any room that turns out to be no-UID would break for every user. It should be safe to carry a UID that the live game happens not to report.

## Fix

Establish a single invariant: **a stored UID must never make a room less resolvable than an otherwise identical room with no UID.**

- When `XMLData.room_id` is zero (no UID exposed), skip UID disambiguation and trust the title/description/paths match, regardless of any UID stored on the room.
- When the game **does** expose a UID, behavior is unchanged: a UID'd room matches only when the live UID is among its stored UIDs, so day/night variants and look-alike maze cells stay disambiguated.

## Refactor (DRY)

The UID-resolution logic was duplicated in four places (the exact- and regex-description branches of both `match_current` and `match_fuzzy`). Those are collapsed into one helper, `resolve_matched_room(room, honor_peer_tags:)`, and the fuzzy-only peer-tag check is factored into `peer_disambiguation_tag?`. The fix now lives in exactly one spot. `match_current`, `match_fuzzy`, and the new helpers get YARD docs.

## Tests

Room matching was previously untested. This adds **43** examples:

- `resolve_matched_room` truth table: no-UID rooms, matching/non-matching/multi UIDs, the `room_id == 0` fallback, and all peer-tag interactions across both exact and fuzzy paths.
- `peer_disambiguation_tag?` pattern coverage (including adversarial near-misses).
- `match_current` / `match_fuzzy` integration: exact match, non-matching UID, the no-UID fix, regex-description fallback, foggy exits, disabled room window, peer honoring (fuzzy) vs. ignoring (exact), and no-match.
- `Map.current` end-to-end: UID fast path, day/night variant disambiguation, an ordinary no-UID room, a no-match, and the **Ancient Tower 9375 regression** (a stray-UID room now resolves when the game reports `room_id 0`).

Full suite: `4966 examples, 0 failures` on the files touched (one pre-existing, unrelated `hmr_spec` failure exists on clean `main` and does not involve `map_dr`). `rubocop` clean on both touched files.

### DIFF
--- a/lib/common/map/map_dr.rb
+++ b/lib/common/map/map_dr.rb
@@ -163,6 +163,74 @@ module Lich
         @@list[id]
       end
 
+      # Pattern identifying a room whose disambiguation depends on a manual
+      # +peer+ action (for example peering through a doorway to read an adjacent
+      # room before committing to a match). Such rooms cannot be told apart
+      # without a running script to perform the peer, so scriptless fuzzy
+      # matching declines to resolve them. Kept verbatim from the historical
+      # inline checks in {match_fuzzy}.
+      PEER_TAG_PATTERN = /^(set desc on; )?peer [a-z]+ =~ \/.+\/$/
+
+      # Whether +room+ carries a peer-disambiguation tag (see {PEER_TAG_PATTERN}).
+      #
+      # @param room [Lich::Common::Map] a room already matched on
+      #   title/description/paths
+      # @return [Boolean] +true+ when the room requires a manual peer to
+      #   disambiguate, +false+ otherwise
+      def self.peer_disambiguation_tag?(room)
+        room.tags.any? { |tag| tag =~ PEER_TAG_PATTERN }
+      end
+
+      # Resolve a room that already matched on title/description/paths down to a
+      # final room id, applying UID disambiguation.
+      #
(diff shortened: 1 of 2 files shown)


MahtraDR:
@coderabbitai review in full

mrhoribu (approved):

PR #1468, opened 2026-07-25
fix(dr): equipmanager.rb - reliable ranged-weapon transitions - stow ammo after unload + verify offhand wield hand

bmcneff:
Updated to address review: added the failure-path success guard, corrected root cause 2, and added regression tests.

## Summary
Fixes combat gear-swapping fumbles when transitioning off a ranged weapon
(crossbow / bow / staff-sling). The ammunition (e.g. a `crossbow bolt`) was left in
a hand after `unload` and never stowed, which cascaded into downstream failures.

## Problem
When `combat-trainer` switched from a loaded ranged weapon to a melee/offhand weapon,
the ammo stayed in hand. That single stuck item caused:
- `skin` / `loot` / `stow` to fail with "You need a free hand" and loop.
- The offhand-weapon logic (which assumed a `get` lands in the right hand, then swaps)
  to scramble, ending in `jab` (no `left`) against the ammo →
  "Wouldn't it be better if you used a melee weapon?".

## Root causes (`EquipmentManager#unload_weapon`)
1. **Acted mid-roundtime** — the ammo-stow ran before the trailing `waitrt?`, i.e.
   during the ~3s unload RT, before hand state settled.
2. **Anchored regex vs. the `<dialogData>` tag** — the in-hand branch used its own
   `^`-anchored `/^(?:You unload|You .* unloading)/`. The game prepends an aim/firing
   timer tag (`<dialogData id='AimTimerDialog'>…</dialogData>You unload …`), so
   `^You unload` no longer matched and the branch was skipped.
   `DRCI::UNLOAD_WEAPON_SUCCESS_PATTERNS` already handle the optional tag; the inline
   regex did not.
3. **Weapon/ammo noun collision** — the check compared against the weapon name, and a
   `crossbow bolt` contains `crossbow`, so a text match treated the bolt as the weapon
   and skipped stowing it.

## Changes (`lib/dragonrealms/commons/equipmanager.rb`)
- `unload_weapon`:
  - move `waitrt?` up so RT/hand state settles before acting;
  - consolidate the two ground-fall branches (feet / tumbles);
  - gate the in-hand stow on a **positive `UNLOAD_WEAPON_SUCCESS_PATTERNS` match**
    (handles the `<dialogData>` tag) instead of a blanket `else`, so an unload failure
    or a `bput` timeout can't stow an unrelated off-hand item;
  - identify weapon vs. ammo by **noun** (`DRC.get_noun`) against actual hand contents.
- `wield_weapon_offhand?`: return success when the weapon is already in the off (left)
  hand instead of a spurious `false`; only swap from the right hand.
- `wield_weapon?`: the offhand branch swaps on `offhand && DRCI.in_right_hand?(weapon)`
  (is the *weapon* in the right hand?) rather than `offhand && DRC.right_hand`
  (is *anything* in the right hand?).

## Tests (`spec/lib/dragonrealms/commons/equipmanager_spec.rb`)
- in-hand stow with a `<dialogData>`-prefixed unload line — asserts the ammo hand is
  stowed and the weapon hand is not (also covers the `crossbow bolt` noun collision);
- unload **failure** does not stow an unrelated off-hand item;
- `bput` **nil/timeout** stows nothing;
- `wield_weapon_offhand?` returns `true` (not a spurious `false`) and does not `swap`
  when already off-handed; and swaps when the weapon landed in the right hand.

## Testing
Existing `equipmanager_spec.rb` suite passes; new regression tests added for the above.
Reproduced and verified in-game with a crossbow → wakizashi → offhand-bola rotation:
before, the bolt was left in hand and blocked skin/loot and flipped `jab left` → `jab`;
after, the ammo is stowed on unload and the rotation completes cleanly. Ground-fall
recovery (ammo falls to feet / tumbles) is preserved.

### DIFF
--- a/lib/dragonrealms/commons/equipmanager.rb
+++ b/lib/dragonrealms/commons/equipmanager.rb
@@ -314,19 +314,23 @@ module Lich
           return false
         end
 
-        if get_item?(weapon)
-          swap_to_skill?(weapon.name, skill) if skill && weapon.swappable
-          if DRCI.in_right_hand?(weapon)
-            case DRC.bput('swap', *DRCI::SWAP_HANDS_SUCCESS_PATTERNS, *DRCI::SWAP_HANDS_FAILURE_PATTERNS)
-            when *DRCI::SWAP_HANDS_SUCCESS_PATTERNS
-              return true
-            else
-              return false
-            end
+        return false unless get_item?(weapon)
+
+        swap_to_skill?(weapon.name, skill) if skill && weapon.swappable
+
+        # We want the weapon in the LEFT (off) hand. If it landed in the right hand
+        # (e.g. get placed it there), swap it over; if it's already in the left
+        # hand, we're done -- previously this returned false in that case.
+        if DRCI.in_right_hand?(weapon)
+          case DRC.bput('swap', *DRCI::SWAP_HANDS_SUCCESS_PATTERNS, *DRCI::SWAP_HANDS_FAILURE_PATTERNS)
+          when *DRCI::SWAP_HANDS_SUCCESS_PATTERNS
+            return true
+          else
+            return false
           end
         end
 
-        return false
+        DRCI.in_left_hand?(weapon)
       end
 
       # @deprecated Use {#wield_weapon_offhand?} instead.
@@ -357,7 +361,7 @@ module Lich
         if get_item?(weapon)
           swap_to_skill?(weapon.name, skill) if skill && weapon.swappable
 
-          if offhand && DRC.right_hand
+          if offhand && DRCI.in_right_hand?(weapon)
             case DRC.bput('swap', *DRCI::SWAP_HANDS_SUCCESS_PATTERNS, *DRCI::SWAP_HANDS_FAILURE_PATTERNS)
             when *DRCI::SWAP_HANDS_SUCCESS_PATTERNS
               return true
@@ -786,11 +790,15 @@ module Lich
       # @see DRCI::UNLOAD_WEAPON_FAILURE_PATTERNS
       def unload_weapon(name)
         result = DRC.bput("unload my #{name}", *DRCI::UNLOAD_WEAPON_SUCCESS_PATTERNS, *DRCI::UNLOAD_WEAPON_FAILURE_PATTERNS)
+        waitrt? # wait out the unload roundtime so the ammo/hand state has settled before we act on it
+
         ammo_match = result&.match(/^(?:Your .*?\b(?<ammo>[\w]+)\b fall.* from your .* to your feet\.)$/)
-        if ammo_match
-          # Ammo fell to ground because hands are full.
-          # Lower weapon, stow ammo, then pick it back up.
-          ammo = ammo_match[:ammo]
+        ground_match = result&.match?(/As you release the string/) ? result.match(/the (?<ammo>\w+) tumbles/) : nil
+
+        if ammo_match || ground_match
+          # Ammo ended up on the GROUND (hands were full, or it tumbled). Lower the
+          # weapon, stow the ammo from your feet, then pick the weapon back up.
+          ammo = (ammo_match || ground_match)[:ammo]
           unless DRCI.lower_item?(name)
             Lich::Messaging.msg("bold", "EquipmentManager: Unable to lower #{name} to pick up ammo")
             return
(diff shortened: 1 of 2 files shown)


MahtraDR (commented):
**Verdict: the fixes are real and the code is sound, but the catch-all rewrite introduces a new failure-path side effect, the PR ships zero tests for three distinct bug fixes, and root-cause #2 in the description is inaccurate.** Fair to merge after those are addressed.

Reviewed by reading the changed method against its helpers (`DRC.get_noun`, `DRC.left_hand`/`right_hand`, `DRCI.in_hand?`, `stow_hand`, the `UNLOAD_WEAPON_*` patterns), tracing all four `unload_weapon` call sites, running the existing 112-example `equipmanager_spec.rb` against this branch's `equipmanager.rb` (all pass), and rubocop (clean).

### Does it do what it says? Mostly yes.
- `wield_weapon_offhand?` -- correctly returns `true` when the weapon is already in the left hand instead of a spurious `false`; behavior-preserving for the right-hand-swap path. Good.
- `wield_weapon?` -- gating the offhand swap on `DRCI.in_right_hand?(weapon)` rather than `DRC.right_hand` (anything in the right hand) is strictly more correct. Good.
- `unload_weapon` -- moving `waitrt?` up is strictly correct; consolidating the two ground-fall branches is clean; the noun-based comparison genuinely fixes the "crossbow bolt contains crossbow" collision. Good.

### Concerns

**1. [Medium -- correctness] The new `else` now fires on failure and on nil/empty results.** The old code gated the in-hand stow behind an explicit success match. Now any result that isn't a ground-fall enters the `else` and stows every held item whose noun differs from the weapon's. So if `unload` fails ("But your `<weapon>` isn't loaded", "You must be holding the weapon to do that") or bput returns nil/empty on timeout, and you are holding a second unrelated item -- precisely the one-handed-ranged + offhand rotation this PR targets -- that legitimate off-hand item gets silently stowed (and may emit a misleading "Unable to stow ammo from left hand"). Suggest early-returning on `UNLOAD_WEAPON_FAILURE_PATTERNS` / blank result, or keeping a positive success guard (matching the dialogData-prefixed SUCCESS patterns) instead of a blanket `else`.

**2. [Low -- accuracy] Root-cause #2 in the description is wrong.** "You stop aiming." is not in `UNLOAD_WEAPON_SUCCESS_PATTERNS`, so bput skips it and returns the real unload line -- it never shadows anything. The actual defect the catch-all fixes is that the old in-hand branch regex `/^(?:You unload|...)/` is `^`-anchored and fails when the game prepends the `<dialogData ...AimTimer...>` tag (which the SUCCESS constants already handle, but that branch did not). The fix is right; the stated reason should be corrected so the next reader is not misled.

**3. [Medium -- testing] No tests, in a file with an extensive spec.** Three separate bugs fixed, zero regression tests added. Existing specs still pass, but none exercises the new behavior. Expected coverage: (a) in-hand stow when `result` carries the dialogData tag; (b) noun-collision (bolt not mistaken for the crossbow); (c) `wield_weapon_offhand?` returning `true` when already left-handed; and especially (d) the failure path not stowing an offhand item (concern #1).

**4. [Nit, pre-existing]** the `the (?<ammo>\w+) tumbles` capture cannot span a hyphenated noun, but ammo is realistically single-word -- not introduced here, just noting.

Net: a solid, well-motivated fix. Would approve after the failure-path guard (#1) and a couple of regression tests (#3), plus the one-line description correction (#2).

MahtraDR:
@coderabbitai review in full

mrhoribu:
@coderabbitai review in full

MahtraDR (approved):
Re-reviewed after commit `9f1a7e9d4` -- all three concerns from the prior pass are resolved. Verified by re-fetching the branch, running the full spec against the PR code, a mutation check on the new guard tests, and rubocop.

**#1 (catch-all `else` firing on failure/timeout) -- Fixed.** In-hand stow is now gated on a positive success match (`elsif result && DRCI::UNLOAD_WEAPON_SUCCESS_PATTERNS.any? { ... }`). None of the `UNLOAD_WEAPON_FAILURE_PATTERNS`, nor a nil result, can satisfy it, so an unload failure or bput timeout can no longer stow an unrelated off-hand item. Feet-fall and release-string cases are still consumed by the earlier `ammo_match`/`ground_match` branches, so ordering is intact.

**#2 (inaccurate root-cause) -- Fixed.** Both the PR body and the in-code comment now correctly attribute the in-hand miss to the `^`-anchored `/^(?:You unload|...)/` regex failing against the `<dialogData ...AimTimer...>`-prefixed line. The "You stop aiming" explanation is gone.

**#3 (no tests) -- Fixed.** Five new, meaningful examples: in-hand stow with a `<dialogData>`-prefixed line + noun-collision ("crossbow bolt" not mistaken for the crossbow); unload failure does not stow the off-hand item; bput nil stows nothing; `wield_weapon_offhand?` returns true (no spurious swap) when already left-handed; and swaps when the weapon landed in the right hand.

**#4 (hyphenated-ammo regex nit)** -- pre-existing and non-blocking; left as-is, which is fine.

Verification:
- Full spec: 117 examples, 0 failures (112 + 5 new).
- Mutation check: reverting the success guard back to a blanket `else` fails exactly the two failure-guard tests -- confirming they genuinely protect #1.
- rubocop: clean on both files.

Tight, well-tested fix. LGTM.

PR #1470, opened 2026-07-26
feat(all): Add supervised child script lifecycle APIs

calael:
### Anonymous and child scripts

- Adds `SubScript`, which runs a Ruby block with its own `Script.current` identity, generated name, buffers, hooks, workers, exit handlers, and completion state.
- A child has one parent, inherits the parent's `hidden`, `no_kill_all`, and `no_pause_all` flags, and unregisters after teardown.
- Self-parenting, ownership cycles, and adoption by multiple parents are rejected.
- Runtime parent teardown stops owned children and waits for their bounded teardown. Process shutdown leaves child ordering to the process-wide drain.
- Script startup is reserved by canonical script name, preventing concurrent non-forced duplicate publication and rejecting new publication after shutdown admission closes.

### Lifecycle and workers

- Adds explicit running, stopping, cleanup, and completion states. Completion includes exit callbacks, descendant-worker teardown, and parent detachment.
- Runtime cleanup is asynchronous by default. The process shutdown coordinator cleans up inline to avoid a thread-allocation burst; shutdown requested from a normal script worker remains asynchronous.
- Interrupted cleanup can be reclaimed by shutdown progress instead of leaving a script permanently stopping.
- `Script#thread_group` returns the script's actual Ruby `ThreadGroup`. Native `add`, `list`, `enclose`, and `enclosed?` behavior is preserved while additions are rejected after teardown begins.
- Threads created by script workers or cleanup callbacks remain owned through teardown, including threads created during worker `ensure` blocks.
- Ordinary scripts, `SubScript`, and `ExecScript` record successful completion and their first execution or teardown error.

### Script libraries

- Library names are normalized to `lib...` names and resolved through normal script discovery.
- Concurrent callers share one library generation and wait for complete execution.
- A library is cached only after successful completion. Failed, interrupted, and timed-out generations are not cached and can be retried.
- Abandoned startup reservations are reaped by owner liveness, so shutdown and later loads cannot wait forever on a dead loader.
- Cyclic library dependencies raise `LoadError`.
- Reloading uses a stable loaded-library snapshot, skips the currently executing library, continues after individual failures, and returns aggregate success.

### Kill-all, shutdown, and buffers

- Normal kill-all preserves hidden scripts and scripts protected by `no_kill_all`; forced kill-all includes them.
- Adds the built-in `;kd` command for forced kill-all.
- Daemon setup makes the current script hidden and protects it from kill-all and pause-all.
- Shutdown closes startup admission before taking its first script snapshot and continues tracking scripts that have left the running registry but are still completing teardown.
- The shutdown drain adopts scripts first seen during polling and reports scripts that remain incomplete.
- The global `clear` helper uses the buffer's atomic clear-and-snapshot operation.
- `Script.run` now waits for full lifecycle completion rather than only running-registry removal.

### Class APIs

- `Script.subscript { ... }` starts a quiet `SubScript` owned by `Script.current`.
- `SubScript.start(parent: Script.current, quiet: true) { ... }` starts an anonymous script and returns its `SubScript`.
- `Script.start_child(*args)` starts a normal script owned by `Script.current`; outside a script it behaves like `Script.start`.
- `Script.run_child(*args)` starts a child, waits for complete teardown, and returns the child or `nil` if teardown does not complete.
- `Script.daemon_me` applies daemon flags to `Script.current`, returning the script or `false`.
- `Script.loadlib(library)` loads one script library and returns `true`, or raises `LoadError`.
- `Script.reloadlibs` reloads the loaded-library snapshot and returns `true` only if every attempted reload succeeds.
- `Script.libs` returns a copy of the loaded-library `Set`.
- `Script.kill_all(force: false, context: :runtime)` requests teardown for eligible scripts and returns the number selected.
- `Script.begin_shutdown` closes startup admission, waits for admitted startup transitions, and returns the shutdown snapshot.
- `Script.shutdown_scripts` returns a side-effect-free snapshot of running and still-stopping scripts.
- `Script.progress_shutdown` reclaims interrupted cleanup/finalization work and returns the remaining shutdown snapshot.

### Instance APIs

- `Script#running?` reports running-registry membership.
- `Script#stopping?` reports whether teardown has been requested.
- `Script#join(timeout = nil)` waits for complete teardown and returns the script, or `nil` on timeout.
- `Script#kill_sync(context: :runtime, timeout: nil)` requests teardown and waits for completion.
- `Script#completed_successfully?` reports successful execution with no execution or teardown error.
- `Script#exit_error` returns the first execution or teardown error, or `nil`.
- `Script#register_child(script)` adopts a running script, or stops it and returns `nil` when adoption is closed.
- `Script#unregister_child(script)` removes an owned child and returns it.
- `Script#child_scripts` returns a snapshot of currently owned children.

## Changed APIs

- `Script#kill(context: :runtime, async: nil)` adds explicit context and asynchronous-control arguments.
- `Script#thread_group` remains a native `ThreadGroup`, but worker additions are lifecycle-aware and are rejected after teardown starts.
- `Script.run` waits through exit callbacks, worker teardown, and parent detachment.

## Summary by CodeRabbit

* **New Features**
  * Added the `kd` command to forcibly stop all scripts, including protected and hidden scripts.
  * Improved script lifecycle handling, including startup, execution, cleanup, library loading, and worker management.

* **Bug Fixes**
  * Shutdown draining now detects and stops scripts that appear during the shutdown process.
  * Improved cleanup reliability, error reporting, and handling of asynchronous script termination.
  * The `clear` command now clears output buffers atomically.

### DIFF
--- a/lib/common/orderly_shutdown.rb
+++ b/lib/common/orderly_shutdown.rb
@@ -43,12 +43,18 @@ module Lich
           ShutdownLog.begin_user_exit_summary!
           coordinator.request(reason: :user_exit, source: source)
 
-          scripts_provider ||= proc { Script.running + Script.hidden }
+          initial_scripts =
+            if scripts_provider
+              scripts_provider.call
+            else
+              Script.begin_shutdown
+            end
+          scripts_provider ||= proc { Script.shutdown_scripts }
           remaining_scripts = proc { scripts_provider.call.reject { |script| script.equal?(current_script) } }
 
-          run(
+          result = run(
             coordinator: coordinator,
-            initial_scripts: remaining_scripts.call,
+            initial_scripts: initial_scripts.reject { |script| script.equal?(current_script) },
             remaining_scripts: remaining_scripts,
             script_drain: script_drain,
             vars: vars,
@@ -56,6 +62,11 @@ module Lich
             active_sessions_lifecycle: active_sessions_lifecycle,
             slow_threshold: slow_threshold
           )
+          if current_script && initial_scripts.include?(current_script)
+            result.scripts_drained = false
+            result.completed = false
+          end
+          result
         end
 
         # Executes orderly user shutdown once.
--- a/lib/common/orderly_shutdown.rb
+++ b/lib/common/orderly_shutdown.rb
@@ -49,7 +49,7 @@ module Lich
             else
               Script.begin_shutdown
             end
-          scripts_provider ||= proc { Script.shutdown_scripts }
+          scripts_provider ||= proc { Script.progress_shutdown }
           remaining_scripts = proc { scripts_provider.call.reject { |script| script.equal?(current_script) } }
 
           result = run(
--- a/lib/common/script.rb
+++ b/lib/common/script.rb
@@ -26,6 +26,7 @@ module Lich
 
     class Script
       VALID_KILL_CONTEXTS = [:runtime, :shutdown].freeze
+      CHILD_MUTEX_INITIALIZER = Mutex.new
 
       @@elevated_script_start = proc { |args|
         if args.empty?
@@ -288,6 +289,8 @@ module Lich
         end
       }
       @@running = Array.new
+      @@library_mutex = Mutex.new
+      @@loaded_libraries = Set.new
       @@kill_metrics_mutex = Mutex.new
       @@kill_metrics = {
         :minute            => nil,
@@ -549,6 +552,33 @@ module Lich
         @@running.dup
       end
 
+      def Script.subscript(&block)
+        SubScript.start(:parent => Script.current, :quiet => true, &block)
+      end
+
+      def Script.start_child(*args)
+        parent = Script.current
+        child = Script.start(*args)
+        return child unless parent && child
+
+        parent.register_child(child)
+      end
+
+      def Script.run_child(*args)
+        child = Script.start_child(*args)
+        child&.join
+        child
+      end
+
+      def Script.daemon_me
+        return false unless (script = Script.current)
+
+        script.hidden = true
+        script.no_kill_all = true
+        script.no_pause_all = true
+        script
+      end
+
       def Script.current
         if (script = @@running.find { |s| s.has_thread?(Thread.current) })
           sleep 0.2 while script.paused? and not script.ignore_pause
@@ -566,6 +596,39 @@ module Lich
         if (s = @@elevated_script_start.call(args))
           sleep 0.1 while @@running.include?(s)
         end
+      end
+
+      def Script.loadlib(library)
+        library_name = library.to_s.downcase
+        library_name = "lib#{library_name}" unless library_name.start_with?('lib')
+        raise ArgumentError, 'library name cannot be empty' if library_name == 'lib'
+        raise LoadError, "script library not found: #{library_name}" unless Script.exists?(library_name)
+
+        script = @@library_mutex.synchronize do
+          running = @@running.find { |candidate| candidate.name.casecmp?(library_name) }
+          if @@loaded_libraries.include?(library_name)
+            running
+          else
+            @@loaded_libraries.add(library_name)
+            started = running || Script.start(library_name)
+            @@loaded_libraries.delete(library_name) unless started
+            started
+          end
+        end
+
+        raise LoadError, "failed to start script library: #{library_name}" unless @@loaded_libraries.include?(library_name)
+
+        script&.join
+        true
+      end
+
+      def Script.reloadlibs
+        @@library_mutex.synchronize { @@loaded_libraries.dup }.each { |library| Script.run(library) }
+        true
+      end
+
+      def Script.libs
+        @@library_mutex.synchronize { @@loaded_libraries.dup }
       end
 
       def Script.running?(name)
@@ -619,6 +682,16 @@ module Lich
         else
           false
         end
+      end
+
+      def Script.kill_all(force: false, context: :runtime)
+        unless VALID_KILL_CONTEXTS.include?(context)
+          raise ArgumentError, "invalid script kill context: #{context.inspect}"
+        end
+
+        scripts = force ? Script.list : Script.running.reject(&:no_kill_all)
+        scripts.each { |script| script.kill(context: context) }
+        scripts.length
       end
 
       def Script.paused?(name)
@@ -992,6 +1065,7 @@ module Lich
           raise ArgumentError, "invalid script kill context: #{context.inspect}"
         end
 
+        @kill_requested = true
         source = @kill_source || caller[0..2]
 
         if context == :shutdown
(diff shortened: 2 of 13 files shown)


MahtraDR:
What kind of use case do you envision here?

MahtraDR:
Took a careful pass through this. Summary first: it does what it claims, the spec coverage is genuinely strong, and I do not think merging breaks existing scripts. My pushback is on the packaging and on a couple of the new code's own-merit choices, not on compatibility.

## On the "what use case" question

The pattern this generalizes is already all over DragonRealms scripting, just hand-rolled. The clearest in-tree example is `DRC.wait_for_script_to_complete` (dr-scripts `common.rb`), which is a poll-based join with no ownership: it calls `start_script`, then spins on `Script.running.include?(handle)` until the child leaves the registry. On top of that sit long-lived orchestrators like `t2.lic`, which launches other scripts by mindstate thresholds, tracks every one it started in an array, and on its own death walks that array to stop only its own children (with an opt-out list so some children survive). That is a supervision tree built entirely by convention: name bookkeeping, `Script.running?` polling, `stop_script`, and a `before_dying` hook.

So the envisioned consumers are real and plural: `run_child` + `join` is the native version of `wait_for_script_to_complete`, and `start_child` + automatic bounded teardown is the native version of what t2 and similar launchers reimplement by hand. Diverging from those existing idioms is fine and not a blocker in my view. The only thing that would be a hard blocker is breaking them, and I do not believe this does (see below).

## Compatibility check (not a blocker)

I traced the consumer entry points against the PR head:

- `start_script` still returns the handle or nil, so poll-based waiters are unaffected.
- Runtime `kill` was already async in base (cleanup in `Thread.new`, returns `@name` immediately), so the PR does not introduce a new stop-then-restart timing window, and the old duplicate check already rejected a restart while the name was still in `@@running`.
- `Script.running` / `Script.list` keep the same contract; a script under teardown stays in `@@running` until its cleanup thread finishes, same as before.
- `clear` delegating to `LimitedArray#clear_snapshot` preserves the old dup-then-clear return value, atomically.
- `;ka` preserves the old set and `;kd` is additive.

## Findings on the new code's own merits

**1. `join` / `run_child` / `kill_sync` can block indefinitely (high).** Runtime `__complete_stop` calls `__wait_for_worker_shutdown(wait: true)`, which does an untimed `thread.join` on every worker (`script.rb:1861-1862`), and `@cleanup_complete` is only set once that returns. A worker whose `ensure` / `before_dying` never returns parks the caller of `join` forever. `run_child` does not pass the `CHILD_JOIN_TIMEOUT` the class already defines for this. Nothing in-tree calls `join` yet, so this only bites adopters of the new API, but it works against the robustness this PR is selling. Suggest a documented default bound on `join`, and `run_child` passing one.

**2. Class-level registry guarded by a per-instance lock (medium-high).** `@@running` / `@@stopping` are shared by all scripts. Publication serializes on the class-level `@@startup_mutex` via `__publish_to_registry`, but the removals run under only the per-instance `lifecycle_mutex` (`@@running.delete` at `script.rb:1823`, `@@stopping.delete` at `script.rb:1906`). Two scripts tearing down at once hold different mutexes touching the same array. On CRuby the GVL makes each `Array#push`/`#delete` atomic so it does not corrupt in practice, which is why tests pass, but there is no single lock owning registry transitions and the per-instance mutex reads as protection it does not provide. Nearly free to fix by moving the delete paths under the same class-level lock as publish.

**3. `shutdown_scripts` is a query with teardown side effects (medium).** It calls `__refresh_deferred_stop` on every script (`script.rb:654-657`), which can re-issue a kill and finalize, and it is wired in as the drain's `remaining_scripts` proc and called up to ~200 times per shutdown. It works, and it is confined to shutdown, but a getter-named method that finalizes scripts on each call hides real behavior. Consider renaming or splitting the "advance deferred teardown" step from the "list what remains" read.

**4. Four near-identical rescue ladders that already disagree (medium).** The trusted and untrusted bodies (`script.rb:145-267`), `ExecScript` (`script.rb:2245-2301`), and `SubScript` (`script.rb:2110-2123`) are four copies of the same exit-handling ladder, and they have already diverged (SubScript collapses it into one rescue list, ExecScript repeats `script.kill` per branch instead of an `ensure`). Since the point of the PR is uniform exit recording via `__record_exit_error` / `__record_successful_exit`, threading that through four hand-maintained copies is the most likely future correctness gap. Worth extracting one runner.

**5. Lock topology is large and undocumented (low / design).** Eight-plus mutexes, acquired nested and across instances (for example `loadlib` holds `@@library_mutex` across a full `Script.start` that itself takes `@@startup_mutex` and spawns threads, `script.rb:802-818`; child finalization reaches into the parent's `child_scripts_mutex`, `script.rb:1900-1901`). I traced what I could and did not find a deadlock, but that is "could not disprove in the time available," not "proven safe." A documented lock hierarchy would help a lot here.

**6. Smaller items (low).** `loadlib` calls `__finish_start` twice on the success path (`script.rb:822` and the `ensure` at `850`); idempotent today, brittle if that method ever changes. `join` busy-polls at 100 Hz during finalization (`script.rb:1456`). `Script.run` still polls `@@running` rather than using the new `join` (`script.rb:765`), so the obvious first consumer of the primitive was left unconverted.

## Packaging

The biggest practical objection is scope: this is a ~1300-line rewrite of the single most critical file in Lich that also bundles child supervision, the kill/teardown model, a library-loading subsystem, kill-all / `;kd`, shutdown accounting, and the `clear` delegation. It cannot be reviewed, bisected, or reverted piecewise, and it ships no in-tree caller exercising the new APIs. The library-loading and kill-metrics / shutdown-accounting pieces look separable and would be much easier to land on their own.

## Recommendation

Nothing here requires changing the public API surface. If `join` gets a bound (1), the registry deletes move under the class-level lock (2), and the non-lifecycle subsystems split into their own PRs, I think this becomes a clean set of changes with an obvious set of consumers waiting for it.

calael:
Addressed the review findings:

- `join` is condition-driven and honors an explicit timeout; once teardown starts, an unbounded join also gets a bounded cleanup deadline. `run_child` and `Script.run` now wait for complete lifecycle teardown.
- Running/stopping registry transitions use the class registry mutex.
- `shutdown_scripts` is a pure snapshot; interrupted teardown recovery is explicit in `progress_shutdown`.
- Ordinary, anonymous, and exec scripts use shared execution/error handling.
- Library startup no longer runs while holding the library mutex, startup reservations are owned and recoverable, and failed reloads make `reloadlibs` return `false`.
- The lock order is documented in `script.rb`.
- `Script#thread_group` remains the actual native `ThreadGroup`, including native enclosure semantics, while cleanup descendants remain internally supervised.

The follow-up adversarial reviews also covered cancellation during startup/finalization, cleanup executor loss, rejected worker registration, child-launch teardown, callback-created descendants, and native ThreadGroup compatibility.

Verification on the pushed head: `5030 examples, 0 failures`; Rubocop, Ruby syntax, and `git diff --check` all pass.

MahtraDR:
@coderabbitai review in full

calael:
This monolithic PR has been split into a reviewable stack rebased onto current `main`:

1. #1474 - supervised child script APIs
2. #1475 - lifecycle transition synchronization
3. #1476 - shutdown coordination
4. #1477 - execution and library loading
5. #1478 - lifecycle ownership hardening
6. #1479 - forced teardown command and atomic buffer clear

The top branch has the same feature-file contents as this PR. The reconstructed stack passes 5,035 RSpec examples, RuboCop, Ruby syntax checks, and `git diff --check`. Closing this PR as superseded; its review history remains available here.

Issue #1471, opened 2026-07-26
`--help` documents `--temp-dir=`/`--script-dir=`/`--data-dir=`, but they are silently ignored


simtel12:
`lich --help` documents `--temp-dir=`, `--script-dir=` and `--data-dir=`, but nothing implements them. Lich silently ignores unrecognised path arguments, so following the help text gets you a Lich that quietly uses default directories.

Found in 5.19.1 while debugging why a client tailing `TEMP_DIR` was watching an empty path.

## What `--help` claims

`lib/main/help_text.rb:180-199` (paths topic):

```
Path options:
  --home=PATH
  --script-dir=PATH
  --data-dir=PATH
  --temp-dir=PATH
  --hosts-dir=PATH
  --hosts-file=PATH

Examples:
  lich --script-dir=/my/scripts
  lich --data-dir=/my/data --temp-dir=/tmp/lich
```

## What's implemented

The directory constants are set in one place only — the pre-`constants.rb` loop at `lich.rbw:9-28` — using different names:

```ruby
if arg =~ /^--(?:home)=(.+)[\\\/]?$/i      then LICH_DIR   = $1
elsif arg =~ /^--temp=(.+)[\\\/]?$/i       then TEMP_DIR   = $1
elsif arg =~ /^--scripts=(.+)[\\\/]?$/i    then SCRIPT_DIR = $1
elsif arg =~ /^--data=(.+)[\\\/]?$/i       then DATA_DIR   = $1
...
```

`grep -rn "TEMP_DIR\s*=" lib/ lich.rbw` → one hit, `lich.rbw:14`. `lib/main/argv_options.rb` handles only `--hosts-dir` (space form, line 167). Nothing reads `--temp-dir=`, `--script-dir=` or `--data-dir=`.

`NON_INSTANCE_OPTION_KEYS` in `lib/common/authentication/login_helpers.rb:50-56` lists both spellings (`temp` and `temp-dir`, `data` and `data-dir`), which may be where the help text's names came from — but that list only classifies flags as non-instance-selectors, it never assigns a directory.

## Reproduction

`--no-gui --without-frontend` so it exits without a game connection; watch where `debug-*.log` lands:

```sh
ruby lich.rbw --no-gui --without-frontend --temp=/tmp/tA        # => /tmp/tA/debug-*.log
ruby lich.rbw --no-gui --without-frontend --temp-dir=/tmp/tB    # => {LICH_DIR}/temp, /tmp/tB never created
[...]

PR #1472, opened 2026-07-26
fix(all): correct path-flag aliases and document all path options

mrhoribu:
## Summary

Fixes problem 1 & 2 of #1471 and aligns the CLI path-flag interface with its own documentation. Lich exposes two independent argument subsystems:

- the early constant loop in `lich.rbw` (runs before `constants.rb`) seeds the directory constants `LICH_DIR`, `SCRIPT_DIR`, `DATA_DIR`, `TEMP_DIR`, `MAP_DIR`, `LOG_DIR`, `BACKUP_DIR`, `LIB_DIR`;
- `lib/main/argv_options.rb` builds the runtime `@argv_options` (including `--hosts-file` and `--hosts-dir`).

The two had drifted: `--help paths` documents `--script-dir` / `--data-dir` / `--temp-dir`, but the loop only recognized the legacy `--scripts` / `--data` / `--temp` short forms, so following the help text produced flags that were silently ignored (the constant fell back to its default). This PR makes the loop accept both spellings, documents the full set of path options, and fixes a separate `--hosts-dir` parsing mismatch.

### `lich.rbw` — accept `-dir` aliases (backward compatible)

Each directory flag now accepts both its legacy short form and a canonical `-dir` form (alias = constant name, lowercased, `_` -> `-`):

| Constant     | Accepted flags            |
| ------------ | ------------------------- |
| `TEMP_DIR`   | `--temp`, `--temp-dir`     |
| `SCRIPT_DIR` | `--scripts`, `--script-dir`|
| `MAP_DIR`    | `--maps`, `--map-dir`      |
| `LOG_DIR`    | `--logs`, `--log-dir`      |
| `BACKUP_DIR` | `--backup`, `--backup-dir` |
| `DATA_DIR`   | `--data`, `--data-dir`     |
| `LIB_DIR`    | `--lib`, `--lib-dir`       |

`--home` (-> `LICH_DIR`) is unchanged. Implemented by widening the existing `(?:home)` non-capturing-group pattern already used on the first branch; no other change to the loop. Every previously accepted flag still works.

### `lib/main/help_text.rb` — document all path options

`paths_help` now lists `--map-dir`, `--log-dir`, `--backup-dir`, and `--lib-dir` alongside the existing entries. These four supported directories were previously undocumented. The help shows the canonical `-dir` spelling (matching how `--script-dir` etc. are already presented); the legacy short forms remain accepted but intentionally undocumented.

### `lib/main/argv_options.rb` — parse `--hosts-dir` with `=`

`handle_hosts_dir` previously matched a bare `--hosts-dir` token and consumed the *next* argv element as its value (`--hosts-dir /path`), while the help text and its sibling `--hosts-file` both use the `=` form. It now parses `--hosts-dir=PATH`, consistent with `--hosts-file=PATH` and the documentation. The existence check, backslash normalization, trailing-slash handling, and warning message are unchanged.

> Behavioral note: the previously-working `--hosts-dir <path>` space form is no
> longer accepted. Use the documented `--hosts-dir=<path>`.

## Testing

- `spec/lib/main/help_text_spec.rb`: new example asserting `--help paths` renders all ten path options.
- `spec/lib/main/argv_options_spec.rb` (new): covers `handle_hosts_dir` — equals-form parse + normalization, trailing-slash idempotence, the missing-directory warning, and a regression guard that the legacy space-separated form is now ignored. `argv_options.rb` auto-runs `process_argv` at load and can't be required standalone, so the spec lifts the real method body from source rather than duplicating logic (keeps it bound to shipping code).

## Summary by CodeRabbit

* **New Features**
  * Added support for directory-specific command-line aliases, including `--temp-dir`, `--script-dir`, `--maps-dir`, `--logs-dir`, `--backup-dir`, and `--data-dir`.
  * Expanded path-related help output to document available directory options.

* **Bug Fixes**
  * Standardized `--hosts-dir` usage to the `--hosts-dir=PATH` format.
  * Normalized trailing slashes in valid hosts directory paths and preserved invalid arguments for correction.

### DIFF
--- a/lib/main/argv_options.rb
+++ b/lib/main/argv_options.rb
@@ -164,11 +164,9 @@ module Lich
         end
 
         def self.handle_hosts_dir(argv_options)
-          if (arg = ARGV.find { |a| a == '--hosts-dir' })
-            i = ARGV.index(arg)
-            ARGV.delete_at(i)
-            hosts_dir = ARGV[i]
-            ARGV.delete_at(i)
+          if (arg = ARGV.find { |a| a =~ /^--hosts-dir=(.+)$/i })
+            hosts_dir = arg[/^--hosts-dir=(.+)$/i, 1]
+            ARGV.delete(arg)
             if hosts_dir && File.exist?(hosts_dir)
               hosts_dir = hosts_dir.tr('\\', '/')
               hosts_dir += '/' unless hosts_dir[-1..-1] == '/'
--- a/lib/main/help_text.rb
+++ b/lib/main/help_text.rb
@@ -190,6 +190,10 @@ module Lich
             --script-dir=PATH
             --data-dir=PATH
             --temp-dir=PATH
+            --map-dir=PATH
+            --log-dir=PATH
+            --backup-dir=PATH
+            --lib-dir=PATH
             --hosts-dir=PATH
             --hosts-file=PATH
 
--- a/lich.rbw
+++ b/lich.rbw
@@ -10,19 +10,19 @@
 for arg in ARGV
   if arg =~ /^--(?:home)=(.+)[\\\/]?$/i
     LICH_DIR = $1
-  elsif arg =~ /^--temp=(.+)[\\\/]?$/i
+  elsif arg =~ /^--(?:temp|temp-dir)=(.+)[\\\/]?$/i
     TEMP_DIR = $1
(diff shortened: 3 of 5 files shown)


mrhoribu:
@coderabbitai review

OSXLich-Doug (approved):

PR #1474, opened 2026-07-27
feat(all): Add supervised child script APIs (01 of 06)

calael:
## Summary

Adds the public script supervision surface:

- `SubScript.start(parent: Script.current, quiet: true, &block)` starts an anonymous block-backed script with independent runtime buffers. It returns the script or `false` when startup cannot be admitted.
- `Script.subscript(&block)` starts an anonymous child and returns it. It raises `ArgumentError` without a block and `ThreadError` when shutdown or parent teardown rejects startup.
- `Script.start_child(*args)` starts and adopts a named child; outside a script worker it behaves like `Script.start`.
- `Script.run_child(*args)` starts a named child, waits for teardown, and returns the child or `nil` when startup fails.
- `Script.daemon_me` marks the current script hidden and exempt from pause-all and kill-all.
- `Script#running?`, `Script#stopping?`, `Script#join(timeout = nil)`, and `Script#kill_sync(context: :runtime, timeout: nil)` expose lifecycle state and synchronous teardown.
- `Script#register_child`, `Script#unregister_child`, and `Script#child_scripts` expose parent-child ownership.
- `Script.kill_all(force: false, context: :runtime)`, `Script.loadlib(library)`, `Script.reloadlibs`, and `Script.libs` add the initial teardown and library APIs used by later stack layers.
- `Script#join` rejects a worker joining its own script with `ThreadError`.

## Validation

## Stack

1 of 6. The next PR targets this branch and synchronizes lifecycle transitions.

## Summary by CodeRabbit

* **New Features**
  * Added support for creating, starting, running, and managing child scripts.
  * Added script lifecycle controls, including status checks, synchronous shutdown, joining, and daemon mode.
  * Added library loading, reloading, caching, and status reporting.
  * Child scripts now stop cleanly when their parent script is terminated.

* **Bug Fixes**
  * Improved cleanup and shutdown coordination for scripts and their child processes.

* **Tests**
  * Added comprehensive coverage for script lifecycles, child scripts, library loading, shutdown behavior, and protection settings.

### DIFF
--- a/lib/common/script.rb
+++ b/lib/common/script.rb
@@ -26,6 +26,7 @@ module Lich
 
     class Script
       VALID_KILL_CONTEXTS = [:runtime, :shutdown].freeze
+      CHILD_MUTEX_INITIALIZER = Mutex.new
 
       @@elevated_script_start = proc { |args|
         if args.empty?
@@ -288,6 +289,8 @@ module Lich
         end
       }
       @@running = Array.new
+      @@library_mutex = Mutex.new
+      @@loaded_libraries = Set.new
       @@kill_metrics_mutex = Mutex.new
       @@kill_metrics = {
         :minute            => nil,
@@ -549,6 +552,33 @@ module Lich
         @@running.dup
       end
 
+      def Script.subscript(&block)
+        SubScript.start(:parent => Script.current, :quiet => true, &block)
+      end
+
+      def Script.start_child(*args)
+        parent = Script.current
+        child = Script.start(*args)
+        return child unless parent && child
+
+        parent.register_child(child)
+      end
+
+      def Script.run_child(*args)
+        child = Script.start_child(*args)
+        child&.join
+        child
+      end
+
+      def Script.daemon_me
+        return false unless (script = Script.current)
+
+        script.hidden = true
+        script.no_kill_all = true
+        script.no_pause_all = true
+        script
+      end
+
       def Script.current
         if (script = @@running.find { |s| s.has_thread?(Thread.current) })
           sleep 0.2 while script.paused? and not script.ignore_pause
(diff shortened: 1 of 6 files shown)


MahtraDR:
## Stack orchestration (6 PRs)

This PR is the base of a six-part stack that supersedes the monolithic #1470 (now closed). I reviewed the whole stack end to end.

Merge order (each branch is based on the one before it):

1. #1474 supervised child script APIs (this PR)
2. #1475 lifecycle transition synchronization
3. #1476 shutdown coordination
4. #1477 execution and library loading
5. #1478 lifecycle ownership hardening
6. #1479 forced teardown command

### The split is faithful
The top of the stack is byte-identical on the five lifecycle feature files to the previously reviewed #1470 head. None of the feature code changed during the split; only the packaging did.

### Verification (at every layer of the stack)
Checked each of the six commits in isolation:

- `ruby -c`: clean
- `rubocop`: clean on touched files
- `git diff --check`: clean
- `rspec` (common + global lifecycle specs): pass at each step
- Full suite at the top of the stack: 5,035 examples, 0 failures, aside from one pre-existing flake (`hmr_spec.rb:32`) that also fails on clean `main` and is unrelated to this work.

### One caveat: the stack is not monotonically clean
Several subsystems are introduced in a naive form early and hardened later, so the middle PRs are fixes to the PR below them rather than independent features:

- registry synchronization and a bounded `join`: introduced flawed in #1474 / #1475, fixed in #1476
- `loadlib` holding the library mutex across `Script.start`: present in #1474 / #1475, fixed in #1477
- the four duplicated exit rescue ladders: present through #1475, unified in #1477

Practical implication: treat #1474 through #1478 as a single approve-and-merge unit. Review them separately (that is what the split buys), but do not merge the lower PRs in isolation, because #1474 alone would publish the new public API in its pre-hardening state. #1479 is additive and the only genuinely optional PR.

MahtraDR:
Base of the stack. Introduces the public surface: `SubScript`, `start_child`, `run_child`, `join`, `kill_sync`, `register_child`, `daemon_me`, `loadlib`, `kill_all`.

Worth flagging that this PR ships the API in its pre-hardening form:

- the registry (`@@running` / `@@stopping`) is mutated without the class-level registry mutex (added in #1476)
- `join` has no cleanup deadline, so it can wait unbounded (bounded in #1476)
- `loadlib` holds `@@library_mutex` across a full `Script.start`, and `reloadlibs` always returns true (both fixed in #1477)
- `SubScript` startup uses a poll-based handshake (replaced by the `start_gate` queue in #1475)

Coherent and tested as an API sketch, but for that reason it should not land on its own ahead of the rest of the stack. See the orchestration note on this PR.

MahtraDR:
## Re-review update (changes since the 07-27 review)

Confirmed the commits pushed after the original review resolve the items flagged on this PR. `subscript` now raises `ThreadError` on a rejected startup instead of returning `nil` (documented with `@raise`), and the doc/guard commits are in. The pre-hardening caveats I noted here (unsynchronized registry, unbounded `join`, `loadlib` holding the library mutex, poll-based handshake) are all addressed downstream in the stack.

Verified: `ruby -c` clean, rubocop clean, full suite shows only pre-existing failures that also reproduce on clean `main` (`hmr_spec:32` flake plus random-order `enhancive_spec` pollution; neither is touched by this stack).

Re-review approval stands: merge this through [#1478](https://github.com/elanthia-online/lich-5/pull/1478) as a unit.

MahtraDR:
@coderabbitai review in full

MahtraDR (approved):

OSXLich-Doug (approved):

PR #1476, opened 2026-07-27
fix(all): Coordinate script shutdown draining (03 of 06)

calael:
## Summary

Connects lifecycle state to process shutdown and makes wait bounds explicit:

- Closes startup admission before the initial shutdown snapshot.
- Tracks scripts through stopping and cleanup rather than only registry removal.
- Reclaims deferred teardown and discovers scripts first observed during drain polling.
- Reports incomplete teardown when the current script remains in the shutdown set.
- Makes `Script#join(timeout = nil)` wait indefinitely unless the caller supplies a timeout.
- Adds `Script.run_child(*args, timeout: nil)`; a timeout returns `nil` if complete teardown does not finish in time.
- Keeps process shutdown bounded at the `ShutdownScriptDrain` call site instead of imposing a global timeout on every join.

## Validation

## Stack

3 of 6. Depends on synchronized lifecycle transitions.

### DIFF
--- a/lib/common/orderly_shutdown.rb
+++ b/lib/common/orderly_shutdown.rb
@@ -43,12 +43,18 @@ module Lich
           ShutdownLog.begin_user_exit_summary!
           coordinator.request(reason: :user_exit, source: source)
 
-          scripts_provider ||= proc { Script.running + Script.hidden }
+          initial_scripts =
+            if scripts_provider
+              scripts_provider.call
+            else
+              Script.begin_shutdown
+            end
+          scripts_provider ||= proc { Script.shutdown_scripts }
           remaining_scripts = proc { scripts_provider.call.reject { |script| script.equal?(current_script) } }
 
-          run(
+          result = run(
             coordinator: coordinator,
-            initial_scripts: remaining_scripts.call,
+            initial_scripts: initial_scripts.reject { |script| script.equal?(current_script) },
             remaining_scripts: remaining_scripts,
             script_drain: script_drain,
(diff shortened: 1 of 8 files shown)


MahtraDR:
Fixes four of the original review findings.

- introduces `@@registry_mutex` and routes every `@@running` / `@@stopping` transition through it (finding 2)
- bounds `join` with `SCRIPT_CLEANUP_TIMEOUT` once teardown starts (finding 1)
- splits `shutdown_scripts` into a pure snapshot plus an explicit `progress_shutdown` (finding 3)
- converts `Script.run` to use `join` instead of polling `@@running` (finding 6)
- wires `orderly_shutdown`, `main`, and the drain to `begin_shutdown` / `progress_shutdown`, and the drain now adopts scripts first observed mid-drain

One note: `join` still calls `__refresh_deferred_stop` on each loop here, so it is not yet a pure wait; that side effect is removed in #1478.

MahtraDR:
## Re-review update

`__script_owning_thread_group` now resolves owners in `@@stopping` as well as `@@running`, which is required so cleanup can borrow a stopping script's thread group. That is consistent with the borrowing rework in [#1478](https://github.com/elanthia-online/lich-5/pull/1478). The four findings this PR targeted remain resolved, and the `join` side effect I noted is removed in #1478 as planned.

Verified green (only the pre-existing `main` failures, which reproduce on a clean checkout).

OSXLich-Doug (approved):

PR #1477, opened 2026-07-27
refactor(all): Unify script execution and library loading (04 of 06)

calael:
## Summary

Consolidates script execution and makes library startup independent of loader locks:

- Routes ordinary scripts, `ExecScript`, and `SubScript` through shared execution and error handling while preserving type-specific reporting and jump behavior.
- Runs library startup outside the library coordination mutex.
- Shares one in-flight library generation among concurrent callers.
- Caches a library only after successful completion and permits failed loads to retry.
- Adds `Script.loadlib(library, timeout: nil)`; the timeout bounds coordination and teardown waits, and raises `Script::LibraryJoinTimeout` when exceeded.
- Preserves the single-flight marker after a timed-out caller so retries rejoin the active generation.
- Reloads from a stable library snapshot.

## Validation

## Stack

4 of 6. Depends on shutdown coordination.

### DIFF
--- a/lib/common/script.rb
+++ b/lib/common/script.rb
@@ -168,128 +168,34 @@ module Lich
                 eval('script = Script.current', script_binding, script.name)
                 Thread.current.priority = 1
                 respond("--- Lich: #{script.custom? ? 'custom/' : ''}#{script.name} active.") unless script.quiet
-                if trusted
-                  begin
-                    eval(script.labels[script.current_label].to_s, script_binding, script.name)
-                    script.__send__(:__record_successful_exit)
-                  rescue SystemExit => e
-                    if e.success?
-                      script.__send__(:__record_successful_exit)
+                begin
+                  Script.__send__(
+                    :__execute,
+                    script,
+                    :on_error => proc { |error| Script.__send__(:__report_script_error, script, error, :untrusted => !trusted) }
+                  ) do
+                    if trusted
+                      eval(script.labels[script.current_label].to_s, script_binding, script.name)
                     else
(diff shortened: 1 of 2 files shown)


MahtraDR:
The cleanest, most self-contained PR in the stack, and a net reduction in size.

- unifies the four exit rescue ladders into a single `Script.__execute` runner plus per-type error reporters (finding 4)
- moves `Script.start` out from under `@@library_mutex`, so library startup no longer runs while holding the mutex (finding 5)
- turns a library teardown timeout into a `LoadError` instead of a silent success

This is the piece the original review called out as separable, and it holds up as a standalone refactor.

MahtraDR:
## Re-review update

This resolves the two items carried into the original review. `reloadlibs` now returns a real Boolean instead of always `true` and takes a timeout, and `loadlib` converts `timeout` into a single monotonic deadline so the total wait is bounded rather than per operation, rejects negative timeouts, and tracks `@@loaded_library_owners` so a reload only drops the loading entry when it matches. `__take_completed_start` is folded into `__begin_library_start`. Library-wait dependency edges are now keyed by `[token, thread]` and pruned by liveness, so a killed caller cannot leave a phantom edge that trips a false cyclic-dependency error.

One non-blocking flag for the author: `reloadlibs` now defaults to a `1.0s` per-library timeout. A genuinely slow reload would report `false` where it previously blocked and returned `true`. Callers can pass `timeout: nil` to restore wait-forever behavior, but a comment or a larger default would avoid a surprise.

OSXLich-Doug (approved):

PR #1478, opened 2026-07-27
fix(all): Harden script lifecycle ownership (05 of 06)

calael:
## Summary

Hardens lifecycle ownership and interruption handling under concurrent starts, stops, child launches, callbacks, and library operations:

- Preserves native `ThreadGroup` behavior while registering script workers with lifecycle ownership.
- Tracks callback and worker descendants until cleanup completes, including cleanup invoked from another script group.
- Uses interruption-safe launch gates for ordinary scripts, `ExecScript`, `SubScript`, watch callbacks, and asynchronous cleanup.
- Recovers abandoned startup, child-launch, cleanup, and library bookkeeping when an owning thread exits at any handoff point.
- Prevents ownership cycles, cross-parent adoption races, overlapping same-name generations, stale loader publication, and dependency records owned by dead waiters.
- Restores an at-exit callback to its exact queue position when cleanup is interrupted before dispatch; dispatched callbacks retain at-most-once semantics.
- Tracks concurrent reload attempts with independent tokens so one timeout or completion cannot erase another active invalidation.
- Adds `Script.reloadlibs(timeout: 1.0)`; the timeout applies per library and the return value reports whether every reload succeeded.
- Bounds child cleanup at ownership call sites, isolates kill metric timing, documents lock ordering, and prunes completed lifecycle records.

## Validation

Focused lifecycle suite: 159 examples, 0 failures under six deterministic seeds. Branch suite: 5,091 examples, 0 failures.

RuboCop: 1,031 files inspected, no offenses. Ruby syntax validation and `git diff --check` pass.

## Stack

5 of 6. Depends on unified execution and library loading.

### DIFF
--- a/lib/common/script.rb
+++ b/lib/common/script.rb
@@ -37,10 +37,13 @@ module Lich
       RAW_THREAD_GROUP_ENCLOSED = ThreadGroup.instance_method(:enclosed?)
       JOIN_WAIT_INTERVAL = 0.05
 
-      # Lock order:
-      #   startup -> registry -> per-script lifecycle -> child ownership
-      # Library coordination never holds its mutex while starting or joining a
-      # script. No lock in this hierarchy may be held while invoking script code.
+      # Nested lock order:
+      #   parent child list -> startup -> registry -> generated name
+      #   -> per-script lifecycle -> child relationship
+      # Finalization releases the relationship lock before taking a parent's
+      # child-list lock. Library coordination releases its mutex before starting
+      # or joining a script. Registry and lifecycle locks are never held while
+      # invoking script code or cleanup callbacks.
 
       module ThreadGroupHandle
         def __attach_script(script)
--- a/lib/common/script.rb
+++ b/lib/common/script.rb
@@ -866,8 +866,8 @@ module Lich
       private_class_method :__startup_in_progress_locked?
 
       def Script.__library_handoff_reserved_locked?(name)
-        @@startup_reservations.any? do |_reservation, (reserved_name, _owner, generation)|
-          reserved_name == name && generation.nil?
+        @@startup_reservations.any? do |_reservation, (reserved_name, owner, generation)|
+          reserved_name == name && owner == Thread.current && generation.nil?
         end
       end
       private_class_method :__library_handoff_reserved_locked?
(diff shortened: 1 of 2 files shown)


MahtraDR:
The highest-risk PR in the stack, and where the ownership model becomes correct. Deserves the closest human review.

- replaces the `ThreadGroup` wrapper with native semantics via bound `UnboundMethod`s (`add` / `list` / `enclose` / `enclosed?`), so `thread_group` stays a real `ThreadGroup`
- adds cleanup-thread group borrowing and a `CLEANUP_SCRIPT_THREAD_KEY` thread local, so `Script.current` resolves during cleanup
- shifts each `die_with` / `at_exit` callback before invoking it, so an interrupted executor can resume with the remaining callbacks
- removes `join`'s `__refresh_deferred_stop` side effect, making `join` a pure wait
- reaps abandoned startup reservations by owner liveness, so a dead loader cannot block shutdown

Concentrated, subtle concurrency logic; the borrowing and cancellation-recovery paths are the parts to scrutinize.

MahtraDR:
## Re-review update (highest-risk PR, closest read)

Every concern I raised is addressed:

- The `kill` orchestration is now interrupt-atomic: `handle_interrupt(Exception => :never)` around bookkeeping, `:immediate` only around real waits.
- The naive `@die_with.shift` / `@at_exit_procs.shift` loops are replaced by `__run_cleanup_queue` plus `__restore_cleanup_queue_claim`, giving at-most-once callback dispatch even when the executor is cancelled inside a callback. This is the densest new code in the stack (the three-branch restore heuristic takes effort to follow) but it is well commented and now heavily spec'd.
- `__borrow_worker` drops the `group` parameter and the `@borrowed_thread_groups` / `__return_borrowed_group` dance, which fixes the borrow/return asymmetry; the guard relaxes to `@cleanup_complete` only, so a stopping script can still lend its group.
- `@child_launches` becomes a `{token => owner_thread}` map pruned by liveness, so a dead launcher can no longer wedge `__wait_for_child_launches`.
- An unowned cleanup group now raises instead of silently running callbacks in an unmanaged group, and abandoned-cleanup recovery re-dispatches the kill.

Verified at the top of the stack: `ruby -c` clean, rubocop clean on the changed files, the new lifecycle examples pass, and the full suite shows only the pre-existing `main` failures (confirmed identical on a clean `main` checkout). Approve.

OSXLich-Doug (approved):

PR #1479, opened 2026-07-27
feat(all): Add forced script teardown command (06 of 06)

calael:
## Summary

Adds the user-facing command and buffer changes separately from lifecycle internals:

- Adds `;kd` to force teardown of hidden and kill-all-protected scripts.
- Routes ordinary kill-all through `Script.kill_all(force: false, context: :runtime)`.
- Uses the script buffer atomic clear-and-snapshot operation from the global `clear` helper.
- Documents `;kd` in built-in command help.
- Exercises `;kd`, `;help`, and `clear` through actual command dispatch.

## Validation

Complete suite: 5,094 examples, 0 failures.

RuboCop: 1,031 files inspected, no offenses. Ruby syntax validation and `git diff --check` pass.

## Stack

6 of 6. Depends on lifecycle ownership hardening.

### DIFF
--- a/lib/global_defs.rb
+++ b/lib/global_defs.rb
@@ -1423,9 +1423,7 @@ end
 
 def clear(_opt = 0)
   unless (script = Script.current) then respond('--- clear: Unable to identify calling script.'); return false; end
-  to_return = script.downstream_buffer.dup
-  script.downstream_buffer.clear
-  to_return
+  script.clear
 end
 
 def match(label, string)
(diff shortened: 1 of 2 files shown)


MahtraDR:
Additive, and the only genuinely optional PR in the stack. Adds the `;kd` forced kill-all command and delegates the `clear` helper to the atomic `Script#clear`.

The specs here are source-string assertions (they check that the code reads a certain way, not that `;kd` behaves at runtime). The underlying `kill_all(force: true)` behavior is covered by the spec in #1474, so this is acceptable, but a behavioral test would be stronger. Safe to land last, or to defer without affecting the rest of the stack.

MahtraDR:
## Re-review update

This addresses the one note from the original review. The specs were source-string assertions before; there is now a behavioral test that spawns a subprocess, stubs `Script.kill_all`, runs `do_client(';kd')`, and asserts `kill_all` is invoked with `force: true` and `context: runtime`, plus tests that `;help` documents `kd` and that `clear` delegates to the atomic `Script#clear`. `;ka` was also refactored onto the count-based `Script.kill_all`.

Verified green. Safe to land last, or to defer without affecting the rest of the stack.

OSXLich-Doug (approved):

PR #1480, opened 2026-07-27
fix(all): surface the underlying LoadError in dependency alerts

Nisugi:
## Problem

When `require 'gtk3'` (or `require 'sqlite3'`) fails at startup, the alert dialog always blames a "missing gem" and never shows why the require actually failed:

- `lib/init.rb` discarded the exception on two paths: the initial `rescue LoadError` didn't capture it, and the no-self-healing branch called `GemCheck.alert` without `error:`.
- `GemCheck.build_alert_body` only rendered the error when the missing-gems list was empty — which it never is, because init.rb always passes the gem name literally.

So a user whose gems are all installed but whose native DLL failed to load (transient LoadLibrary failure, broken dependency deep in the gtk3 require chain, etc.) gets told "gtk3 is not installed" with no way to diagnose the real fault after the fact — nothing is on screen and the debug log isn't open yet at that point in startup.

## Fix

- Capture the original `LoadError` in both rescues and pass it through to `alert` on the paths that previously dropped it.
- `build_alert_body` now always appends `Underlying error: <first line of message>` when an error is present, alongside the missing-gems list instead of hidden behind it. (`write_log` already recorded the full class + message whenever the error was provided; now it is provided on every failure path.)

## Testing

- `bundle exec rspec spec/lib/gemcheck_spec.rb` — 82 examples, 0 failures (updated the three `build_alert_body` examples that codified the old either/or behavior)
- `bundle exec rubocop lib/gemcheck.rb lib/init.rb spec/lib/gemcheck_spec.rb` — no offenses

### DIFF
--- a/lib/gemcheck.rb
+++ b/lib/gemcheck.rb
@@ -282,7 +282,7 @@ module Lich
 
     # @param missing [Array<String>] gem names identified by our detector
     # @param groups [Array<Symbol>] groups being verified
-    # @param error [Exception, nil] the Bundler exception, if any
+    # @param error [Exception, nil] the Bundler or require exception, if any
     # @return [void]
     def alert(missing: [], groups: [:default], error: nil)
       write_log(missing: missing, groups: groups, error: error)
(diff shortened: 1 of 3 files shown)


MahtraDR:
@coderabbitai review in full

mrhoribu (approved):

PR #1481, opened 2026-07-27
fix(all): deflake PasswordCipher wrong-key specs

Nisugi:
## Problem

`spec/lib/common/gui/password_cipher_spec.rb` has two intermittently failing examples ("fails to decrypt with wrong account name" / "wrong master password"). They expect `DecryptionError` on every wrong-key decrypt, but AES-256-CBC only raises when the final block's PKCS#7 padding is invalid — about 1 in 256 wrong-key decryptions yields garbage ending in `0x01` (valid one-byte padding), which OpenSSL returns without error.

Measured locally: **9 non-raises in 2000 wrong-key decrypts** (~1/256, as predicted). With two such examples in the suite, roughly 0.8% of CI runs fail here at random — it just hit PR #1480 on a change nowhere near this code.

## Fix

Assert the property the cipher actually guarantees: the original password is never recovered with a wrong key. Each example now accepts either `DecryptionError` (the overwhelmingly common outcome) or a plaintext that does not match the original password.

The neighboring corrupted/truncated-data examples are deterministic (invalid strict Base64) and unchanged, and the `validate_master_password`/`validate_current_password` specs already compare round-tripped plaintext, so they were never affected.

## Testing

- `bundle exec rspec spec/lib/common/gui/password_cipher_spec.rb` — 18 examples, 0 failures
- `bundle exec rubocop spec/lib/common/gui/password_cipher_spec.rb` — no offenses

### DIFF
--- a/spec/lib/common/gui/password_cipher_spec.rb
+++ b/spec/lib/common/gui/password_cipher_spec.rb
@@ -29,11 +29,18 @@ RSpec.describe Lich::Common::GUI::PasswordCipher do
         expect(described_class.decrypt(encrypted2, mode: :standard, account_name: account_name)).to eq(password)
       end
 
-      it 'fails to decrypt with wrong account name' do
+      it 'never recovers the password with wrong account name' do
         encrypted = described_class.encrypt(password, mode: :standard, account_name: account_name)
-        expect do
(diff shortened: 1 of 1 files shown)


mrhoribu:
duplicate of #1473

PR #1483, opened 2026-07-28
feat(dr): DRCMM.moon_weapon_duration - read a moon weapon's remaining life

MahtraDR:
## Summary

Adds two `DRCMM` (Moon Mage commons) methods for reading how long a summoned moon weapon will last, so scripts can refresh a moonblade/moonstaff **before** it dissipates instead of reacting after it is already gone.

### `DRCMM.moon_weapon_duration`
FOCUSes the summoned moon weapon you are holding or wearing and returns its remaining duration in **roisaen** (1 roisaen == 60 seconds), or `nil` if you have no moon weapon or the message can't be parsed.

- Parses `You judge that the moonblade will last for roughly <count> roisaen.`
- Checks `moonblade` then `moonstaff` (mirrors `moon_used_to_summon_weapon`).
- Waits out the FOCUS roundtime only on a successful read.
- Documented as unsuitable for per-tick polling (FOCUS has roundtime). Moonblade duration ranges 12-41 minutes ([Elanthipedia](https://elanthipedia.play.net/Moonblade)).

### `DRCMM.parse_roisaen`
Converts the spelled-out count the game reports (`"twenty-nine"`, `"forty-one"`, `"five"`) into an `Integer`, or `nil` if any word is unrecognized. Reuses the canonical `Lich::DragonRealms::NUM_MAP`, so compounds may be hyphen- or space-separated and are summed; case-insensitive.

## Why

The game reports a moon weapon's remaining life via FOCUS, but there is currently no commons helper to read it. Scripts that use a summoned moonblade as a weapon can only tell *whether* one exists (`glance`/`moon_used_to_summon_weapon`), not *how long* it will last, so they tend to recast on a fixed timer that doesn't match the variable (12-41 min) duration and can let the blade expire mid-use. This exposes the actual remaining time as a primitive; the consuming logic lives in a separate dr-scripts change.

## Tests

`spec/lib/dragonrealms/commons/common_moonmage_spec.rb`:

- **`parse_roisaen` across the full 0-41 range** - each value is generated by an independent English-number oracle built in the spec (deliberately *not* the production `NUM_MAP`), so the test does not merely mirror the implementation. Plus boundary values (12/41/0/1), formatting tolerance (space vs hyphen, case, surrounding whitespace), and adversarial/malformed input (nil, empty, whitespace, non-number words, digits, partial words, mixed known/unknown -> `nil`).
- **`moon_weapon_duration` end-to-end** - moonblade read, single-word and max durations, moonstaff fallback, moonblade-before-moonstaff ordering with early stop, not-found, `Focus on what`, unparseable, and roundtime handling (waits on success, not on miss).

`drvariables.rb` (pure constants, provides `NUM_MAP`) is required in the spec; it is already required by several existing specs. Full `commons` + `drinfomon` suites pass (2338 examples, 0 failures) and `rubocop` is clean on both changed files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved moon weapon duration detection for moonblades and moonstaffs.
  * Correctly interprets durations written as words, including hyphenated values.
  * Recognizes weapons with less than one roisan remaining as expiring immediately.
  * Improved fallback behavior when a weapon cannot be identified or focused.
* **Reliability**
  * Handles invalid duration messages and roundtime more consistently.

### DIFF
--- a/lib/dragonrealms/commons/common-moonmage.rb
+++ b/lib/dragonrealms/commons/common-moonmage.rb
@@ -350,17 +350,21 @@ module Lich
 
       # Converts a spelled-out roisaen count (as reported by FOCUS, e.g. "five",
       # "twenty", "twenty-nine", "forty-one") into an Integer, or nil if any word
-      # is not a recognized number. Reuses the canonical DR number word map, so
-      # compounds may be hyphen- or space-separated and are summed
-      # ("forty-one" -> 41). Case-insensitive.
+      # is not a recognized number. Compounds may be hyphen- or space-separated
+      # and are summed ("forty-one" -> 41). Case-insensitive.
+      #
+      # Uses the global $NUM_MAP published by drinfomon/drvariables.rb (the same
+      # number word map DRC.text2num uses) rather than the Lich::DragonRealms
+      # constant, to avoid any namespace-resolution surface.
       def parse_roisaen(text)
         return nil if text.nil?
+        return nil if $NUM_MAP.nil?
 
         words = text.downcase.tr('-', ' ').split
         return nil if words.empty?
-        return nil unless words.all? { |word| NUM_MAP.key?(word) }
+        return nil unless words.all? { |word| $NUM_MAP.key?(word) }
 
-        words.sum { |word| NUM_MAP[word] }
+        words.sum { |word| $NUM_MAP[word] }
       end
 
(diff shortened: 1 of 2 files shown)


MahtraDR:
@coderabbitai review in full

mrhoribu (approved):

PR #1484, opened 2026-07-28
fix(all): extract game-agnostic CreatureBase into lib/common

MahtraDR:
## What

Extracts the game-agnostic core of the GemStone creature model out of `lib/gemstone/creature.rb` into a new shared `lib/common/creature/creature_base.rb`, following the `common/map/map_base.rb` convention: a `Lich::Common::CreatureBase` mixin whose `self.included` hook wires `ClassMethods` (id-keyed registry, current-room roster, target/room queries) and `InstanceMethods` (`<crtrStatus>` status/classification flag tracking) into the host class.

## Why

DragonRealms now emits the same `<crtrStatus exist=... hostile=1 .../>` feed GemStone gets, with an identical flag vocabulary. The registry, roster, `sync_crtr_status` reconciliation and target filtering are all game-neutral and belong in `lib/common`. This PR is the safe, behavior-preserving first step; a follow-up will add the DragonRealms `Creature` layer on top of this base.

## Scope / safety

- **Behavior-preserving.** No public API changes, no new runtime behavior.
- `Lich::Gemstone::CreatureInstance` / `Creature` keep their exact signatures (`register(name, id, noun)`, `new(id, noun, name)`), so the combat tracker/processor and every other consumer are untouched. GemStone-only concerns (bestiary templates, UCS tracking, HP/injury modelling, GemStone `valid_target?` exclusions) stay in `lib/gemstone/creature.rb`, which now just `include`s the base.
- Each extending class gets an independent registry (per-class instance-var storage), so GemStone and DragonRealms never share a roster.

## Tests

- New DAMP spec `spec/lib/common/creature/creature_base_spec.rb` exercises the shared base directly through a minimal stand-in creature class (22 examples).
- Two existing GemStone spec cases that reached into the private `@@instances` class variable now use the equivalent public API (`register` + `clear_room`), which is less brittle and matches the new per-class storage.
- Full suite green (no new failures); touched files are rubocop-clean.

## Summary by CodeRabbit

* **New Features**
  * Added shared creature tracking across game environments.
  * Added creature registry and room roster management.
  * Added status, classification, expiration, and filtering support.
  * Added configurable automatic registration, capacity limits, cleanup, and debug controls.
  * Improved target and in-room queries, including negated and unknown-filter handling.

* **Bug Fixes**
  * Improved room roster isolation and protection against unintended changes.
  * Corrected cleanup behavior and facade delegation for creature management.

### DIFF
--- a/lib/common/creature/creature_base.rb
+++ b/lib/common/creature/creature_base.rb
@@ -275,6 +275,11 @@ module Lich
 
         # Configures registry limits.
         #
+        # Every call resets omitted options to their defaults - max_size to 1000
+        # and auto_register to true - so calling this with no arguments restores
+        # all defaults. Facade callers (e.g. `Creature.configure(**options)`)
+        # should pass every option they mean to keep.
+        #
         # @param max_size [Integer] maximum number of retained instances.
         # @param auto_register [Boolean] whether {#register} creates instances.
         # @return [void]
--- a/lib/common/creature/creature_base.rb
+++ b/lib/common/creature/creature_base.rb
@@ -1,0 +1,518 @@
+# frozen_string_literal: true
+
+# =============================================================================
+# common/creature/creature_base.rb
+#
+# Game-agnostic core of Lich's runtime creature model, shared by the GemStone
+# and DragonRealms `Creature` implementations. Mirrors the structure of
+# common/map/map_base.rb: a single {Lich::Common::CreatureBase} mixin whose
+# `self.included` hook wires the shared {CreatureBase::ClassMethods} (id-keyed
+# registry, room roster and target queries) and {CreatureBase::InstanceMethods}
+# (transient status / classification tracking) into the including class.
+#
+# Both games receive the same structured, id-keyed creature feed:
+#
+#   <crtrStatus exist="..." hostile="1" .../>
+#
+# a full snapshot of a creature's transient combat status and classification
+# flags. The flag vocabulary is identical across the two games (see
+# {CRTR_STATUS_FLAGS} and {CRTR_CLASSIFICATION_FLAGS}), so status/flag
+# reconciliation, the registry, the room roster and target filtering all live
+# here and are mixed into each game's concrete `CreatureInstance`.
+#
+# Game-specific concerns stay out of this file:
+#   * GemStone: bestiary templates, UCS (Unarmed Combat System) tracking,
+#     HP/injury modelling and GemStone `valid_target?` exclusions
+#     (lib/gemstone/creature.rb).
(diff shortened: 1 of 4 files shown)


MahtraDR:
@coderabbitai review in full

MahtraDR:
@coderabbitai review in full

MahtraDR:
@coderabbitai — addressed all three comments in `587fdecb`. Summary of what changed and one clarification:

**1. `#flag_active?` spec description (nitpick — valid, fixed)**
The description overstated coverage. `flag_active?` has no negation handling (that lives in `apply_filters`) and no assertion here exercised `not_`. Trimmed to match:
```diff
-    it 'matches a status, a classification flag, or a not_ negation' do
+    it 'matches either a status or a classification flag' do
```
Assertions/behavior unchanged.

**2. `configure` YARD doc (nitpick — valid, fixed)**
Made the "omitted options reset to defaults" behavior explicit, since the specs rely on the no-arg-restores-defaults form and facade callers (`Creature.configure(**options)`) shouldn't be surprised:
```diff
         # Configures registry limits.
         #
+        # Every call resets omitted options to their defaults - max_size to 1000
+        # and auto_register to true - so calling this with no arguments restores
+        # all defaults. Facade callers (e.g. `Creature.configure(**options)`)
+        # should pass every option they mean to keep.
+        #
         # @param max_size [Integer] maximum number of retained instances.
         # @param auto_register [Boolean] whether {#register} creates instances.
```
Behavior unchanged.

**3. `Time.now` self-referential stub (potential issue — applied, but the claimed failure does not occur)**
Adopted the hoist for clarity:
```diff
-      # Fast-forward past the expiry rather than sleeping.
-      allow(Time).to receive(:now).and_return(Time.now + 6)
+      # Fast-forward past the expiry rather than sleeping. Compute the target
+      # time before stubbing so it never reads back through the stub.
+      later = Time.now + 6
+      allow(Time).to receive(:now).and_return(later)
```
For the record, the described `nil + 6` failure mode isn't reachable in this form: Ruby parses `allow(Time).to receive(:now).and_return(Time.now + 6)` as `allow(Time).to(receive(:now).and_return(Time.now + 6))`, so the `Time.now + 6` argument is evaluated while building the `receive(...).and_return(...)` matcher — before `.to` installs the stub. The existing test also passed, which would not be the case if it were computing `nil + 6`. Applied the change anyway to remove the self-referential pattern.

Validation: `creature_base_spec` + `creature_spec` green (105 examples, 0 failures), rubocop clean on both touched files.

OSXLich-Doug (approved):

PR #1485, opened 2026-07-28
feat(dr): id-based creature tracking from crtrStatus + assess

MahtraDR:
## Stacked on #1484

This PR is **stacked on #1484** (the behavior-preserving `CreatureBase` extraction) and contains that commit as its base. Please review #1484 first; the net-new surface here is the single `feat(dr): ...` commit. Once #1484 merges, this PR's diff reduces to just the DragonRealms layer.

## What

Adds a DragonRealms runtime creature layer (`lib/dragonrealms/creature.rb`) on top of the shared `Lich::Common::CreatureBase`, giving DR the id-keyed, flag-aware creature model GemStone already has. DR previously had no ids for room creatures — only name strings in `DRRoom` — so this is **purely additive** and leaves `DRRoom` and its ~120 script consumers untouched.

## Why now

DragonRealms recently began emitting `<crtrStatus exist=... hostile=1 immobile=1/>` — the same structured, id-keyed feed GemStone gets, with an identical flag vocabulary. DR splits creature data across three streams (vs GemStone's single `<a exist noun>` room-object tag):

| Stream | Gives | Missing |
|---|---|---|
| `<crtrStatus>` (automatic, batched after room-objs) | `exist` id + `hostile`/`immobile`/`disengaged` | name |
| room-objs bold text | name | id (left to `DRRoom` as before) |
| `assess` stream | id **+** name + assess number + relation + range | — |

So creatures are registered **id-first** from `<crtrStatus>` (name-less) and have their name/position **backfilled from `assess`** — the only feed that ties an `exist` id to a name.

## Parser wiring (`lib/common/xmlparser.rb`) — all guarded by game + `defined?`

- The `<crtrStatus>` handler routes DR tags to `Creature.sync` immediately (DR has no `<a>` text path to defer to), while GemStone keeps its existing deferred `@pending_crtr_status` path unchanged.
- `nav` and room-objs refresh now also clear the DR roster, so the following `<crtrStatus>` batch rebuilds a clean snapshot.
- Completed `assess` creature entries are fed to `Creature.feed_assess`.
- `gameloader` loads the DR creature file for DR sessions.

## Tests

- DAMP unit spec for the DR layer (`spec/lib/dragonrealms/creature_spec.rb`).
- End-to-end spec (`spec/lib/common/xmlparser_dr_creature_spec.rb`) driving the **real Ox pipeline** with fragments captured from a live DR session (five jeol moradu, exist 99353095..99355351): crtrStatus roster + flags, roster rebuild on refresh, and assess name/position backfill.
- The GemStone crtrStatus spec's `XMLData` double now also reports `game` (the handler consults it to route GS vs DR).
- Full suite green (no new failures); touched files rubocop-clean.

## Follow-up commits (on this branch, after the base `feat(dr)` commit)

These build on the layer above; the "assess is the *only* id→name tie" note in **Why now** is superseded by the first of these (room-objs order is now a second, every-refresh id→name path).

- **`feat(dr): name creatures from room-objs stream order + future-proof <a>`** — DR emits `<crtrStatus>` in the same order as the bold room-objs names (one per bold NPC, items skipped, counts equal, cross-checked vs assess id-order). Pair the Nth bold name → Nth `crtrStatus` id, so creatures are **named every refresh without waiting for `assess`**. Adds a derived `noun` (`"a jeol moradu"` → `"moradu"`; DRRoom `CREATURE_NAME` convention) making the layer a superset of `DRRoom.npcs`. Bounds-checked (surplus `crtrStatus` → nil, never a mis-pair); `assess` stays authoritative. Also: future-proofs the room-objs bold-`<a>` path (if DR ever emits GemStone-style `<a exist noun>`, route to DR's `Creature`, guarded — the stream-order branch then goes dormant), and fixes a latent DR crash — `GameObj.npcs&.last&.status=` guard, since `GameObj.npcs` is nil in DR so a `(dead)`/`(immobile)` room run raised and reset the parser mid-fragment.

- **`feat(dr): parse assess enrichment (balance, conditions, target)`** — expands the assess backfill: `target`/`target_number` (already parsed, previously dropped); `balance` (a `DR_BALANCE_VALUES` descriptor) + `off_balance?`; an open-ended `conditions` array with `condition?(name)`/`cursed?`; and `enriched?`/`enriched_at` so consumers can **poll-before-act** (assess only when a mob is un-enriched or a depended-on field is stale — assess fields are a staleable *pull* snapshot, unlike the *push* `crtrStatus` flags refreshed every room update).

- **`fix(dr): keep crtrStatus flag words out of assess conditions`** — live assess puts crtrStatus states in the parenthetical too (e.g. `(1: immobile and slightly off balance)`, `(1: stunned and very badly balanced)`). Those words are dropped from `conditions` (rejected via `CreatureBase::ALL_CRTR_FLAGS`) and tracked via `crtr_flag?`/`has_status?` (fresh), so the two sources never duplicate or disagree; `conditions` holds only assess-only afflictions (`cursed`, `poisoned`, …).

- **`fix(dr): track hidden crtrStatus flag, harden noun derivation, expand adversarial specs`** — adversarial review found two real bugs: (1) `hidden` was in none of the flag maps, so DR's `<crtrStatus ... hidden="1"/>` was silently dropped (`has_status?('hidden')` always false) — added to `CRTR_STATUS_FLAGS` (touches the shared `CreatureBase`; arguably belongs in #1484); (2) `derive_noun` was end-anchored, returning a nil noun on any trailing whitespace/punctuation — switched to `scan+last`. Plus edge/boundary specs: multi-affliction, no-balance, nil/empty status, the balance ladder both sides of "solidly", apostrophe/multi-word nouns.

- **`fix(dr): all-or-nothing count gate for room-objs name pairing`** — the per-tag pairing could mis-name on a *mid-list gap* (a bold room entity that emits no `crtrStatus`, e.g. a gelapod-like object, shifts every creature after it). Replaced with a batched gate: collect crtrStatus ids in order (flags still applied id-first immediately), then at the following `<prompt>` apply bold names by position **only when the counts match exactly**; on any mismatch apply **no** names (assess backfills by id) rather than risk a shifted mis-pair. `Creature.sync` dropped its now-unused `name` param.

- **`fix(dr): parse "imbalanced" balance phrasing; add friendly?`** — harvested ~2M assess parentheticals from live logs: the worst balance levels are phrased `extremely imbalanced` (not `...balanced`), which the pattern missed → `balance` nil and `off_balance?` false for the *most* vulnerable creatures; now accepts the optional `im`. The harvest also showed assess exposes friend/foe via a `friendly` marker (`friendly, cursed, and nimbly balanced`) that crtrStatus does **not** carry — added `friendly?`. Confirmed Oxford-comma separators too.

Each ships with specs — stream-order naming + noun, count-gate (both mismatch directions + a gelapod-like no-crtrStatus mob never mis-named), GemStone-`<a>` future-proofing, enrichment fields, flag-word exclusion, and parser edge cases drawn from the real log vocabulary (`extremely imbalanced`, `friendly`/Oxford-comma, `hidden`/`immobile`/`stunned` exclusion). Full suite green apart from two pre-existing flaky specs (enhancive order-dependence, hmr worktree-path); touched files rubocop-clean. Verified live in a DR combat room: names/nouns populate before any `assess`, enrichment fills in on assess and persists by id.

### DIFF
--- a/lib/common/creature/creature_base.rb
+++ b/lib/common/creature/creature_base.rb
@@ -84,7 +84,8 @@ module Lich
         'blind'       => nil,
         'sunburst'    => nil,
         'webbed'      => nil,
-        'poisoned'    => nil
+        'poisoned'    => nil,
+        'hidden'      => nil
       }.freeze
 
       # Maps `<crtrStatus>` transient XML flags to canonical status names.
@@ -107,7 +108,8 @@ module Lich
         'prone'       => 'prone',
         'sitting'     => 'sitting',
         'flying'      => 'flying',
-        'hovering'    => 'hovering'
+        'hovering'    => 'hovering',
+        'hidden'      => 'hidden'
       }.freeze
 
       # Maps `<crtrStatus>` classification XML flags to predicate keys.
--- a/lib/common/gameloader.rb
+++ b/lib/common/gameloader.rb
@@ -58,6 +58,7 @@ module Lich
         require File.join(LIB_DIR, 'dragonrealms', 'dependency', 'settings_config.rb')
         require File.join(LIB_DIR, 'dragonrealms', 'drinfomon.rb')
         require File.join(LIB_DIR, 'dragonrealms', 'commons.rb')
+        require File.join(LIB_DIR, 'dragonrealms', 'creature.rb')
         DRInfomon.watch!
         self.common_after
       end
--- a/lib/common/xmlparser.rb
+++ b/lib/common/xmlparser.rb
@@ -378,8 +378,10 @@ module Lich
             GameObj.clear_pcs
             GameObj.clear_room_desc
             # Creature tracks its own room roster independently of GameObj
-            # (see lib/gemstone/creature.rb) - not loaded for DR sessions.
+            # (see lib/gemstone/creature.rb / lib/dragonrealms/creature.rb).
+            # Only one game's Creature is ever loaded per session.
             Lich::Gemstone::Creature.clear_room if defined?(Lich::Gemstone::Creature)
+            Lich::DragonRealms::Creature.clear_room if defined?(Lich::DragonRealms::Creature)
             # Any <crtrStatus> cached for the room being left is scoped to
             # that room - don't let it survive to misapply if the id gets
             # reused elsewhere.
@@ -417,6 +419,9 @@ module Lich
             if attributes['id'] == 'room objs'
               GameObj.begin_room_objs
               Lich::Gemstone::Creature.clear_room if defined?(Lich::Gemstone::Creature)
+              # DR rebuilds the roster from the <crtrStatus> batch that follows
+              # this component; clearing here gives that batch a clean snapshot.
+              Lich::DragonRealms::Creature.clear_room if defined?(Lich::DragonRealms::Creature)
               @pending_crtr_status.clear
             elsif attributes['id'] == 'room players'
               GameObj.begin_room_players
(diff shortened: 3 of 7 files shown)


OSXLich-Doug (approved):

MahtraDR:
Ran the CodeRabbit CLI (`coderabbit review --committed`) locally against `origin/main` (in sync with `upstream/main`). Result: **No findings** across all 7 changed files:
- `lib/common/creature/creature_base.rb`
- `lib/common/gameloader.rb`
- `lib/common/xmlparser.rb`
- `lib/dragonrealms/creature.rb`
- `spec/lib/common/xmlparser_crtr_status_spec.rb`
- `spec/lib/common/xmlparser_dr_creature_spec.rb`
- `spec/lib/dragonrealms/creature_spec.rb`

PR #1488, opened 2026-07-30
fix(all): bound GemCheck alert dialogs with a timeout so they can't hang forever

simtel12:
## Summary
- `alert_linux`'s `system('zenity', ...)` call had no timeout, so on a display with no one present to click the dialog (e.g. a dead or forwarded X session), it blocked `lich.rbw`'s startup indefinitely.
- `alert_windows` had the same latent issue: `WScript.Shell.Popup`'s timeout was hardcoded to `0` (wait forever).
- Adds a `run_with_timeout` helper that spawns the dialog process directly and kills it if it outlives `ALERT_TIMEOUT_SECONDS` (120s), since wrapping a blocking `Kernel#system` call in `Timeout` can't safely reclaim an orphaned child process.

## Test plan
- [x] `bundle exec rspec spec/lib/gemcheck_spec.rb`
- [x] Manually confirmed `run_with_timeout` reclaims control from a hung `zenity`/`sleep` process within the timeout window

### DIFF
--- a/lib/gemcheck.rb
+++ b/lib/gemcheck.rb
@@ -1,5 +1,6 @@
 # lib/gemcheck.rb
 require 'bundler'
+require 'timeout'
 require_relative 'bundler_recovery'
 require_relative 'dependency_recovery'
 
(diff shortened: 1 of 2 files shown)


simtel12:
Would it make sense for GemCheck to write to stderr or similar so that there is a console log of the error as well?  At the moment, if you don't have the UI available/visible it just hangs for 2 minutes and then exits with a code 1.

OSXLich-Doug (approved):

PR #1490, opened 2026-07-31
fix(all): identify Genie as the frontend for headless launches

simtel12:
## Problem

`--headless PORT` normalizes to `--without-frontend --detachable-client=PORT` (`lib/main/arg_normalization.rb`), and the `--without-frontend` branch in `lib/main/main.rb` resolved `Frontend.client` like this:

```ruby
Frontend.client = if ARGV.any? { |a| a =~ /^--saga$/i }
                    'saga'
                  elsif @argv_options[:detachable_client_port] && !ARGV.any? { |a| a =~ /^--genie$/i }
                    'profanity'
                  else
                    'unknown'
                  end
```

The `!--genie` guard correctly keeps a Genie client from being labelled `profanity`, but there is no branch that labels it `genie`, so `lich.rbw --login Char --dragonrealms --genie --headless 8000` ends up as `'unknown'`.

`'unknown'` isn't in the capability registry, so `Frontend.supports_mono?` is false and `respond` skips both the `<output class="mono"/>` wrapper and the XML escaping:

```ruby
if Frontend.supports_mono?
  str = "<output class=\"mono\"/>\r\n#{str.gsub('&', '&amp;').gsub('<', '&lt;').gsub('>', '&gt;')}<output class=\"\"/>\r\n"
elsif Frontend.client.eql?('profanity')
```

Genie then receives script output with raw angle brackets and its XML parser consumes anything between `<` and `>` as a tag. `;validate` is where I hit this — it wraps every message as `ERROR:< #{message}  >`, so the entire body vanished and the run reported counts with no text:

```
  CHECKING: 108 different potential errors in [Char-setup.yaml]
[validate: ERROR:]
[validate: WARNING:]
[validate: WARNING:]
[validate: WARNING:]
[validate: WARNING:]
[validate: ERROR:]
  WARNINGS:4 ERRORS:2
```

The same session also loses monospace for all script output, since the mono bracket is never emitted. A profanity user is unaffected — the `elsif` branch escapes for them.

`handle_detachable_client` in `lib/global_defs.rb` already special-cases `--genie` for detachable clients, so the combination is supported elsewhere; this branch just never got the case.

## Change

Move the resolution into `LoginHelpers.resolve_headless_frontend` (next to `resolve_lookup_frontend`, which already reasons about the frontend a headless launch intends) so it's unit-testable, and add the missing genie branch.

Behavior is otherwise unchanged: saga still wins outright, a detachable client with no frontend flag is still assumed to be profanity, and everything else is still `'unknown'`.

## Test plan

- [x] `bundle exec rspec spec/lib/common/authentication/login_helpers_spec.rb` — 80 examples, 0 failures (4 new, confirmed failing before the fix)
- [x] `bundle exec rspec` — 4932 examples, 0 failures
- [x] `bundle exec rubocop` on the changed files — no offenses
- [ ] Live: `lich.rbw --login Char --dragonrealms --genie --headless 8000` with Genie attached; `;validate` prints its warning and error text, and script output renders monospaced again

### DIFF
--- a/lib/common/authentication/login_helpers.rb
+++ b/lib/common/authentication/login_helpers.rb
@@ -543,6 +543,28 @@ module Lich
           requested_frontend
         end
 
+        # Resolves the frontend identity for a headless (`--without-frontend`)
+        # launch.
+        #
+        # Lich spawns no frontend here, but a detachable client still attaches
+        # and renders the stream, so the identity has to name whatever is
+        # reading it: `respond` uses the frontend's registered capabilities to
+        # decide whether to XML-escape script output and wrap it in mono
+        # brackets. A Genie client left as 'unknown' receives unescaped output,
+        # so any script text containing angle brackets is swallowed by Genie's
+        # XML parser as if it were a tag.
+        #
+        # @param argv [Array<String>] command line arguments
+        # @param detachable_client [Boolean] whether a detachable client port is configured
+        # @return [String] frontend identity for Frontend.client
+        def self.resolve_headless_frontend(argv, detachable_client: false)
+          return 'saga' if argv.any? { |arg| arg.match?(/^--saga$/i) }
+          return 'unknown' unless detachable_client
+          return 'genie' if argv.any? { |arg| arg.match?(/^--genie$/i) }
+
+          'profanity'
+        end
+
         # Formats the game instance launch flag for Lich based on version.
         #
         # Older versions of Lich (pre-5.12) only recognize specific lowercase flags
--- a/lib/main/main.rb
+++ b/lib/main/main.rb
@@ -235,13 +235,9 @@ reconnect_if_wanted = proc {
       Lich.log "info: Current WINE working directory is #{custom_launch_dir}"
     end
     if ARGV.include?('--without-frontend')
-      Frontend.client = if ARGV.any? { |a| a =~ /^--saga$/i }
-                          'saga'
-                        elsif @argv_options[:detachable_client_port] && !ARGV.any? { |a| a =~ /^--genie$/i }
-                          'profanity'
-                        else
-                          'unknown'
-                        end
+      Frontend.client = Lich::Common::Authentication::LoginHelpers.resolve_headless_frontend(
+        ARGV, detachable_client: !@argv_options[:detachable_client_port].nil?
+      )
       unless (game_key = @launch_data.find { |opt| opt =~ /KEY=/ }) && (game_key = game_key.split('=').last.chomp)
         $stdout.puts "error: launch_data contains no KEY info"
         Lich.log "error: launch_data contains no KEY info"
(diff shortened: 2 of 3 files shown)


MahtraDR (approved):
Approving. I traced this against the code rather than the description, and it holds up on every axis.

## Behavior preservation (verified by truth table)
Reconstructed both the old inline `if/elsif` and the new `resolve_headless_frontend` across all three inputs (saga / detachable / genie). The **only** cell that changes is `saga=0, detachable=1, genie=1`, which goes from `'unknown'` (the bug) to `'genie'` (the fix). Every other path -- saga wins outright, detachable + no flag = profanity, everything else unknown -- is identical to before. "Behavior otherwise unchanged" is exactly right.

## The fix actually reaches the symptom
- `front-end.rb` registers `genie` with `%i[xml mono]`.
- `respond` gates XML-escaping + the `<output class="mono"/>` wrapper on `Frontend.supports_mono?`, i.e. `has_capability?('genie', :mono)` -> true (the registry normalizes `to_s.downcase`, so the `'genie'` string resolves).

So setting `'genie'` instead of `'unknown'` genuinely restores both the escaping and the mono font -- the `;validate` swallowing and the session-wide mono loss both trace back to this one identity string.

## Load-order footgun -- checked, it's clear
The new cross-module call in `main.rb`'s startup branch would `NameError` if `LoginHelpers` weren't loaded yet. It's safe: unconditional require chain `lich.rbw` -> `argv_options.rb` -> `cli_orchestration.rb` -> `authentication/cli.rb` -> `login_helpers.rb`, all before `main.rb` is required.

## Boolean coercion -- checked the one way it could bite
`detachable_client: !@argv_options[:detachable_client_port].nil?` only diverges from the old truthiness check if the port were literally `false`. It's only ever `nil` or an Integer (`0` being the documented OS-assign case), so `nil`->false / Integer->true matches the old behavior in every reachable state. No regression, and using `.nil?` (rather than truthiness) is the more correct guard given `0` is valid.

## SOLID / YARD / tests
- Pulling the logic out of a monolithic untested startup branch into a side-effect-free, single-responsibility function is a clear improvement; the keyword boolean avoids the positional-boolean trap.
- YARD is complete and, more usefully, explains *why* the identity matters.
- Specs cover all four return paths and each description matches its assertion; the `--genie`-without-detachable case is a nice touch, proving genie alone doesn't force `'genie'`. Locally: 80 examples / 0 failures, rubocop clean on all three files.

## Two non-blocking follow-up thoughts
1. The extracted function is fully covered, but the `main.rb` call-site wiring (the `Frontend.client =` assignment and the `!...nil?` conversion) isn't -- that's the one spot a future change to the port's type could slip through untested. Fine to leave to the manual live check for now.
2. Frontend-flag detection regexes (`--saga` / `--genie`) are now scattered across `resolve_headless_frontend`, `handle_detachable_client`, and the `argv_options.rb` ternaries. This file already has `gemstone_flag?` / `dragonrealms_flag?`; `saga_flag?` / `genie_flag?` helpers would be a tidy consolidation someday. Pre-existing, not this PR's job.

Nice, tight fix with a genuinely useful commit message and test plan.

PR #1491, opened 2026-07-31
feat(dr): capture room UID from <nav> tag; stop enforcing ShowRoomID; opt-in room-id placement

MahtraDR:
## Summary

DragonRealms now emits native `<nav rm='NNNN'/>` tags on every arrival (a bare `<nav/>` with no `rm` for a room that has no UID), matching GemStone. This PR adopts the nav tag as the authoritative DR room-UID source instead of scraping the `streamWindow` subtitle title bar, which only carried the UID when the game's `ShowRoomID` flag was on. On the back of that it stops Lich from fighting the flag, and adds an opt-in choice of *where* the room id/UID is displayed.

**Defaults are unchanged** for existing players: nothing is embedded in the title unless the player opts in; `display_lichid`/`display_uid` stay off for DR; the new placement setting defaults to the historical below-room line.

### Core: nav becomes the DR room-UID source (`lib/common/xmlparser.rb`)
- The `<nav>` handler now sets `room_id`/`previous_nav_rm` for DR too (the `unless XMLData.game =~ /^DR/` guard is removed). This gives early room knowledge before the room text streams and finally populates `Map.previous_uid` for DR (previously always nil).
- The subtitle UID scrape is removed. `<nav>` precedes the subtitle in the stream, and a ShowRoomID-off subtitle carries no UID, so scraping it would write `0` and clobber the good nav value. The title parse stays, so `XMLData.room_title` remains UID-free (scripts depend on exact title matching).
- `@room_id` initializes to `0` (not `nil`) so a read before the first `<nav>` is a valid "no UID" value instead of a `NoMethodError` on `room_id.zero?` / `room_id > N` in the map layer (hardens GS too).

### Stop enforcing ShowRoomID (nav works with the flag on or off)
- `lib/dragonrealms/drinfomon/drparser.rb`: remove the handler that forced `flag showroomid on` and nagged the player.
- `lib/dragonrealms/drinfomon/startup.rb`: drop `ShowRoomID` from the one-time forced flags. `MonsterBold` is still enforced.

### Opt-in display + placement control
- `lib/lich.rb`: new `Lich.display_roomid_location` (`title` | `line` | `both`, DR default `line`, invalid input ignored, DB-persisted) beside the other `display_*` settings.
- `lib/global_defs.rb`: new `;display roomid <title|line|both>` command, help entry, and a note that `;display flaguid` is largely unnecessary now (a player can just `flag showroomid off`).
- `lib/games.rb`: for the `title`/`both` placements, inject the id/UID into the inline `<style id="roomName"/>` line GS-style (visible on every front-end, including ProfanityFE/Genie); gate the below-room `Room Number:` line to `line`/`both`. Three private DRY helpers (`room_number_display`, `room_id_in_title?`, `room_id_in_line?`) are shared by `modify_room_display` and `process_room_display` so their formatting can't drift; the shared helper also guards a latent `Map.current` nil crash. **GS code is untouched.**

### Why this is script-safe
The injection edits only the outbound client string. `process_xml_data` (which populates `XMLData` via Ox SAX) runs before `process_downstream_hooks` (where the display rewrite happens), so `XMLData.room_title`/`room_id` - and therefore every script - are unaffected.

## Player-facing

| Want | Do |
|---|---|
| Nothing (default) | leave defaults |
| Lich id in the title only | `;display lichid` + `;display roomid title` |
| Room number below the room (classic) | `;display lichid` (and/or `;display uid`) |
| Both places | `;display roomid both` |

`;display uid` now works even with ShowRoomID off, since the UID comes from `<nav>`.

## Tests

DAMP, gap-free coverage added/updated across `xmlparser_spec`, `map_dr_spec`, `drparser_spec`, `startup_spec`, `lich_room_display_settings_spec`, `games_spec`, and `spec_helper`:
- nav capture, including the clobber-prevention regression and the bare-`<nav/>` zero case;
- `display_roomid_location` default / lowercase-normalize / invalid-ignored / persistence, exercised against the real `lib/lich.rb` in a subprocess;
- ShowRoomID de-enforcement;
- placement rendering (title/line/both, `(**)` for a no-UID room, strip-before-inject, unmapped nil-guard, no `XMLData` mutation) and a GS-unaffected guard.

Full suite green (`ruby -c` clean, `rubocop` 0 offenses on all touched files). The only failing example is the pre-existing `enhancive_spec` flake tracked by a separate PR.

### DIFF
--- a/lib/common/map/map_dr.rb
+++ b/lib/common/map/map_dr.rb
@@ -339,6 +339,19 @@ module Lich
           return set_current(room.id)
         end
 
+        # Guard against a blank/incomplete arrival frame. DR occasionally streams a room whose
+        # <nav> UID is delayed or absent (room_id 0) before the room text populates: the
+        # description is only the "pitch dark" placeholder and there are no exits. Minting a
+        # room here creates a junk stub (empty title, no UID) that orphans or duplicates the
+        # real room. Keep the current room instead; the real room resolves by UID once the
+        # delayed nav (or a re-look) provides it.
+        if XMLData.room_id.zero? &&
+           XMLData.room_exits_string.to_s.strip.empty? &&
+           XMLData.room_description.to_s.strip == "It's pitch dark and you can't see a thing!"
+          echo 'Map: skipped blank/incomplete room frame (no uid, pitch-dark, no exits)'
+          return set_current(@@current_room_id)
+        end
+
         id = get_free_id
         title = [XMLData.room_title]
         description = [XMLData.room_description.strip]
--- a/lib/common/map/map_dr.rb
+++ b/lib/common/map/map_dr.rb
@@ -349,7 +349,13 @@ module Lich
            XMLData.room_exits_string.to_s.strip.empty? &&
            XMLData.room_description.to_s.strip == "It's pitch dark and you can't see a thing!"
           echo 'Map: skipped blank/incomplete room frame (no uid, pitch-dark, no exits)'
-          return set_current(@@current_room_id)
+          # Keep the current room - but only if one has actually resolved.
+          # @@current_room_id is the -1 sentinel before the first match; passing
+          # that to set_current would index @@list[-1] (the last room) and make an
+          # unrelated room current, so fall back to nil in that case instead.
+          return set_current(@@current_room_id) if @@current_room_id.is_a?(Integer) && @@current_room_id >= 0
+
+          return nil
         end
 
         id = get_free_id
--- a/lib/common/xmlparser.rb
+++ b/lib/common/xmlparser.rb
@@ -137,6 +137,11 @@ module Lich
         # rather than a NoMethodError on room_id.zero? (DR) or room_id > N (GS) in the map layer.
         @room_id = 0
         @previous_nav_rm = nil
+        # True once <nav> has supplied a real (non-zero) UID for the current
+        # arrival and the streamWindow subtitle has not yet consumed it. Keeps a
+        # ShowRoomID-on subtitle from overwriting an authoritative nav UID in the
+        # same arrival (see the streamWindow handler).
+        @nav_uid_pending = false
 
         # Lich::Claim update
         @arrival_pcs = []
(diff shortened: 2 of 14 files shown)


MahtraDR:
## Local CodeRabbit review run

Ran the CodeRabbit CLI locally against this branch using the repo's `.coderabbit.yaml` settings before requesting human review:

```
coderabbit review --agent --committed -c .coderabbit.yaml --base main
```

**Initial run (commit `92ea35e1`): 2 findings, both minor test-coverage tightenings — no code-logic or correctness issues.**

1. `spec/lib/dragonrealms/drinfomon/drparser_spec.rb` — the "ShowRoomID no longer enforced" example asserted no warning fires, but not that no command is sent to re-enable the flag.
2. `spec/lib/games_spec.rb` — the `"both"` placement example only checked the below-room `Room Number:` line, not the inline title injection.

**Addressed in commit `9e7b6009`:**

1. Added `expect(described_class).not_to receive(:put)` so the test proves the removed handler no longer sends `put("flag showroomid on")`, in addition to the existing no-warning assertion.
2. The `"both"` example now asserts both surfaces — `modify_room_display` (inline title injection) and `process_room_display` (below-room line).

Both verified locally: 244 examples in the two touched specs pass, `rubocop` clean, `ruby -c` clean.

**Re-review (commit `9e7b6009`): 0 findings** across all 13 changed files.

Full suite is green apart from the pre-existing `spec/lib/gemstone/enhancive_spec.rb` order-dependent flake (unrelated to this change; tracked separately).

MahtraDR:
Ran the CodeRabbit CLI (`coderabbit review --committed`) against `origin/main` (in sync with `upstream/main`). It surfaced two `major` (functional-correctness) findings:

1. **`lib/common/xmlparser.rb`** - the streamWindow subtitle UID could clobber the `<nav>` UID within the same arrival when the two differed.
2. **`lib/common/map/map_dr.rb`** - the blank/incomplete-frame guard in `current_or_new` called `set_current(@@current_room_id)` even when `@@current_room_id` was still the `-1` sentinel, which indexes `@@list[-1]` (the last room) and makes an unrelated room current.

Both are addressed in commit 5c84d3d3:

1. A per-arrival `@nav_uid_pending` flag makes `<nav>` authoritative - the subtitle marker is applied only as a fallback when nav did not supply a real UID this arrival (late/absent/bare `<nav/>`), and it never overwrites a good nav UID. Added a spec where the nav and subtitle UIDs differ (nav wins).
2. The guard now only falls back to `set_current` for a real (`>= 0`) id and returns `nil` otherwise. Added specs for the `-1` case.

Full suite green (5386 examples, 0 failures) and rubocop clean on the touched files. **Re-ran the CodeRabbit CLI on the updated branch: No findings** - both issues are now clear.

PR #1492, opened 2026-07-31
fix(all): restore detachable client disconnect notice to stdout

MahtraDR:
## Summary

Since #1459 made the detachable listener persistent, `Frontend.create_session_file` (which printed `character/host/port`) runs **once at startup** instead of on every re-listen. As a side effect, closing a detachable frontend no longer prints any session info to the shared controlling terminal.

Concretely: a wrapper script that spawns Lich with `--detachable-client` and then `exec`s ProfanityFE in the same terminal used to see the session descriptor (character/host/port) reappear when Profanity closed and released the screen. After #1459 that line is gone, so an operator no longer sees which session they just disconnected from.

This restores a **symmetric disconnect notice**, mirroring the existing connect-side line:

```
--- Lich: detachable client listening on 127.0.0.1:8000        <- connect (existing)
--- Lich: detachable client Mahtra disconnected from 127.0.0.1:8000 (0 attached)   <- disconnect (new)
```

## Changes

- **New `lib/main/detachable_client_notice.rb`** — `Lich::Main::DetachableClientNotice`, a pure formatter (`.address`, `.listening`, `.disconnected`) with IPv6-safe bracketing and YARD docs. Mirrors the existing `DetachableClientTarget` pattern so the notice wording lives in one place (DRY) and is unit-testable without sockets.
- **`lib/main/main.rb`** — the connect-side listening line now routes through the module (single source for the wording). Behavior-preserving; IPv6 bracketing moves from `server.local_address.ipv6?` to a host-string check, which is equivalent for real addresses and is covered by the spec.
- **`lib/global_defs.rb`** — `handle_detachable_client` now emits the disconnect notice to `$stdout` in its `ensure`, after `detachable_client_unregister` so the attached count reflects the clients that remain. Character comes from `SessionLifecycle.resolve_session_name` (the same derivation the old session descriptor used).
- **New `spec/lib/main/detachable_client_notice_spec.rb`** — 12 examples covering IPv4/IPv6/hostname/nil endpoints, the listening notice, and the disconnect notice with/without a name and varying attached counts.

## Testing

- `rubocop -A` on all touched files: no offenses.
- New spec: 12 examples, 0 failures. Related `detachable_client_target_spec` + `session_lifecycle_spec`: 36 examples, 0 failures.
- `ruby -c` clean on all edited files.

## CodeRabbit

CodeRabbit CLI (repo `.coderabbit.yaml`) was already run locally against this change: **0 findings on the code in this PR** (the new module, both wired lines, and the spec all clean).

It did surface **2 findings in pre-existing adjacent code from #1459** that this PR does not touch, so they are intentionally left out and will be addressed in a separate follow-up PR:

1. *(major)* `detachable_clients_respond` does not isolate per-client failures — a raise from one client's `puts_main_stream` aborts the loop and blocks output to the remaining attached clients.
2. *(minor)* `detachable_client_send_init` concatenates raw `GameObj`/`XMLData` values into the XML init payload without escaping `&`/`<`/`>`.

### DIFF
--- a/lib/global_defs.rb
+++ b/lib/global_defs.rb
@@ -2382,6 +2382,21 @@ rescue StandardError => e
 ensure
   client.close rescue nil
   detachable_client_unregister(client)
+  # Mirror the connect-side "listening on" line so an operator sharing the
+  # controlling terminal (for example a wrapper script that exec's a frontend)
+  # sees which session dropped and from where. Emitted after unregister so the
+  # attached count reflects the clients that remain.
+  if $_DETACHABLE_LISTENER_
+    session_name = Lich::Common::SessionLifecycle.resolve_session_name(
+      argv: ARGV, account_character: (Lich::Common::Account.character rescue nil)
+    )
+    $stdout.puts Lich::Main::DetachableClientNotice.disconnected(
+      name: session_name,
+      host: $_DETACHABLE_LISTENER_[:host],
+      port: $_DETACHABLE_LISTENER_[:port],
+      attached: detachable_client_count
+    ) rescue nil
+  end
   Lich::Common::ShutdownLog.info("detachable client cleaned up (#{detachable_client_count} attached)")
 end
 
--- a/lib/main/detachable_client_notice.rb
+++ b/lib/main/detachable_client_notice.rb
@@ -1,0 +1,57 @@
+# frozen_string_literal: true
+
+module Lich
+  module Main
+    # Builds the human-facing notices Lich prints to $stdout for detachable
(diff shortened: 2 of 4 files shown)


MahtraDR:
Ran the CodeRabbit CLI (`coderabbit review --committed`) against `origin/main` (in sync with `upstream/main`). Result: **No findings** across all 4 changed files:
- `lib/global_defs.rb`
- `lib/main/detachable_client_notice.rb`
- `lib/main/main.rb`
- `spec/lib/main/detachable_client_notice_spec.rb`

PR #1493, opened 2026-07-31
fix(all): harden detachable client teardown and escape init payload

MahtraDR:
## Summary

Follow-up to #1492. While reviewing that PR, CodeRabbit flagged two issues in **pre-existing** detachable-client code (introduced by #1459) that #1492 intentionally left untouched to stay focused. This PR fixes both, plus a small DRY refactor.

## Changes

1. **`detachable_clients_respond` isolates per-client failures** *(was: major)*
   A raise from one client's `puts_main_stream` aborted the `each` loop, so a single broken detachable connection starved every other attached client of that output. The write is now wrapped in a per-client `rescue`, mirroring the per-client rescue already used by `detachable_clients_close`.

2. **`detachable_client_send_init` escapes its free-text values** *(was: minor)*
   The init payload embedded raw game-state strings (`prepared_spell`, `mind_text`, `encumbrance_text`, and both hand item names) into XML-like markup. Unescaped `&`/`<`/`>` produced malformed markup for a newly attached frontend, the same class of issue `respond` already guards against. These values are now escaped.

3. **New `Lich::Common::XmlEntities.encode`** (DRY)
   A mirror of the existing `.decode`, giving XML entity encoding a single home (`&amp;` encoded first so nothing double-encodes). The two previously inlined encode chains in `respond` are routed through it as well, behavior-identical.

## Testing

- New `spec/lib/common/xml_entities_spec.rb`: 7 examples (encode incl. double-encode + `to_s` coercion, decode, and a decode(encode(x)) round-trip), 0 failures.
- `rubocop -A` on all touched files: no offenses.
- `ruby -c` clean on edited files.

## CodeRabbit

CodeRabbit CLI (repo `.coderabbit.yaml`) was run locally against this change, scoped to the base commit: **No findings.**

The two functional fixes above are the exact findings CodeRabbit raised during the #1492 review; this PR resolves them.

### DIFF
--- a/lib/common/xml_entities.rb
+++ b/lib/common/xml_entities.rb
@@ -15,6 +15,14 @@ module Lich
 
         str.gsub('&lt;', '<').gsub('&gt;', '>').gsub('&quot;', '"').gsub('&apos;', "'").gsub('&amp;', '&')
       end
+
+      # Encodes the three markup-significant characters so a raw value can be
+      # safely embedded in an XML-like payload sent to a frontend. &amp; is
+      # encoded first so the ampersands introduced by encoding < and > are not
+      # themselves double-encoded.
+      def self.encode(str)
+        str.to_s.gsub('&', '&amp;').gsub('<', '&lt;').gsub('>', '&gt;')
+      end
     end
   end
 end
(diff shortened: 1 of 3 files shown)


MahtraDR:
Ran the CodeRabbit CLI (`coderabbit review --committed`) locally against `origin/main` (in sync with `upstream/main`). Result: **No findings** across all 3 changed files:
- `lib/common/xml_entities.rb`
- `lib/global_defs.rb`
- `spec/lib/common/xml_entities_spec.rb`

OSXLich-Doug (approved):

PR #1499, opened 2026-08-06
fix(all): scrub startup credentials once login completes

mrhoribu:
## Summary
 
Startup credentials stayed readable for the entire session. Three of the four
containers holding them are reachable from any running script:
 
| Container | Holder | Contents | Reachable from a script |
|---|---|---|---|
| `@launch_data` | ivar on top-level `main` | eaccess key | yes |
| `@argv_options` | ivar on top-level `main` | account password | yes |
| `ARGV` | global constant | account password | yes |
| `game_key` | block-local in `@main_thread` | eaccess key | no |
 
Both are live credentials. The eaccess key is **not** single-use: it stays valid
for some period after issue, and the exact invalidation window is not documented
on our side. A key read out of `@launch_data` well into a session may still be
usable, which is the main reason this change is worth making rather than being
cosmetic cleanup.
 
`@launch_data` is set on `main` because `main.rb` is required from `lich.rbw:130`
and the `@main_thread` block closes over that same `self`. `@argv_options` lands
on `main` the same way, via `argv_options.rb:435`. Neither is ever cleared, and
`SideEffects.execute` / `GameConnection.execute` both return the Hash they
receive without duplicating it, so `OptionParser`'s module ivar is a second
reference to the same object.
 
This adds `Lich::Common::CredentialScrub` and calls it once every connection path
has finished with the credentials, plus makes the temp `.sal` removal failsafe.
 
## Both exposed credentials are live
 
The account password is reusable and sits in two script-readable places, so it is
the larger blast radius. But the eaccess key is not the spent artifact it might
look like: it remains valid for a period after issue, so `@launch_data` is holding
a working credential for the rest of the session, not a receipt. Reviewers should
treat both as live.
 
Checked while tracing this: nothing writes `@launch_data` or a raw `ARGV` to the
log file. `main.rb:323` already logs a scrubbed `launcher_cmd`, and the many
`ARGV.include?("--debug")` call sites are flag checks that this change leaves
intact. So the exposure is script-level and process-memory-level, not log-level.
 
## `game_key` is hardening, not a hole
 
`game_key` is a block-local in the `@main_thread` closure. Locals in a required
file are not present in `TOPLEVEL_BINDING.local_variables`, so a script binding
cannot reach it. The two `game_key = nil` additions shorten how long a live key
sits in memory; they do not close an access path. Worth doing precisely because
the key does not expire at login, but do not review this as a hole closure.
Placement:
 
- once after the `client_thread` branch chain closes, rather than in each branch.
  Same thread as the mudlet send at `main.rb:631`, so there is no race on the
  shared local.
- once on the `$offline_mode` early return, which `next nil` would otherwise skip.
## Reconnect regression trap
 
`reconnect_if_wanted` does `args.concat ARGV` then `exec` (`main.rb:36`).
Redacting `--password=` in place would have silently broken `--reconnect --login`,
because the re-exec'd process would authenticate with `[scrubbed]`.
 
Resolved by snapshotting `ARGV.dup` into a `main.rb` top-level local before the
scrub and pointing the proc's four ARGV reads at the snapshot. Locals in a
required file are not script-reachable, so the snapshot does not reopen the hole
the scrub closes.
 
The scrub preserves flag prefixes and overwrites only values, so
`ARGV.include?('--without-frontend')` at `main.rb:539` and the other flag checks
behave identically.
 
## `.sal` removal
 
`CredentialScrub.shred_file` overwrites with nulls, `fsync`s, then deletes with a
bounded retry, and never raises. Three improvements over the previous code:
 
- the timeout-path `File.delete` had no `rescue` and could raise
- Windows refuses deletion while the spawned frontend still holds the handle,
  hence the retry
- an `at_exit` backstop registers immediately after the file is written, so an
  exception between write and the normal removal cannot leave the key in
  `TEMP_DIR`
`at_exit` does not run through `exec`, but the normal removal already precedes
`reconnect_if_wanted.call` on both paths.
 
Overwrite-then-delete is best effort. Copy-on-write filesystems, journaling, and
flash translation layers may retain the original blocks. Documented in the module.
This matters more than it would for a spent key, since a `.sal` file recovered
from disk after the fact may still contain a usable credential.
 
## Design decisions worth a reviewer's opinion
 
- **`--account=` is left alone.** It is an account identifier, not a credential.
  One entry in `ARGV_SECRET` if you disagree.
- **New file rather than inline in `main.rb`.** `main.rb` has no unit seam, and
  specs are mandatory; this keeps `main.rb` to five one-line call sites. It
  follows the existing small-focused-module pattern (`bind_host_resolver.rb`,
  `shutdown_intent.rb`).
- **In-place `String#replace` with a `FrozenError` fallback**, so the login GUI's
  reference to the same launch data array loses the secret too, rather than being
  left pointing at an unscrubbed copy.
## Not fixed
 
The key still reaches the frontend's argv via `launcher_cmd` (`main.rb:322` into
`spawn`), so it is visible to `ps` and Task Manager for as long as the frontend
runs. That is how the Simutronics frontends accept a key and is not fixable on our
side. Given the key stays valid after login, this is the largest remaining
exposure, and it is worth tracking as a separate issue rather than being quietly
accepted.
 
## Files
 
- `lib/common/credential_scrub.rb` (new, 162 lines)
- `spec/lib/common/credential_scrub_spec.rb` (new, 25 examples)
- `lib/main/main.rb` (+43 / -9)
## Verification
 
- `ruby -c` clean on all three files
- non-ASCII scan: zero lines across all three files
- `rspec`: 48 examples, 0 failures across `credential_scrub_spec`,
  `launch_data_spec`, and `limitedarray_spec`
Two caveats from my environment. RuboCop could not run locally, so the one thing
to watch in CI is `Lint/UselessAssignment` on the new `game_key = nil` lines; the
identical pattern already exists at `main.rb:632` and passes today.
`session_launcher_spec.rb` fails to load in my sandbox through the
`account_manager_ui.rb` GTK require chain, and I confirmed that failure is
identical on the unmodified tree.
 
## Manual checks requested
 
Startup paths I could not exercise without a live login:
 
- [ ] GUI login, Wizard or Wrayth frontend
- [ ] `--login` with `--password=` on the command line, then confirm
      `ARGV.join(' ')` in a script shows `--password=[scrubbed]`
- [ ] `--reconnect --login`, confirm reconnect still authenticates
- [ ] `--without-frontend`
- [ ] `--sal` launch from SGE
- [ ] mudlet, confirm the login key still sends

## Summary by CodeRabbit

* **Security**
  * Added credential redaction during startup to prevent passwords and master passwords from remaining in launch data, command-line arguments, or options.
  * Temporary launch files are securely overwritten and removed after connection attempts, including when connections time out.
  * Login credentials are cleared after transmission or when offline mode is used.

### DIFF
--- a/lib/common/credential_scrub.rb
+++ b/lib/common/credential_scrub.rb
@@ -7,8 +7,10 @@ module Lich
     # Lich keeps startup credentials where a running script can read them:
     # +@launch_data+ and +@argv_options+ are instance variables on the top-level
     # +main+ object, and +ARGV+ is a global constant. All three outlive login, so
-    # the single-use eaccess key and the reusable account password stay readable
-    # for the rest of the session.
+    # both the eaccess key and the account password stay readable for the rest of
+    # the session. Both are live credentials: the key is not single-use and stays
+    # valid for some period after issue, so a key read hours into a session may
+    # still be usable.
     #
     # Values are overwritten in place rather than reassigned because other
     # objects hold references to the same Array, Hash, and String instances (the
@@ -24,7 +26,7 @@ module Lich
       # Replacement text for a redacted value.
       REDACTED = '[scrubbed]'
 
-      # launch_data entry holding the single-use eaccess key.
+      # launch_data entry holding the eaccess key.
       LAUNCH_DATA_SECRET = /\AKEY=/i
 
       # argv entries of the form --flag=secret. Deliberately excludes
@@ -81,14 +83,9 @@ module Lich
 
           OPTION_SECRET_KEYS.select do |key|
             value = options[key]
-            next false unless value.is_a?(String) && value != REDACTED
+            next false unless value.is_a?(String)
 
-            begin
-              options[key] = REDACTED
-              true
-            rescue FrozenError
-              false
-            end
+            overwrite(options, key, REDACTED)
           end
         end
 
@@ -123,8 +120,9 @@ module Lich
 
         private
 
-        # Rewrites container[index], mutating the existing String when possible so
-        # that other holders of that String also lose the secret.
+        # Rewrites container[key], mutating the existing String when possible so
+        # that other holders of that String also lose the secret. Works for both
+        # Array indices and Hash keys.
         #
         # @return [Boolean] true when the value changed
         def overwrite(container, index, value)
--- a/lib/common/credential_scrub.rb
+++ b/lib/common/credential_scrub.rb
@@ -1,0 +1,160 @@
+# frozen_string_literal: true
+
+module Lich
+  module Common
+    # Redacts startup credentials once the login sequence no longer needs them.
+    #
+    # Lich keeps startup credentials where a running script can read them:
+    # +@launch_data+ and +@argv_options+ are instance variables on the top-level
+    # +main+ object, and +ARGV+ is a global constant. All three outlive login, so
+    # the single-use eaccess key and the reusable account password stay readable
+    # for the rest of the session.
+    #
+    # Values are overwritten in place rather than reassigned because other
+    # objects hold references to the same Array, Hash, and String instances (the
+    # login GUI keeps the launch data array; the argv option pipeline threads one
+    # options hash through three stages). Replacing a container would leave those
+    # references pointing at unscrubbed copies.
+    #
+    # Scope: this closes script-readable and log-readable paths. It does not
+    # guarantee the bytes are gone from process memory. Earlier copies may
+    # survive in socket write buffers, in a spawned frontend's argv, or in
+    # unreclaimed heap.
+    module CredentialScrub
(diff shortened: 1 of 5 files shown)


mrhoribu:
@coderabbitai can you look over the latest commit and verify if that resolves and meets the findings you proposed (including the nitpicks).

OSXLich-Doug (approved):
Very needful, thank you.  Reviewed, no blockers.  Two small nits (polish):

- The comment saying CredentialScrub “only ships” a mutating helper is now stale because the same line calls the new non-mutating helper.
- Tests cover redact_argv thoroughly but still do not behaviorally exercise the actual reconnect log-to-exec wiring.

Approved for merge.

PR #1500, opened 2026-08-06
fix(dr): preserve game RealID in room title (GS parity for ";display roomid title")

MahtraDR:
Follow-up to #1491.

## Problem

In DragonRealms, `;display roomid title` + `;display lichid` (with the game's `ShowRoomID` flag on) **stripped** the game-supplied RealID before injecting the lich id, so the room name read `[Bosque Deriel, Hermit's Shacks - 2071]` and the RealID `(230008)` was lost. To see it, a player had to also enable `;display uid`, which renders a different, u-prefixed form *inside* the brackets (`[... - 2071 - (u230008)]`).

GemStone already **preserves** the game RealID in this exact case:

| game | `;display lichid` + roomid title, uid off | result |
|---|---|---|
| **GS** (today) | preserves game RealID | `[Room - 1234] (230008)` ✅ |
| **DR** (before) | strips game RealID | `[Room - 2071]` ❌ |

So DR diverged from GS. As a secondary symptom, front-ends that re-wrap a bracketed title that lacks a trailing `(number)` (e.g. ProfanityFE's room window) turned the leftover `]` into a doubled `]]`.

## Fix

Bring DR into parity with GemStone:

- `Lich::DragonRealms::GameInstance#modify_room_display` now strips the game RealID **only** when Lich renders its own UID (`;display uid`) or the player hid it (`;display flaguid` / `hide_uid_flag`). Otherwise it preserves the game RealID and injects the lich id before the bracket → `[Bosque Deriel, Hermit's Shacks - 2071] (230008)`.
- The room-window subtitle (window title bar) is reconstructed to read **identically** to the room-name line. Because the subtitle is rebuilt from the always-UID-free `XMLData.room_title`, a new `XMLData#show_room_id` flag records whether the game displayed the RealID this arrival (ShowRoomID on), and `#preserved_realid_suffix` synthesizes the marker from the authoritative `<nav>` UID.
- Two shared private helpers (`#suppress_game_realid?`, `#preserved_realid_suffix`) keep the room-name line and the subtitle from drifting.
- **GemStone code is untouched.**

## Tests

DAMP + edge coverage, full suite green.

- `games_spec`: preserve-vs-strip matrix (lichid only preserves; `display_uid`/`hide_uid_flag` strip), `(**)` no-uid marker preserved, room-window subtitle parity including a byte-identical room-name/subtitle assertion.
- `xmlparser_spec`: `show_room_id` set true for numeric/`(**)` markers, false when ShowRoomID is off, and cleared when a later arrival turns the flag off.
- `spec_helper`: `show_room_id` added to the `XMLData` mock.

Result: **5531 examples, 0 failures**; `rubocop` clean on all touched files.

### DIFF
--- a/lib/common/xmlparser.rb
+++ b/lib/common/xmlparser.rb
@@ -17,7 +17,7 @@ module Lich
                   :roundtime_end, :cast_roundtime_end, :last_pulse, :level, :next_level_value,
                   :next_level_text, :society_task, :stow_container_id, :name, :game, :in_stream,
                   :player_id, :prompt, :current_target_ids, :current_target_id, :room_window_disabled,
-                  :dialogs, :room_id, :previous_nav_rm, :concentration, :max_concentration,
+                  :dialogs, :room_id, :previous_nav_rm, :show_room_id, :concentration, :max_concentration,
                   :arrival_pcs, :room_player_hidden, :field_exp, :max_field_exp,
                   :ascension_exp, :exp, :until_next, :fashlonae, :lumnis, :rpa,
                   :room_climate, :room_terrain, :room_weather, :room_bonfire,
@@ -149,6 +149,14 @@ module Lich
         # ShowRoomID-on subtitle from overwriting an authoritative nav UID in the
         # same arrival (see the streamWindow handler).
         @nav_uid_pending = false
+        # True when the game itself embedded the RealID marker in the room title
+        # this arrival (i.e. the DragonRealms ShowRoomID flag is ON), e.g. the
+        # subtitle " - [Room] (230008)" or the no-UID " - [Room] (**)". Distinct
+        # from a UID merely known via <nav>: it records whether the *game* chose
+        # to display it, so Lich's room-name/subtitle rewrite can preserve that
+        # choice instead of second-guessing the flag (see the DR streamWindow
+        # branch and Lich::DragonRealms::GameInstance#modify_room_display).
+        @show_room_id = false
 
         # Lich::Claim update
         @arrival_pcs = []
@@ -617,6 +625,11 @@ module Lich
                     @room_title = "[#{room[:roomtitle]}]"
                     @room_id = (room[:uid] =~ /\A\d+\z/ ? room[:uid].to_i : 0) if room[:uid] && !@nav_uid_pending
                     @nav_uid_pending = false
+                    # Record whether the game displayed the RealID this arrival (a UID
+                    # marker present == ShowRoomID ON) so the outbound room-name/subtitle
+                    # rewrite can preserve that choice GS-style. A marker-less subtitle
+                    # (flag OFF) leaves the previous room's flag stale, so clear it here.
+                    @show_room_id = !room[:uid].nil?
                   end
                 else
                   @room_title = String.new
--- a/lib/games.rb
+++ b/lib/games.rb
@@ -1458,21 +1458,26 @@ module Lich
       # Rewrites the room-name line (the "<style id=\"roomName\" />[Title]" chunk) on its way
       # to the client. Two independent concerns:
       #   1. When the player wants the room id/UID in the title (";display roomid title|both"),
-      #      strip any RealID the game appended (present only under the ShowRoomID flag) and
-      #      inject Lich's configured id/UID, so the name reads e.g.
-      #      "[Town Square - 1234 - (u230008)]".
+      #      inject Lich's configured id/UID before the closing bracket. The game-supplied
+      #      RealID (present only under the ShowRoomID flag) is preserved GemStone-style so
+      #      ";display lichid" alone reads "[Town Square - 1234] (230008)" - matching how
+      #      Lich::Gemstone::GameInstance#modify_room_display already behaves. It is stripped
+      #      only when Lich is rendering its own UID instead ({#suppress_game_realid?}), so a
+      #      ";display uid" title reads "[Town Square - 1234 - (u230008)]" with no duplicate.
       #   2. Otherwise preserve the historical behavior of hiding the game's inline RealID when
       #      the player asked to (";display uid" or ";display flaguid").
       # Edits only this outbound client string; XMLData (and therefore scripts) are untouched,
       # because process_xml_data has already parsed the inbound stream before this runs.
       # @param alt_string [String] the outbound room-name server string, modified in place
       # @return [String] the rewritten string
+      # @see #suppress_game_realid?
+      # @see Lich::Gemstone::GameInstance#modify_room_display
       def modify_room_display(alt_string)
         if room_id_in_title?
-          alt_string.sub!(/] \((?:\d+|\*\*)\)/) { "]" }
+          alt_string.sub!(/] \((?:\d+|\*\*)\)/) { "]" } if suppress_game_realid?
           room_number = room_number_display
           alt_string.sub!(']') { " - #{room_number}]" } unless room_number.empty?
-        elsif Lich.display_uid == true || Lich.hide_uid_flag == true
+        elsif suppress_game_realid?
           alt_string.sub!(/] \((?:\d+|\*\*)\)/) { "]" }
         end
 
@@ -1483,8 +1488,10 @@ module Lich
       # room id/UID. The "Room Number:" line below the room appears for the "line" and "both"
       # placements (honoring the mono toggle). The room-window subtitle (room-window front-ends
       # only) reflects the id/UID for every placement, since the window title bar is itself a
-      # title surface. The inline room-name injection for the "title"/"both" placements is
-      # handled separately in #modify_room_display.
+      # title surface, and re-appends the preserved game RealID ({#preserved_realid_suffix}) so
+      # the window title bar reads identically to the room-name line rewritten by
+      # #modify_room_display. The inline room-name injection for the "title"/"both" placements
+      # is handled separately in #modify_room_display.
       # @param alt_string [String] the server string being rewritten
       # @return [String] the rewritten server string
       def process_room_display(alt_string)
@@ -1496,8 +1503,9 @@ module Lich
             alt_string = "#{room_styled("Room Number: #{room_number}")}\r\n#{alt_string}"
           end
           if Frontend.supports_room_window?
-            alt_string = "<streamWindow id='main' title='Story' subtitle=\" - [#{XMLData.room_title[2..-3]} - #{room_number}]\" location='center' target='drop'/>\r\n#{alt_string}"
-            alt_string = "<streamWindow id='room' title='Room' subtitle=\" - [#{XMLData.room_title[2..-3]} - #{room_number}]\" location='center' target='drop' ifClosed='' resident='true'/>#{alt_string}"
+            subtitle = " - [#{XMLData.room_title[2..-3]} - #{room_number}]#{preserved_realid_suffix}"
+            alt_string = "<streamWindow id='main' title='Story' subtitle=\"#{subtitle}\" location='center' target='drop'/>\r\n#{alt_string}"
+            alt_string = "<streamWindow id='room' title='Room' subtitle=\"#{subtitle}\" location='center' target='drop' ifClosed='' resident='true'/>#{alt_string}"
           end
         end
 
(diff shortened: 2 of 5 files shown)


mrhoribu (approved):
# PR Review: fix(dr): preserve game RealID in room title (elanthia-online/lich-5#1500)
 
> **Draft.** This is written to be adapted, not pasted. Tone, ordering, and which
> items are worth raising are your call -- you know the author and what this repo
> spends review capital on. Whether to disclose how the review was produced is
> also your call; it is not obscured here either way.
 
**Reviewed:** 5 files, +169/-12 at `b715f7b` (merge base `84ce0fb`).
Based on: full clone with the PR branch checked out; Ruby 3.2.3 + RSpec 3.13; full
suite run on both head and merge base.
 
**Verdict:** Approve with comments.
 
**Summary:** Brings DragonRealms `modify_room_display` into line with GemStone by
preserving the game-supplied RealID instead of stripping it unconditionally, and adds
an `XMLData#show_room_id` flag so the reconstructed room-window subtitle can
re-synthesize that marker. Cleanly factored into two shared private helpers. The one
substantive item is that the subtitle change also lands on the default placement,
where the parity rationale that motivates it does not apply.
 
---
 
## Mechanical verification
 
| Check | Result |
|---|---|
| `ruby -c` on all 5 changed files | Syntax OK |
| Non-ASCII byte scan on all 5 changed files | 0 non-ASCII lines |
| Full suite at merge base `84ce0fb` | 5518 examples, 0 failures |
| Full suite at PR head `b715f7b` | 5531 examples, 0 failures |
| Net example delta | +13, reconciles exactly with the diff |
| PR specs against merge-base `lib/` | 9 failures -- the new specs are real regression tests |
 
The 9 failures break down as 5 `xmlparser_spec` examples failing with
`NoMethodError: undefined method 'show_room_id'`, plus 4 `games_spec` examples
asserting the preserve-instead-of-strip behavior. Four other new examples pass on
both sides; those are deliberate guard rails (strip-when-uid-on, hide_uid_flag,
subtitle-off, no-doubling) rather than regression tests, which is fine.
 
RuboCop could not be run -- see Coverage notes.
 
---
 
## Minor
 
### 1. The subtitle change also alters the `line` and default placements, where the parity rationale does not hold -- `lib/games.rb:1500-1510`
 
```ruby
room_number = room_number_display
unless room_number.empty?
  if room_id_in_line?
    alt_string = "#{room_styled("Room Number: #{room_number}")}\r\n#{alt_string}"
  end
  if Frontend.supports_room_window?
    subtitle = " - [#{XMLData.room_title[2..-3]} - #{room_number}]#{preserved_realid_suffix}"
```
 
The subtitle block is gated only on `room_number_display` being non-empty and on
`Frontend.supports_room_window?`, not on placement -- deliberately, per the doc
comment ("reflects the id/UID for every placement"). Folding
`preserved_realid_suffix` into that same string therefore changes the window title
bar for every placement value, including `line` and `nil`.
 
Measured on both sides with DR, `;display lichid` on, the game's ShowRoomID flag on,
and Wrayth as the front-end, varying only `Lich.display_roomid_location`:
 
| placement | subtitle at `84ce0fb` | subtitle at `b715f7b` | room-name line at `b715f7b` |
|---|---|---|---|
| `nil` (default) | `" - [Test Room - 1234]"` | `" - [Test Room - 1234] (230008)"` | `[Test Room] (230008)` |
| `line` | `" - [Test Room - 1234]"` | `" - [Test Room - 1234] (230008)"` | `[Test Room] (230008)` |
| `title` | `" - [Test Room - 1234]"` | `" - [Test Room - 1234] (230008)"` | `[Test Room - 1234] (230008)` |
| `both` | `" - [Test Room - 1234]"` | `" - [Test Room - 1234] (230008)"` | `[Test Room - 1234] (230008)` |
 
The rendered result is defensible on its own terms: ShowRoomID on means the player
asked to see the RealID. But the justification given in both the code comment and the
commit message is parity with "the room-name line rewritten by
`#modify_room_display`", and under `line`/`nil` that method rewrites nothing at all --
`room_id_in_title?` is false and `suppress_game_realid?` is false, so both branches
fall through. The existing spec at `spec/lib/games_spec.rb:445` pins exactly that:
 
```ruby
it 'does not inject for the "line" placement (room name left as sent)' do
  Lich.display_lichid = true
  Lich.display_roomid_location = 'line'
  expect(dr.modify_room_display(+'[Test Room] (230008)')).to eq('[Test Room] (230008)')
end
```
 
So under the default placement the two surfaces do not match (last two columns above
differ), and the change lands there for a reason that does not apply.
 
**Coverage gap:** all five examples in the new
`#process_room_display (room-window subtitle parity)` block
(`spec/lib/games_spec.rb:504-552`) set `Lich.display_roomid_location = 'title'`.
Nothing exercises `line` or `nil` with `show_room_id` true, which is the combination
most DR players will actually be in given `line` is the documented DR default from
\#1491.
 
**Suggested fix:** either gate `preserved_realid_suffix` on `room_id_in_title?`, or
keep the behavior and add one example asserting it for the default placement plus a
line in the PR body noting that the default path changes. Leaning toward the second,
but as a stated decision rather than a side effect.
 
Confidence: **confirmed** (measured on both branches).
 
---
 
## Nits (non-blocking)
 
- `lib/games.rb:202` declares the abstract
  `modify_room_display(alt_string, uid_from_string, lichid_from_uid_string)` with
  three parameters, while both concrete implementations take one. Pre-existing, not
  caused by this PR, but this is the natural moment to drop the two dead parameters.
- `preserved_realid_suffix` (`lib/games.rb:1573`) tests `XMLData.room_id.zero?` while
  its sibling `room_number_display` (`lib/games.rb:1530`) tests
  `XMLData.room_id == 0` for the identical condition. Harmless -- `@room_id` has been
  a guaranteed Integer since \#1491 -- but the two helpers sit twelve lines apart and
  are meant to read as a matched pair.
---
 
## Open questions
 
- **`modify_room_display` preserves the game's literal RealID digits, while
  `preserved_realid_suffix` re-synthesizes them from `XMLData.room_id` (the nav UID).**
  The \#1491 comment at `lib/common/xmlparser.rb:608-622` establishes that nav is
  authoritative precisely because the subtitle marker can be late, absent, or stale,
  so the sources are known to be capable of disagreeing. Forcing
  `XMLData.room_id = 999999` against a roomName line carrying `(230008)` produces:
```
  roomName line : "[Test Room - 1234] (230008)"      <- literal, preserved
  window title  : " - [Test Room - 1234] (999999)"   <- synthesized from nav
```
 
  which is exactly the drift the shared helpers were introduced to prevent.
  **Settles it:** can the RealID on the `<style id="roomName"/>` line ever disagree
  with `<nav rm=...>` for the same arrival? If it cannot, this closes cleanly.
 
---
 
## Considered and dismissed
 
Recorded so these are not re-raised later.
 
- **Genie/Frostbite subtitle strippers.** `lib/games.rb:1077-1083` strips
  `] (\d+|\*\*)` from server-sent `streamWindow` subtitles for those two front-ends,
  and runs *before* `process_room_display` at line 1087, so a newly synthesized
  subtitle would escape it. Does not apply: `lib/common/front-end.rb:108-112`
  registers `genie` as `%i[xml mono]` and `frostbite` as `%i[xml]`, neither carrying
  `room_window`, so `Frontend.supports_room_window?` is false and no subtitle is
  emitted for them. The same check clears ProfanityFE (`%i[xml streams]`).
- **`@show_room_id` staleness.** The assignment at `lib/common/xmlparser.rb:632` sits
  inside `if room`, so an identity-less subtitle (the `" - "` case covered by
  `xmlparser_spec.rb:559`) leaves a stale `true`. Confirmed mechanically, but
  `room_id` and `room_title` are equally stale in that same window, so the subtitle
  would already be rendering the wrong room's name for a pre-existing reason. A stale
  flag adds nothing to a string that is already wrong. Not this PR's problem.
- **`@show_room_id` missing from `#reset`.** `XMLParser#reset`
  (`lib/common/xmlparser.rb:206-228`) is parser-state only and deliberately does not
  touch room fields, so initializing the flag in `initialize` alongside `@room_id` and
  `@nav_uid_pending` is consistent, not an oversight.
- **`attr_reader` vs `attr_accessor` split.** Production exposes `show_room_id`
  read-only (`lib/common/xmlparser.rb:10-24`) with only the parser writing it, while
  the spec mock in `spec/spec_helper.rb` makes it writable. That is the same split the
  suite already uses for `room_id` and `room_title`.
---
 
## What this PR gets right
 
The GS-parity claim holds up against the code rather than just the commit message:
`Lich::Gemstone::GameInstance#modify_room_display` (`lib/games.rb:1335-1337`) injects
the lich id before `]` and leaves the trailing game RealID alone, so DR really was
the divergent one.
 
Routing both surfaces through `suppress_game_realid?` is the right shape. The old
code spelled the same three-way condition out twice with different bracket handling,
which is how these two drifted in the first place -- worth protecting during any
revision.
 
Gating the synthesized marker on a parser-tracked flag rather than inferring it from
`room_id != 0` is the correct call, since `room_id` is now populated from `<nav>`
whether or not the game chose to display it. Inferring would have produced a marker
under ShowRoomID-off.
 
The byte-identical parity assertion at `spec/lib/games_spec.rb:544` is a good
invariant test. It happens to pass on the old code too, so it is not a regression
guard for this fix, but it will catch the next person who changes one surface and
forgets the other.
 
---
 
## Coverage notes
 
**Access:** full clone with the PR branch checked out, so call sites and history were
greppable. `gh` was unavailable and the plain-HTTP `.diff` path was blocked by a
redirect to a non-allowed host, so the diff came from
`git fetch origin pull/1500/head`.
 
**PR metadata** came from the rendered HTML page, not the API, which stayed
rate-limited (0/60) for the entire session: PR is open, not a draft, single commit,
and CodeRabbit auto-review is disabled repo-wide, so there was no prior automated
review to duplicate. **CI check results and any human review comments were not
seen.**
 
**RuboCop never ran.** It is not packaged in this environment's apt repositories and
rubygems.org is blocked, so the repo's own lint gate -- including the custom
`Custom/AsciiOnlySource` cop and the `TargetRubyVersion` override -- is unverified. A
manual non-ASCII byte scan was substituted. Ruby here is 3.2.3, not the repo's
declared target.
 
**Suite caveat:** the full suite required stubbing the unavailable `os` gem so
`spec/lib/common/authentication/*` could load. That stub touches no code path in this
PR. `concurrent`, `ffi`, `kramdown`, and `terminal-table` came from apt.
 
**Not verified:** the ProfanityFE doubled-`]]` symptom cited as motivation in the
commit message (needs the ProfanityFE repo, and ProfanityFE has no `room_window`
capability, so the subtitle half of this fix does not reach it); whether the game can
emit disagreeing nav and roomName RealID values (see Open questions); and all
real-front-end rendering behavior.

MahtraDR:
Thanks for the thorough review, @mrhoribu — measuring both branches across every placement was exactly the right lens. Addressed in f9650735:

**Minor #1 (subtitle on the `line`/`nil` placements) — kept the behavior, fixed the rationale.** You're right that the "parity with the room-name rewrite" justification doesn't hold where `modify_room_display` rewrites nothing. The behavior itself is intended, though: `;display roomid` governs only where *Lich's* id goes, not whether the *game's* RealID is shown. When ShowRoomID is on the game displays the RealID in the room title regardless of placement — and it already rides the room-name line for `line`/`nil` (passthrough), so putting it in the reconstructed window title makes the two surfaces consistent rather than divergent. I rewrote the `process_room_display` doc comment to say that, and closed the coverage gap with examples for `line` and `nil` with ShowRoomID on.

**Open question (roomName-line RealID vs `<nav>`) — nav is authoritative, and that's the GS standard too.** For a complete arrival they can't disagree: the parser sets `room_id` from `<nav rm=...>` and the DR subtitle marker defers to it (`... if room[:uid] && !@nav_uid_pending`). They diverge only in the late/absent/stale-subtitle case #1491 introduced nav-authority for, and there the nav-derived value is the correct one. This mirrors GemStone — `Lich::Gemstone::GameInstance#modify_room_display` derives its `;display uid` marker from `XMLData.room_id` (nav), not the inbound digits. So the subtitle synthesizing from nav is the *more* authoritative surface, not a drift risk. Documented this, and added a regression test that forces `room_id` to disagree with any room-name literal and asserts the window title follows nav.

**Nits:**
- Dropped the two dead params from the abstract `GameBase::GameInstance#modify_room_display` signature (both concrete impls take one) and updated its arity spec.
- Aligned `room_number_display` on `.zero?` to match its sibling `synthesized_realid_suffix`.
- Renamed `#preserved_realid_suffix` → `#synthesized_realid_suffix` since it synthesizes from nav rather than preserving inbound digits.

Full suite green (5534 examples, 0 failures, +3 new); rubocop clean on the touched files.

mrhoribu (approved):

PR #1501, opened 2026-08-07
feat(all): teardown and distinguishable errors for Script.run_child

calael:
Follow-up to the supervised child script API (#1474–#1479).

`Script.run_child` previously overloaded `nil` across two outcomes with opposite liveness — "never started" (nothing running) and "timed out" (a live child the caller now unknowingly owns) — and a returned script still required knowing to inspect exit state. A field incident (a helper script that outlived its supervisor's patience and kept manipulating the character's hands) motivated tightening the contract.

New contract:

- **Timeout tears the child down**: on `join` timeout the child is `kill_sync`ed before `Script::TimeoutError` (carrying the script via `#script`) is raised — no unsupervised child ever survives the call.
- **Startup refusal raises** `Script::StartError` instead of returning `nil`.
- **A returned script has always terminated**; callers distinguish success from crash via `completed_successfully?` / `exit_error`.
- Negative timeouts still raise `ArgumentError` before the child starts.

The error classes are deliberately generic (`StartError` / `TimeoutError`, not child-specific) so future supervision APIs can reuse them.

There are no in-repo callers besides the definition; external callers that invoke `run_child` without a timeout only see startup refusal change from a (typically ignored) `nil` to a raise.

https://claude.ai/code/session_01QMEQExmjRiUdtWvbVPH9XD

### DIFF
--- a/lib/common/script.rb
+++ b/lib/common/script.rb
@@ -348,6 +348,21 @@ module Lich
 
       class JumpError < StandardError; end
       class LibraryJoinTimeout < LoadError; end
+
+      # Raised when a script cannot be started (shutdown in progress,
+      # duplicate name, or launch failure).
+      class StartError < StandardError; end
+
+      # Raised when a supervised script outlives its timeout. The script has
+      # already been torn down (kill_sync) by the time this raises.
+      class TimeoutError < StandardError
+        attr_reader :script
+
+        def initialize(script)
+          @script = script
+          super("script timed out: #{script.name}")
+        end
+      end
       JUMP = JumpError.exception('JUMP')
       JUMP_ERROR = JumpError.exception('JUMP_ERROR')
 
@@ -966,16 +981,29 @@ module Lich
 
       # Starts a child script and waits for it to finish.
       #
+      # The returned script has always terminated: inspect
+      # {#completed_successfully?} / {#exit_error} to distinguish success from
+      # a crash. Startup refusal and timeout raise instead of overloading the
+      # return value, and a timed-out child is torn down before the raise, so
+      # no unsupervised child ever survives this call.
+      #
       # @param timeout [Numeric, nil] maximum seconds to wait, or nil to wait indefinitely
-      # @return [Script, nil] the completed child, or nil on startup failure or timeout
+      # @return [Script] the terminated child
       # @raise [ArgumentError] when the timeout is negative, before the child starts
+      # @raise [StartError] when the child cannot be started
+      # @raise [TimeoutError] when the child outlives the timeout, after its teardown
       def Script.run_child(*args, timeout: nil)
         raise ArgumentError, 'timeout must be non-negative' if timeout && timeout.negative?
 
         child = Script.start_child(*args)
-        return nil unless child
+        unless child
+          name = args.first.is_a?(Hash) ? args.first[:name] : args.first
+          raise StartError, "script failed to start: #{name}"
+        end
+        return child if child.join(timeout)
 
-        child.join(timeout) ? child : nil
+        child.kill_sync
+        raise TimeoutError.new(child)
       end
 
       # Marks the current script as hidden and protected from kill-all and
--- a/spec/lib/common/script_lifecycle_spec.rb
+++ b/spec/lib/common/script_lifecycle_spec.rb
@@ -565,12 +565,25 @@ RSpec.describe 'Lich::Common::Script lifecycle extensions' do
       expect(script_class.run_child('child')).to equal(child)
     end
 
-    it 'applies an explicit Script.run_child timeout' do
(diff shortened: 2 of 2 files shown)


mrhoribu (approved):
# PR Review: feat(all): teardown and distinguishable errors for Script.run_child (elanthia-online/lich-5#1501)

**Reviewed:** 2 files, +51/-8 at `a77fdce` on 2026-08-09. Based on: full clone (branch checked out, history available, repo-wide grep performed).
**Verdict:** Approve
**Summary:** Changes `Script.run_child` from returning `nil` on either "never started" or "timed out" to raising distinct `StartError` / `TimeoutError`, and guarantees a timed-out child is torn down (`kill_sync`) before the raise. The change is small, self-contained, has no other in-repo callers, and its tests and CI (RSpec + RuboCop) match the new behavior.

## Blockers

None.

## Major

None.

## Minor

None.

## Nits

### 1. `StartError` message can end with a blank name — `lib/common/script.rb:999-1002`

```ruby
unless child
  name = args.first.is_a?(Hash) ? args.first[:name] : args.first
  raise StartError, "script failed to start: #{name}"
end
```

If `run_child` is called with no positional args (e.g. `run_child(timeout: 5)`), `args.first` is `nil`, so the message reads `"script failed to start: "` with nothing after the colon. This mirrors the same call shape that `Script.start`'s `@@elevated_script_start` proc already treats as an error (`args.empty?`), so it's not a new failure mode — just a slightly unhelpful message in an edge case that's already a misuse of the API. Not worth blocking on.

## Open questions

None — the one case worth checking (whether anything outside this repo calls `run_child` in a way that depends on the old `nil` return) is addressed directly in the PR description: no in-repo callers exist besides the definition and its specs (confirmed via repo-wide `grep -rn "run_child"`), and external callers that don't pass a timeout only see the startup-refusal path change from an easily-ignored `nil` to a raise.

## What this PR gets right

- **Closes a real liveness hole.** Previously, a `run_child` timeout returned `nil` while the child was still running and now unsupervised — exactly the "field incident" the PR describes. The new code calls `child.kill_sync` (which, per `lib/common/script.rb:1928-1933`, blocks on `join` until teardown fully completes) before raising `TimeoutError`, so the "no unsupervised child ever survives this call" claim in the doc comment (`lib/common/script.rb:984-988`) is actually true by construction, not just documented.
- **`StartError` vs `TimeoutError` are genuinely distinguishable now**, replacing an overloaded `nil` that conflated "nothing running" with "a live child you now unknowingly own" — the exact ambiguity called out in the PR description.
- **Tests were updated in lockstep with the behavior change**, not just patched to stop failing: `spec/lib/common/script_lifecycle_spec.rb:568-579` asserts both the raise *and* that `kill_sync` was invoked, and `:1089-1096` does the same via `instance_double` with `have_received(:kill_sync)`. The `instance_double(script_class, ..., :name => 'worker')` addition at `:1090` is necessary and correct — `TimeoutError#initialize` calls `script.name` to build its message, and `name` is a real public `attr_reader` (`lib/common/script.rb:342`), so the doubles remain verified.
- **CI is green on both fronts that matter for a change like this**: `RSpec Tests` and `Rubocop` both succeeded on `a77fdce`, confirming syntax, lint, and the updated spec expectations all pass.
- **No blast radius.** Repo-wide grep confirms `run_child` has zero in-repo callers outside its own definition and specs, so this is a pure library-surface change with no call sites to break.

## Coverage notes

- Full local clone of `elanthia-online/lich-5`, PR branch (`pull/1501/head`) checked out at `a77fdce`, diffed against merge-base `84ce0fb`. Repo history available for `git blame`/`git log -S`.
- Traced `run_child` → `start_child` → `@@elevated_script_start`, `join`, `kill`, and `kill_sync` in full to verify the name-extraction logic and the teardown-before-raise guarantee.
- Confirmed via `git log -S "def Script.run_child"` that the original `nil`-returning version was added in #1474, and that this PR's stated motivation (a field incident with an orphaned helper script) is a genuine hardening rather than a change of established, deliberately-chosen behavior.
- No Ruby toolchain available in this environment (no `ruby`/`gem`/`bundle`, and rubygems.org isn't reachable from this sandbox), so tests and RuboCop were **not** re-run locally. Instead I read the actual CI job results for this commit via the PR's Checks tab: both `RSpec Tests` and `Rubocop` workflows report **succeeded** on `a77fdce`, which is the evidence this review relies on for syntax/lint/test correctness rather than local execution.
- `gh` CLI was unavailable and the GitHub REST API was rate-limited unauthenticated, so PR metadata (title, body, CI links) was pulled via `web_fetch` on the PR's HTML pages instead of the API.

OSXLich-Doug (approved):
Two non-blockers, approved to merge.

PR #1502, opened 2026-08-08
fix(all): bound EAccess auth with a timeout so a stalled SGE backend can't hang Lich silently

simtel12:
## Summary

While debugging a `--detachable-client` hang (process alive, no crash, no log output past character resolution, no listener ever opened), I traced it to `EAccess.auth`: the TCP connect, TLS handshake, and every K/A/M/F/G/P/C/L round-trip in the SGE protocol exchange are bare blocking calls with no connect or read timeout anywhere. In this instance the SGE backend itself was unreachable (confirmed via `lsof` showing the socket stuck in `SYN_SENT`, and a bare `/dev/tcp` connect from the same host independently timing out on port 7910 while other outbound traffic worked fine) — but regardless of cause, Lich has no way to surface that: the thread just blocks forever, silently.

This is the same failure mode `Game.open_with_timeout` already guards against for the *game* connection (`lib/games.rb`) — the SGE connection just never got the same treatment.

## Changes

- **`EAccess.auth_with_timeout`** (`lib/common/authentication/eaccess.rb`) — wraps `EAccess.auth` in the same `Thread.new { }.join(timeout)` pattern as `Game.open_with_timeout`, defaulting to 30s. Re-raises whatever `auth` raises; raises a clear "timed out authenticating with EAccess after Ns" on timeout. Killing the thread on timeout still runs `auth`'s `ensure`, so the stuck socket gets closed instead of leaked.
- **`authenticator.rb`** — its three `EAccess.auth(...)` call sites now call `EAccess.auth_with_timeout(...)` instead, so a stuck auth flows through the existing retry/backoff/logging path (`with_retry`, `MAX_AUTH_RETRIES`) instead of hanging silently.

## Testing

- New specs in `eaccess_spec.rb` for `.auth_with_timeout` (success passthrough, timeout raises, exception passthrough), written and confirmed failing before implementing.
- Existing `authenticator_spec.rb` specs (which stub `EAccess.auth`) needed no changes — the stub is visible inside the wrapper thread since it's a singleton-method redefinition.
- Full suite: `bundle exec rspec` — 5521 examples, 0 failures.
- `rubocop` on all touched files: no offenses.

## Test plan

- [x] Point `--login` at an account while the SGE backend is unreachable/unresponsive; confirm Lich now logs a timeout error and exits/retries instead of hanging with no listener ever opening.
- [x] Normal login still succeeds and behaves identically under normal network conditions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Authentication now enforces a timeout across supported authentication methods.
  * Expired authentication attempts fail with a timeout error instead of waiting indefinitely.
  * Authentication errors continue to be reported correctly.

### DIFF
--- a/lib/common/authentication/authenticator.rb
+++ b/lib/common/authentication/authenticator.rb
@@ -36,7 +36,7 @@ module Lich
       def self.authenticate(account:, password:, character: nil, game_code: nil, legacy: false, generator: false)
         with_retry do
           if game_code && (character || generator)
-            EAccess.auth(
+            EAccess.auth_with_timeout(
               account: account,
               password: password,
               character: character,
@@ -44,13 +44,13 @@ module Lich
               generator: generator
             )
           elsif legacy
-            EAccess.auth(
+            EAccess.auth_with_timeout(
               account: account,
               password: password,
               legacy: true
             )
           else
-            EAccess.auth(
+            EAccess.auth_with_timeout(
               account: account,
               password: password
             )
--- a/lib/common/authentication/eaccess.rb
+++ b/lib/common/authentication/eaccess.rb
@@ -232,6 +232,35 @@ module Lich
         def self.read(conn)
           conn.sysread(PACKET_SIZE)
         end
+
+        # Bounds how long the full SGE authentication exchange may block.
+        #
+        # {.auth} has no connect or read timeouts of its own -- every step (TCP
+        # connect, TLS handshake, and each K/A/M/F/G/P/C/L round-trip) is a bare
+        # blocking call. An unresponsive SGE backend (e.g. a stalled connect that
+        # never gets a SYN-ACK) hangs the caller indefinitely with no exception
+        # and no log output. This wraps the whole exchange the same way
+        # {Lich::GameBase::Game.open_with_timeout} bounds the game connect.
+        #
+        # @param timeout [Integer, Float] seconds to wait for the full exchange
+        # @param kwargs [Hash] forwarded to {.auth}
+        # @return [Hash, Array] see {.auth}
+        # @raise [RuntimeError] if the exchange does not complete within +timeout+
+        # @raise [StandardError] re-raises whatever {.auth} raises
+        # @see .auth
+        def self.auth_with_timeout(timeout: 30, **kwargs)
+          auth_thread = Thread.new {
+            # report_on_exception off: a failed auth is surfaced by the join below
+            # (which re-raises it), not by an auto-printed thread warning.
+            Thread.current.report_on_exception = false
+            auth(**kwargs)
+          }
+          if auth_thread.join(timeout).nil?
+            auth_thread.kill rescue nil
+            raise "error: timed out authenticating with EAccess after #{timeout}s"
+          end
+          auth_thread.value
+        end
       end
     end
   end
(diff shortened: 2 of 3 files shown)


mrhoribu (approved):
# PR Review: fix(all): bound EAccess auth with a timeout so a stalled SGE backend can't hang Lich silently (elanthia-online/lich-5#1502)

**Reviewed:** 3 files, +58/-3 at `ff83e0b` on 2026-08-19. Based on: full clone (branch checked out, history and callers greppable).
**Verdict:** Approve
**Summary:** Wraps `EAccess.auth` in a background thread with a join-timeout (`EAccess.auth_with_timeout`) so a stalled SGE (Simutronics game entry) backend can no longer hang the login flow indefinitely with no exception and no log line. All three call sites in `authenticator.rb` are switched over.

## What this PR gets right

This is a near-exact structural copy of an already-shipped, in-production pattern: `Lich::GameBase::Game.open_with_timeout` (`lib/games.rb:492`), down to the `report_on_exception = false` comment, the `join(timeout).nil?` check, and the `kill rescue nil` cleanup. The PR's own doc comment says as much and it checks out — same shape, same reasoning about why `join`'s three-way return (thread/nil/exception) is used instead of the old, ambiguous `.status` check. Reusing a reviewed pattern rather than inventing a new concurrency primitive is exactly right here.

`EAccess.auth`'s `ensure conn&.close unless conn&.closed?` (`lib/common/authentication/eaccess.rb`) means a killed thread that has an open socket will generally still close it as the `Thread#kill` unwinds the stack — MRI's `Thread#kill` raises through `ensure` blocks rather than hard-terminating, so this isn't a silent resource leak in the common case.

The plain `RuntimeError` raised on timeout (`raise "error: timed out authenticating..."`) is caught by `Authenticator.with_retry`'s generic `rescue StandardError` branch and retried with exponential backoff, which is the correct behavior for what's framed as a transient backend stall — verified by reading `with_retry` end-to-end rather than assuming.

Existing test compatibility was checked, not assumed: `spec/lib/common/authentication/authenticator_spec.rb` (unmodified by this PR) stubs `EAccess.auth` directly with `expect(...).to receive(:auth).with(...)`. Since `auth_with_timeout` still calls `self.auth(**kwargs)` inside its thread and RSpec method stubs aren't thread-scoped, those pre-existing tests continue to exercise the same contract and should still pass unmodified.

## Open questions

- **Native blocking calls may not be truly interruptible.** `auth_thread.kill` asks MRI to unwind the thread, but a thread blocked in a C-level blocking call that MRI can't preempt (this varies by what exactly is blocking — DNS resolution via the system resolver is the classic example) can in principle keep running in the background after `kill` returns, eventually completing the connection and writing into the module-level `Lich::Common::Account.name/game_code/character` state (`eaccess.rb`, `self.auth`) well after the caller has moved on. In the common retry case this is low-risk because a retry re-sends the *same* account/character, so a late write and a fresh write would carry identical values — but it's a real, if narrow, latent risk. **This is not a new risk introduced by this PR** — it's the identical risk already accepted in the merged `Game.open_with_timeout` precedent this code copies, so it doesn't block this PR, but it's worth naming since it's now duplicated in a second place. **Settles it:** nothing to settle from the diff alone; this is an inherent MRI limitation, not a code defect, and matches an already-accepted tradeoff.

## What this PR gets right (tests)

`eaccess_spec.rb`'s three new cases (`succeeds within timeout`, `raises when auth hangs past timeout`, `re-raises whatever auth raises`) map directly onto `Thread#join`/`Thread#value` semantics and are the right cases to cover for this kind of wrapper.

PR #1505, opened 2026-08-09
fix(all): Honor Simu-side teardown for 10 seconds

OSXLich-Doug:
## Summary by CodeRabbit

* **New Features**
  * Orderly shutdown now requests a clean game-server exit and waits for confirmation before closing.
  * Exit commands are forwarded consistently during client shutdown, including detachable sessions.

* **Bug Fixes**
  * Improved shutdown handling for timeouts, connection failures, and unexpected disconnections.
  * Prevented library reload cleanup errors when no reload state exists.
  * Improved handling of missing bounty descriptions.
  * Game connection status now accurately detects remote disconnects.

### DIFF
--- a/lib/common/orderly_shutdown.rb
+++ b/lib/common/orderly_shutdown.rb
@@ -2,6 +2,7 @@
 
 require_relative 'shutdown_log'
 require_relative 'shutdown_result_predicates'
+require_relative 'shutdown_intent'
 
 module Lich
   module Common
@@ -12,6 +13,10 @@ module Lich
     # the stored result. Connection-loss cleanup can add its own runner without
     # overloading the state object.
     module OrderlyShutdown
+      SERVER_EXIT_TIMEOUT_SECONDS = 10
+
+      class ServerExitTimeout < StandardError; end
+
       # Result for one orderly-shutdown attempt.
       Result = Struct.new(
         :completed,
@@ -38,8 +43,11 @@ module Lich
         #
         # @param source [#to_s] source recorded in shutdown logs
         # @param current_script [Object, nil] script to exclude from shutdown drain
+        # @param server_exit_command [String, nil] validated command sent before close
+        # @param server_exit_timeout [Numeric] seconds to wait for remote EOF
         # @return [Result] stored orderly-shutdown result
-        def request_user_exit(source:, current_script: nil, coordinator: ShutdownCoordinator, scripts_provider: nil, script_drain: ShutdownScriptDrain, vars: Vars, game: Game, active_sessions_lifecycle: nil, slow_threshold: 1.5)
+        def request_user_exit(source:, current_script: nil, coordinator: ShutdownCoordinator, scripts_provider: nil, script_drain: ShutdownScriptDrain, vars: Vars, game: Game, active_sessions_lifecycle: nil, slow_threshold: 1.5, server_exit_command: nil, server_exit_timeout: SERVER_EXIT_TIMEOUT_SECONDS)
+          validate_server_exit!(server_exit_command, server_exit_timeout)
           ShutdownLog.begin_user_exit_summary!
           coordinator.request(reason: :user_exit, source: source)
 
@@ -60,7 +68,9 @@ module Lich
             vars: vars,
             game: game,
             active_sessions_lifecycle: active_sessions_lifecycle,
-            slow_threshold: slow_threshold
+            slow_threshold: slow_threshold,
+            server_exit_command: server_exit_command,
+            server_exit_timeout: server_exit_timeout
           )
           if current_script && initial_scripts.include?(current_script)
             result.scripts_drained = false
@@ -83,10 +93,13 @@ module Lich
         # @param game [#close] game connection facade
         # @param active_sessions_lifecycle [#update_connected, nil] optional session registry
         # @param slow_threshold [Float] seconds before script drain reports slow scripts
+        # @param server_exit_command [String, nil] validated command sent before close
+        # @param server_exit_timeout [Numeric] seconds to wait for remote EOF
         # @return [Result] stored orderly-shutdown result
         # @raise [ArgumentError] when caller input is invalid or reason is not :user_exit
-        def run(coordinator:, initial_scripts:, remaining_scripts:, script_drain:, vars:, game:, active_sessions_lifecycle: nil, slow_threshold: 1.5)
-          validate!(coordinator: coordinator, remaining_scripts: remaining_scripts, script_drain: script_drain, vars: vars, game: game)
+        def run(coordinator:, initial_scripts:, remaining_scripts:, script_drain:, vars:, game:, active_sessions_lifecycle: nil, slow_threshold: 1.5, server_exit_command: nil, server_exit_timeout: SERVER_EXIT_TIMEOUT_SECONDS)
+          validate!(coordinator: coordinator, remaining_scripts: remaining_scripts, script_drain: script_drain, vars: vars, game: game, server_exit_command: server_exit_command)
+          validate_server_exit!(server_exit_command, server_exit_timeout)
           raise ArgumentError, "orderly user exit requires reason=:user_exit" unless coordinator.orderly_user_exit?
 
           result = Result.new(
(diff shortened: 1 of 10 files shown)


OSXLich-Doug:
@coderabbitai review

mrhoribu (approved):
# PR Review: fix(all): Honor Simu-side teardown for 10 seconds (elanthia-online/lich-5#1505)

**Reviewed:** 10 files, +191/-22 at `eff2b87` (base `84ce0fb`) on 2026-08-09.
**Based on:** full contents of all 10 changed files at head and base (fetched individually via `raw.githubusercontent.com`), diffed locally. No `gh` CLI / API access (rate-limited), so **no git history, blame, or full-repo grep was available** — see Coverage notes.
**Summary:** Adds a bounded (default 10s) wait for the game server to acknowledge an `exit`/`quit` command and close its side of the socket before Lich closes the connection locally, on top of the existing `OrderlyShutdown` sequence. The core mechanism (new `Game#remote_eof?`, a real success return from `Game#_puts`, and `OrderlyShutdown.request_server_exit`) is correctly built and thoroughly tested against every failure mode (send failure, timeout, unrelated reader exit). Two small, unrelated one-line changes are bundled in that don't actually change behavior.

## Blockers

None found.

## Major

None found.

## Minor

### 1. Two bundled changes have no behavioral effect and are unrelated to this PR's purpose

`lib/gemstone/bounty/parser.rb:156` and `lib/common/script.rb:1165,1173`:

```ruby
# lib/gemstone/bounty/parser.rb
-          if desc&.empty?
+          if desc && desc.empty?
```
```ruby
# lib/common/script.rb (both occurrences)
-              @@reloading_libraries.delete(library) if reloaders&.empty?
+              @@reloading_libraries.delete(library) if reloaders && reloaders.empty?
```

CodeRabbit's summary describes these as bug fixes ("Prevented library reload cleanup errors when no reload state exists", "Improved handling of missing bounty descriptions"), but they aren't: `x&.empty?` and `x && x.empty?` evaluate identically in an `if` — both are falsy when `x` is `nil` and both call `.empty?` otherwise. `desc` comes from `checkbounty`, which only ever returns `XMLData.bounty_task` or `nil` (never `false`), and `@@reloading_libraries` is a plain `{}` with no default proc, so `reloaders` is only ever `nil` or a `Set`. Neither side of either diff can behave differently from the other.

This isn't a bug, just unrelated churn (likely a rubocop autocorrect or leftover from another branch) riding along in a shutdown-focused PR. Worth confirming it was intentional to include, since it makes the diff slightly harder to review for what actually changed.
**Confidence:** confirmed (traced both call sites and their inputs).

### 2. `@remote_eof` is written from the reader thread and read from other threads with no synchronization

`lib/games.rb:754` sets `@remote_eof = true` inside `start_socket_reader_thread`'s background thread; `remote_eof?` (`lib/games.rb:536`) is read from whichever thread calls `OrderlyShutdown.request_server_exit` (a script thread, or the primary-frontend thread). Every other piece of cross-thread game state in this class (`@socket` writes in `_puts`) is protected by `@mutex`; this new flag isn't.

In practice this is very likely safe on MRI: a single `Boolean` ivar write/read has no torn-read hazard under the GIL, and the existing `@last_recv`, `@thread`, `@reader_thread` ivars are already read cross-thread without a mutex, so this matches the file's existing (implicit) concurrency model rather than introducing a new pattern.
**Confidence:** suspected only in the sense that I can't rule out a non-MRI Ruby target for this project; **settles it:** does this codebase ever run under JRuby/TruffleRuby, or is MRI (with its GIL) a hard assumption? `lich5-environment.md` and the Gemfile give no indication otherwise, so this is very low risk.

## Nits

- `lib/common/orderly_shutdown.rb:50,102` — `validate_server_exit!` is called once directly in `request_user_exit` and again inside the `run` it delegates to, so a `request_user_exit` call validates the same two arguments twice. Harmless (the check is pure/stateless) but slightly redundant; likely deliberate so `run` stays independently safe to call, which is a reasonable trade-off to leave as-is.
- `lib/main/user_exit_dispatch.rb:56-60` — `run_orderly_user_shutdown(source: :detachable_frontend, **request_options, server_exit_command: prefixed)`: if a future caller ever passed `server_exit_command` inside `request_options`, the trailing explicit `server_exit_command: prefixed` would silently win (no error, no warning) rather than the two being reconciled. No current caller does this (checked `global_defs.rb`'s one call site), so this is speculative, not a live bug — just worth keeping in mind if `request_options` ever gets threaded through from a caller closer to the user.
- CI's docstring-coverage check is failing (8.33% vs. 60% required) per the PR's automated pre-merge checks. The two new private helpers this PR adds (`request_server_exit`, `validate_server_exit!`) are undocumented, but that matches the file's existing convention of not YARD-documenting private helpers (`close_game`, `save_vars`, etc. are likewise undocumented) — the public-facing methods (`request_user_exit`, `run`) do get updated `@param` docs. Not a new problem this PR introduces, but flagging since CI is red on it.

## Open questions

- **Settled, no action needed:** I checked whether the new blocking `reader_thread.join(server_exit_timeout)` (up to 10s) could race the `ShutdownWatchdog`'s force-exit (`lib/common/shutdown_watchdog.rb`, default 60s) and cut a graceful exit short. It can't under default settings — 60s comfortably covers the 10s wait plus the rest of teardown — but an operator who has lowered `shutdown_watchdog_timeout` below ~10-15s via the `lich_settings` override would now be more likely to get force-killed mid-exit than before this PR. Worth a one-line callout in the PR description for anyone who has customized that setting.
- `lib/main/main.rb:688-689` forwards whatever `client_string` the primary-frontend loop already normalized (via `Frontend.supports_gsl?` / `fb_to_sf`) straight through to `game._puts` as the literal exit command sent to the game server. I traced this as consistent with how `Game.puts` already prefixes every other outgoing command the same way, so I'm treating it as correct rather than flagging it — but I don't have a captured Wizard/Stormfront/Frostbite session log to confirm the exact wire format for every frontend. **Settles it:** if QA testing on this PR (the `test` label was added) exercises a Frostbite or Wizard-mode exit specifically, that closes this out completely.

## What this PR gets right

- `Game#_puts` previously returned `nil` unconditionally (`IO#puts` always returns `nil`), so a truthiness check on its result could never distinguish success from failure. Adding the explicit `true` return (`lib/games.rb:632-640`) is the fix that makes `request_server_exit`'s send-failure check actually work, and it's backward-compatible — every other call site in the changed files discards `_puts`'s return value already.
- `request_server_exit` distinguishes three distinct failure modes (command couldn't be sent, timeout waiting for the reader thread, reader thread stopped for a reason other than remote EOF) and each has its own test in `orderly_shutdown_spec.rb` that exercises the real implementation code (not just re-asserting mocks), including the subtle "reader died from something other than EOF" case that a less careful implementation would have silently treated as success.
- `ShutdownIntent.user_exit_command?` is reused (rather than re-implemented) to validate `server_exit_command` before it's ever sent to the socket, so `OrderlyShutdown` can't be made to forward an arbitrary string to the game server — `orderly_shutdown_spec.rb`'s "rejects a non-exit server command" test confirms this.
- The new server-exit step is opt-in (`if server_exit_command`), and the extra `validate!` checks for `_puts`/`reader_thread`/`remote_eof?` are gated behind `server_exit_command` too, so any existing caller of `OrderlyShutdown.run`/`request_user_exit` that doesn't pass the new keyword is completely unaffected by this change.

## Coverage notes

- **What was reviewed:** full contents of all 10 changed files at both the PR's head commit (`eff2b87`) and merge-base (`84ce0fb`), diffed locally; the PR description/CodeRabbit walkthrough; the "Files changed" page for the file list and stat totals; and the non-diff files needed to ground the review (`shutdown_intent.rb`, `shutdown_result_predicates.rb`, `shutdown_watchdog.rb`).
- **What was not available:** no local clone, so no `git blame`/`git log -L`/`git log -S` on the touched lines — I could not check whether the two no-op changes (Minor #1) or the `@remote_eof` synchronization question (Minor #2) were already discussed in a prior commit or review thread. The `gh` CLI wasn't present and the GitHub REST API was rate-limited (unauthenticated), so PR comments/review-thread history beyond what rendered on the page (CodeRabbit's auto-review, no human comments yet) weren't fetched.
- **Not run:** Ruby syntax check / RuboCop / RSpec — no working Ruby interpreter was available in this environment (an `apt-get install ruby` attempt failed on a missing `security.ubuntu.com` package). The new/changed specs were read and manually traced against the implementation rather than executed.
- **Blast radius:** I could only grep for other callers of `Game#_puts`, `OrderlyShutdown.run`/`.request_user_exit`, and the `run_orderly_user_shutdown` proc within the 10 changed files plus the few files fetched for grounding — not the full repo (e.g. `scripts/`, `dr-scripts/`, or other `lib/` files not touched by this PR). Everything found within that scope is consistent with the change, but a caller outside these files that relies on `Game#_puts` returning literal `nil` on success, or that calls `OrderlyShutdown.run` directly with a `game` double lacking `_puts`/`reader_thread`/`remote_eof?` **and** a truthy `server_exit_command`, would not have been visible to this review.

PR #1513, opened 2026-08-11
feat(dr): user-extensible creature name normalization (02 of 06)

MahtraDR:
> **Stacked on #1512 (01 of 06).** This branch includes #1512's commit; review/merge #1512 first, then this rebases cleanly onto `main`. Depends on `Lich::DragonRealms::CustomSubstitutions` introduced there.

## Problem

`normalize_creature_names` hardcodes the creatures whose variant descriptions collapse to a canonical name (`alfar warrior`, `sinewy leopard`, `lesser naga`). A player whose own pet/creature description confuses room parsing must wait for a Lich release to have it normalized.

## Fix

Source the canonical list from `CustomSubstitutions.resolve(:custom_creature_normalizations, DRDefsPattern::CREATURE_NORMALIZATIONS, type: :names)`, merging the player's own additions on top of the built-in defaults (which remain the fallback when nothing is configured).

- The three per-creature `Regexp` constants are replaced by one `CREATURE_NORMALIZATIONS` name list.
- `normalize_creature_names` collapses via a plain substring check (`text.include?(name) ? name : text`) — behaviorally identical to the old `/.*name.*/` subs, but without compiling a regex per name per call. This matters because `find_npcs` runs on **every room update**; the merged list is memoized by `CustomSubstitutions`, so the hot path never re-reads settings.

**Note on GELAPOD:** intentionally left as-is. On inspection it is *not* a creature normalization — it unwraps `pushBold` XML around a room object inside `find_objects`, a different transform at a different point — so folding it into this list would be incorrect.

## Tests

Full suite: **5569 examples, 0 failures.** Existing alfar/leopard/naga/other cases stay green; adds a player-added creature case and a malformed-entry case asserting the warning fires (with key, index, reason) and the built-in defaults still apply.

### DIFF
--- a/lib/dragonrealms/commons/common.rb
+++ b/lib/dragonrealms/commons/common.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require_relative '../custom_substitutions'
+
 module Lich
   module DragonRealms
     module DRC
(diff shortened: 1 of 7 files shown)


mrhoribu (approved):

mrhoribu:
 
|---|---|---|---|
 
### Final status
 
| PR | Before | After |
|---|---|---|
 

PR #1514, opened 2026-08-11
feat(dr): user-extensible boxes and trash receptacles (03 of 06)

MahtraDR:
> **Stacked on #1513 (02 of 06) → #1512 (01 of 06).** This branch includes those commits; review/merge them first. Depends on `Lich::DragonRealms::CustomSubstitutions` from #1512.

## Problem

`box_list_to_adj_and_noun` only recognizes a hardcoded set of box woods and container nouns (`BOX_REGEX`), and `TRASH_STORAGE` is a hardcoded set of trash receptacles. A player who finds a box made of an unlisted wood, or whose town has an unlisted trash receptacle, silently can't loot/trash it — and must wait for a Lich release.

## Fix

**Boxes:** extract `BOX_WOODS` and `BOX_CONTAINERS` in `drvariables` and rebuild `BOX_REGEX` from them (byte-identical source — verified; `$box_regex` unchanged for third-party scripts). `box_list_to_adj_and_noun` now builds its match regex from those defaults merged with the player's `custom_box_woods` / `custom_box_containers` (each term `Regexp.escape`d), and applies `DEFAULT_BOX_SUBSTITUTIONS` (`ironwood → iron`) merged with `custom_box_substitutions`.

**Trash:** add `DRCI.trash_storage` returning `TRASH_STORAGE` merged with the player's `custom_trash_storage`, used in `dispose_trash`.

All merges go through `CustomSubstitutions` (#1512): per-entry validation, verbose per-entry rejection messages (key, index, reason, consequence), memoization, and defaults-as-fallback.

## Tests

Full suite: **5576 examples, 0 failures.** Adds box specs (player-added wood, container, substitution, and a malformed-wood case that still parses built-in boxes and warns) and trash specs (default, player-added receptacle, malformed). A `ruby` check confirms the rebuilt `BOX_REGEX` source is identical to the original.

### DIFF
--- a/lib/dragonrealms/commons/common-items.rb
+++ b/lib/dragonrealms/commons/common-items.rb
@@ -27,6 +27,8 @@
 # @see DRC Core common module
 # @see EquipmentManager Higher-level gear management
 
+require_relative '../custom_substitutions'
+
 module Lich
   module DragonRealms
     module DRCI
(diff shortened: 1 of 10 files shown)


mrhoribu (approved):

mrhoribu:
 
|---|---|---|---|
 
### Final status
 
| PR | Before | After |
|---|---|---|
 

PR #1516, opened 2026-08-11
feat(dr): user-extensible flavor-text stripping (05 of 06)

MahtraDR:
> **Stacked on #1515 (04) → #1514 (03) → #1513 (02) → #1512 (01).** Review/merge those first. Depends on `Lich::DragonRealms::CustomSubstitutions` from #1512.

## Problem

`remove_flavor_text` strips descriptive flavor (`... adorned with ...`) using a single large hardcoded regex, `FLAVOR_TEXT_PATTERN`. When the game introduces a new flavor phrasing the pattern doesn't cover, the flavor leaks into item nouns (and `get_noun` mis-parses) until a Lich release extends the regex. Flavor is open-ended grammar, so a literal-name list can't help — this case genuinely needs pattern power.

## Fix

`remove_flavor_text` applies the built-in `FLAVOR_TEXT_PATTERN` first, then folds any player-defined `custom_flavor_text_patterns` (regular expressions) over the result via `CustomSubstitutions.apply_regexes`. A player can strip a new flavor phrasing without a Lich release; a plain phrase also works since it's a valid regex.

**Strict guards** (this is user-supplied regex on a control-relevant path — all via `CustomSubstitutions`, #1512):
- each pattern is validated and compiled with a per-pattern `Regexp.timeout` before use;
- an invalid pattern is reported with the exact compile error and skipped;
- a pattern that times out at apply time is caught, reported once (deduped), and skipped — never hanging Lich.

The built-in pattern always runs, so behavior is unchanged when nothing is configured.

## Tests

Full suite: **5585 examples, 0 failures.** Existing flavor cases stay green; adds a case where a user pattern strips flavor the built-in misses, and an invalid-regex case asserting the built-in still applies and the warning fires. (Compile/timeout guard internals are covered in the `CustomSubstitutions` unit spec from #1512.)

### DIFF
--- a/lib/dragonrealms/commons/common-items.rb
+++ b/lib/dragonrealms/commons/common-items.rb
@@ -27,6 +27,8 @@
 # @see DRC Core common module
 # @see EquipmentManager Higher-level gear management
 
+require_relative '../custom_substitutions'
+
 module Lich
   module DragonRealms
     module DRCI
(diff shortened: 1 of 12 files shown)


mrhoribu (approved):

mrhoribu:
 
|---|---|---|---|
 
### Final status
 
| PR | Before | After |
|---|---|---|
 
OSXLich-Doug (approved):
Reviewed - no blockers, correctness issues or nits.  Clean run from local CodeRabbit AI for the whole stack.  Ready to merge.

PR #1517, opened 2026-08-12
fix(all): Linux WINE and Wrayth FE detection

OSXLich-Doug:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of Wine-managed Windows frontends on Linux, including valid files without Unix execute permissions.
  * Made Wine registry lookups case-insensitive for more reliable frontend discovery.
  * Preserved validation for other executable paths and continued handling system-call errors safely.

* **Tests**
  * Added coverage for non-executable Windows frontend files and case-insensitive registry keys.

### DIFF
--- a/lib/common/frontend_locator.rb
+++ b/lib/common/frontend_locator.rb
@@ -161,7 +161,7 @@ module Lich
         candidates.concat(path_candidates(definition).map { |path| [path, :path] })
 
         candidates.each do |path, source|
-          next unless executable?(path)
+          next unless executable?(path, definition)
 
           resolved = resolution(definition, path, source)
           return resolved if resolved
@@ -171,7 +171,7 @@ module Lich
 
       def resolve_override(definition, override)
         path = expand_path(override)
-        unless executable?(path)
+        unless executable?(path, definition)
           raise ArgumentError, "frontend override is not executable: #{override}"
         end
 
@@ -352,11 +352,20 @@ module Lich
         nil
       end
 
-      def executable?(path)
-        !path.to_s.empty? && File.file?(path) && File.executable?(path)
+      def executable?(path, definition)
+        return false if path.to_s.empty? || !File.file?(path)
+
+        wine_managed_executable?(path, definition) || File.executable?(path)
       rescue SystemCallError => e
         log_discovery_error(path, e)
         false
+      end
+
+      def wine_managed_executable?(path, definition)
+        platform_key == :linux &&
+          @wine &&
+          definition.dig(:metadata, :launcher_adapter) == :simutronics &&
+          File.extname(path).casecmp?('.exe')
       end
 
       def resolution(definition, path, source)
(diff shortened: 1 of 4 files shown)


mrhoribu (approved):
# PR Review: fix(all): Linux WINE and Wrayth FE detection (elanthia-online/lich-5#1517)

**Reviewed:** 4 files, +48/-8 at `39089f4` on 2026-08-19. Based on: full clone (branch checked out, history and callers greppable).
**Verdict:** Approve
**Summary:** Fixes frontend discovery on Linux+Wine two ways: (1) `.exe` frontend binaries under a Wine prefix are recognized as launchable even without the Unix executable bit set (PE executability isn't governed by that bit), and (2) a Wine `system.reg` registry-key lookup that was comparing subkey paths case-sensitively is made case-insensitive.

## What this PR gets right

The fix is consistent with logic the codebase already relies on elsewhere: `native_launcher_supported?` (`lib/common/frontend_locator.rb:290`) already treats a `:simutronics`-adapter frontend as supported on Linux whenever `!@wine.nil?`, with no Unix-executable-bit requirement — so `wine_managed_executable?` extending that same "Wine + `:simutronics` adapter" recognition down into the `executable?` gate is closing a real gap between "we say this frontend is supported here" and "we can actually find its binary here," not introducing new logic.

The new `wine_managed_executable?` is properly scoped: it requires `platform_key == :linux`, an injected `@wine` provider, `launcher_adapter == :simutronics` specifically (confirmed against `lib/common/front-end.rb:238`, where `stormfront`/`Wrayth` is the only frontend registered with that adapter and a `.exe` discovery list), and a `.exe` extension — so it can't broaden matching for unrelated frontends or platforms.

`File.file?(path)` is still checked before either branch of the `||`, so the change only ever affects the *permission* check, never the *is-this-a-real-file* check.

The `casecmp?` fix in `wine.rb` addresses a real, separate case-sensitivity bug in how `Wine.registry_gets` matches subkey lines against `system.reg` content — well-isolated, single-line change, and it does not alter the surrounding `hkey`/`thingie` parsing.

Both fixes are covered by tests that assert the specific pre-fix failure mode directly: the frontend-locator test explicitly confirms `File.executable?(wrayth)` is `false` before asserting the locator still resolves it, and the wine test exercises the mismatched-case lookup that would have returned `nil` before this fix.

## Minor

**The Unix-executable check is fully bypassed for any `.exe` file once the Wine+`:simutronics` conditions hold, not just for known frontend binaries** (`lib/common/frontend_locator.rb:364`). Candidate paths are already scoped to specific discovery locations (`path_candidates`, registry-derived paths), so this isn't opening arbitrary user input to execution — but it does mean any `.exe` sitting in one of those scanned locations, executable or not, genuinely a Simutronics launcher or not, will now pass the "is this executable" gate. Given the pre-existing scope of candidate discovery, this is a narrow, low-severity widening rather than a new attack surface, but worth the author's awareness since it wasn't spelled out in the PR description.

PR #1518, opened 2026-08-12
fix(all): update Gemstone game-code validations

### LINKED ISSUE #1507: GameSelection::GAME_MAPPING is missing GS4, which LoginHelpers::VALID_GAME_CODES treats as valid
\`LoginHelpers::VALID_GAME_CODES\` (\`lib/common/authentication/login_helpers.rb:14\`) includes \`GS4\`, but \`GameSelection::GAME_MAPPING\` (\`lib/common/gui/game_selection.rb:11-20\`) does not.

\`CLIPassword#add_character\` (\`lib/common/authentication/cli_password.rb:339\`) derives \`game_name\` via \`GameSelection.get_game_name(game_code)\`, which falls back to \`'Unknown'\` for any code not present in \`GAME_MAPPING\`.

Result: \`--add-character ACCOUNT Char --game-code GS4\` passes CLI validation (added in #1504) but still persists \`game_name: 'Unknown'\` in \`entry.yaml\`.

Pre-existing, confirmed unrelated to #1504's diff - neither file was touched by that PR.

**Fix direction:** either add \`GS4\` to \`GAME_MAPPING\` with its real display name, or have the CLI validator's allow-list derive from \`GameSelection::GAME_MAPPING.keys\` instead of a separately-maintained list, so the two can't drift again.

OSXLich-Doug:
## Summary by CodeRabbit

* **Bug Fixes**
  * Updated supported game-code validation across login and character creation.
  * Retired and unknown codes are now rejected with clear error messages.
  * Improved game selection handling and display names for supported games.
  * Invalid login selections no longer proceed or produce misleading resolution logs.

* **Tests**
  * Expanded coverage for supported, retired, missing, and invalid game codes.
  * Added coverage for game-selection mappings, lookups, and refresh behavior.

### DIFF
--- a/lib/common/authentication/cli.rb
+++ b/lib/common/authentication/cli.rb
@@ -17,7 +17,7 @@ module Lich
         # Executes CLI login flow for a specified character
         #
         # @param character_name [String] Character name to login with
-        # @param game_code [String, nil] Game code/instance (GS3, GS4, DR, etc.)
+        # @param game_code [String, nil] Game code/instance (GS3, GST, DR, etc.)
         # @param frontend [String, nil] Frontend type (stormfront, avalon, wizard)
         # @param custom_launch [String, nil] Custom launch filter (if provided, frontend is ignored for matching)
         # @param data_dir [String] Directory containing saved login entries
@@ -127,7 +127,7 @@ module Lich
             return nil
           end
 
-          unless game_code && LoginHelpers::VALID_GAME_CODES.include?(game_code.to_s)
+          unless game_code && LoginHelpers.valid_game_code?(game_code.to_s)
             Lich.log "error: A valid game code is required for new character creation (e.g. --dr, --gemstone). Got: #{game_code.inspect}"
             return nil
           end
--- a/lib/common/authentication/login_helpers.rb
+++ b/lib/common/authentication/login_helpers.rb
@@ -55,7 +55,7 @@ module Lich
         # CLI flags that should never be interpreted as game-instance selectors.
         NON_INSTANCE_FLAGS = %w[
           login gui no-gui without-frontend headless reconnect reconnected save
-          genie frostbite wrayth saga
+          genie frostbite wrayth saga suks
         ].freeze
 
         # CLI options (key portion before '=') that are non-instance modifiers.
--- a/lib/common/authentication/login_helpers.rb
+++ b/lib/common/authentication/login_helpers.rb
@@ -13,8 +13,17 @@ module Lich
         # Load up / require gem 'os' for operating system detection work
         Lich::Util.install_gem_requirements({ 'os' => true })
 
-        # Valid game codes
-        VALID_GAME_CODES = %w[GS3 GS4 GSX GSF GST DR DRX DRF DRT].freeze
+        # Game codes accepted for login and character creation.
+        VALID_GAME_CODES = %w[GS3 GST GSF DR DRX DRT DRF].freeze
+
+        # Returns whether a game code is accepted for login and character creation.
+        # Callers remain responsible for any input normalization they require.
+        #
+        # @param game_code [String, nil] game instance code
+        # @return [Boolean]
+        def self.valid_game_code?(game_code)
+          VALID_GAME_CODES.include?(game_code)
+        end
 
         # CLI login name that requests the character generator instead of
         # selecting an existing character (e.g. `--login NEW`). This is a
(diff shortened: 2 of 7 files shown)


OSXLich-Doug:
@coderabbitai confirm latest commit addresses `--suks` finding satisfactorily.

mrhoribu (approved):
# PR Review: fix(all): update Gemstone game-code validations (elanthia-online/lich-5#1518)

**Reviewed:** 7 files, +193/-30 at `164cb5c` on 2026-08-19. Based on: full clone (branch checked out, history and callers greppable).
**Verdict:** Approve
**Summary:** Retires `GS4` and `GSX` (GemStone Platinum) as valid login/character-creation game codes across CLI parsing, the login-helpers validator, and the GUI game-selection dropdown, while deliberately keeping their legacy realm/name mappings intact for normalizing already-persisted saved-login entries. Also fixes an unrelated pre-existing bug where `--suks` was misclassified as a probable invalid instance selector.

## What this PR gets right

**The GSX/Platinum removal is a deliberate decommission, not a regression — verified by tracing the full resolution path, not just by reading the diff.** My first pass through this PR raised a specific concern: `REALM_TO_GAME_CODE['platinum']` still maps to `'GSX'` (`login_helpers.rb`), so `--gemstone --platinum` still *resolves* to `'GSX'` internally — the question was whether that value then leaks past the new `valid_game_code?` gate. Tracing `resolve_instance` (`login_helpers.rb:420-476`) end-to-end confirms it doesn't: the method's final line changed from `return resolved_instance unless resolved_instance.nil?` to `return resolved_instance if valid_game_code?(resolved_instance)`, so a resolved-but-invalid code like `'GSX'` now falls through to the `instance_flags_seen`-driven `nil` return instead of being handed back as if it were valid. This is confirmed by an explicit, intentionally-renamed test: `'rejects the retired GemStone Platinum selector'` (previously `'returns GSX for --gemstone --platinum'`), which now asserts `resolve_instance(['--gemstone', '--platinum'])` is `nil`. The comment above `GAME_CODE_TO_REALM` — "Legacy realm and name mappings intentionally retain retired codes for persisted-entry normalization" — is consistent with what the code actually does: `realm_from_game_code('GSX')` still returns `'platinum'` (still tested), so an old saved-entry YAML file with `game_code: GSX` from before the realm's retirement is still normalized correctly, it just can't be freshly selected for a new login.

**The new hard-fail in `lib/main/main.rb:122-131` closes a real unguarded gap, confirmed by reading what used to happen.** Before this PR, `resolve_login_args` could return `nil` for "probable invalid instance" (the doc comment even said so: "Invalid game codes are not rejected here; call site can choose to validate") and `main.rb`'s CLI login path had no such check — a `nil` `requested_instance` would flow straight into `SagaManagedLogin.cli_decision(game_code: nil, ...)` unguarded. The new check fails fast with a clear stderr message and log line before that happens, and correctly distinguishes `nil` (invalid) from `:__unset` (no instance flags given, defer to saved-entry logic) — the guard only fires on `.nil?`, not on the symbol.

**The `--suks` fix is real and independently grounded.** `suks` (`lib/common/front-end.rb:257`, `launcher_adapter: :embedded`, "SUKS has no client socket") was missing from `NON_INSTANCE_FLAGS` before this PR. Since it isn't a `VALID_GAME_CODES` entry, isn't in `NON_INSTANCE_OPTION_KEYS`, and (before this fix) wasn't in `NON_INSTANCE_FLAGS` either, the argv-scanning loop in `resolve_instance` would fall to its `else` branch and set `instance_flags_seen = true`, causing any `--suks` invocation to be reported as a probable invalid instance — a real, user-facing bug for anyone using the SUKS frontend, now fixed and covered by a dedicated test.

**`GameSelection::GAME_MAPPING` is now derived from `VALID_GAME_CODES` instead of hand-duplicated** (`lib/common/gui/game_selection.rb`), via `VALID_GAME_CODES.to_h { |code| [code, GAME_NAMES.fetch(code)] }.freeze`. This is a good DRY move with a nice side effect: `GAME_NAMES.fetch` (not `[]`) means a future edit that adds a code to `VALID_GAME_CODES` without a matching display name fails loudly at require-time instead of silently rendering `nil`/`"Unknown"` in the GUI. Verified the two lists match 1:1 today (7 entries each) so this doesn't crash on load.

## Nits

**Scope creep.** The PR title is "update Gemstone game-code validations," but it also fixes the unrelated `--suks` classification bug. Both changes are small, well-tested, and genuinely related to the same `resolve_instance` code path, so this isn't a real problem — just worth a one-line mention in the PR description for anyone bisecting later.

**Debug-log suppression on the invalid path is a silent behavior change.** `resolve_login_args` used to always log `"Resolved instance: ..."` at debug level; it's now wrapped in `unless instance.nil?` (`login_helpers.rb`), so a failed resolution no longer produces that debug line. This is intentional and tested (`'does not report an invalid game code as a resolved instance'`), and makes sense given the new caller-side error message replaces it — just flagging it as a deliberate, not accidental, logging change for anyone grepping old debug logs for this line.

PR #1520, opened 2026-08-12
feat(all): add --active-session-dir to decouple Active Sessions coordination from --temp

simtel12:
## Summary

- Active Sessions coordination (the lock + discovery files that let independent Lich processes find each other) previously always lived under `TEMP_DIR`. Anyone overriding `--temp` per-character to keep debug logs separate ends up with an isolated Active Sessions registry per character instead of one shared one, defeating the point of a cross-process session list.
- Adds `--active-session-dir=PATH` to point coordination at a shared directory independently of `--temp`. Unset, behavior is unchanged (falls back to `TEMP_DIR`, or `Dir.tmpdir` if that's undefined).
- Fixes two gaps caught in review before this landed: `coordination_dir` had no boot-time `mkdir` the way `TEMP_DIR` does, so a not-yet-existing `--active-session-dir` path silently degraded to "service unavailable" with no trace of the real cause; and `SessionLauncher`'s optional-flag forwarding assumed every path flag has a `constants.rb` default a spawned child re-derives on its own, which isn't true for this one and caused it to be silently dropped when forwarded value matched the parent's.
- Second commit: an explicit `--active-session-dir` also implicitly enables the Active Sessions service for that launch, without persisting anything to `lich_settings` — removing the flag on a later launch reverts to the persisted default (disabled). Kept as a separate commit in case the "implicit enable" behavior specifically needs to be reverted independently.

### Review round 2 (@OSXLich-Doug)

Four follow-up commits, one per finding:

- **[P1] Propagate the parent's active-session directory to GUI children** — confirmed, and broader than the report: *no* production caller populates any of the nine `OPTIONAL_PATH_FLAGS` context keys, since GUI launch contexts are login entries (account/frontend), not directory overrides. For the other eight that's harmless — a child re-derives them from `constants.rb`. `--active-session-dir` has no such default by design, so it was dropped on every GUI-launched child. Marked `:inherit` and now falls back to this process's `ACTIVE_SESSION_DIR` when the context omits the key; an explicitly empty value stays an opt-out.
- **[P2] Correct the "writes nothing to disk" claim** — the service does create the coordination dir, the lock, and the discovery record. Help text now says it persists no *setting*, and spells out that coordination files are still written. Corrected in the summary above too.
- **[P3] Unused `:active_session_dir` launch-context path** — resolved by the P1 fix, which makes the path reachable from production.
- **[P3] Release the ownership lock before deleting its temp dir** — done in the example's own `ensure`, ahead of `rm_rf`.
- **[P3] Remove the uncalled `spawn_login` helper** — removed, along with its now-unused `ruby_executable` require (both remaining `RubyExecutable` callers require it themselves). Predates this branch; folded in here since it's a second child-launch path already drifted from `SessionLauncher`. Happy to split it out if you'd rather keep the branch narrow.

## Test plan

- [x] `bundle exec rspec` (full suite): 6346 examples, 0 failures (rebased onto `main` through #1532 / 5.20.1)
- [x] `bundle exec rubocop` on all touched files: no offenses
- [x] Manually reproduced both review findings (ENOENT on a missing coordination dir; flag silently dropped when forwarded value matches the parent's own) before and after the fix
- [x] Manually verified the implicit-enable behavior against a real `lich_settings` row forced to `false`, confirming it stays untouched on disk
- [x] Round 2: new specs cover inheriting `ACTIVE_SESSION_DIR` when the context omits it, when no context is passed at all, the explicit-empty opt-out, and the no-constant case
- [x] Re-ran every manual verification against the post-rebase base, driving the real library code rather than specs:
  - missing coordination dir is created and the lock acquired; with the `mkdir_p` removed the same call raises `Errno::ENOENT` and degrades to "service unavailable"
  - real detached child processes spawned via `SessionLauncher`, asserting on the child's actual `ARGV`: inherits the parent's `--active-session-dir` from a production-shaped GUI context, still forwards it when the explicit value equals the parent's, honours the explicit-empty opt-out, and emits nothing when the parent has no `ACTIVE_SESSION_DIR`. `--data` matching its `constants.rb` default is still suppressed in the same run, confirming the suppression logic is live and only bypassed for this flag
  - against a real SQLite `lich_settings` row forced to `false`: `ActiveSessions.enabled?` returns true with the flag set, the row still reads `false`, no rows are added, the db file is byte-identical, and dropping the flag reverts to the persisted `false`

### DIFF
--- a/lib/common/authentication/login_helpers.rb
+++ b/lib/common/authentication/login_helpers.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require_relative '../front-end'
-require_relative '../ruby_executable'
 
 # login_helpers.rb: Core lich file for collection of utilities to extend Lich capabilities.
 # Entries added here should always be accessible from Lich::Common::Authentication::LoginHelpers.method namespace.
@@ -603,52 +602,6 @@ module Lich
             when 'DRT' then '--drt'
             else nil
             end
-          end
-        end
-
-        # Spawns a Lich login session using a saved entry.
-        #
-        # This constructs and launches a Ruby + Lich command line with proper login arguments.
-        # It is aware of the Lich version and formats launch flags (e.g., `--gst`, `--GSX`) accordingly.
-        # Only the character name and game instance are passed - all sensitive data is handled by Lich internally.
-        #
-        # @param entry [Hash] the login entry (must include :char_name and :game_code)
-        # @param lich_path [String, nil] optional path to lich.rbw; defaults to LICH_DIR/lich.rbw
-        # @param startup_scripts [Array<String>] optional scripts to autostart post-login
-        # @param instance_override [String, Symbol, nil] optional instance override (e.g., 'GST', 'GSX')
-        # @param frontend_override [String, nil] optional frontend (e.g., 'avalon', 'wizard')
-        # @param custom_launch_filter [String, nil] optional custom launch filter for entry selection
-        # @return [Process::Waiter, nil] detached process handle if successful, nil otherwise
-        def self.spawn_login(entry, lich_path: nil, startup_scripts: [], instance_override: nil, frontend_override: nil, custom_launch_filter: nil)
-          ruby_path = Lich::Common::RubyExecutable.resolve
-          lich_path ||= File.join(LICH_DIR, 'lich.rbw')
-
-          spawn_cmd = [
-            "#{ruby_path}",
-            "#{lich_path}",
-            '--login', entry[:char_name]
-          ]
-          if instance_override
-            flag = format_launch_flag(instance_override)
-            spawn_cmd << flag if flag
-          end
-          spawn_cmd << "--#{frontend_override}" unless frontend_override.nil?
-          spawn_cmd << "--custom-launch=#{custom_launch_filter}" if custom_launch_filter
-          spawn_cmd << "--start-scripts=#{startup_scripts.join(',')}" if startup_scripts.any?
-
-          Lich::Messaging.msg('info', "Spawning login: #{spawn_cmd}")
-
-          begin
-            pid = Process.spawn(*spawn_cmd)
-            Process.detach(pid)
-          rescue Errno::ENOENT => e
-            Lich::Messaging.msg('error', "Executable not found: #{e.message}")
-            Lich.log "error: Executable not found: #{e.message}"
-            nil
-          rescue StandardError => e
-            Lich::Messaging.msg('error', "Failed to launch login session: #{e.class} - #{e.message}")
-            Lich.log "error: Failed to launch login session: #{e.class} - #{e.message}"
-            nil
           end
         end
       end
--- a/lib/common/session_launcher.rb
+++ b/lib/common/session_launcher.rb
@@ -16,7 +16,8 @@ module Lich
         { option: 'maps', key: :map_dir, constant: :MAP_DIR },
         { option: 'logs', key: :log_dir, constant: :LOG_DIR },
         { option: 'backup', key: :backup_dir, constant: :BACKUP_DIR },
-        { option: 'lib', key: :lib_dir, constant: :LIB_DIR }
+        { option: 'lib', key: :lib_dir, constant: :LIB_DIR },
+        { option: 'active-session-dir', key: :active_session_dir, constant: :ACTIVE_SESSION_DIR }
       ].freeze
 
       class << self
@@ -26,7 +27,8 @@ module Lich
         # @param launch_context [Hash, nil] Optional context keys:
         #   :char_name, :game_code, :frontend, :custom_launch, :dark_mode
         #   and optional directory overrides (:home_dir, :data_dir, :script_dir,
-        #   :temp_dir, :map_dir, :log_dir, :backup_dir, :lib_dir).
+        #   :temp_dir, :map_dir, :log_dir, :backup_dir, :lib_dir,
+        #   :active_session_dir).
         # @return [Hash] Structured result:
         #   - success: { ok: true, pid: Integer }
         #   - failure: { ok: false, error: String }
@@ -177,6 +179,14 @@ module Lich
           if option_name == 'home'
             return Object.const_defined?(:LICH_DIR) ? File.expand_path(Object.const_get(:LICH_DIR).to_s) : nil
           end
+
+          # active-session-dir has no constants.rb default of its own (see
+          # ActiveSessions.coordination_dir): an unset child falls back to its
+          # own TEMP_DIR, not to this process's ACTIVE_SESSION_DIR. There is no
+          # default here to compare against, so never suppress this flag --
+          # doing so would silently drop it whenever the requested value
+          # happens to match the parent's own coordination dir.
+          return nil if option_name == 'active-session-dir'
 
           home_override = context[:home_dir]
           if !home_override.to_s.empty?
--- a/lib/common/session_launcher.rb
+++ b/lib/common/session_launcher.rb
@@ -17,7 +17,10 @@ module Lich
         { option: 'logs', key: :log_dir, constant: :LOG_DIR },
         { option: 'backup', key: :backup_dir, constant: :BACKUP_DIR },
         { option: 'lib', key: :lib_dir, constant: :LIB_DIR },
-        { option: 'active-session-dir', key: :active_session_dir, constant: :ACTIVE_SESSION_DIR }
+        # :inherit marks a flag with no constants.rb default of its own, which a
+        # child therefore cannot re-derive from its own ARGV. Such a flag falls
+        # back to this process's constant when the launch context omits it.
+        { option: 'active-session-dir', key: :active_session_dir, constant: :ACTIVE_SESSION_DIR, inherit: true }
       ].freeze
 
       class << self
@@ -96,8 +99,10 @@ module Lich
         end
 
         # Builds optional CLI flags for child launches.
-        # Emits path flags only when explicitly overridden to a non-default value.
-        # This keeps child ARGV concise and avoids passing parent default directories.
+        # Emits path flags only when explicitly overridden to a non-default value,
+        # or, for :inherit flags, when this process holds a value the child cannot
+        # re-derive. This keeps child ARGV concise and avoids passing parent
+        # default directories.
         #
         # @param context [Hash] Optional launch context from GUI callbacks.
         # @return [Array<String>] Optional flags (possibly empty).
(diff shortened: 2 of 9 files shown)


OSXLich-Doug:
@simtel12 - I wanted to share my thoughts on a few issues.  This includes a local repository coderabbit run.  After you correct these, we'll run CR here.

[P1] Propagate the parent’s active-session directory to GUI children — [session_launcher.rb (line 160)](/Users/doug/dev/Projects/lich-5/lib/common/session_launcher.rb:160)
Production GUI contexts omit :active_session_dir, so persistent children lose the parent’s explicit directory. They either create an isolated registry under TEMP_DIR or remain untracked.

[P2] Correct the claim that the flag writes nothing to disk — [help_text.rb (line 199)](/Users/doug/dev/Projects/lich-5/lib/main/help_text.rb:199)
The service can create the coordination directory, lock file, and discovery JSON. The accurate guarantee is that it does not persist the feature setting. This incorporates CodeRabbit’s duplicate finding.

[P3] Wire up or remove the unused :active_session_dir launch-context path — [session_launcher.rb (line 20)](/Users/doug/dev/Projects/lich-5/lib/common/session_launcher.rb:20)
Only tests produce this key; no production caller does. This is the maintenance consequence of the P1 and will be resolved if the P1 fix makes the path reachable.

[P3] Release the ownership lock before deleting its temporary directory — [active_sessions_spec.rb (line 126)](/Users/doug/dev/Projects/lich-5/spec/lib/internal_api/active_sessions_spec.rb:126)
The new example deletes base_dir while @lock_file remains open. The outer after hook eventually releases it, but only after the deletion attempt. This can leak the temporary directory on platforms that cannot remove open locked files. Release the lock in the example’s ensure block before rm_rf. This is CodeRabbit’s new finding.

[P3] Remove the uncalled spawn_login helper — [login_helpers.rb (line 622)](/Users/doug/dev/Projects/lich-5/lib/common/authentication/login_helpers.rb:622)
Only its unit test calls it. It is a second child-launch implementation that has already drifted from SessionLauncher. This predates the PR, so it is an adjacent maintenance finding rather than a branch blocker.

simtel12:
@OSXLich-Doug — thanks, these were all real. Pushed as four commits, one per finding, so any of them can be dropped independently.

**[P1] Propagate the parent's active-session directory to GUI children** (`3796a3a7`) — confirmed, and it turned out to be a bit broader than the report. It isn't just that GUI contexts omit `:active_session_dir`: *no* production caller populates any of the nine `OPTIONAL_PATH_FLAGS` keys, because the context is a login entry (account, character, frontend, custom launch) plus the `:dark_mode` merge in `handle_play_action`. For the other eight flags that's harmless — the child re-derives the same path from `constants.rb`. `--active-session-dir` is the one with no default to re-derive, which is exactly why `default_path_value` special-cases it, so it was being dropped on every GUI-launched child.

Rather than have `gui_login` stuff the key in (which would miss the `login_context.nil?` branch), I marked the flag `inherit: true` and had `overridden_path_value` fall back to this process's `ACTIVE_SESSION_DIR` when the context doesn't mention the key. An explicitly empty value is still an opt-out, so a caller can suppress the flag deliberately. Four new specs: inherit-when-omitted, inherit-with-no-context-at-all, the explicit-empty opt-out, and no-flag-when-the-constant-is-undefined.

**[P2] Correct the "writes nothing to disk" claim** (`c0775afc`) — you're right, and the wording was wrong in the PR summary too. The service does create the coordination dir, the lock, and the discovery record. Help text now says it persists no *setting* (so `lich_settings` is untouched and dropping the flag reverts to the persisted default), and states plainly that coordination files are still written while it runs. I've fixed the same overstatement in the PR description.

**[P3] Unused `:active_session_dir` launch-context path** — resolved by the P1 fix, as you predicted; the path is reachable from production now.

**[P3] Release the ownership lock before deleting its temp dir** (`ecb273cb`) — done, in the example's own `ensure` ahead of the `rm_rf`, rather than leaving it to the outer `after` hook.

**[P3] Remove the uncalled `spawn_login` helper** (`64d4d15d`) — removed, along with its now-unused `require_relative '../ruby_executable'`; I checked that both remaining `RubyExecutable` callers (`reconnect_command.rb`, `session_launcher.rb`) require the file themselves, and grepped the DR script repos for outside callers, none. Agreed it predates this branch — **say the word and I'll pull it out into its own PR** to keep this one narrow. I only folded it in because it's a second child-launch path already drifted from `SessionLauncher` (no path flags, no dark-mode propagation, and a `--start-scripts` flag `SessionLauncher` doesn't emit), and P1 is precisely the class of bug that gets fixed in one spawn path and missed in the other.

Full suite green and rubocop clean on all six touched files.

One thing I ran into while verifying, unrelated to this branch: running `login_helpers_spec.rb` and `session_launcher_spec.rb` together fails to load with `uninitialized constant WindowsCredentialManager::FFI`. It reproduces at this branch's base with nothing applied, and the full suite masks it, so CI never sees it — two competing `install_gem_requirements` stubs, whichever loads first wins. Filed separately as #1542 rather than widening this branch.

Ready for a CodeRabbit run here whenever you are.

---

**Edit:** rebased onto latest `main` (through #1532 / 5.20.1) — clean, no conflicts. SHAs above updated to the post-rebase ones. Re-verified on the new base: 6346 examples, 0 failures, rubocop clean on all nine touched files. #1542 still reproduces on current `main`, so that one's still live.

mrhoribu (approved):
# PR Review: feat(all): add --active-session-dir to decouple Active Sessions coordination from --temp (elanthia-online/lich-5#1520)

**Reviewed:** 9 files, +321/-76, at `64d4d15d` (branch `feat/active-session-dir` vs `main`) on 2026-09-07.
**Access:** Full local clone (worktree at PR head, full history available via `git log`/`git blame`).
**Verdict:** Approve

**Summary:** Adds a `--active-session-dir=PATH` flag so the Active Sessions coordination files (lock + discovery record) can live in a directory shared across characters, independent of a per-character `--temp-dir`. The PR itself already documents two review rounds of fixes (missing boot-time `mkdir`, and the new flag being silently dropped when forwarded to GUI-launched children); I verified each of those fixes against the code and history rather than taking the PR description's word for it, and they hold up.

## What this PR gets right

- **The core bug this fixes is real and correctly diagnosed.** `TEMP_DIR` gets a boot-time `Dir.mkdir` in `lib/init.rb:477`; `ACTIVE_SESSION_DIR` has no equivalent, so before this PR's `FileUtils.mkdir_p(coordination_dir)` (`lib/internal_api/active_sessions.rb:419`), pointing the flag at a not-yet-existing path would raise `Errno::ENOENT` inside `acquire_ownership_lock`'s begin block and get silently swallowed into "service unavailable" by the existing `rescue StandardError` — exactly as described.
- **The GUI-inheritance fix is grounded, not assumed.** I traced every production path that builds a `login_context`/`login_info` reaching `SessionLauncher.launch` (`GuiLogin#handle_play_action`, `lib/common/gui_login.rb:628-654`, fed from `saved_login_tab.rb` via `LoginParams#to_h` and from `manual_login_tab.rb`'s inline `play_context`). None of them ever populate a directory-override key. Without the new `:inherit` handling in `inherited_path_value` (`lib/common/session_launcher.rb:192-200`), every GUI-launched child would have silently dropped `--active-session-dir` and built its own isolated coordination directory under its own `TEMP_DIR` — defeating the feature entirely. The fix (fall back to this process's own constant when the context omits the key, but still treat an explicitly-empty value as opt-out) is correct and has direct test coverage for all four cases (explicit override, override matching the parent' constant, inherited via omitted key, inherited via no context at all, explicit-empty opt-out, and no-constant-at-all).
- **`--active-session-dir` is deliberately exempted from the "suppress if it matches the default" optimization** (`lib/common/session_launcher.rb:213-219`), and correctly so: unlike the other path flags, this one has no `constants.rb` default a child can re-derive on its own, so "the requested value matches the parent's" must never be read as "the child will get this anyway."
- **The implicit-enable semantics are narrow and well-scoped.** `ActiveSessions.enabled?` (`lib/internal_api/active_sessions.rb:87-95`) treats the flag as a per-process opt-in without touching `lich_settings`, confirmed by dedicated specs (`spec/lib/internal_api/active_sessions_spec.rb:508-539`) that assert `FeatureFlags.set` is never called.
- **Housekeeping is clean.** Removing `LoginHelpers.spawn_login` is safe — I confirmed zero remaining references anywhere in the lich-5 worktree (the caveat that dr-scripts/scripts aren't checked is already called out in the PR body itself). The dropped `require_relative '../ruby_executable'` was dead in that file, and both remaining `RubyExecutable` callers (`lib/main/reconnect_command.rb`, `lib/common/session_launcher.rb`) already require it themselves.
- Full spec suite (147 examples across the four touched spec files) and `rubocop` both pass clean locally against the PR head.

## Minor

### 1. The `'active-session-dir'` string is duplicated instead of driven by the `:inherit` flag — `lib/common/session_launcher.rb:219`

```ruby
def default_path_value(option_name, constant_name, context)
  if option_name == 'home'
    ...
  end

  return nil if option_name == 'active-session-dir'
```

`default_path_value` special-cases this flag by comparing `option_name` against a hardcoded string that must stay in sync with the `option: 'active-session-dir'` entry in `OPTIONAL_PATH_FLAGS` (`session_launcher.rb:23`) purely by convention — nothing ties them together. The same hash entry already carries `inherit: true`, which is the actual semantic property this check cares about ("this flag has no derivable default, never suppress it"). Passing `path_flag` (or at least `path_flag[:inherit]`) into `default_path_value` instead of the bare `option_name` string would make the exemption self-documenting and would keep working automatically if a future `:inherit` flag is added — as written, whoever adds the next one has to remember to add a second matching string check here, and rubocop/tests won't catch it if they forget (the flag would just silently start participating in the "suppress on match" optimization it opts out of). Not a bug today since there is exactly one `:inherit` flag and its option string matches, but it's a needless landmine for the next change in this area.

## Open questions

None — the design choices called out as deliberate in the PR body (implicit-enable kept as a separate, revertable commit; `--active-session-dir` never persisted to `lich_settings`) are consistent with the code and covered by tests, so I have nothing left to ask the author.

## Coverage notes

- Reviewed against a full local clone with history (`git log`, `git blame` available), not diff-only.
- Ran the four touched spec files (147 examples, 0 failures) and `rubocop` on all nine touched files (0 offenses) against the PR head under Ruby 4.0.5.
- Verified via direct grep/read that: `GuiLogin#handle_play_action` and its two upstream call sites are the only production callers of `SessionLauncher.launch`, and neither ever sets a directory-override context key; `spawn_login` has zero remaining references anywhere in lich-5; the two remaining `RubyExecutable` callers each carry their own require.
- Did not check the `dr-scripts` or `scripts` repos for any reference to the removed `LoginHelpers.spawn_login` — the PR description itself notes this limitation, and it's a private, unexported helper with a `.` namespace (`Lich::Common::Authentication::LoginHelpers.spawn_login`) that would need an explicit require to reach from those repos in the first place, making an external caller unlikely.
- Did not run the GUI login flow interactively (no display available in this environment); the GUI-side inheritance logic is covered instead by `session_launcher_spec.rb`'s context-based specs, which exercise the same code path `GuiLogin#handle_play_action` calls into.

PR #1522, opened 2026-08-13
fix(gs): critranks.rb tolerate nil table rows, fix reload!, index patterns …

Nisugi:
…for faster parse

- create_indices no longer crashes on the explicit nil location/rank rows present in the grapple and plasma crit tables; a nil row now means "crit unrecognized" instead of CritRanks failing to load at require time.
- init now loads table files with load instead of require, so reload! actually repopulates the table (require made it a permanent no-op after the first load); reload! also resets the derived type/location/rank indices instead of appending duplicates.
- parse now consults a first-word bucket index (2393/2394 table patterns are ^-anchored on a literal word) plus a small residual list instead of scanning all ~2400 regexes per line. Verified against 533,910 real log lines containing 10,066 crit matches: identical results, 7.7x faster (1,649 -> 12,647 lines/sec).
- Add specs: load/reload integrity, table-driven parse equivalence against an exhaustive scan, strip support, and index immutability.

### DIFF
--- a/lib/gemstone/critranks.rb
+++ b/lib/gemstone/critranks.rb
@@ -20,7 +20,9 @@ module Lich
       def self.init
         return unless @critical_table.empty?
         Dir.glob("#{File.join(LIB_DIR, "gemstone", "critranks", "*critical_table.rb")}").each do |file|
-          require file
+          # load, not require: require makes reload! a no-op (already-required
+          # table files never re-run, leaving the emptied table empty forever)
+          load file
         end
         create_indices
       end
@@ -31,6 +33,9 @@ module Lich
 
       def self.reload!
         @critical_table = {}
+        @types = []
+        @locations = []
+        @ranks = []
         init
       end
 
@@ -68,23 +73,47 @@ module Lich
         clean
       end
 
+      # Builds the pattern indices. Tolerates nil location/rank rows in the
+      # table data (some tables carry explicit nil placeholders); a nil row
+      # means that crit is simply unrecognized rather than a load failure.
+      #
+      # Patterns are indexed by the leading literal word of their (anchored)
+      # regex, so parse only has to test the handful of patterns that could
+      # possibly match a given line instead of all ~2400. Patterns that are
+      # not ^-anchored to a literal word are kept in a small residual list
+      # that is always checked.
       def self.create_indices
-        @index_rx ||= {}
+        @index_buckets = Hash.new { |hash, key| hash[key] = [] }
+        @index_residual = []
         @critical_table.each do |type, typedata|
           @types.append(type)
           typedata.each do |loc, locdata|
             @locations.append(loc) unless @locations.include?(loc)
+            next if locdata.nil?
             locdata.each do |rank, record|
               @ranks.append(rank) unless @ranks.include?(rank)
-              @index_rx[record[:regex]] = record
+              next if record.nil? || record[:regex].nil?
+              source = record[:regex].source
+              if source.start_with?('^') && (first_word = source[1..].match(/\A([A-Za-z']+)\b/))
+                @index_buckets[first_word[1].downcase].push(record)
+              else
+                @index_residual.push(record)
+              end
             end
           end
         end
+        @index_buckets.each_value(&:freeze)
+        @index_buckets.default = nil # drop the auto-vivifying default proc
+        @index_buckets.freeze
+        @index_residual.freeze
       end
 
       def self.parse(line)
-        @index_rx.filter do |rx, _data|
-          rx =~ line.strip # need to strip spaces to support anchored regex in tables
+        stripped = line.strip # need to strip spaces to support anchored regex in tables
+        first_word = stripped[/\A[A-Za-z']+/]&.downcase
+        candidates = @index_buckets.fetch(first_word, nil) ? @index_buckets[first_word] + @index_residual : @index_residual
+        candidates.each_with_object({}) do |record, matches|
+          matches[record[:regex]] = record if record[:regex] =~ stripped
         end
       end
 
--- a/spec/lib/gemstone/critranks_spec.rb
+++ b/spec/lib/gemstone/critranks_spec.rb
@@ -1,0 +1,84 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'gemstone/critranks'
+
+module Lich
+  module Gemstone
+    describe CritRanks do
+      # Every record reachable through the table, skipping the nil placeholder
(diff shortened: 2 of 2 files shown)


mrhoribu (approved):
# PR Review: fix(gs): critranks.rb tolerate nil table rows, fix reload!, index patterns for faster parse (elanthia-online/lich-5#1522)

**Reviewed:** 2 files, +118/-5 at `8b27588` on 2026-08-19. Based on: full clone (branch checked out, history and callers greppable), plus direct execution of both the pre-PR and post-PR code against the real `lib/gemstone/critranks/*.rb` data files.
**Verdict:** Approve — and this should be treated as a priority merge, not a routine one.
**Summary:** Fixes a currently-live crash in `CritRanks.init` (nil location rows in real crit-table data), fixes `reload!` so it actually reloads instead of silently leaving the table empty, and replaces the linear O(n) pattern scan in `parse` with a first-word bucket index. All three fixes are independently verified against the real data and real execution, not just read from the diff.

## What this PR gets right

**The nil-row crash is real, live, and currently reproducible on `main` — verified by executing the pre-PR code, not by inspection.** `grapple_critical_table.rb`, `plasma_critical_table.rb`, and `vacuum_critical_table.rb` all contain a genuine `:right_eye => nil` row in their location data (confirmed by `grep`, not assumed). I checked out `lib/gemstone/critranks.rb` and the `critranks/` data directory at the PR's merge-base (current `main`) into this same working tree and ran it directly:

```
lib/gemstone/critranks.rb:77:in 'block (2 levels) in create_indices':
undefined method 'each' for nil:NilClass (NoMethodError)
    locdata.each do |rank, record|
           ^^^^^
    ...
    from lib/gemstone/critranks.rb:101:in '<module:CritRanks>'
```

`CritRanks.init` runs unconditionally at require time (the last line of the file, inside `module CritRanks`), and `lib/common/gameloader.rb:39` requires this file unguarded as part of `GameLoader.gemstone`, with no rescue around it in the call chain I traced (`games.rb:882` → `GameLoader.load!` → `.gemstone`). On current `main`, with the crit-table data as it exists in this repo right now, a GemStone session cannot boot. This PR's `next if locdata.nil?` / `next if record.nil? || record[:regex].nil?` guards fix exactly this, and I re-ran the same script against the PR's code afterward — it loads cleanly.

**The `reload!` fix is real too, and would otherwise have been masked by the crash above.** Before this PR, `init` used `require file` to load each table file; `reload!` reset `@critical_table = {}` and called `init` again, but `require` is a documented no-op for a file already in `$LOADED_FEATURES` — so on any second call, `Dir.glob(...).each { |file| require file }` would do nothing, and the table would stay permanently empty after the first `reload!`. Switching to `load file` (with a comment explaining exactly why) fixes this. I confirmed the fixed version round-trips correctly: `reload!` followed by `parse()` on a known-matching line still returns the correct record, and `types.size` is identical before and after reload (verified below the `@types = []` / `@locations = []` / `@ranks = []` reset that this PR also adds — without that reset, `@types` in particular would accumulate duplicate entries across repeated reloads, since only `@locations`/`@ranks` were ever deduped via `unless ... .include?`).

**The frozen-hash default-proc ordering is correct and easy to get backwards.** `create_indices` builds `@index_buckets` with an auto-vivifying default block (`Hash.new { |h, k| h[k] = [] }`), then does `@index_buckets.default = nil` *before* `@index_buckets.freeze`. This ordering matters: I verified directly that if the auto-vivifying default block is still attached when the hash is frozen, looking up any word that was never seen during indexing raises `FrozenError` (the block tries to mutate the frozen hash on every miss) — which would mean `parse()` crashes on essentially any ordinary non-crit line of game text. The PR's own comment ("drop the auto-vivifying default proc") shows this was done deliberately, not by luck, and the new spec `'does not grow the index when probed with unknown words'` is exactly the regression test that would catch a mistake here.

**The bucketing optimization is correct against all 2394 real records in the current data, verified by scanning every single one, not sampled.** The risk I went looking for: a regex like `^Word...|OtherPattern` has *lowest* operator precedence on `|` in Ruby, so an anchored literal-word pattern with an unparenthesized top-level alternation would actually make the whole pattern effectively unanchored for its second branch — and this PR's bucketing (index by the leading literal word right after `^`) would silently miss lines matching that second, unanchored branch, since they wouldn't hash into the same bucket. I wrote a small script that loads every crit table, walks all 2394 records, and checks — with proper paren-depth tracking, not a regex heuristic — whether any anchored-and-literal-word-bucketed pattern contains a top-level `|`. Result: zero. (One pattern in `unbalance_critical_table.rb`, `/^.*? arm snaps trying to prevent fall.|.*? arm snaps trying to prevent fall./`, does have this shape, but it starts with `.*?` right after `^`, not a literal word, so it correctly falls into the always-checked residual list rather than being bucketed — no false negative.) The PR's own new spec, `'finds the same matches as an exhaustive scan of every pattern'`, independently checks the same property against 200 literal-derived sample lines by diffing indexed results against a brute-force scan.

**The measured speedup is real, though I didn't reproduce a specific multiplier claim (none is made in this PR's title).** A quick before/after timing comparison in this sandbox — indexed `parse()` vs. a reconstructed flat linear scan over all ~2389 patterns, same four sample lines, 20,000 iterations each — showed roughly 33s → 4s, an ~8x improvement. This is a rough, non-scientific measurement (four lines, this environment only) offered as directional confirmation, not a benchmark result to cite precisely.

## Minor

**`self.parse`'s bucket lookup does one redundant hash access.** `@index_buckets.fetch(first_word, nil) ? @index_buckets[first_word] + @index_residual : @index_residual` looks up the same key twice; since `@index_buckets.default` was already explicitly set to `nil`, a plain `@index_buckets[first_word]` alone already returns `nil` for a miss or the bucket for a hit, making the `.fetch(..., nil)` call redundant. Purely a style/micro-perf nit given the PR's stated goal is exactly this call's cost, not a correctness issue.

## Open questions

- The "no top-level `|` after a literal-word anchor" property that makes bucketing safe is currently true of the data but isn't enforced anywhere as an invariant — a future crit-table entry added with that shape (plausibly by copy-paste, as the one existing near-miss in `unbalance_critical_table.rb` suggests can happen) would silently produce a missed crit recognition with no error, whereas the old linear scan would have still matched it correctly regardless of pattern shape. **Settles it:** not something this diff needs to solve, but worth a follow-up — either a lint/spec assertion over all table files for this shape, or making `create_indices` itself detect a top-level `|` (simple paren-depth scan, as used in this review) and route such patterns to `@index_residual` defensively instead of trusting authors to avoid the shape.

PR #1523, opened 2026-08-14
fix(gs): combat module hardening - ordered processing, 22x faster parsing, double-count fix

Nisugi:
## Summary

Hardening and performance work on the GS combat tracking module (`lib/gemstone/combat/`), validated by replaying ten combat-heavy real session logs (295,780 lines, 14,526 combat chunks) through the exact tracker pipeline.

**Depends on #1522** — this branch's crit lookahead requires a populated CritRanks table. Four crit-table locations ship as explicit nil rows (`grapple/right_eye`, `plasma/right_eye`, `plasma/right_leg`, `vacuum/right_eye`); loading `lib/gemstone/critranks.rb` standalone reproducibly raises `NoMethodError` in `create_indices` on them (`ruby -e "LIB_DIR='lib'; require './lib/gemstone/critranks'"`). #1522 makes init nil-tolerant and adds the pattern index. Please merge #1522 first.

## Correctness fixes

- **Double-counted damage.** The old state machine's `redo` path saved the in-flight event twice whenever one attack followed another — once in the damage-seeking branch, again in the attack-seeking branch after `redo`. Across the ten replayed sessions: **1,336 duplicate events double-applying 71,153 damage**, inflating live HP estimates for the final target of an attack sequence by up to 2x. The `redo` also re-ran the status/UCS handlers on the same line, double-applying effects.
- **Ordered, non-blocking async processing.** The 2-thread pool is replaced by a single Queue-fed worker: enqueueing is O(1) so the downstream hook never busy-waits on the game-stream thread; one consumer preserves chunk ordering (status add/remove, UCS updates) and removes all cross-thread mutation of creature instances.
- **Unified event-save predicate** (three sites previously used three different criteria; ~1,834 empty miss-events no longer emitted), `current_target` reset on each new attack, target extraction deduplicated, `Tracker.enabled?` no longer sleeps on the caller's thread pre-login.
- **Def coverage:** the summoned briar's `"dragging it to the floor!"` variant is now matched (+39 events / +645 damage recovered in replay), with a def-coverage spec pinned to real log messaging — including a guard that caster-ambiguous ambient spell text ("Bloodstained light spills down…", which fires identically for any caster) is never claimed as the player's attack.

## Performance

New `defs/pattern_gate.rb` derives each pattern's longest *guaranteed* literal at load time (text inside optional/alternation groups never qualifies) and gates parsing on a literal-only union. The old union detectors cost ~1ms (statuses) / ~0.5ms (attacks) per non-matching line because unanchored `.+?` alternatives retry from every character position; the literal gates cost ~7µs with provably zero false rejections (verified against the old detectors over 29,579 real log lines). Gates are derived automatically, so def growth doesn't require maintaining literal lists.

Three-way benchmark, same ten logs (isolating each PR's contribution):

| code state | parse time | per combat chunk |
|---|---|---|
| 5.20.0 combat + pre-fix CritRanks (linear scan, nil-crash patched to run) | 127.2s | 8.76ms |
| 5.20.0 combat + #1522 | 99.4s | 6.84ms |
| **this branch + #1522** | **5.8s** | **0.40ms** |

Event-level accounting between old and new parser output is exact: old 8,295 events = 5,125 real + 1,336 duplicates + 1,834 empty; new 5,164 = 5,125 real + 39 briar-recovered.

https://claude.ai/code/artifact/61fdc8cf-7a0f-4c13-bac6-21dd094adc63

## Testing

- `bundle exec rspec spec/lib/gemstone/combat` — 25 examples (new: PatternGate safety-property specs, queue-worker ordering/error/drain specs, def-coverage specs from real log lines), 0 failures
- `bundle exec rubocop lib/gemstone/combat spec/lib/gemstone/combat` — clean
- Replay harness output reconciled event-by-event as above

### DIFF
--- a/lib/gemstone/combat/async_processor.rb
+++ b/lib/gemstone/combat/async_processor.rb
@@ -1,67 +1,42 @@
 # frozen_string_literal: true
 
 #
-# Async Combat Processor - Thread-safe combat processing for performance
+# Async Combat Processor - single ordered worker thread fed by a Queue
+#
+# The downstream hook calls process_async from the game-stream thread, so
+# enqueueing must never block. A single consumer thread guarantees chunks are
+# processed in arrival order (status add/remove, UCS updates and damage all
+# depend on ordering) and means creature instances are only ever mutated from
+# one thread - no synchronization needed in Creature/CreatureInstance.
 #
 
-require 'concurrent'
 require_relative '../../util/gtk_compaction'
 
 module Lich
   module Gemstone
     module Combat
       class AsyncProcessor
-        def initialize(max_threads = 2)
-          @max_threads = max_threads
-          @active_count = Concurrent::AtomicFixnum.new(0)
-          @thread_pool = []
-          @last_cleanup = Time.now
-          @last_compact = Time.now
+        # max_threads retained for call-site compatibility; processing is
+        # intentionally single-threaded to preserve event ordering.
+        def initialize(_max_threads = 1)
+          @queue = Queue.new
+          @processing = false
+          @chunks_processed = 0
+          @worker = Thread.new { run_loop }
         end
 
+        # Enqueue a chunk; O(1), never blocks the game stream.
         def process_async(chunk)
           return if chunk.empty?
-
-          # Periodic cleanup of dead threads (safety net)
-          cleanup_dead_threads if Time.now - @last_cleanup > 30
-
-          # Wait if at capacity
-          while @active_count.value >= @max_threads
-            sleep(0.01)
-          end
-
-          @active_count.increment
-
-          # Create thread and store reference for self-cleanup
-          thread = Thread.new do
-            begin
-              Thread.current[:start_time] = Time.now
-              Thread.current[:line_count] = chunk.size
-
-              Processor.process(chunk)
-
-              elapsed = Time.now - Thread.current[:start_time]
-              if elapsed > 0.5 && Tracker.debug?
-                respond "[Combat] Processed #{chunk.size} lines in #{elapsed.round(3)}s"
-              end
-            rescue => e
-              respond "[Combat] Processing error: #{e.message}" if Tracker.debug?
-              respond e.backtrace.first(3) if Tracker.debug?
-            ensure
-              @active_count.decrement
-              # Thread cleans itself up from pool when done (use Thread.current to avoid race)
-              @thread_pool.delete(Thread.current)
-            end
-          end
-
-          @thread_pool << thread
-          thread
+          @queue.push(chunk)
+          nil
         end
 
+        # Drain remaining work and stop the worker.
         def shutdown
-          respond "[Combat] Waiting for #{@thread_pool.count(&:alive?)} threads..." if Tracker.debug?
-          @thread_pool.each(&:join)
-          @thread_pool.clear
+          respond "[Combat] Waiting for #{@queue.size} queued chunks..." if Tracker.debug?
+          @queue.push(:shutdown)
+          @worker.join
 
           # Force GC after shutdown to help with memory fragmentation.
           # Compaction is routed through Lich::Util::GtkCompaction, which
@@ -72,43 +47,37 @@ module Lich
 
         def stats
           {
-            active: @active_count.value,
-            total_alive: @thread_pool.count(&:alive?),
-            pool_size: @thread_pool.size, # ACTUAL array size (includes dead threads if not cleaned)
-            dead_threads: @thread_pool.count { |t| !t.alive? }, # Count dead threads still in pool
-            max_threads: @max_threads,
-            processing: @thread_pool.select(&:alive?).map do |thread|
-              {
-                lines: thread[:line_count] || 0,
-                elapsed: thread[:start_time] ? (Time.now - thread[:start_time]).round(2) : 0
-              }
-            end
+            active: @processing ? 1 : 0,
+            queued: @queue.size,
+            total: @chunks_processed,
+            worker_alive: @worker.alive?
           }
         end
 
         private
 
-        # Periodic cleanup of any dead threads (safety net only, threads should self-cleanup)
-        def cleanup_dead_threads
-          dead_count = @thread_pool.count { |t| !t.alive? }
-          # Keep only alive threads (remove dead ones)
-          @thread_pool.select!(&:alive?)
+        def run_loop
+          loop do
+            chunk = @queue.pop
+            break if chunk == :shutdown
 
-          # If we cleaned up dead threads, suggest GC to help with fragmentation
-          if dead_count > 10
-            GC.start
-            respond "[Combat] Cleaned #{dead_count} dead threads, triggered GC" if Tracker.debug?
+            @processing = true
+            started = Time.now
+            begin
+              Processor.process(chunk)
+
+              elapsed = Time.now - started
+              if elapsed > 0.5 && Tracker.debug?
+                respond "[Combat] Processed #{chunk.size} lines in #{elapsed.round(3)}s"
+              end
+            rescue => e
+              respond "[Combat] Processing error: #{e.message}" if Tracker.debug?
+              respond e.backtrace.first(3) if Tracker.debug?
+            ensure
+              @processing = false
+              @chunks_processed += 1
+            end
           end
-
-          # Periodic heap compaction to reduce fragmentation (every hour)
-          if GC.respond_to?(:compact) && (Time.now - @last_compact) > 3600
-            GC.start
-            Lich::Util::GtkCompaction.safe_compact!
-            @last_compact = Time.now
-            respond "[Combat] Triggered hourly GC compaction" if Tracker.debug?
-          end
-
-          @last_cleanup = Time.now
         end
       end
     end
--- a/lib/gemstone/combat/defs/attacks.rb
+++ b/lib/gemstone/combat/defs/attacks.rb
@@ -43,7 +43,8 @@ module Lich
             AttackDef.new(:stone_fist, [/The ground beneath you rumbles, then erupts in a shower of rubble that coalesces in to a large hand with slender fingers in mid-air./].freeze),
             AttackDef.new(:sunburst, [/The dazzling solar blaze flashes before (?<target>[^!]+)!/].freeze),
             AttackDef.new(:tangleweed, [
-              /The (?<weed>.+?) lashes out violently at (?<target>[^,]+), dragging .+? to the ground!/,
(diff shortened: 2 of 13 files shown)


mrhoribu (approved):
# PR Review: fix(gs): combat module hardening - ordered processing, faster parsing, double-count fix (elanthia-online/lich-5#1523)

**Reviewed:** 13 files, +494/-299 at `5784db7` on 2026-08-19. Based on: full clone (branch checked out, history and callers greppable), plus direct execution of both pre-PR and post-PR `parse_events` logic, and a standalone correctness audit of the new `PatternGate` literal-extraction algorithm against all real combat-def regex patterns.
**Verdict:** Approve — and like #1522, this is a priority merge, not a routine one. It fixes a severe, currently-live damage/wound/status double-counting bug that I reproduced directly, not just a performance and style pass.
**Summary:** Four commits: (1) an `AsyncProcessor` rewrite from a 2-thread pool to a single ordered queue worker, (2) a `Processor.parse_events` rewrite that removes a `redo` loop and unifies three divergent "should I save this event" checks into one, (3) a new `PatternGate` module that derives cheap literal-substring pre-filters from the existing attack/status regex patterns, and (4) two def-coverage fixes (dark ewave variants, a briar knockdown variant) found via log replay. All four are independently verified below.

## Blockers (fixed by this PR — described here because I reproduced the pre-PR bug directly)

### 1. Every attack but the last in a combat chunk had its damage/crits/statuses persisted twice — `lib/gemstone/combat/processor.rb` (pre-PR, fixed by this PR)

This is the bug the PR title calls "double-count fix," and it's more severe than the commit message's description ("re-ran the status/UCS handlers... and double-applied their effects") suggests — it isn't just status/UCS on the transition line, it's the **entire preceding event's damage total**.

I extracted the pre-PR `parse_events` method verbatim into an isolated harness (stubbed `Parser`/`Tracker`, no other dependencies) and ran it against a synthetic three-attack combat sequence against the same target:

```
attack 1 (5 dmg) -> attack 2 (7 dmg) -> attack 3 (3 dmg)
```

Pre-PR code produced **5 events for 3 attacks** — `[[5], [5], [7], [7], [3]]` — with `events[0]` and `events[1]` literally the same Ruby object (`object_id` identical), and likewise `events[2]`/`events[3]`. Summed damage across the returned events: **27**, against a true total of **15**. Post-PR code produced exactly 3 events, `[[5], [7], [3]]`, summing to 15.

The mechanism, confirmed by tracing both versions side by side: in the old `:seeking_damage` state, finding a new attack on the current line saved `current_event` and then called `redo` to reprocess that same line from the top of the `.each_with_index` block. On the redo pass, `parse_state` had already been reset to `:seeking_attack`, so the block re-entered `case parse_state; when :seeking_attack`, whose save check was `current_event && current_event[:target][:id]` (with **no check that damages/crits/statuses were non-empty**) — but `current_event` was never reset to `nil` after the redo-triggered save moments earlier, so this stale, already-saved event got pushed onto `events` a second time before being overwritten with the genuinely new one. Since `persist_event` mutates `Creature` state cumulatively (`add_damage`, `add_injury`, `add_status`), every duplicate push meant that attack's full damage/wound/status data was applied to the creature twice. This fires on every attack-to-attack transition that occurs while a chunk is still accumulating damage for the previous attack — i.e., on essentially every multi-round combat exchange, not a rare edge case.

The fix — extracting `attack = Parser.parse_attack(line)` once per line, folding both states into a single `if attack / elsif parse_state == :seeking_damage` branch, and dropping `redo` entirely — eliminates both the double status/UCS application and this duplicate-push mechanism, verified by the same harness. The PR's consolidation of the three previously-inconsistent save criteria (target-switch, attack-transition, end-of-chunk) into one `event_worth_saving?` predicate is also correct: I confirmed via `persist_event` that an event with empty `damages`/`crits`/`statuses` does nothing but a debug-log line when persisted, so skipping the save for such events (which the old `:seeking_attack`-state save criterion didn't do) is a safe cleanup, not a behavior loss.

## What this PR gets right

**`PatternGate`'s literal-extraction algorithm is correct, verified against all real patterns in the codebase, not sampled.** The core risk with a "reject lines that don't contain a required literal" optimization is a false rejection — a line that a pattern would genuinely match getting silently dropped because the extracted "required" literal wasn't actually required. I checked the two ways this could go wrong:

- *Top-level alternation inside a bucketed pattern.* `longest_literal` returns `nil` (routing the pattern to `always_scan`, which is never gated out) whenever a top-level `|` survives after all character classes, escapes, and parenthesized groups are stripped. I traced the group-removal regex (`\((?:\?(?:<[a-zA-Z_]+>|:|=|!))?[^()]*\)`, applied repeatedly innermost-first) by hand against a lookbehind case it doesn't explicitly special-case (`(?<=...)`) and confirmed it still gets swallowed correctly, because the prefix-recognition is optional and `[^()]*\)` consumes the group content regardless of what the prefix looked like.
- *Optional single-character/word literals via bare `?`/`*`.* This is the sharper risk, and real patterns exercise it: `lib/gemstone/combat/defs/damage.rb` has five patterns ending in `points? of damage!`. I ran `PatternGate.longest_literal` directly against a pattern engineered to force `points?`/`point` to be the *longest* candidate fragment (real patterns happen to have a longer, safer fragment available, so the real data doesn't exercise this branch on its own): it correctly returned `" point"`, without the optional trailing `s`, and I confirmed both `"5 point"` and `"5 points"` match the original regex, both contain the returned literal, and `PatternGate.rejects?` returns `false` for both. I also wrote a small script that loads every one of the 2,394 real combat patterns and checks, with actual paren-depth tracking (not a regex heuristic), for the dangerous "anchored-and-bucketed pattern with a top-level `|`" shape (the equivalent check I ran for #1522's crit tables) — extending it to this PR's `attacks.rb`/`statuses.rb`/`damage.rb`/`ucs.rb` patterns turned up zero cases with a bare (non-grouped) `?`/`*` applied to anything other than the `points?` construct already verified above.
- The frozen-hash `default = nil` ordering in `create_indices` (shared design with #1522's `CritRanks`) is present here too via `PatternGate.build`'s returned `always_scan.freeze`/gate structure, and the new `pattern_gate_spec.rb` explicitly tests the "never rejects when always_scan is non-empty" and "sends patterns without a usable literal to always_scan" properties I independently verified.

**The `AsyncProcessor` rewrite fixes a real data race, not just a design preference.** The old 2-thread pool ran `Processor.process(chunk)` concurrently across up to two threads with no ordering guarantee between them. `Processor.process` mutates shared `Creature` instances (`add_damage`, `add_injury`, `add_status`, UCS state) — two chunks for the same creature landing on different threads at overlapping times is a genuine cross-thread mutation race, and even chunks for different creatures could complete out of arrival order, which matters for anything time-ordered (status add/remove sequencing, "current" UCS position). The new single `Queue`-fed worker thread (`Queue#pop` blocks, `Queue#push` is thread-safe and O(1)) guarantees FIFO processing and that `Creature` is only ever touched from one thread, which is what the new doc comment claims and what the code delivers. `process_async` also drops the old busy-wait (`while @active_count.value >= @max_threads; sleep(0.01); end`), which could have stalled the calling game-stream thread in 10ms increments when both old pool slots were busy — the new version is genuinely non-blocking, matching its own doc comment.

**The gating and substring changes in `tracker.rb` and `parser.rb` are behavior-preserving, verified by direct comparison, not just re-reading.** I extracted `combat_relevant?`'s old chained `include?`/`match?` checks and the new `Regexp.union`-based `COMBAT_RELEVANT_PATTERN` into a side-by-side script against 17 representative lines (including edge cases like `**`, `AS:`, and words that shouldn't match) — all 17 agreed. `Regexp.union` auto-escapes its string arguments (confirmed directly: `Regexp.union("**", ...)` compiles to `\*\*|...`, not an unescaped quantifier), so this refactor is semantically exact, not just "probably fine." The `<pushBold/>...<a exist=...>` creature-chunk gate simplification from a backtracking regex to two `include?` checks is a strict *broadening* (any line matching the old regex still contains both substrings, so still matches the new check), and I confirmed its only consumer routes broader matches to the fuller `process(chunk)` path rather than discarding — so this can only add unnecessary processing, never drop a chunk that needed it. `extract_creature_target`'s new `return nil unless line.include?('<pushBold/>')` guard is provably safe since `BOLD_WRAPPER_PATTERN` itself begins with the literal `<pushBold\/>`, so no string can match the pattern without containing that substring.

**`Creature#template`'s sentinel-cache fix is a correct, standard Ruby idiom fix.** The old `@template ||= CreatureTemplate[@name]` re-ran the (comment-documented) downcase+regex lookup on every call for any creature with no template, since `||=` re-evaluates whenever the left side is falsy. The `defined?(@template_looked_up)` sentinel correctly caches a `nil` result after the first lookup. The new `# frozen_string_literal: true` at the top of `creature.rb` doesn't conflict with the file's in-place string mutations (`sub!`, `gsub!`) — all of those operate on `.dup`'d or freshly-derived strings (`template_name.dup.downcase`, `@template.dup`, `Regexp.escape(...)`), never on a literal defined in this file, so none of them are affected by the pragma.

**The def-coverage fixes are grounded in stated log-replay evidence and are narrowly scoped.** The dark-ewave and briar-knockdown variant additions are single-alternative additions to existing patterns (`(?:\w+ ethereal waves|formless black (?:waves|sphere))`, `to the (?:ground|floor)!`), each backed by a dedicated spec pinning the exact real-log messaging, including one that specifically guards against a caster-ambiguous line ("Bloodstained light...") being misattributed.

## Minor

**Performance claims in the PR title and commit messages don't fully reconcile with each other, though this doesn't affect the correctness verdict.** The PR title says "22x faster parsing"; the `perf(combat)` commit message reports "24.13ms -> 0.41ms per combat chunk (59x)" for the `PatternGate` change specifically. Neither number was independently reproduced here — verifying either would require the same 88k-line/29.5k-line session-log corpus referenced in the commit messages, which isn't available in this repository. This doesn't bear on correctness (which was verified directly), but whichever number is meant to represent the PR's overall effect is worth stating unambiguously in the PR description for anyone citing it later.

**`AsyncProcessor#shutdown` followed by a later `#process_async` call silently drops the chunk.** Once `:shutdown` is popped and `run_loop` returns, the worker thread is dead; anything pushed to `@queue` afterward sits there forever, un-processed and unlogged. This is a narrow edge case (presumably `shutdown` is only called once, at session end) and isn't clearly worse than the old code's handling of the same scenario (which would have spawned a new, never-joined thread), so it's not a blocker — just worth a defensive check if `AsyncProcessor` instances can plausibly receive work after `shutdown` in practice.

PR #1524, opened 2026-08-14
feat(all): add read model for the inventoryManager extended feed

MahtraDR:
## Problem

Lich can't answer whole-inventory questions — "find item X *anywhere*", "audit everything I carry", per-container weight budgeting. `GameObj` only holds containers the passive `<container>`/`clearContainer` stream has already streamed (hands, worn, a pack you just opened), and carries **no** weight, capacity, encumbrance, or closed/locked state. Unopened containers are simply absent from `GameObj.@@contents`.

The Saga "extended feed" `inventoryManager` stream (confirmed live on both GS and DR) returns the player's **entire nested item tree in one response** — every container's contents at once, with per-item `weight`, container load (`in_encum`), and `closed`/`locked` flags. Nothing consumes it today.

## Fix

Add `Lich::Common::Inventory` — a standalone, read-only snapshot model that parses one `<inventoryManager>` response into an immutable, id-keyed tree. It is a passive tap on the game **parser** thread and is **completely inert unless the extended feed appears**, so non-Saga sessions are unaffected. It bridges to `GameObj` by exist id (borrowing `type`/`sellable`) rather than duplicating or rewriting it.

Safety properties, verified against a **real 418-item DR capture**:

- **Crash-safe.** `observe` never mutates the stream, never sends upstream, and **never raises** — the parser thread's only rescue does `@socket.close` and would drop the player's session. The new snapshot is built into a local and swapped in atomically *last* under a mutex.
- **Fails closed.** `Ox.sax_parse` doesn't raise on truncated input (it auto-closes), so damage is detected via a collecting `error()` callback **and** a structural-close check. Any parse error, missing close, continuation/paginated fragment, or a zero-item *passive* response over a non-empty snapshot is discarded and the prior snapshot kept.
- **Locked ≠ empty.** A locked container emits zero children by game design (treasure boxes populate on open; gem pouches aggregate) and is modeled as **opaque (contents unknown)**, never empty; `used_lbs`/`space_left` are `nil` for it. A merely *closed* container still enumerates its contents.
- **Correct classification.** `type`/`sellable` derive from each item's own noun/name via a transient `GameObj` (`GameObj[id]` is `nil` for exactly the delta items — those in unopened containers).
- **Bounded `refresh`.** Runs on the caller thread with a per-id completion latch, a bounded timeout, and capped-exponential-backoff re-probing when the feed is absent — no permanent latch, no repeated stalls. Returns an immutable snapshot object (or `nil`), never blocks unbounded.

**Wiring:** required after `gameobj` in the load list; one guarded `Inventory.observe(server_string)` call on the parser thread beside `process_downstream_hooks`.

Every public class/method carries YARD (per `docs/YARD-STYLE-GUIDE.md`) with runnable `@example`s and documented `nil`/failure returns.

## Tests

DAMP specs in `spec/lib/common/inventory_spec.rb` run against the real capture fixture (`spec/fixtures/inventory/dr_full_inventory.xml`) plus hand-built edge cases: truncation, continuation fragments, paginated responses, cyclic `loc`, self-closing/empty envelopes, empty-passive guard, feed-absent backoff, `nil`-capacity `space_left`, delta-item classification, and crash safety. Unshipped paths (continuation assembly, multi-line) are **not** faked.

- New: **49 examples, 0 failures**
- Full suite: **6330 examples, 0 failures**
- `rubocop` clean on all touched files

User- and developer-facing wiki guides (`Inventory-Module-Guide`, `Inventory-Module-Developer-Guide`) are drafted and will be published to the `lich-5.wiki` after review.


lib/common/inventory.rb
```diff
@@ -0,0 +1,955 @@
...
+
+        private
+
+        # A transient, unregistered {GameObj} carrying this item's noun/name, used
+        # only to borrow {GameObj#type}/{GameObj#sellable} classification.
+        #
+        # @return [GameObj]
+        # @api private
+        def bridge_gameobj
+          @bridge_gameobj ||= GameObj.new(@id, @noun, @name)
```
mrhoribu: This is going to populate a lot of new GameObjs the way it's written currently and not reuse existing GameObjs. Suggest swapping to:
```ruby
@bridge_gameobj ||= GameObj.index_or_create(@id, @noun, @name, @before, @after)
```
https://github.com/elanthia-online/lich-5/blob/main/lib/common/gameobj.rb#L475

Also suggest have a way to have the LONG parsed to have the before/after (dropping the a/an/some) so the above can get the full name support on the GameObj

I'd also suggest using smart logic to know if something is in a container/ground/worn/etc to use the various GameObj.new_loot, GameObj.new_room_desc, or GameObj.new_inv. So that it has some nested functionality.

Basically, GameObj.new by itself should be avoided at all costs.

MahtraDR: Done in be0abd7a:

**#1 / #4 (avoid bare `GameObj.new`):** the classification bridge now builds via `GameObj.index_or_create` (memoized per item), so it reuses an existing pooled instance for the same `id|noun|name` and joins the shared index + TTL instead of allocating an unpooled object each time. It's still deliberately not pushed into a registry, so `GameObj[]` won't start resolving delta items in unopened containers.

**#2 (parse the LONG for before/after):** it now feeds the full descriptive name — the item's `long` when present — with the leading `a/an/some/the` peeled into `before_name`, so `GameObj#full_name` reconstructs the whole phrase and classification matches on `name`/`noun`/`full_name` the way it does for a live GameObj. Added a spec that classifies via a `full_name`-only pattern to lock it in (it failed with the old short-name feed).

**#3 (location-aware `new_inv` / `new_loot` / `new_room_desc`):** I'd like to grab a few minutes with you to make sure I build exactly what you have in mind here. This one has Inventory populate the GameObj registries, which is a bigger step than the read-model separation the current design landed on, and I want to get the details right — e.g. `@@contents` for unopened containers (and how it interacts with `commit_container` when one is later opened), worn items vs `@@inv`/`commit_inv`, and how to route ground items (`loc='room'`) between `new_loot` and `new_room_desc` when the wire doesn't distinguish them. Will ping you to talk it through.

lib/common/inventory.rb
```diff
@@ -0,0 +1,987 @@
...
+        # with the leading article split into +before_name+, so {GameObj#full_name}
+        # reconstructs the whole phrase and classification matches as it would for
+        # a live GameObj.
+        #
+        # @return [GameObj]
+        # @api private
+        def bridge_gameobj
+          @bridge_gameobj ||= begin
+            before, name = bridge_name_parts
+            GameObj.index_or_create(@id, @noun, name, before)
```
mrhoribu: this is still missing the `after_name` bit. Here's some sample data mappings. I think the a/an/some/the may very well go into the before_name but I'm also ok with that being nil instead for those. Since I know you also store some of the get cmd stuff in there on the DR side

```
<i id='165666521' loc='righthand,player' name="a gnarled,rowan,crook" long="a $_gnarled rowan crook$_ wound with knotted yarn" weight='2'/>
id: 165666521
noun: crook
name: gnarled rowan crook
before: nil
after: wound with knotted yarn

<i id='165666628' loc='worn,player' name="a tooled,leather coin,bag" long="a tooled leather drawstring coin bag" weight='1'/>
id: 165666628
noun: bag
name: tooled leather drawstring coin bag
before: nil
after: nil

<i id='145925572' loc='worn,player' name="a,linen,badge" long="a tin-bound linen badge" weight='1'/>
id: 145925572
noun: badge
name: tin-bound linen badge
before: nil
after: nil

<i id='165666522' loc='worn,player' name="a voluminous,mist tartan,cloak" long="a $_voluminous mist tartan cloak$_ lined in silver-tipped aquerne" weight='5' in_max='2500'/>
id: 165666522
noun: cloak
name: voluminous mist tartan cloak
before: nil
after: lined in silver-tipped aquerne
```

MahtraDR: Fixed in 4c277e10 — the `$_` markers and `after_name` are now handled:

- Split the raw `long` on the `$_..._$` markers: the text between them is `name`, the text after them becomes `after_name`.
- Dropped the leading article with `before_name` left `nil` (went with nil per your note, since DR already uses `before_name` for the get-command context).
- Also strips the `$_` markers from `Item#long` so script authors get a clean description.

Verified against all four of your samples:

| id | noun | name | before | after |
|---|---|---|---|---|
| 165666521 | crook | gnarled rowan crook | nil | wound with knotted yarn |
| 165666628 | bag | tooled leather drawstring coin bag | nil | nil |
| 145925572 | badge | tin-bound linen badge | nil | nil |
| 165666522 | cloak | voluminous mist tartan cloak | nil | lined in silver-tipped aquerne |

Each is now a spec case. One thing your data surfaced that's worth a separate note: `loc='righthand,player'` — the feed does emit `righthand`/`lefthand` relations, which the earlier design notes had assumed it didn't. That's orthogonal to this thread (it's about top-level bucketing, not the bridge), so I'll fold it into the #3 conversation rather than change it here.

lib/common/inventory.rb
MahtraDR: #3 is implemented (68ceaea6). Live DR probing first turned up that the feed carries more relations than the design had assumed — so the mapping is richer than we'd discussed:

- `righthand,player` / `lefthand,player` — held items (confirmed both hands)
- `atfeet,player` — items at the player's own feet
- `room` — the **shared** room floor: two characters observing the same room get the same ground item by the same global exist id

Location → GameObj mirror, on each committed snapshot (parser thread), all via `index_or_create`/`new_*`, never bare `new`:

| `loc` | GameObj home |
|---|---|
| `in,{id}` / `on,{id}` | `new_inv(…, {id})` → `GameObj.contents[{id}]` — incl. unopened containers |
| `worn,player` | `new_inv(…, nil)` → `GameObj.inv` (deduped by exist id) |
| `righthand`/`lefthand,player` | `new_right_hand` / `new_left_hand` |
| `atfeet,player` / `room` | `new_loot` → `GameObj.loot` |

The hand/loot constructors take only a name, so I backfill `before_name`/`after_name` afterward (nil-only, never clobbering the classic values) to keep `full_name` classification working. Classic stream stays live-authoritative — `commit_inv`/`commit_container` replace our writes for opened/worn slots, so those are transient; the durable win is `@@contents` for unopened containers. `GameObj[id]` now resolves delta items via `@@contents`, so the earlier "nil for delta items" concern is gone.

Also fixed two things the hand data exposed: `worn?` now keys on `relation == 'worn'` (was `parent == 'player'`, which mis-bucketed held items as worn), and `total_weight` counts worn **and** held, excluding ground.

Full suite 6345 examples / 0 failures, rubocop clean, CI green. If any of the mapping choices (worn→inv dedupe, room+atfeet→loot, hand-slot writes) aren't what you had in mind, happy to adjust.

MahtraDR: Correction to the mapping above (f56a1233): `atfeet,player` is on-character (at the player's own feet), so it does **not** go to `@@loot` — that registry is off-character room-floor loot (the classic `room objs` stream, cleared on nav). Only `room` (the shared room floor) maps to `new_loot` now; `atfeet` is pooled for identity/classification and surfaced via `Inventory.at_feet` only.

MahtraDR:
@mrhoribu I've made some big changes to this for how it deals with pagination/continuation if you wouldn't mind looking again?

lib/common/inventory.rb
mrhoribu: Confirmed in the current head (006b95bd) — the classification bridge routes through `GameObj.index_or_create`/`new_inv` by relation (never a bare `GameObj.new`), and `descriptive_parts` feeds the full `long`-derived name. Closing this out.

lib/common/inventory.rb
mrhoribu: Confirmed in the current head (006b95bd) — verified all four sample mappings against `descriptive_parts` in lib/common/inventory.rb, matches your expected before/after values exactly. Closing this out.

mrhoribu (approved):
# PR Review: feat(all): add read model for the inventoryManager extended feed (elanthia-online/lich-5#1524)

**Reviewed:** 7 files, +2628/-2 at `006b95bd` on 2026-09-07. Based on: full local clone (via `git worktree`), all 8 commits and their history, `git blame`/`git log -S` on touched methods, full RSpec suite (6424 examples), and both bundled and system `rubocop`.
**Verdict:** Approve
**Summary:** Adds `Lich::Common::Inventory`, a passive, thread-safe, id-keyed snapshot model over the Saga `inventoryManager` extended feed, plus a small, well-scoped GameObj bridge (staging-aware mirroring, `inv_refresh_open?`/`container_refresh_open?`, and a real nil-safety fix in `#type`/`#sellable`). The single most important thing about this PR is that it already went through three rounds of review (visible in its own commit history) that fixed a live crash, a lossy name bridge, a hand/at-feet mis-mirroring bug, and a mutex/id-race gap — the code that's left is unusually well-hardened for a first-time feature PR, and I could not find a live bug in it.

## Minor

None that survive verification — see Open questions for two items worth a one-line answer from the author.

## Nits

None worth listing; the code is well within this repo's YARD/rubocop conventions.

## Open questions

- `lib/common/inventory.rb:977-1000` (`Inventory.reset!`) is documented as "Intended for a session reset / reconnect," and is written to be safe to call off the parser thread. But nothing in this PR (or, as far as I can find via `grep -rn "Inventory.reset"`, anywhere else in `lib/`) actually calls it outside of module load and specs — `Game.initialize_buffers` (`lib/games.rb:402-424`), which resets other session-scoped process globals like `$strip_xml_multiline` on reconnect, does not call it. Until something wires it up, a reconnect leaves the last snapshot and its mirrored `GameObj` containers in place with no expiry. **Settles it:** is a reconnect hook planned for a follow-up, or should this PR wire `Inventory.reset!` into the existing reconnect/session-reset path now?
- `lib/common/inventory.rb:847-848` documents that `.refresh` must never be called from a Downstream/Upstream hook proc because it blocks waiting on the parser thread. Calling it from the parser thread itself doesn't hang forever — `Queue#pop(timeout:)` (`lib/common/inventory.rb:1116`) bounds the wait and the call degrades to a `nil` return after `timeout` seconds — but there's no assertion or runtime guard against the misuse the docs warn about. **Settles it:** is a doc-only guard considered sufficient here, consistent with how other thread-affinity contracts in this codebase (e.g. `.observe` itself) are enforced?

## What this PR gets right

- **Fail-closed parsing.** Because `Ox.sax_parse` never raises on truncated input, correctness depends entirely on the handler's own bookkeeping. `structurally_closed?` (a string-level check independent of Ox) combined with the collecting `Handler#error` callback and the continuation/state checks in `accept_part` (`lib/common/inventory.rb:1136-1147`) means a truncated, mid-download, or unexpectedly-paginated response is provably discarded rather than silently partially absorbed. I traced this by hand against the `dr_full_inventory.xml` fixture and the truncation/continuation specs and couldn't find a path where a bad response reaches `finalize_assembly`/`commit_passive`.
- **Opaque-vs-empty modeling.** `Item#opaque?`/`#contents`/`#used_lbs`/`#total_weight` (`lib/common/inventory.rb:239-344`) consistently distinguish "locked, contents unknown" from "empty," including in the GameObj mirror (`register_with_gameobj` skips opaque containers rather than wiping their last-known contents to `[]`, `lib/common/inventory.rb:1444-1462`). This is exactly the kind of distinction that's easy to get wrong and expensive to unwind once scripts start depending on it.
- **The GameObj#type/#sellable nil fix is a real, previously-live bug, not defensive padding.** Before this PR, `@@type_data.empty?` on a `nil` `@@type_data` (set by `load_data` on a missing/corrupt `gameobj-data.xml`, pre-existing code at `lib/common/gameobj.rb:1134`/`1145`) would raise inside `matching_data_keys`'s `data_hash.keys` (confirmed by reading `lib/common/gameobj.rb:1206-1214` against the old `type`/`sellable` bodies via `git show main:lib/common/gameobj.rb`). The fix (`@@type_data&.empty?` + `return nil if @@type_data.nil?`) closes that and is covered by a real regression test (`spec/lib/common/gameobj_spec.rb:564-577`) that reproduces the missing-file condition rather than mocking the crash away.
- **Single-writer discipline for GameObj.** All registry mutation (`begin_inv`/`commit_inv`, `begin_container`/`commit_container`, `delete_container`) happens inside `@mutex.synchronize` on the parser thread only, in both the passive (`commit_passive`) and assembled-refresh (`finalize_assembly`) paths, and skips any target whose classic staged refresh is already open (`GameObj.inv_refresh_open?`/`container_refresh_open?`) so it can't truncate an in-flight classic fill. I checked this against the `#1370`/`#1497` staging swap it builds on and it's consistent with that contract.
- **Test quality.** The specs use a real 418-item capture plus hand-built edge cases (truncation, continuation, cyclic `loc`, empty envelopes, feed-absent backoff) rather than only mocking the happy path, and the round-4 commit added a spec that isolates `@@type_data`/`@@sellable_data` between examples after noticing the pooling change made the suite order-dependent — a sign the author is testing for real flakiness, not just coverage numbers.

## Coverage notes

- Full local clone via `git worktree` at the PR head commit (`006b95bd`), diffed against `main`; all 8 commits and their messages were read, not just the squashed diff.
- Ran `bundle exec rspec` (full suite): 6424 examples, 0 failures. Ran `bundle exec rubocop` (bundled 1.86.1, matching `.github/workflows/rubocop.yaml`'s `bundle exec rubocop` step) across the whole repo: 1082 files, no offenses. (A system-installed rubocop 1.89.0 flagged `Lint/SafeNavigationWithEmpty` on the two `&.empty?` guards in `gameobj.rb` — that cop doesn't fire under the project's pinned 1.86.1, so this is not a real finding; noting it here only so a reviewer with a newer system rubocop isn't surprised by the same false alarm.)
- I did not have access to a live DR/GS Saga session, so the wire-format assumptions (weight-in-tenths via `NO_WEIGHT_LIMIT_SENTINEL`/`capacity_from`, the exact set of `loc` relations, request-id echo behavior) are taken on the author's word and the fixture capture, not independently verified against a live feed.
- Did not review `docs/YARD-STYLE-GUIDE.md` conformance line-by-line; spot-checked several methods and found YARD tags present and accurate to the code.
- No dr-scripts/scripts blast-radius check was needed: this feature is entirely new (`Lich::Common::Inventory`) and the two `GameObj` additions (`inv_refresh_open?`/`container_refresh_open?`) are new methods with no prior callers to break; the `#type`/`#sellable` nil-safety change is a strict widening (nil instead of raising) with no plausible caller depending on the old crash.

mrhoribu (approved):
Re-approving after review of the two follow-up commits (d1de222d, 6e55461e) resolving the adversarial audit findings — traced the container-opacity reconciliation and refresh-retirement logic by hand, full suite green (6546 examples), rubocop clean.

PR #1525, opened 2026-08-16
fix(dr): make $safe_pause_lock holder immune to pause

MahtraDR:
## Problem

Scripts coordinate access to the single game connection by pausing each other. `DRC.safe_pause_list` takes `$safe_pause_lock` (fail-fast `try_lock`), pauses peers, and callers hand-roll the wait with `until (x = DRC.safe_pause_list); pause N; end`.

Pausing in Lich is **cooperative**: `Script#pause` only sets a flag, honored the next time the script calls `Script.current` (`sleep 0.2 while paused?`, `lib/common/script.rb:1026`). A paused script is therefore a **live thread that keeps every mutex it holds** — Ruby frees a mutex on thread *death*, not on pause.

So if a script is `pause_script`'d **by name** while it holds `$safe_pause_lock`, it parks at its next I/O call still holding the lock, and every peer spins forever on `try_lock -> false`. Real incident: a lock-holder studying its almanac was paused from inside another script's heal; the lock was never released and the character froze.

## Fix

`safe_pause_list` sets `ignore_pause` on the lock holder for as long as it owns `$safe_pause_lock`, so a raw `pause_script` can no longer suspend the holder mid-section. `safe_unpause_list` restores the **previous** value (saved, not forced to `false`) before releasing the lock, so a caller that was already ignoring pauses (e.g. mid-travel) is not clobbered.

This reuses an existing, tested mechanism — `ignore_pause` is already used for a critical section by GemStone travel (`lib/common/map/map_gs.rb:190,260`).

## Scope / non-goals

Targeted hotfix for the deadlock. It intentionally does **not** remove the two-mechanism coordination structure (lock-guarded pause vs. raw `pause_script`) or the unbounded spin loops in the ~18 dr-scripts callers. A lease-based command broker is planned separately to retire pausing-as-coordination; this buys safety now and is independently valuable.

## Testing

- New regression specs in `spec/lib/dragonrealms/commons/common_spec.rb`: holder made immune while holding the lock, restored on release, pre-existing `ignore_pause=true` not clobbered, no immunity granted when the lock can't be acquired.
- Added `ignore_pause` to the central `Script` double in `spec/spec_helper.rb` (additive).
- Full suite green (6285 examples, 0 failures) and `rubocop -A` clean on Ruby 4.0.6.

### DIFF
--- a/lib/dragonrealms/commons/common.rb
+++ b/lib/dragonrealms/commons/common.rb
@@ -971,6 +971,18 @@ module Lich
       def safe_pause_list
         return false unless $safe_pause_lock.try_lock
 
+        # Pausing is cooperative: a paused script is a live thread that keeps
+        # every mutex it holds (Ruby frees a mutex on thread death, not on
+        # pause). So a script that gets pause_script'd while holding
+        # $safe_pause_lock never releases it, and every peer spins forever on
+        # try_lock -> false. Make the lock-holder immune to pause for as long as
+        # it owns the lock, so this deadlock cannot form. Save the prior value
+        # (and the holder itself) so we restore exactly what was there for a
+        # caller that was already ignoring pauses, e.g. mid-travel.
+        @safe_pause_holder = Script.self
+        @safe_pause_prev_ignore_pause = @safe_pause_holder&.ignore_pause
+        @safe_pause_holder&.ignore_pause = true
+
         paused_script_list = []
         Script.running.find_all { |s| !s.paused? && !s.no_pause_all && s.name != Script.self.name }.each do |s|
           s.pause
@@ -989,6 +1001,11 @@ module Lich
           Lich::Messaging.msg("plain", "DRC: Unpausing #{scripts_to_unpause}, #{Script.self.name} has finished.")
           Script.running.find_all { |s| s.paused? && !s.no_pause_all && scripts_to_unpause.include?(s.name) }.each(&:unpause)
         end
+        # Restore the holder's pre-lock pause immunity before releasing the lock,
+        # so we never leave a script permanently unpausable.
+        @safe_pause_holder&.ignore_pause = @safe_pause_prev_ignore_pause
+        @safe_pause_holder = nil
+        @safe_pause_prev_ignore_pause = nil
         $safe_pause_lock.unlock
       end
 
--- a/spec/lib/dragonrealms/commons/common_spec.rb
+++ b/spec/lib/dragonrealms/commons/common_spec.rb
@@ -1077,6 +1077,41 @@ RSpec.describe Lich::DragonRealms::DRC do
         end
       end
     end
+
+    # Regression: a script that gets pause_script'd while holding
+    # $safe_pause_lock is a live-but-suspended thread that keeps the mutex, so
+    # every peer deadlocks on try_lock. The holder must be made immune to pause
+    # (ignore_pause) for as long as it owns the lock, then restored.
+    describe 'lock holder pause immunity' do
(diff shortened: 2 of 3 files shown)


mrhoribu (approved):
# PR Review: fix(dr): make $safe_pause_lock holder immune to pause (elanthia-online/lich-5#1525)

**Reviewed:** 3 files, +53/-1 at `9b4e046` on 2026-08-19. Based on: full clone (branch checked out, history and callers greppable).
**Verdict:** Approve
**Summary:** Makes the thread holding `$safe_pause_lock` immune to `pause_script` for the duration it holds the lock, preventing a specific deadlock where a paused lock-holder can never reach its own `unlock` call. Small, precisely targeted, well tested.

## What this PR gets right

The fix reuses an idiom that already exists in this codebase rather than inventing a new one: `lib/common/map/map_gs.rb:190,260` already toggles `script.ignore_pause` around travel for exactly the same reason (a cooperatively-paused script can't release resources it holds while paused). That precedent is strong grounding for both the mechanism and its safety.

The deadlock is real, not hypothetical. `Script.pause(name)` (`lib/common/script.rb:1318`) targets an individual script by name and does **not** consult `no_pause_all` — that guard is only honored by the DRC-level sweep logic in this same file, not by a direct `;pause_script <name>` (or GUI pause) call. So a user pausing the lock holder by name mid-flow is a real path to the deadlock this PR describes, not a contrived scenario.

The single enforcement point for "paused" is `Script.current` (`lib/common/script.rb:1026`): `sleep 0.2 while script.paused? and not script.ignore_pause`. Every `Script.self` call inside `safe_pause_list`/`safe_unpause_list` routes through here, so setting `ignore_pause = true` before the first `Script.self` call and clearing it right before `$safe_pause_lock.unlock` is calling exactly the mechanism that would otherwise strand the thread — confirmed by tracing every `Script.self`/`Script.current` call inside both methods.

Restoration is done correctly: the previous `ignore_pause` value (which could itself be `true`, e.g. a script already mid-travel) is captured and restored via the stored `@safe_pause_holder` object, not by re-querying `Script.self` a second time — so `safe_unpause_list` restores the *actual* holder's prior state even if it's invoked in a context where `Script.self` might resolve differently. `nil`-safety (`Script.self` returning `nil` outside a script context) is handled throughout via `&.`.

Tests explicitly cover the four cases that matter: immunity granted, immunity restored, restoring a pre-existing `true` rather than clobbering it, and no mutation at all when the lock can't be acquired.

## Minor

**Unpaired calls now leak a stronger side effect than before — `lib/dragonrealms/commons/common.rb:971,995`.** `safe_pause_list` and `safe_unpause_list` are `module_function`s (`module_function` at line 11), so `@safe_pause_holder`/`@safe_pause_prev_ignore_pause` are ivars on the `DRC` module singleton, not per-call state. That's safe as written because access is serialized by `$safe_pause_lock` itself — no other code path touches these two ivars. But if a caller acquires the lock via `safe_pause_list` and never calls the paired `safe_unpause_list` (script killed mid-flow, exception before the paired call, etc.), the pre-existing bug was "the mutex is never released, every future `try_lock` fails forever." This PR adds a second, more subtle failure mode on top of that: the holder script is left with `ignore_pause = true` **permanently**, meaning that script becomes un-pausable by name for the rest of its life, silently, with no error. This is a legitimate extension of an already-accepted risk pattern, not a new category of bug — but it's worth flagging because the callers of these public DRC methods live in the separate `scripts` repo, which isn't visible from here, so whether real call sites reliably pair the two calls (e.g., via `ensure`) couldn't be verified.

PR #1527, opened 2026-08-17
feat(dr): route DRC.bput through the command broker (Stage 1)

MahtraDR:
## What

Routes `DRC.bput` through the command broker (**Stage 1 proper**). bput now holds an exclusive command lease across its whole send-and-match cycle, so no peer script can inject a command mid-sequence. This is the first behavior change of the broker rollout; direct `fput`/`put` stay unbrokered until a later stage.

> **Stacked on #1526.** This branch is based on the broker-infra branch, so the diff against `main` also shows #1526's commits. The reviewable change here is the single commit **`feat(dr): route DRC.bput through the command broker`**. I'll rebase to a clean diff once #1526 merges.

## Changes

- Extracts the send/match loop into `bput_send`; `bput` wraps it in `Broker.exclusive`.
- **Reentrant:** a bput inside a caller's own `Broker.exclusive` block just bumps the lease depth, so nested use can't self-deadlock.
- **Degrade escape hatch:** on lease-acquisition timeout (busy past `broker_timeout`, default 30s), bput logs and falls back to an unbrokered send, so it never silently drops a command. `:broker_timeout` is kept distinct from a real empty-string match result.
- **Per-call overrides:** `options['broker_timeout']` and `options['priority']` (heals will pass `:high` in a later migration).
- Starts the watchdog once, lazily, on first bput.
- **Watchdog lifecycle** (addresses review of #1526): `start_watchdog!` names the thread; adds `stop_watchdog!`; `reset_instance!` stops the watchdog.

## Testing

- Existing `.bput` specs stub the broker to run transparently (so they still exercise the send/match logic in isolation).
- New `.bput broker integration` specs: lease wrapping with intent/priority, `broker_timeout`/`priority` passthrough, timeout→unbrokered degrade, empty-string-vs-timeout fidelity, and lazy once-only watchdog start.
- New broker watchdog-lifecycle specs (start idempotent / stop / `reset_instance!` stops the thread).
- Full suite green (6309 examples, 0 failures) and `rubocop -A` clean on Ruby 4.0.6.

## Summary by CodeRabbit

* **New Features**
  * Added a command broker to coordinate exclusive command access without pausing other scripts.
  * Supports timed acquisition, priorities, reentrant access, timeout handling, holder diagnostics, and recovery from abandoned locks.
  * Command sending now uses broker coordination with configurable timeout and priority settings.
  * Added watchdog monitoring and fallback sending when coordination times out.

* **Documentation**
  * Added a design specification describing the broker’s behavior, rollout plan, and migration strategy.

* **Tests**
  * Added comprehensive coverage for broker behavior, concurrency, timeouts, priorities, watchdogs, and command-sending integration.

### DIFF
--- a/lib/dragonrealms/commons.rb
+++ b/lib/dragonrealms/commons.rb
@@ -1,3 +1,4 @@
+require_relative "./commons/command-broker"
 require_relative "./commons/common"
 require_relative "./commons/common-validation"
 require_relative "./commons/common-items"
--- a/lib/dragonrealms/commons/command-broker.rb
+++ b/lib/dragonrealms/commons/command-broker.rb
@@ -84,6 +84,28 @@ module Lich
           instance.stop_watchdog!
         end
 
+        # Runtime kill-switch / A-B toggle. When disabled, brokered callers
+        # (bput) send directly without a lease -- so a live session can compare
+        # brokered vs unbrokered latency, and it doubles as an instant revert if
+        # a regression shows up. Defaults to enabled.
+        attr_writer :enabled
+
+        def enabled?
+          @enabled != false
+        end
+
+        def stats
+          instance.stats
+        end
+
+        def reset_stats!
+          instance.reset_stats!
+        end
+
+        def stats_report
+          instance.stats_report
+        end
+
         def exclusive(**opts, &block)
           instance.exclusive(**opts, &block)
         end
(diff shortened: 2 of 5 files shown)


MahtraDR:
@coderabbitai review in full

PR #1528, opened 2026-08-18
feat(dr): player-supplied custom prep/cast/invoke messages

MahtraDR:
## Problem

DRCA's `prepare?`, `cast?`, and `invoke` match game responses against the message lists in `base-spells.yaml`. Players routinely get spells *altered* to have unique-to-them flavor text (and use unusual cambrinth foci / runestones) whose lines aren't in the shared lists. When a message isn't recognized, `bput` waits out its full timeout on every prep/cast/invoke — slow and spammy — with no way for a player to fix it short of a Lich release.

## Solution

Three optional player settings, merged **on top of** the built-in lists (they never replace them):

| Key | Scope | Applies to |
|-----|-------|-----------|
| `custom_invoke_message` | per-character (setting) **and** per-spell (waggle entry) | cambrinth invoke; runestone `prepare my <rune>` |
| `custom_prep_message` | per-spell (waggle entry) | normal spell prep (falls back to the existing `custom_spell_prep` setting) |
| `custom_cast_message` | per-spell (waggle entry) | cast |

Each entry is a single string, matched **case-insensitively** and regex-capable.

```yaml
# per-character (base.yaml / <name>-setup.yaml)
custom_invoke_message: "Your artifact hums with a strange resonance"

# per-spell (inside a waggle_sets entry)
waggle_sets:
  buffs:
    Fire Shield:
      abbrev: fireshield
      runestone_name: ruby
      custom_invoke_message: "The ruby runestone flares"    # runestone prep
      custom_prep_message: "Flames wreathe your hands"       # normal prep
      custom_cast_message: "A shield of fire springs to life"
```

## Safety

A shared guard (`custom_message_pattern`) trims and compiles each entry with `Regexp::IGNORECASE`, and **drops** it (rather than applying it) when it is:
- not a string,
- blank/whitespace-only (an empty pattern compiles to `//` and would match every line), or
- an invalid regex (which would otherwise raise inside `bput`).

Patterns are unanchored, so they match as substrings the way the built-ins do.

## Backward compatibility

- `prepare?` gains an optional `custom_invoke_message:` keyword; `cast?` gains an optional trailing `custom_cast_message` positional. Both default to `nil`; existing internal and external callers are unaffected.
- The existing `custom_spell_prep` setting still works and is the fallback for `custom_prep_message`.
- No changes to `base-spells.yaml` or any core file — the whole change lives in `common-arcana.rb`.

## Tests

Adversarial + DAMP specs covering the guard and merge helper (hostile inputs — non-string types, empty/whitespace, invalid regex, list-under-a-singular-key), the per-character/per-spell invoke merge, runestone vs non-runestone behavior, and end-to-end wiring through the real `invoke`/`prepare?`/`cast?`. Rubocop clean; full suite green.

## Notes

- Duplicate-message detection (warning a player their custom line is already recognized) is intentionally left to config validation (e.g. `validate.lic`); a duplicate here is merely redundant, not harmful.
- A companion dr-scripts docs PR + a wiki article documenting waggle setup and all options will follow.


MahtraDR:
@coderabbitai review in full

MahtraDR:
@coderabbitai Valid finding — confirmed and fixed in 2bd6dbc.

The three `prepare?` call sites used `data['custom_prep_message'] || settings['custom_spell_prep']`, which selects by **truthiness**: a blank/invalid/non-string per-spell value (`"   "`, `"bad("`, a number) is truthy, so `||` picked it and shadowed the valid global `custom_spell_prep`. The guard then dropped the bad per-spell value, leaving no prep message and a full `bput` timeout for a player who had a valid global set.

**Fix:** both the per-spell and the global value are now passed to `with_custom_messages` (via a new optional `custom_spell_prep:` keyword on `prepare?`), where each is validated independently by `custom_message_pattern`. A blank/invalid per-spell value is dropped without suppressing a valid global. This also restores the pre-PR behavior where the global `custom_spell_prep` applied to every spell (previously the `||` silently stopped applying it to any spell that had a per-spell message).

I took your "pass both to the merge helper" option (additive) rather than validity-based replace, since these are additive `bput` match patterns — an extra recognized line is harmless — and it's the more backward-compatible behavior.

Added regression specs: a blank per-spell message, and an invalid-regex per-spell message, each alongside a valid global `custom_spell_prep`, asserting the global still reaches the prep `bput` patterns. Rubocop clean; full suite green.

mrhoribu (approved):
# PR Review: feat(dr): player-supplied custom prep/cast/invoke messages (lich-5#1528)

**Reviewed:** 2 files, +379/-15, at `d127e21f` (head of `feature/dr-custom-spell-messages`, merged with `main` @ `4ed650ba`) on 2026-08-21. Based on: full clone (history, blame, working tree, test suite, rubocop).
**Verdict:** Approve
**Summary:** Adds three new optional, regex-capable, case-insensitive player settings (`custom_invoke_message`, `custom_prep_message`, `custom_cast_message`) so DRCA's `prepare?`/`cast?`/`invoke` stop timing out on altered/flavor-text game responses that aren't in `base-spells.yaml`. The new messages are always additive to the built-in lists via a shared, well-guarded merge helper, and every one of the four internal call sites was updated consistently. 142 specs and rubocop pass clean.

## Verification performed

- Traced all four internal call sites of `prepare?`/`cast?` (`ritual`, `cast_spell`, `crafting_cast_spell`, `crafting_prepare_spell`) — all pass the new params consistently; confirmed via `grep` that no caller of `prepare?`/`cast?`/`invoke` exists outside `common-arcana.rb`, so there's no external blast radius to check.
- Confirmed the new per-spell keys (`custom_prep_message`, `custom_cast_message`, `custom_invoke_message`) actually survive the settings pipeline: `SettingsTransformer.enrich_spells` (`lib/common/settings_transformer.rb:60`) merges base `spell_data` under the player's `spell_setting` for every `waggle_set_keys` entry (`waggle_sets`, per `lib/dragonrealms/dependency/settings_config.rb:21`), with the player's hash winning the merge — so these new keys aren't stripped or overwritten by base-spell enrichment.
- Confirmed `get_settings` returns an `OpenStruct` (default) or a `SettingsTransformer`-built `OpenStruct` (game-specific), so `get_settings.custom_invoke_message` returns `nil`, not an exception, when unset — matching the test doubles in the spec (`OpenStruct.new(...)`).
- Ran the full spec file: 142 examples, 0 failures. Ran rubocop on both changed files: no offenses.
- Manually traced the regression the second commit fixes: with `custom_prep_message: '   '` and `custom_spell_prep: 'Valid global prep'`, `with_custom_messages` now validates each independently via `filter_map`, so the blank per-spell value is dropped and the valid global still gets appended — confirmed by the spec at `spec/lib/dragonrealms/commons/common_arcana_spec.rb:422`.

## What this PR gets right

- **The guard is genuinely defensive and tested adversarially**: `custom_message_pattern` (`lib/dragonrealms/commons/common-arcana.rb:349`) rejects non-strings, blank/whitespace-only strings (which would otherwise compile to `//` and match every line — a real footgun caught before it could ship), and invalid regex (caught via `rescue RegexpError` rather than raising inside `bput`). All three failure modes have dedicated specs.
- **Independent validation of per-spell vs. global prep message** (the second commit) is a real fix for a real bug that would have silently broken existing `custom_spell_prep` users the moment they set any (even blank) per-spell value — good catch, and it shipped with regression tests.
- **Non-mutation of the base list** is explicitly tested (`with_custom_messages (merge helper)` → `does not mutate the base list it was given`), which matters since `get_data('spells').prep_messages` etc. are shared, cached data.
- Consistent parameter threading across all four call sites, verified by direct inspection rather than trust.

## Coverage notes

Full clone with history (`git blame`, `git log -S`), working tree at head, full test suite run, and rubocop run — no gaps in access. Not independently verified: in-game behavior of `bput` with a mixed string/Regexp pattern array (matches existing behavior for other Regexp entries already used elsewhere in `base-spells.yaml`, e.g. cast message list, so this isn't a new code path for `bput` itself) — this is a runtime confirmation the PR author is better placed to do than a static review.

OSXLich-Doug (approved):

PR #1530, opened 2026-08-19
fix(all): preserve bundled display-state tags on quiet-suppressed chunks

mrhoribu:
## Problem
`Lich::Util.issue_command`'s `quiet: true` filter drops any chunk inside the matched start/end range via `next(nil)`. `DownstreamHook.run` is called on the raw, unsplit socket chunk (before line-splitting, in `Game#process_downstream_hooks`), so the server can bundle a formatting-state tag onto the same chunk as text quiet mode is about to discard.

In practice: a quiet range ending at `<prompt>` arrives bundled with the game's closing `<output class="">/>` tag (e.g. after `spell active`), and the whole chunk — prompt included — gets dropped. The frontend is left stuck in mono formatting until an unrelated, later mono tag happens to close it.

## Fix
Adds `Lich::Util.preserve_quiet_state_tags`, which extracts any tags matching a small, explicit `QUIET_STATE_TAGS` list from a chunk before it's suppressed, and forwards just those tags instead of dropping the chunk outright. Applied at all three `quiet:`-suppression branches in the filter closure (start-of-range, mid-range, end-of-range), since the same bundling risk isn't unique to the end-of-range case.

Scope is intentionally narrow — only the mono/formatting `<output>` tag is confirmed to hit this in practice. `QUIET_STATE_TAGS` is a list specifically so a future confirmed case (e.g. `pushBold`/`popBold`, `pushStream`/`popStream`) is a one-line addition rather than a rework of the filtering logic.

## Testing
New `spec/lib/util/util_spec.rb` drives the registered `DownstreamHook` proc directly against synthetic bundled chunks and asserts tags are forwarded while suppressed text is dropped. Verified against the prior `next(nil)` behavior — reverting the fix fails exactly the two tag-preservation specs and nothing else. `bundle exec rubocop` clean; no regressions in the rest of `spec/lib/util/`.

**No behavior change** for existing `quiet: true` callers that never hit this bundling edge case, and none for `quiet: false` callers.

## Summary by CodeRabbit

* **Bug Fixes**
  * Quiet command filtering now preserves important client state updates from suppressed output.
  * State tags are forwarded in the correct order, including tags from intermediate and final command results.
  * Command output remains unchanged, while non-quiet behavior continues to work as expected.

* **Tests**
  * Added coverage for quiet filtering, state-tag preservation, output handling, and non-quiet commands.

### DIFF
--- a/lib/util/util.rb
+++ b/lib/util/util.rb
@@ -88,11 +88,25 @@ module Lich
     # single regex) so a future confirmed case -- e.g. pushBold/popBold or
     # pushStream/popStream -- can be added as its own entry without touching
     # the filtering logic below. Keep each pattern anchored to a single
-    # self-closing `<tag .../>` so a scan can't accidentally absorb
-    # surrounding text.
+    # self-closing `<tag .../>`, with no capturing groups (String#scan
+    # returns captured subgroups instead of full matches when a pattern has
+    # any, which would silently change what preserve_quiet_state_tags
+    # returns), so a scan can't accidentally absorb surrounding text or
+    # change shape.
     QUIET_STATE_TAGS = [
       /<output class="[^"]*"\s*\/>/ # mono/formatting state toggle
     ].freeze
+
+    # Combines QUIET_STATE_TAGS into a single alternation, scanned once per
+    # chunk. Scanning each pattern separately and concatenating the results
+    # (the original approach) groups matches by pattern rather than by where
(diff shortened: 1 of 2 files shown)


OSXLich-Doug (approved):
All clean, no actionable findings from independent review and local CodeRabbit review returned 0 findings as well.  Good to merge.

PR #1532, opened 2026-08-20
fix(all): terminate preserved quiet-state tags with a  newline

mrhoribu:
Every chunk that normally reaches DownstreamHook.run is newline/CRLF- terminated, since GS4's stream is fundamentally line-oriented. preserve_quiet_state_tags returned just the matched tag text with no terminator at all -- a shape of line this pipeline never produced before this method existed.

For sentinel-supporting frontends (currently only Saga -- see Frontend::ORIGIN_SENTINEL / Frontend.supports_sentinel? / Game#prefix_origin_sentinel in lib/games.rb), every forwarded line gets a leading \x1f origin-marker byte that the client consumes as routing metadata and strips before display. Confirmed in production: with an unterminated preserved-tag chunk, that byte was rendering on screen as a literal, visible 'US' character during quiet-suppressed output -- reported as a regression starting with the tag-preservation fix, since previously these chunks returned nil and never reached prefix_origin_sentinel at all.

Append a trailing newline to the returned tag string, matching the termination every other line in this pipeline already has. Confirmed this resolves the reported symptom.

Updated all affected spec assertions in spec/lib/util/util_spec.rb to expect the trailing newline, and added a dedicated regression case for the termination itself. Verified against the prior no-newline implementation: reverting produces the same 6 targeted failures this patch fixes, and no others.

### DIFF
--- a/lib/util/util.rb
+++ b/lib/util/util.rb
@@ -114,17 +114,17 @@ module Lich
     #
     # The returned string is newline-terminated even though the matched tags
     # themselves never include one. Every other chunk that reaches this
-    # pipeline is newline/CRLF-terminated as part of GS4's line-oriented
-    # stream framing; a bare, unterminated tag is not a shape of line this
-    # pipeline produced before this method existed. For sentinel-supporting
-    # frontends (currently only Saga -- see Frontend::ORIGIN_SENTINEL and
-    # Game#prefix_origin_sentinel), every forwarded line gets a leading
-    # origin-marker byte that the client is expected to consume as routing
-    # metadata and strip before display. That worked here once the segment
-    # was given the same line termination as everything else the client
-    # already handles; without it, the client had nothing to delimit an
-    # otherwise content-only segment, and the marker byte fell through to
-    # the display as a literal, visible character.
+    # pipeline is newline/CRLF-terminated as part of the game server's
+    # line-oriented stream framing; a bare, unterminated tag is not a shape
+    # of line this pipeline produced before this method existed. For
+    # sentinel-supporting frontends (currently only Saga -- see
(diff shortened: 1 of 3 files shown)


OSXLich-Doug:
The implementation appears behaviorally correct, with one regression-coverage issue and one documentation correction to resolve.
Findings:
- [P2] The test does not exercise the Saga sentinel path that caused the regression. The new assertions verify that preserve_quiet_state_tags returns a newline, but the reported failure occurs downstream in Game#prefix_origin_sentinel and process_downstream_hooks. Add a regression test that enables sentinel support, routes the preserved tag through that pipeline, captures send_to_client, and expects "\x1f<output class=\"\"/>\n". This ensures the visible marker bug cannot return through ordering or framing changes.
- [P3] The documentation incorrectly describes the shared stream as GS4-specific. [lib/util/util.rb (line 117)](/Users/doug/dev/Projects/lich-5/lib/util/util.rb:117) says the framing is part of “GS4’s line-oriented stream,” but Game#read_server_string is shared and uses socket.gets for both supported games. Replace “GS4’s” with “the game server’s” or equivalent shared-runtime wording.

A local bunny hop produced the 0 additional items.

OSXLich-Doug:
@coderabbitai review and determine if prior findings are closed, in addition to your own

mrhoribu:
@coderabbitai thanks for pointing out that P2, but the very next test (outside of this PR) sets the `$frontend` to `wrayth` already, So do we really need to correct this test when the $frontend is already being changed later in the same test suite?

OSXLich-Doug (approved):
Downgrading P2 to nit level, random seed impact here is negligible risk.  Approved.

PR #1538, opened 2026-08-21
feat(creatures): measured combat data and spawn locations from GST forge campaign

Nisugi:
## Summary

Fills the creature template library with **measured** combat data gathered on the GST test server by forge (automated measurement campaign): 204 recorded sessions, 93k combat samples, 13.8k measured kills across ~420 creatures. Wiki-derived figures are replaced by measurements; every written value traces to the session and timestamp of the endroll behind it.

Three commits:
1. **Saga spawn import** — `spawns:` blocks (zone + game-UID ranges) on each template, converted to Lich ids at time of use so mapdb renumbering can't stale them.
2. **15 new templates** for creatures encountered in the campaign: twelve Grimswarm warcamp variants (level nil — they scale with the camp; all link the shared gswiki Grimswarm page), the Kraken's Fall zone-7131 pair (barbed cavern urchin, luminous worm), and the greater fetid corpse.
3. **The measured-data fill** across 377 templates.

## Measurement policy

| system | how it's measured |
|---|---|
| melee DS | `(offensive floor..guarded ceiling)` of clean standing samples; ever-stunned individuals quarantined (stun leaves DS depressed after standing); both ends need a confirming second observation |
| bolt/ranged DS | one pooled measurement, both fields |
| UDF | clean standing UAC endrolls only |
| per-circle TDs | stated values; range spans the ±5 individual level jitter; debuffed readings excluded |
| max_hp | the most any individual demonstrably survived (jab kills only, for single-target damage attribution); raise-only against curated values; troll families exempt — regeneration inflates parsed damage |

## Validation

Cross-checked against Saga's bestiary (33 creatures): levels match on all 33, HP floors converge 0–6 points under Saga's values on 25, and the data corrects four upstream errors with deep sample support (goblin 40→49, greater spider 83→90 at n=45, greater orc 110→112 at n=31, tomb wight 135→140).

All 626 template files parse as valid Ruby after the write.


Nisugi:
Opened prematurely - will resubmit through the intended flow.

PR #1541, opened 2026-08-22
fix(gs): Infomon xmlparser additional npc death messaging spec

jscholly:
Additional death messages to go along with additions to xmlparser.rb


mrhoribu:
are you able to merge this into #1540 as it should be one PR with one branch so that the spec tests runs against that.

jscholly:
combined pull request with #1540

PR #1544, opened 2026-08-23
fix(all): always report missing gems on stderr; only wait on a dialog without a tty

### LINKED ISSUE #1543: GemCheck blocks 120s on a GUI dialog and exits silently for CLI invocations on Linux (--help/--version/--active-sessions unusable when gems are missing)
## Summary

On Linux, when any bundled gem is missing, `GemCheck.verify!` blocks for a full 120 seconds on a GUI dialog and then exits `1` having printed **nothing** to the terminal. Every CLI-only invocation is affected, including `--help` and `--version`, because the gem check runs at `lich.rbw:39` — long before the CLI dispatcher is even loaded at `lich.rbw:133`.

The net effect from a terminal is that Lich appears to hang for two minutes and then do nothing, with no indication of why.

## Reproduce

On a Linux box with `zenity` installed (any desktop distro), using a Ruby that doesn't yet have Lich's gems — e.g. right after following the 4.0.3 → 4.0.5 bump from #1529 with a fresh `rbenv install`:

```sh
ruby ~/code/dragonrealms/lich-5/lich.rbw --active-sessions
```

It sits for 120 seconds, prints nothing, exits 1. `--help` and `--version` behave identically. Interrupting shows where it is:

```
^C/home/user/code/dragonrealms/lich-5/lib/gemcheck.rb:541:in 'Process.wait': Interrupt
        from lib/gemcheck.rb:541:in 'block in Lich::GemCheck.run_with_timeout'
        from .../timeout.rb:306:in 'Timeout.timeout'
        from lib/gemcheck.rb:541:in 'Lich::GemCheck.run_with_timeout'
        from lib/gemcheck.rb:514:in 'Lich::GemCheck.alert_linux'
        from lib/gemcheck.rb:295:in 'Lich::GemCheck.alert'
        from lib/gemcheck.rb:73:in 'Lich::GemCheck.verify!'
        from lich.rbw:39:in '<main>'
```

## Cause

`alert_linux` (`lib/gemcheck.rb:512-522`) picks a GUI dialog whenever one is on `PATH`:

```ruby
if cmd_available?('zenity')
  run_with_timeout(['zenity', '--info', '--title', TITLE, '--text', body], ALERT_TIMEOUT_SECONDS)
elsif cmd_available?('kdialog')
  ...
else
  warn "!!ALERT!! #{body}"      # <-- the branch that would actually be visible
end
```

`run_with_timeout` spawns it with `out: File::NULL, err: File::NULL` and waits up to `ALERT_TIMEOUT_SECONDS` (120). So:

- The dialog may be on another workspace, behind other windows, or simply not rendering — the user launched from a terminal and isn't looking for one.
- Nothing reaches the terminal, because the message only ever goes into the dialog and the child's output is discarded.
- After 120s the child is TERMed and `verify!` exits 1, still silently.

The information isn't lost, just never surfaced: `alert` calls `write_log` *before* opening the dialog, so `TEMP_DIR/lich5-missing-gems.log` already has the details. And the message it was trying to show is the actually-useful `UNIX_MESSAGE` — "You're missing required Ruby gems! Please run 'bundle install' from your Lich5 folder."

## Why it bites harder than it looks

`lich.rbw:39` runs `GemCheck.verify!` before `lib/main/argv_options.rb` is required at line 133. So the blocking dialog fires ahead of *every* early-exit CLI command — `--help`, `--version`, `--active-sessions`, `--session-info`, `--link-to-sge`, `--install`, and the rest. A user whose gems are missing cannot even run `lich.rbw --help` to find out what to do, which is exactly when they would.

## Suggested fix

Prefer the terminal when there is one. `alert_linux` already has a perfectly good `warn` branch; it just isn't reachable when a dialog binary exists. Something like:

```ruby
def alert_linux(body)
  return warn("!!ALERT!! #{body}") if $stdout.isatty
  # ... existing zenity/kdialog/xmessage chain
end
```

`$stdout.isatty` is already the established idiom for this in the codebase — `lib/main/argv_options.rb` uses it for `--link-to-sge`, `--unlink-from-sge`, and friends.

Options, roughly in order of preference:

1. Use the `warn` fallback whenever `$stdout.isatty`, so terminal launches get an instant readable error and desktop-launcher starts keep the dialog.
2. Additionally, print the log path unconditionally before opening any dialog, so there is always a breadcrumb.
3. Optionally drop `ALERT_TIMEOUT_SECONDS` for the non-interactive case — two minutes is a long time to block a startup path that has already decided to exit 1.

Happy to put up a PR for (1) plus (2) if that's the direction you'd want.

## Environment

- Linux, `zenity` present
- Ruby 4.0.5 via rbenv (fresh install, gems not yet installed for it — `bundle install` resolves the underlying missing-gem condition and is the correct user action; this issue is about the two-minute silent block that hides that fact)

simtel12:
## Summary

Fixes #1543.

On Linux, `alert_linux` opened a GUI dialog whenever `zenity`/`kdialog`/`xmessage` was on `PATH`, and `run_with_timeout` then waited on it for up to `ALERT_TIMEOUT_SECONDS` (120) with the child's stdout and stderr sent to `File::NULL`. The message existed *only* inside that dialog. A dialog raised from a terminal launch is easy to miss — another workspace, behind other windows, or not rendering at all — so from the terminal Lich appeared to hang for two minutes and then exit 1 having printed nothing.

It lands hardest on the early-exit CLI commands: `GemCheck.verify!` runs at `lich.rbw:39`, but `lib/main/argv_options.rb` is not required until `lich.rbw:133`. So `--help`, `--version`, `--active-sessions`, `--session-info`, `--link-to-sge` and the rest are all unreachable while the dialog blocks — including the `--help` that would have told the user to run `bundle install`.

Easy to hit right now: following the 4.0.3 → 4.0.5 bump from #1529 with a fresh `rbenv install` leaves you on an empty gem home, which is exactly the triggering condition.

## Change

The reporting and the waiting used to be the same act. Separating them is the whole fix:

- **Always `warn` on stderr**, unconditionally. It costs nothing, it carries the whole message rather than a pointer to it, and when stderr is captured (a launcher's output, a service log) it is the only record that survives. This is the part that makes the failure legible.
- **Only wait on a dialog when there was no terminal that report could reach.** A launcher start has no other channel, so it keeps the dialog and its timeout exactly as before. A terminal launch already has the message and no longer blocks two minutes to deliver it a second time.

So nobody loses information relative to today, and the GUI path is untouched for the launches that actually depend on it. Both `$stdout` and `$stderr` are probed, since a terminal launch may redirect either; `$stdout.isatty` is already the idiom for this in `lib/main/argv_options.rb` (`--link-to-sge`, `--unlink-from-sge`, …).

The trailing no-dialog-tool `else warn` branch is removed — the unconditional warn is now the fallback it used to provide.

One deliberate non-change: a terminal launch no longer *pops* the dialog at all, rather than popping it and not waiting. Firing it off unwaited would leave an orphaned window behind an exited process, which is precisely what `run_with_timeout` was written to avoid ("the spawned child keeps running as an orphan; spawning it ourselves lets us kill it directly"). If you'd rather terminal launches still get a window, the natural shape is a short bounded wait rather than fire-and-forget — happy to switch if you prefer that trade.

Scoped to Linux, matching the issue. `alert_macos` has the same shape and would benefit from the same treatment — happy to extend it here or in a follow-up. I left `alert_windows` alone deliberately: it is bound up with the Ruby4Lich5 self-healing flow and I have no way to test it.

## Test plan

- [x] `bundle exec rspec` (full suite): 6338 examples, 0 failures
- [x] `bundle exec rubocop` on both touched files: no offenses
- [x] New specs: warn happens in both branches, a tty on stdout suppresses the dialog wait, a tty on stderr alone counts as interactive, and the no-dialog-tool case still reports. Confirmed the three key examples **fail against unmodified `main`** and pass with the fix
- [x] Existing `alert_linux` specs now stub both streams explicitly rather than inheriting the runner's — rspec's stdout is a tty when run directly in a terminal but a pipe under CI, so they would otherwise pass or fail by accident
- [x] End-to-end against a simulated desktop (`cmd_available?` always true, `run_with_timeout` sleeping the full timeout): under a pty the message prints in 0.000s and no dialog is spawned; with output redirected the full alert lands on stderr **before** the dialog wait begins, and the process still blocks on the dialog — confirming the launcher path is genuinely unchanged

### DIFF
--- a/lib/gemcheck.rb
+++ b/lib/gemcheck.rb
@@ -507,18 +507,44 @@ module Lich
       body.split("\n").map(&:inspect).join(' & return & ')
     end
 
+    # Reports missing gems on stderr, and additionally in a GUI dialog when
+    # there is no terminal that report could have reached.
+    #
+    # The dialog is only ever *waited on*, never merely shown: a launcher start
+    # has no other channel, so {run_with_timeout} keeps the process alive long
+    # enough for the dialog to be read. From a terminal that wait is what makes
+    # the failure look like a hang -- the dialog is easily missed (another
+    # workspace, behind other windows, or not rendering at all), it blocks for
+    # {ALERT_TIMEOUT_SECONDS}, and the child's output is discarded, so the user
+    # waits two minutes and never learns why. Worst for the early-exit CLI
+    # commands (`--help`, `--version`, `--active-sessions`, ...): they are
+    # dispatched well after this check runs, so while it blocks they cannot be
+    # reached at all.
+    #
+    # Both streams are probed because a terminal launch may redirect either
+    # one.
+    #
     # @param body [String]
     # @return [void]
     def alert_linux(body)
+      # Always say it on stderr, whether or not a dialog follows. It costs
+      # nothing, it is the whole message rather than a pointer to it, and when
+      # stderr is captured (a launcher's output, a service log) it is the only
+      # record that survives -- the dialog text goes nowhere else.
+      warn("!!ALERT!! #{body}")
+
+      # Only wait on a dialog when there was no terminal to have said it in.
+      return if $stdout.isatty || $stderr.isatty
+
       if cmd_available?('zenity')
         run_with_timeout(['zenity', '--info', '--title', TITLE, '--text', body], ALERT_TIMEOUT_SECONDS)
       elsif cmd_available?('kdialog')
         run_with_timeout(['kdialog', '--title', TITLE, '--msgbox', body], ALERT_TIMEOUT_SECONDS)
       elsif cmd_available?('xmessage')
         run_with_timeout(['xmessage', '-center', body], ALERT_TIMEOUT_SECONDS)
-      else
-        warn "!!ALERT!! #{body}"
       end
+      # No dialog tool available needs no fallback branch any more -- the
+      # unconditional warn above is the fallback.
     end
 
     # @param cmd [String] executable name to probe
--- a/spec/lib/gemcheck_spec.rb
+++ b/spec/lib/gemcheck_spec.rb
@@ -692,6 +692,58 @@ RSpec.describe Lich::GemCheck do
       allow(described_class).to receive(:cmd_available?).with(tool).and_return(true)
     end
 
+    # The dialog chain is only reached with no terminal attached. Stub both
+    # streams explicitly rather than inheriting whatever the runner happens to
+    # have -- rspec's stdout is a tty when run directly in a terminal and a
+    # pipe under CI, which would otherwise make these pass or fail by accident.
+    before do
+      allow($stdout).to receive(:isatty).and_return(false)
+      allow($stderr).to receive(:isatty).and_return(false)
+      allow(described_class).to receive(:warn)
+    end
+
+    it 'always warns, whether or not a dialog follows' do
+      stub_only_available('zenity')
+      allow(described_class).to receive(:run_with_timeout)
+
+      expect(described_class).to receive(:warn).with(/!!ALERT!!.*body text/m).twice
+
+      described_class.alert_linux('body text')             # no tty: warn + dialog
+      allow($stdout).to receive(:isatty).and_return(true)
+      described_class.alert_linux('body text')             # tty: warn only
+    end
+
+    context 'when a terminal is attached' do
+      it 'does not wait on a dialog, even when zenity is available' do
+        stub_only_available('zenity')
+        allow($stdout).to receive(:isatty).and_return(true)
+
+        expect(described_class).not_to receive(:run_with_timeout)
+        expect(described_class).to receive(:warn).with(/!!ALERT!!.*body text/m)
+
+        described_class.alert_linux('body text')
+      end
+
+      it 'treats a terminal on stderr alone as interactive' do
+        stub_only_available('zenity')
+        allow($stderr).to receive(:isatty).and_return(true)
+
+        expect(described_class).not_to receive(:run_with_timeout)
+
+        described_class.alert_linux('body text')
+      end
+    end
+
+    context 'when no dialog tool is available and no terminal is attached' do
+      it 'still reports on stderr' do
+        allow(described_class).to receive(:cmd_available?).and_return(false)
+
+        expect(described_class).to receive(:warn).with(/!!ALERT!!.*body text/m)
+
+        described_class.alert_linux('body text')
+      end
+    end
+
     context 'when zenity is available' do
       before { stub_only_available('zenity') }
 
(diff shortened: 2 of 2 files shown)


mrhoribu (approved):
# PR Review: fix(all): always report missing gems on stderr; only wait on a dialog without a tty (elanthia-online/lich-5#1544)

**Reviewed:** 2 files, +80/-11 at `6e4a6d48` on 2026-09-07. Based on: full local clone (worktree checkout of the PR branch against `main`), including project history.
**Verdict:** Approve
**Summary:** Fixes #1543 (a 120s silent hang on `--help`/`--version`/etc. when Linux gems are missing and a GUI dialog tool is on `PATH`) by unconditionally `warn`-ing the alert body on stderr and skipping the blocking dialog wait whenever either `$stdout` or `$stderr` is a tty. The change is small, precisely scoped, and the launcher (no-tty) path is verifiably unchanged.

## Blockers
None.

## Major
None.

## Minor
None.

## Nits
None worth raising — style, naming, and doc comments are consistent with the surrounding file and rubocop reports no offenses.

## What this PR gets right

- **Ordering is safe.** `GemCheck.verify!` runs at `lich.rbw:39`, before `lib/init.rb` (`$stderr = File.open(...)`, `lib/init.rb:494`) and `lib/main/main.rb` (`$stdout = $_CLIENT_`, `lib/main/main.rb:686`) ever reassign the standard streams. So `$stdout`/`$stderr` are guaranteed to still be real process IO objects (which always respond to `isatty`) at the point `alert_linux` checks them — there's no risk of a `NoMethodError` from a reassigned stream lacking `isatty`, which would otherwise be a plausible-looking landmine for this kind of check.
- **Single call site.** `alert_linux` is only ever invoked from `alert` (`lib/gemcheck.rb:295`), gated by `RUBY_PLATFORM`, so there's no blast radius to chase beyond this file.
- **Checking both streams is the right call**, not just cosmetic caution — `lib/main/argv_options.rb` already tests only `$stdout.isatty` for its interactivity checks, but a shell redirecting just one of stdout/stderr (`cmd 2>err.log` or `cmd >out.log`) is a completely ordinary invocation shape, and treating that as "no terminal" would resurrect the two-minute hang for a real subset of users.
- **Test coverage matches the described behavior exactly**: unconditional warn on both paths, dialog suppressed when either stream is a tty, dialog still fires with no tty and no dialog tool present (still reports via warn). The existing `zenity`/`kdialog`/`xmessage` specs correctly inherit the new `before` block's `isatty` stubs rather than needing their own, and stubbing both streams explicitly (rather than relying on the runner's actual tty state) correctly avoids the CI/local-terminal flakiness the PR description calls out.
- Full suite (6338 examples via PR description; `bundle exec rspec spec/lib/gemcheck_spec.rb` reproduced locally: 87 examples, 0 failures) and `rubocop` on both touched files are clean, confirmed independently in this review.
- The PR body is unusually well-grounded for a self-review — it explains the one deliberate behavior change (terminal launches no longer pop an unwaited dialog) and ties it back to the exact reasoning in the `run_with_timeout` doc comment about orphaned processes, rather than silently making that call.

## Open questions

None — the diff is self-contained, the reasoning in the PR description checks out against the code and issue, and the scoping decisions (leaving `alert_macos`/`alert_windows` untouched) are explicitly justified rather than left ambiguous.

## Coverage notes

Reviewed via a full local clone (git worktree of `pr-1544` against `origin/main`), not just the diff — read the complete `lib/gemcheck.rb` (call graph, `run_with_timeout`, `cmd_available?`, `verify!`), the file's git history, `lich.rbw`'s require order, and the reassignment sites for `$stdout`/`$stderr` elsewhere in the codebase. Ran `ruby -c`, `rubocop`, and the full `gemcheck_spec.rb` suite locally against the PR branch — all pass. Did not execute an end-to-end manual repro with an actual missing gem / real `zenity` binary; relying on the PR author's stated pty/redirect end-to-end test plan for that portion, which is consistent with the code as written.

PR #1546, opened 2026-08-24
fix(all): dispatch --help and --version before the GTK require

### LINKED ISSUE #1545: Early-exit CLI commands (--help, --version, --active-sessions) require gtk3 and abort without it
## Summary

The early-exit CLI commands — `--help`, `--version`, `--active-sessions`, `--session-info`, `--link-to-sge`, `--install` and the rest — all require GTK to be installed, even though none of them draws a window. On a machine without `gtk3` they abort before the CLI dispatcher ever runs.

It is chicken-and-egg: `--no-gtk` avoids the problem, but you cannot run `--help` to discover that `--no-gtk` exists. The alert that appears instead says *"Please run 'bundle install' from your Lich5 folder"*, which does not fix it when `:gtk` is excluded via `BUNDLE_WITHOUT` — so the one piece of guidance offered is actively misleading for this case.

## Reproduce

On a checkout at `4ed650ba` (5.20.1), with the default group installed but `gtk3` absent — e.g. `bundle config set --local without 'gtk:vscode:profanity'`, which is a normal setup for a headless/CLI or proxy-only install:

```console
$ ruby lich.rbw --help
!!ALERT!! You're missing required Ruby gems!

Please run 'bundle install' from your Lich5 folder.

Missing gems:
  - gtk3

Underlying error:
  cannot load such file -- gtk3

See .../temp/lich5-missing-gems.log for details.

$ ruby lich.rbw --no-gtk --help
Lich 5
Usage:
  lich [command] [options]
...
```

`--active-sessions` behaves the same way: it needs `--no-gtk` to run at all, then works fine.

## Cause

It is a load-order problem in `lich.rbw`:

| line | what happens |
| --- | --- |
| 37 | `Lich::GemCheck.verify!` |
| **68** | `require .../init.rb` → **`require 'gtk3'`** |
| **131** | `require .../main/argv_options.rb` → CLI dispatch, `CLIOrchestration.execute` |

`lib/init.rb:417` requires GTK for every launch that did not explicitly opt out:

```ruby
unless ARGV.any? { |arg| arg.match?(/^--no-(?:gtk|gui)$/i) }
  begin
    require 'gtk3'
```

and on Linux a `LoadError` lands at `lib/init.rb:445-447`, which alerts and exits 1:

```ruby
# GTK is required unless the user explicitly selected a headless launch.
# Do not infer that choice from DISPLAY, TTY, or cron environment state.
Lich::GemCheck.alert(missing: ['gtk3'], groups: [:gtk], error: gtk_load_error)
exit 1
```

Because that runs at line 68 and the dispatcher is not loaded until line 131, a command like `--help` never gets the chance to handle itself.

## Regression from #1439

This is a behaviour change from #1439 (`feat(all): self-healing gem requires`, 2026-07-16). Before it, `lib/init.rb:520` read:

```ruby
unless (ARGV.grep(/^--no-(?:gtk|gui)$/).any? || RUBY_PLATFORM !~ /mingw/ && (ENV['DISPLAY'].nil? && !ARGV.include?('--gtk')))
```

On Linux with no `DISPLAY`, `gtk3` was never required at all, and the old `LoadError` branch additionally checked `not $stdout.isatty` before nagging — so CLI use on a headless box simply worked.

I appreciate the intent behind the change: the new comment is explicit that inferring headless from `DISPLAY`/TTY/cron state was deliberately dropped, and for an actual *session* launch that seems right — silently falling back to no-GUI because `DISPLAY` happened to be unset is its own bug. The gap is that the rule is now applied to commands that will never open a window either way.

## Suggested fix

Decide GTK by what the launch is going to do, not by a blanket rule applied before we know:

1. **Preferred:** dispatch the early-exit CLI commands before `init.rb` requires GTK. They are self-contained and exit; nothing they touch needs a toolkit. This keeps #1439's strictness exactly as-is for real session launches.
2. Or treat the presence of an early-exit command flag in `ARGV` as an implicit `--no-gtk`, which is a smaller change but leaves the ordering oddity in place.
3. Independently: when `gtk3` is missing *and* `:gtk` is in `BUNDLE_WITHOUT`, "run `bundle install`" is the wrong instruction. Mentioning `--no-gtk` in that alert would have made this self-service.

Happy to put up a PR for (1) if that is the direction you would want.

## Related

- #1543 / #1544 — the same alert path blocks 120s on a GUI dialog with no terminal output. Independent of the ordering problem reported here, but they compound: the transcript above is from unmodified `main`, and the message is only visible because the box I reproduced on has no `zenity`/`kdialog`/`xmessage`, so `alert_linux` fell through to its existing `warn` fallback. On a desktop with `zenity` installed, `ruby lich.rbw --help` instead hangs for 120 seconds and prints nothing at all.

simtel12:
Fixes #1545.

## Problem

`ruby lich.rbw --help` aborts on any runtime without the `gtk3` gem. The cause is load order in `lich.rbw`:

| line | what happens |
| --- | --- |
| 37 | `Lich::GemCheck.verify!` |
| 68 | `require .../init.rb` -> `require 'gtk3'` at `lib/init.rb:417` |
| 131 | `require .../main/argv_options.rb` -> CLI dispatch |

The GTK require sits 60 lines above the dispatcher, so an early-exit command never gets to handle itself. That makes it chicken-and-egg: `--no-gtk` avoids the problem, but `--help` is the command that reports `--no-gtk` exists.

## Approach

This takes option (1) from the issue, limited to the two commands that qualify.

`--help` and `--version` move into a new `Lich::Main::EarlyExit` module. `lich.rbw` dispatches it before **both** `GemCheck.verify!` and the `init.rb` require. Both commands read `ARGV`, print, and exit. They touch no gem, no directory, no log file, and no database, so they are safe that early.

Dispatch runs before `verify!` as well as before the GTK require. On Windows `startup_groups` adds `:gtk` to the verified groups, so a `--help` run on a machine with no GTK would otherwise stop at the gem check instead. Placing the dispatch first makes `--help` work on any incomplete runtime, which is the point of the report.

### The allow-list stays syntactic

`EarlyExit.requested?` reads `ARGV` and nothing else. It is not a "is this a GUI launch" classifier, and that is deliberate: #1439 removed the inference of a headless launch from `DISPLAY`, `TTY`, and cron state, and this must not bring it back. A spec asserts the answer does not change when `DISPLAY` is set or unset.

GTK stays required for every session launch that does not pass `--no-gtk` or `--no-gui`. #1439's strictness is unchanged.

### Why only two commands

`lib/init.rb` creates the Lich directory, performs `Dir.chdir`, opens the debug log, and calls `Lich.init_db` **after** the GTK require, at lines 456-570. The remaining early-exit commands depend on that state:

- `--install` and `--uninstall` call `Lich.log`, which needs the file opened at `lib/init.rb:492`.
- The account and password handlers reach `EntryStore`, which needs `DATA_DIR` from `lib/init.rb:515`.

Those keep their current behaviour and still require GTK. This also matches @OSXLich-Doug's read in the issue thread: `--link-to-sge` and `--install` target Windows, where GTK is present anyway.

### Removed rather than left unreachable

`lich.rbw:131` is the only caller of `ArgvOptions`, so the `-h`/`--help` and `-v`/`--version` branches there became dead once the dispatch moved. This deletes them and the two now-unused print methods instead of leaving a second copy of the version text to drift. `HelpText` is now required by `early_exit.rb`; `argv_options.rb` no longer references it, so that require goes too.

An ordering spec reads `lich.rbw` and asserts the dispatch precedes both `verify!` and the `init.rb` require, so a later edit cannot quietly reintroduce the fault.

## Not in scope

Two points from the issue stay open, and I am happy to take either as a separate PR:

- The `UNIX_MESSAGE` "run `bundle install`" text is wrong when `:gtk` sits in `BUNDLE_WITHOUT`. Naming `--no-gtk` there would make this self-service.
- Self-healing on Linux, which @OSXLich-Doug raised in the thread. Worth doing on its own merits, but it does not close this issue: the `gtk3` gem on Linux builds from source against system GTK development packages, and Ruby4Lich5 publishes `x64-mingw-ucrt` assets only. There is nothing for a Linux manifest to point at. macOS is in the same position today; `BundlerRecovery` excludes the `gtk` group at `lib/bundler_recovery.rb:18`, so a Mac without the `gtk3` gem hits this identical `--help` failure. Even with a working Linux repair, `--help` opening a consent dialog to install a toolkit before printing text would be the wrong outcome.

## Test plan

- [x] `bundle exec rspec` passes (6349 examples, 0 failures locally)
- [x] `bundle exec rubocop` reports no offences on the changed files
- [ ] With `bundle config set --local without 'gtk:vscode:profanity'` (no `gtk3` installed):
  - [ ] `ruby lich.rbw --help` prints the overview and exits 0
  - [ ] `ruby lich.rbw --version` prints the version banner
  - [ ] `ruby lich.rbw --help login` and `ruby lich.rbw --help=advanced` print the right topic
  - [ ] `ruby lich.rbw --no-gtk --help` still works
- [ ] Tier-2 behaviour is unchanged: `ruby lich.rbw --change-master-password` still shows the missing-gem alert, and `ruby lich.rbw --no-gtk --change-master-password` still reaches its own usage output
- [ ] On a machine **with** `gtk3`: a normal session launch still loads GTK, and `--no-gtk` still suppresses it

### DIFF
--- a/lib/main/argv_options.rb
+++ b/lib/main/argv_options.rb
@@ -12,7 +12,6 @@ require File.join(LIB_DIR, 'common', 'bind_host_resolver.rb')
 require File.join(LIB_DIR, 'main', 'bind_address_option.rb')
 require File.join(LIB_DIR, 'main', 'arg_normalization.rb')
 require File.join(LIB_DIR, 'main', 'detachable_client_target.rb')
-require File.join(LIB_DIR, 'main', 'help_text.rb')
 require File.join(LIB_DIR, 'main', 'startup_theme.rb')
 
 module Lich
@@ -31,12 +30,6 @@ module Lich
 
           ARGV.each do |arg|
             case arg
-            when '-h', '--help', /^--help=.+$/
-              print_help(HelpText.topic_from_argv(ARGV, arg))
-              exit
-            when '-v', '--version'
-              print_version
-              exit
             when '--link-to-sge'
               result = Lich.link_to_sge
               $stdout.puts(result ? 'Successfully linked to SGE.' : 'Failed to link to SGE.') if $stdout.isatty
@@ -133,22 +126,6 @@ module Lich
         def self.handle_dark_mode(value)
           # Regex returns Integer/nil; force strict boolean for startup handling.
           @argv_options[:dark_mode] = !!(value =~ /^(true|on)$/i)
-        end
-
-        def self.print_help(topic = nil)
-          puts HelpText.render(topic)
-        end
-
-        def self.print_version
-          puts "The Lich, version #{LICH_VERSION}"
-          puts ' (an implementation of the Ruby interpreter by Yukihiro Matsumoto designed to be a \'script engine\' for text-based MUDs)'
-          puts ''
-          puts '- The Lich program and all material collectively referred to as "The Lich project" is copyright (C) 2005-2006 Murray Miron.'
-          puts '- The Gemstone IV and DragonRealms games are copyright (C) Simutronics Corporation.'
-          puts '- The Wizard front-end and the StormFront front-end are also copyrighted by the Simutronics Corporation.'
-          puts '- Ruby is (C) Yukihiro \'Matz\' Matsumoto.'
-          puts ''
-          puts 'Thanks to all those who\'ve reported bugs and helped me track down problems on both Windows and Linux.'
         end
       end
 
--- a/lib/main/early_exit.rb
+++ b/lib/main/early_exit.rb
@@ -1,0 +1,84 @@
+# frozen_string_literal: true
+
+require_relative 'help_text'
+
+module Lich
+  module Main
+    # Commands that print to stdout and exit without starting a session.
+    #
+    # lich.rbw dispatches these before Lich::GemCheck.verify! and before
+    # lib/init.rb requires GTK, so they stay usable on a runtime that cannot
+    # load a toolkit. Without that ordering, `--help` cannot report the very
+    # options that make a headless launch work.
+    #
+    # The allow-list is deliberately syntactic. It reads ARGV and nothing else.
+    # DISPLAY, TTY, and cron state never select a launch mode here, which keeps
+    # the guarantee that #1439 introduced.
+    #
+    # Membership is limited to commands that need no directory, no log file,
+    # and no database. lib/init.rb creates all three after the GTK require, so
+    # a command that needs any of them cannot move into this module.
+    module EarlyExit
+      HELP_FLAGS = ['-h', '--help'].freeze
+      HELP_TOPIC_PREFIX = '--help='
+      VERSION_FLAGS = ['-v', '--version'].freeze
+
+      module_function
+
+      # Prints the first selected command and exits. Returns when the arguments
+      # select no command handled here.
+      #
+      # @param argv [Array<String>] command-line arguments
+      # @return [void]
+      def dispatch!(argv = ARGV)
+        argv = Array(argv)
+        argv.each do |arg|
+          if help?(arg)
+            print_help(HelpText.topic_from_argv(argv, arg))
+            exit
+          elsif version?(arg)
+            print_version
+            exit
+          end
+        end
+      end
+
+      # @param argv [Array<String>] command-line arguments
+      # @return [Boolean] whether the arguments select a command handled here
+      def requested?(argv = ARGV)
+        Array(argv).any? { |arg| help?(arg) || version?(arg) }
+      end
+
+      # @param arg [String] one command-line argument
+      # @return [Boolean]
+      def help?(arg)
+        HELP_FLAGS.include?(arg) || arg.to_s.start_with?(HELP_TOPIC_PREFIX)
+      end
+
+      # @param arg [String] one command-line argument
+      # @return [Boolean]
+      def version?(arg)
+        VERSION_FLAGS.include?(arg)
+      end
+
+      # @param topic [String, nil] optional help topic
+      # @return [void]
+      def print_help(topic = nil)
+        puts HelpText.render(topic)
+      end
+
+      # @return [void]
+      def print_version
+        puts "The Lich, version #{LICH_VERSION}"
+        puts ' (an implementation of the Ruby interpreter by Yukihiro Matsumoto designed to be a \'script engine\' for text-based MUDs)'
+        puts ''
+        puts '- The Lich program and all material collectively referred to as "The Lich project" is copyright (C) 2005-2006 Murray Miron.'
+        puts '- The Gemstone IV and DragonRealms games are copyright (C) Simutronics Corporation.'
+        puts '- The Wizard front-end and the StormFront front-end are also copyrighted by the Simutronics Corporation.'
+        puts '- Ruby is (C) Yukihiro \'Matz\' Matsumoto.'
+        puts ''
+        puts 'Thanks to all those who\'ve reported bugs and helped me track down problems on both Windows and Linux.'
+      end
+    end
+  end
+end
--- a/lich.rbw
+++ b/lich.rbw
@@ -33,6 +33,14 @@ else
   require_relative('./lib/constants.rb')
 end
 require File.join(LIB_DIR, 'version.rb')
+
+# --help and --version print to stdout and exit. They read ARGV, and they touch
+# no gem, no directory, and no database. Dispatch them here, before the gem
+# check and before lib/init.rb's `require 'gtk3'`, so a runtime that cannot
+# load a toolkit can still report how to launch without one.
+require File.join(LIB_DIR, 'main', 'early_exit.rb')
+Lich::Main::EarlyExit.dispatch!
+
 require File.join(LIB_DIR, 'gemcheck.rb')
 Lich::GemCheck.verify!(*Lich::GemCheck.startup_groups)
 
--- a/spec/lib/main/early_exit_spec.rb
+++ b/spec/lib/main/early_exit_spec.rb
@@ -1,0 +1,112 @@
+# frozen_string_literal: true
+
+require 'rspec'
+
+require_relative '../../../lib/main/early_exit'
+
+RSpec.describe Lich::Main::EarlyExit do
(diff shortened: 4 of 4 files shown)


mrhoribu (approved):
# PR Review: fix(all): dispatch --help and --version before the GTK require (elanthia-online/lich-5#1546)

**Reviewed:** 4 files, +204/-23 at `10f365a6` on 2026-09-07. Based on: full clone (local checkout, fetched PR ref into a worktree), full test suite run, and manual runtime verification.
**Verdict:** Approve
**Summary:** Moves `--help`/`--version` dispatch in `lich.rbw` to run before `Lich::GemCheck.verify!` and before `lib/init.rb`'s `require 'gtk3'`, fixing a real bug (confirmed by reproduction) where those two commands abort on a runtime without the `gtk3` gem. The change is small, isolated, well-tested, and I could not make it misbehave.

## What I verified

- **The bug is real.** Checked out `main` (`4ed650ba`), blocked `require 'gtk3'` via a `Kernel#require` shim, and ran `ruby lich.rbw --help`: it aborts with the "You're missing required Ruby gems!" alert, exit code path through `lib/init.rb:445-448`. Same command against this PR's branch prints help text and exits 0. This confirms the issue (elanthia-online/lich-5#1545) and that the fix addresses it.
- **`--login` (a real session launch) still requires GTK post-fix** — verified with the same require-blocking shim; it still hits the missing-gem alert, so #1439's strictness for actual sessions is unchanged.
- **`--change-master-password` (a Tier-2 early-exit command, deliberately left out of the allow-list) still requires GTK** — verified directly. Matches the PR's stated scoping rationale (it depends on `DATA_DIR`, set up in `lib/init.rb` after the GTK require).
- **`--no-gtk --help` still works**, `--help=login` and `--version` both print the right content and exit 0.
- **Full test suite**: `bundle exec rspec` — 6349 examples, 0 failures, on the PR branch.
- **Rubocop**: `rubocop lib/main/early_exit.rb spec/lib/main/early_exit_spec.rb lich.rbw lib/main/argv_options.rb` — 4 files inspected, no offenses.
- **No dangling references**: grepped the whole tree for `ArgvOptions` usages of the removed `print_help`/`print_version`/the `-h`/`-v` `case` branches, and for other consumers of `HelpText` — the only caller was the code being deleted; `HelpText` is now required solely by `early_exit.rb`, matching the diff's claim that it's not referenced elsewhere.
- **Load order**: confirmed `LICH_VERSION` (from `lib/version.rb`, required at `lich.rbw:35`) is defined before the new dispatch call at `lich.rbw:42`, and that `lib/gemcheck.rb`'s own top-level requires (`bundler`, `timeout`, two `require_relative`s) are lightweight — pushing its require below the dispatch call doesn't reorder anything expensive ahead of `--help`.
- **`init.rb` requires `sqlite3` (line 386) before `gtk3` (line 417)** — both now happen after the early-exit dispatch, consistent with the new spec's "load isolation" assertions (`lib/main/early_exit.rb` requires neither `gtk3`, `sqlite3`, `argv_options`, nor `cli_orchestration`).
- **History check on #1439** (`dca20398`, "feat(all): self-healing gem requires") — confirms the PR's characterization: that commit is what introduced the current strict, environment-independent GTK-require behavior the new code is careful not to circumvent.

## What this PR gets right

- The allow-list is provably syntactic (`EarlyExit.requested?`/`help?`/`version?` read only `argv`), and the included spec that toggles `ENV['DISPLAY']` around `--help` is a good regression guard against re-introducing the exact inference #1439 removed.
- Scoping to just `--help`/`--version` rather than every early-exit command in the issue is the right call — the PR body's dependency table on `lib/init.rb:456-570` state (directory, log, `DATA_DIR`) checks out against the current source, and I verified `--change-master-password` still behaves as before.
- Dead code removal is clean: no orphaned `print_help`/`print_version` methods left in `argv_options.rb`, and the `help_text.rb` require moved with the functionality rather than being duplicated.
- The `lich.rbw` boot-ordering spec (asserting `dispatch!`'s call site precedes both `GemCheck.verify!` and the `init.rb` require, by string-searching the actual file) is a solid regression guard for the one thing that matters here: ordering.
- `require_relative 'help_text'` inside `early_exit.rb` is correct and arguably more robust than the codebase's usual `File.join(LIB_DIR, ...)` pattern — it resolves relative to `early_exit.rb`'s own actual location regardless of a custom `--lib-dir`.

## Nits

### 1. `EarlyExit.requested?` has no production caller

`lib/main/early_exit.rb:102-104` defines `requested?`, but nothing in `lich.rbw` or elsewhere calls it — only its own spec (`spec/lib/main/early_exit_spec.rb:173-199`) does. It's a reasonable public predicate to expose alongside `dispatch!` (and the spec's DISPLAY-independence assertion is genuinely useful as a regression guard), so this isn't dead code in the "delete it" sense, but it's worth knowing it's currently test-only.

## Open questions

- `lib/internal_api/active_sessions.rb:445` falls back to `Dir.tmpdir` when `TEMP_DIR` isn't defined yet, which suggests `--active-sessions`/`--session-info` — also named in elanthia-online/lich-5#1545 as affected — might be closer to self-contained than the PR body's "the remaining early-exit commands depend on that state" framing implies for `--install`/`--uninstall`/the account handlers. The PR explicitly scopes to two commands and says it's happy to take the rest as follow-ups, so this isn't a defect in what's submitted — but if `--active-sessions` genuinely doesn't need `lib/init.rb`'s directory/log/db setup, is a follow-up already planned, or was it tried and rejected for a reason not in the PR body (e.g. it also wants `Lich.log`, which does need the debug log file)?

## Coverage notes

Reviewed via a full local clone (`git fetch origin pull/1546/head`), diffed against `main` in an isolated worktree so the reviewer's own uncommitted work on another branch was never touched. Ran the full existing spec suite and rubocop against the PR's actual files, plus manual CLI reproduction of both the pre-fix bug and the post-fix behavior using a `Kernel#require` shim to simulate a missing `gtk3` gem (the real gem is installed in this environment, so this was necessary to actually exercise the failure path rather than just reading the code). Did not test on Windows or macOS, and did not exercise the GTK self-healing/consent-recovery flow (`Lich::GemCheck.recover_with_consent!`) since it's unreached by this change. No blockers or major findings.

PR #1547, opened 2026-08-26
fix(gs): correct messaging drift in 13 crit table patterns

Nisugi:
## Summary

The crit tables were transcribed from the wiki. Live messaging has since drifted, so a number of patterns no longer match what the game actually sends and those crits are silently lost.

Found by scanning **200,341,002 lines** of session logs: build a relaxed variant of each of the 2,394 table regexes (flexible whitespace, optional articles/punctuation), then report lines that match the relaxed form but fail `CritRanks.parse`. Every hit is a message we already have a table entry for, just with drifted wording.

## Drift classes

**1. Two spaces between sentences, table had one**

```
table: /^Burn to the elbow eats through tendons. Muscles snap free./
live:  "Burn to the elbow eats through tendons.  Muscles snap free!"
```

Affects `acid` (elbow tendons, elbow ligaments, arm drenched, eye socket, drop of acid, Acid Flush) and `crush` (Whoosh).

**2. Trailing punctuation absent in live messaging**

```
table: /^Stinging burn to .*? hand./
live:  "Stinging burn to the basalt grotesque's hand"
```

Occurs when the crit is followed by another damage line, e.g. a plasma DoT tick:

```
The whip of plasma continues to wreathe a hapless charmed corsair in its sizzling embrace!
   ... 7 points of damage!
   Light burns to the charmed corsair's leg
   ... 18 points of damage!
```

Present in Prime (1,463 of 4,066 instances) as well as Shattered, so not realm-specific.

**3. Article mismatch**

```
table: /^Perfect strike to the abdomen./
live:  "Perfect strike to abdomen.  The steelwing harpy howls in pain and drops quite dead!"
```

```
table: /^Flame sets a .*? head alight like a torch./
live:  "Flame sets an eyeless black valravn's head alight like a torch."
```

**4. Capitalised variant unreachable**

The tables are case-sensitive, so a leading `.*?` cannot match a capitalised word:

```
table: /^.*? knocked back and stunned./
live:  "Chest strike.  Knocked back and stunned!"   (vs "The triton brawler knocked back and stunned!")
```

## Also included

`crush_critical_table.rb` had a malformed `secondary_wound` key — the string `"wound_rank"` instead of the symbol `:wound_rank`. `Processor#apply_secondary_wound` reads `secondary[:wound_rank]`, so that nerves wound was silently dropped.

## Changes

| Table | Change |
|---|---|
| puncture | `to (?:the )?abdomen` |
| plasma | `leg.?`, `hand.?` |
| lightning | `kidneys.?` |
| acid | two spaces ×6 |
| crush | `Whoosh.  Several`, `secondary_wound` key |
| fire | `Flame sets an?` |
| impact | `Hard blow to(?: the)? left hand` |
| unbalance | `[Kk]nocked back and stunned` |

## Testing

- 19 targeted assertions covering every changed pattern, including **both** the old and new wording where a form was made optional, confirming no existing match was broken
- `bundle exec rubocop lib/gemstone/critranks/` — 21 files, no offenses
- Re-scanning the corpus after the fixes drops near-misses to zero (the only remaining hits are second-person lines, which are out of scope — these tables describe crits against creatures)

## Scope note

Second-person crits (a crit landing on the player) are deliberately **not** addressed. All 22 existing "you" patterns in the tables have the player as *attacker* ("your well placed strike"), never as recipient, so 2p messaging is out of scope by design.

### DIFF
--- a/lib/gemstone/critranks/acid_critical_table.rb
+++ b/lib/gemstone/critranks/acid_critical_table.rb
@@ -434,7 +434,7 @@ module Lich
                                   :slowed          => false,
                                   :wound_rank      => 1,
                                   :secondary_wound => nil,
-                                  :regex           => /^Drop of acid gets in .*? eye. Better flush it out./ },
+                                  :regex           => /^Drop of acid gets in .*? eye.  Better flush it out./ },
                            2 =>
                                 { :type            => "acid",
                                   :location        => "left eye",
@@ -511,7 +511,7 @@ module Lich
                                   :slowed          => false,
                                   :wound_rank      => 3,
                                   :secondary_wound => nil,
-                                  :regex           => /^Hit to eye empties the socket. How can something missing be so painful./ },
+                                  :regex           => /^Hit to eye empties the socket.  How can something missing be so painful./ },
                            6 =>
                                 { :type            => "acid",
                                   :location        => "left eye",
@@ -589,7 +589,7 @@ module Lich
                                   :slowed          => false,
                                   :wound_rank      => 3,
                                   :secondary_wound => { :location => "head", :wound_rank => 3 },
-                                  :regex           => /^Acid Flush! The .*? brains get washed right out the mouth./ } },
+                                  :regex           => /^Acid Flush!  The .*? brains get washed right out the mouth./ } },
           :right_eye  =>
                          { 4 =>
                                 { :type            => "acid",
@@ -1345,7 +1345,7 @@ module Lich
                                   :slowed          => false,
                                   :wound_rank      => 2,
                                   :secondary_wound => nil,
-                                  :regex           => /^Burn to the elbow eats through tendons. Muscles snap free./ },
+                                  :regex           => /^Burn to the elbow eats through tendons.  Muscles snap free./ },
                            6 =>
                                 { :type            => "acid",
                                   :location        => "left arm",
@@ -1364,7 +1364,7 @@ module Lich
                                   :slowed          => false,
                                   :wound_rank      => 2,
                                   :secondary_wound => nil,
-                                  :regex           => /^Acid dissolves the elbow ligaments. Arm swings in a very odd manner./ },
+                                  :regex           => /^Acid dissolves the elbow ligaments.  Arm swings in a very odd manner./ },
                            7 =>
                                 { :type            => "acid",
                                   :location        => "left arm",
@@ -1383,7 +1383,7 @@ module Lich
                                   :slowed          => false,
                                   :wound_rank      => 3,
                                   :secondary_wound => nil,
-                                  :regex           => /^The .*? arm is drenched in acid. Flesh and bones are reduced to a smoking slime./ },
+                                  :regex           => /^The .*? arm is drenched in acid.  Flesh and bones are reduced to a smoking slime./ },
                            8 =>
                                 { :type            => "acid",
                                   :location        => "left arm",
--- a/lib/gemstone/critranks/crush_critical_table.rb
+++ b/lib/gemstone/critranks/crush_critical_table.rb
@@ -700,7 +700,7 @@ module Lich
                                   :slowed          => false,
                                   :wound_rank      => 2,
                                   :secondary_wound => nil,
-                                  :regex           => /^Whoosh. Several ribs driven into lungs./ },
+                                  :regex           => /^Whoosh.  Several ribs driven into lungs./ },
                            6 =>
                                 { :type            => "crush",
                                   :location        => "chest",
@@ -1157,7 +1157,7 @@ module Lich
                                   :silenced        => false,
                                   :slowed          => false,
                                   :wound_rank      => 3,
-                                  :secondary_wound => { :location => "nerves", "wound_rank" => 3 },
+                                  :secondary_wound => { :location => "nerves", :wound_rank => 3 },
                                   :regex           => /^A mighty blow cleaves a swath through .*? back, taking the spine with it./ } },
           :right_arm  =>
                          { 0 =>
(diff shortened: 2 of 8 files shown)


mrhoribu (commented):
# PR Review: fix(gs): correct messaging drift in 13 crit table patterns (elanthia-online/lich-5#1547)

**Reviewed:** 8 files, +15/-15 at `0de8cee8` on 2026-09-07. Based on: full local clone (`origin` + PR ref fetched as `pr-1547`), all history available.
**Verdict:** Approve
**Summary:** Adjusts 13 regex patterns across 7 crit tables (`acid`, `crush`, `fire`, `impact`, `lightning`, `plasma`, `puncture`, `unbalance`) so they match live game text that has drifted from the wiki-transcribed wording (double spaces, missing articles/punctuation, case mismatches), plus one unrelated key-type fix in `crush_critical_table.rb`. Every changed pattern was individually verified against `CritRanks.parse` and produces exactly the intended single match with no new ambiguity; the only real issue is that the PR body's stated justification for the `secondary_wound` key fix doesn't match what's actually in the codebase.

## Minor

### 1. `secondary_wound` fix justification doesn't match the code — `lib/gemstone/critranks/crush_critical_table.rb:1160`

```ruby
:secondary_wound => { :location => "nerves", "wound_rank" => 3 },
```

The PR body says: *"`Processor#apply_secondary_wound` reads `secondary[:wound_rank]`, so that nerves wound was silently dropped."* No method named `apply_secondary_wound` exists anywhere in this repo (`grep -rn "apply_secondary_wound"` across `lib/` and `spec/` returns nothing), nor in `dr-scripts` or `scripts`. More broadly, `:secondary_wound` itself is never read anywhere in any of the four repos today — it's written into every table entry (156 occurrences) but has no consumer yet, so nothing is currently "silently dropped" by this bug.

That said, the fix is still correct and worth keeping: this was the **only** `secondary_wound` hash in the entire table set using a string key (`grep` across all `critical_table.rb` files confirms all 155 other instances use the symbol `:wound_rank`), so it's a genuine data-consistency defect that will bite whenever a consumer is eventually written. The change itself is safe and harmless regardless of the cited mechanism.

**Fix:** correct or drop the specific `Processor#apply_secondary_wound` claim in the PR description before merge, so the merged history doesn't assert a code path that doesn't exist. Worth a one-line ask to the author: is `apply_secondary_wound` planned/in a branch elsewhere, or was this description written from a different (perhaps future or misremembered) context?

## What this PR gets right

- **Every claimed fix was independently reproduced.** I wrote a standalone probe against `CritRanks.parse` for all 17 example lines quoted in the PR body (the drift-class examples plus the untouched control cases) and every one matches exactly the record it's supposed to, with no ambiguous multi-match results — including the trickiest one, the capitalized `Knocked back and stunned!` case, which works because `^` in Ruby anchors per-line and the lazy `.*?` prefix absorbs the leading sentence.
- **No collisions introduced.** The parser buckets patterns by first literal word (`lib/gemstone/critranks.rb:97`) and returns every bucket member that matches, so loosening a regex (adding `?`, `(?:...)?`, or a character class) is exactly the kind of change that could make two table entries match the same line. I checked every first-word bucket touched by this PR (`Massive`, `Hard`, `Perfect`, `Light`, `Stinging`, `Flame`, `Whoosh`, `Drop`, `Hit`, `Acid Flush`, and the `.*?`-prefixed residual bucket for `unbalance`) against sibling entries sharing that word — no new overlap.
- **`crush_critical_table.rb`'s `secondary_wound` key fix is the only non-string-key anomaly in the whole table set** (see Minor #1) — genuinely worth fixing even though its stated rationale doesn't hold up.
- Full spec suite (6335 examples) and `rubocop lib/gemstone/critranks/` both pass clean on the PR branch under Ruby 4.0.5, the version pinned in `.rubocop.yml`.
- The existing `critranks_spec.rb` regression suite (loads/indexes without raising, exhaustive-scan-vs-index consistency, whitespace stripping, empty-hash-for-non-crit-lines) all continue to pass unmodified.
- Precisely scoped: only the tables/entries with actual log-verified drift were touched; the PR resists the temptation to "fix" the unescaped-`.`-as-wildcard style already present throughout every other pattern in these files (pre-existing, out of scope here).

## Open questions

- `lib/gemstone/critranks/impact_critical_table.rb:2039` makes `the` optional for **left** hand (`Hard blow to(?: the)? left hand breaking bones.`) but the sibling **right** hand entry at `impact_critical_table.rb:1848` (`Hard blow to right hand breaking bones.`) is untouched and still requires no `the`. Is that asymmetry because the corpus scan only found drift on the left-hand variant, or is right-hand simply not yet exercised in the 200M-line corpus? **Settles it:** was a "Hard blow to the right hand..." (with article) ever observed in the logs, or just not yet? If it hasn't shown up, no action needed — the PR is deliberately data-driven rather than symmetry-driven, which is a reasonable and stated design choice.

## Coverage notes

Reviewed against a full local clone with complete history (`git log -S`, `git blame` all available) and the PR branch checked out into a worktree. Ran the actual Ruby regex engine against `CritRanks.parse` for every example line quoted in the PR description plus adjacent sibling patterns, to check for regressions the diff itself can't show (cross-table match ambiguity via the first-word index). Ran the full `bundle exec rspec` suite (6335 examples, 0 failures) and `rubocop` (21 files, no offenses) on the PR branch under the pinned Ruby 4.0.5. Did not have access to the original 200M-line session-log corpus the author scanned, so the drift-instance counts and "near-misses drop to zero after the fix" claim in the PR body are taken on trust rather than independently re-verified — that would require the raw logs, which aren't in this repo.

mrhoribu (approved):
looks good, everything now resolved.

PR #1556, opened 2026-09-04
feat(dr): non-destructive upsert for INV SEARCH / category scrapes (03 of 05)

MahtraDR:
**Part 03 of 05** — DR inventory-model series. Stacked on #1554 → #1555; review/merge those first. (Cumulative diff until predecessors merge.)

## What

`INV LIST` is complete, so it clears+repopulates GameObj. `INV SEARCH <word>` and `INV <category> full` are **filtered** listings, yet shared the clear-everything path — so `inv search pouch` wiped the whole model down to the matches. Split the trigger by header completeness; the `looking for …` scrapes now **upsert** via `GameObj.upsert_inv` (removes any prior placement of an id, then re-places) instead of clearing.

## Hardening from adversarial review

- **Anchored triggers**: `InventoryListStart`/`InventorySearchStart` are anchored to the stream-line start (search allows its one real leading `<roundTime/>`), so quoted/chat/book text can't open a mutating scrape. Prefixes verified against logs.
- **Non-destructive nesting (#5)**: a filtered scrape of a deeply-nested item no longer routes the intermediate container through `upsert_inv` with `name=nil` (which blanked its name and pulled a worn parent out of `@@inv`); the middle container is left alone in partial mode.
- **Mutex-guarded removal**: `upsert_inv`'s `reject!` runs under `@@index_mutex` so it can't race the off-thread prune sweep.
- **Interrupted-scrape guard**: an unclosed scrape resets on the next `<prompt>` so stray `<d cmd>` links aren't hijacked/upserted.

## Testing

Adversarial specs: quoted-line rejection, `<roundTime/>` acceptance, deep-nested partial non-destruction, prompt-reset (both flags), partial-vs-full routing, upsert move/dedup/`(closed)`. Header classification verified across all ~90 `inv` verbs. Full suite green; rubocop clean (also fixed two pre-existing `ExtraSpacing` offenses in `gameobj.rb`).

### DIFF
--- a/lib/common/gameobj.rb
+++ b/lib/common/gameobj.rb
@@ -363,6 +363,44 @@ module Lich
         end
       end
 
+      # Upserts an inventory item by id for a PARTIAL (filtered) scrape -- e.g.
+      # +INV SEARCH <word>+ or +INV <category> full+, which list only the
+      # matching items and so must not clear unrelated inventory the way a full
+      # +INV LIST+ refresh does. Any prior placement of +id+ is removed from
+      # +@@inv+ and every +@@contents+ list first, so an item that moved
+      # containers (or whose name changed, e.g. gained "(closed)") is refreshed
+      # in exactly one place with neither a duplicate nor a stale copy left
+      # behind. It is then (re)placed via {.new_inv}: worn (+container+ nil) into
+      # +@@inv+, otherwise into +@@contents[container]+.
+      #
+      # Because a filtered scrape only reports matches, it can add or relocate
+      # items but can never prove an item is gone -- removals are left to the
+      # next full refresh.
+      #
+      # The removal runs under +@@index_mutex+ (the same lock +find_or_create+
+      # holds for registry writes) so it cannot race the off-thread
+      # +prune_index!+ sweep that walks +@@inv+/+@@contents+. It is a separate
+      # acquisition from the +new_inv+ that follows -- Ruby's Mutex is not
+      # reentrant, and +new_inv+ takes the lock itself -- so remove and re-add
+      # are not one atomic step; that is fine because the parser thread is the
+      # sole writer and a reader can at worst momentarily not see the item.
+      #
+      # @param id        [Integer, String]
+      # @param noun      [String, nil]
+      # @param name      [String, nil]
+      # @param container [String, nil]
+      # @param before    [String, nil]
+      # @param after     [String, nil]
+      # @return [GameObj]
+      def self.upsert_inv(id, noun, name, container = nil, before = nil, after = nil)
+        str_id = id.is_a?(Integer) ? id.to_s : id
+        @@index_mutex.synchronize do
+          @@inv.reject! { |obj| obj.id == str_id }
+          @@contents.each_value { |list| list.reject! { |obj| obj.id == str_id } }
+        end
+        new_inv(str_id, noun, name, container, before, after)
+      end
+
       # Creates and registers a new reserve slot item.
       #
       # +@@reserve+ is +nil+ until the first reserve stream is seen; thereafter
@@ -877,8 +915,8 @@ module Lich
           npc = @@npcs.find { |n| n.id == id }
           next unless npc
           next if npc.status.to_s =~ /dead|gone/i
-          next if npc.name  =~ /^animated\b/i && npc.name !~ /^animated slush/i
-          next if npc.noun  =~ /^(?:arm|appendage|claw|limb|pincer|tentacle)s?$|^(?:palpus|palpi)$/i &&
+          next if npc.name =~ /^animated\b/i && npc.name !~ /^animated slush/i
+          next if npc.noun =~ /^(?:arm|appendage|claw|limb|pincer|tentacle)s?$|^(?:palpus|palpi)$/i &&
                   npc.name !~ /(?:amaranthine|ghostly|grizzled|ancient) kraken tentacle/i
           npc
         end
--- a/lib/dragonrealms/drinfomon/drparser.rb
+++ b/lib/dragonrealms/drinfomon/drparser.rb
@@ -37,17 +37,24 @@ module Lich
         Rested_EXP_F2P = %r{^<component id='exp rexp'>\[Unlock Rested Experience}.freeze
         TDPValue_XPWindow = %r{^<component id='exp tdp'>\s*TDPs:\s*(?<tdp>\d+)</component>}.freeze
         FavorValue_XPWindow = %r{^<component id='exp favor'>\s*Favors:\s*(?<favor>\d+)</component>}.freeze
-        # Header line emitted by the DR inventory verbs whose <d> output we
-        # scrape for item IDs. Two verbs share this scrape:
-        #   INV SEARCH <text> => "You rummage about your person, looking for ..."
-        #   INV LIST          => "You take a moment and rummage about your
-        #                          person, taking stock of your possessions..."
-        # Anchored to the start of the stream line, allowing only the optional
-        # leading <roundTime/> the search verbs emit, so quoted/chat/book text
-        # containing the phrase mid-line cannot open a scrape (cf. GameShutdown).
-        # Logs confirm the header is never otherwise prefixed: INV LIST is bare at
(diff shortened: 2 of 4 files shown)


mrhoribu (approved):
# PR Review: DR GameObj inventory-model series (elanthia-online/lich-5 #1555 → #1554 → #1556 → #1557)

**Reviewed:** 4-PR stack, full local clone of `elanthia-online/lich-5`, each PR diffed against its actual stacked parent (not against current `main`, which has advanced past the stack's branch point). Commits: `baf97ae7` (#1555), `ec115641` (#1554), `465aeaae` (#1556), `9b8e318a` (#1557). Reviewed 2026-09-07.
**Verdict:** Approve — all four, in stack order (#1555 → #1554 → #1556 → #1557).
**Summary:** This is a well-sequenced, well-tested fix to DR's `GameObj` inventory model: it makes item-name normalization case-insensitive, teaches the parser to populate `GameObj` from `INV LIST`, stops `INV SEARCH` from wiping the whole inventory model (a real pre-existing bug), and closes the last gap — items picked into a hand no longer leave a stale duplicate in their old container. Each PR's own "hardening from adversarial review" notes turned out to be accurate; I could not find a case they missed. I ran the full suite and rubocop against the stack's head (#1557) and both are clean.

## What this PR gets right

- **The stack order is exactly right and each PR is independently mergeable.** Diffing against the true stacked parent (not `main..head`) confirms #1555 is `+10/-2`, #1554 adds only the `INV LIST` commit, #1556 only the upsert-split commit, #1557 only the hand-reconciliation commit — the "cumulative diff" framing in each PR body is accurate, not a red flag.
- **#1556 catches a real, currently-shipped bug.** On today's `main`, `InventoryGetStart` only matches `INV SEARCH`'s "looking for" header, and that path already unconditionally calls `GameObj.clear_inv` / `clear_all_containers` (verified directly against `main:lib/dragonrealms/drinfomon/drparser.rb:471-473`). So `inv search pouch` wiping the whole model down to the matches is a genuine pre-existing bug, not something #1554 introduces — #1554 only *widens* the same clear-path to also cover `INV LIST` (correctly, since that scrape is complete), and #1556 is the PR that actually fixes the pre-existing `INV SEARCH` data-loss bug by splitting `COMPLETE` vs `PARTIAL` handling. Good sequencing: merging #1554 alone doesn't make anything worse than today.
- **The doubly-nested middle-container reasoning is correct and is exactly the kind of thing that's easy to get wrong.** I traced it by hand: `<d cmd='get #child in #mid in a watery portal'>` only reveals the child→`#mid` edge; `#mid`'s own line (`get #mid in a watery portal`) is what registers it with its real name. Synthesizing the `#mid`→parent edge from the child line would use `name=nil`, and since `find_or_create` (`lib/common/gameobj.rb:1307`) keys on `id|noun|name`, that's a genuine second entry, not a dedup hit — confirmed by the non-mocked regression test in #1554 that drives real `GameObj` (not a `new_inv` mock) and asserts exactly one entry.
- **#1557 fixes a real, currently-observable data-integrity bug for DR** (no passive container stream; containers only rebuilt on explicit `INV LIST`/`SEARCH`, per the PR body and confirmed by there being no other DR-side call that clears a single item's old placement on pickup). Before this PR, `lib/stash.rb` — a game-agnostic helper used by both games for hand/container reasoning (`free_hands`/`empty_hands`) — would see a picked-up item as simultaneously in-hand and still in its old container, since nothing removed the stale entry. `remove_inv_item`'s nil/empty-id guard is correctly reasoned and tested: empty hands stream a hand tag with no `exist` id, so this fires on every empty-hand update, and the guard order (`nil?` short-circuits before `.empty?`) matters given the `NilClass#method_missing` monkeypatch in this codebase.
- **GS gating is correct and idiomatic.** `XMLData.game =~ /^DR/` matches the established pattern used throughout `lib/` (e.g. `lib/lich.rb:995`, `lib/global_defs.rb:2610`, `lib/common/xmlparser.rb:505`), and grepping for other `GameObj.new_inv` call sites in `xmlparser.rb` (lines 1184, 1268) confirms GS's own inventory stream is untouched by this stack.
- **Mutex discipline is sound.** `upsert_inv`/`remove_inv_item`'s `reject!` over `@@inv` and `@@contents` runs under the same `@@index_mutex` that `find_or_create` and the off-thread `prune_index!` sweep use, so the removal can't race the TTL-based GC sweep. The non-atomicity between the removal and the following `new_inv` call is explicitly called out in the comment and is fine given the parser thread is the sole writer — I confirmed there's no other writer of `@@inv`/`@@contents` in this codebase.
- **Verified mechanically clean.** `bundle exec rubocop lib/common/gameobj.rb lib/common/xmlparser.rb lib/dragonrealms/drinfomon/drparser.rb` → no offenses. Full suite at the stack head: `6367 examples, 0 failures`.

## Minor

### 1. Orphaned `@@index` entries after a name-changing upsert are not cleaned up until TTL prune — `lib/common/gameobj.rb:384-388`

```ruby
def self.upsert_inv(id, noun, name, container = nil, before = nil, after = nil)
  str_id = id.is_a?(Integer) ? id.to_s : id
  remove_inv_item(str_id)
  new_inv(str_id, noun, name, container, before, after)
end
```

`find_or_create`/`upsert_inv` key `@@index` on `id|noun|name`. When a partial (`INV SEARCH`) scrape re-observes an item whose name changed (the PR's own example: `"soft gem pouch"` → `"soft gem pouch (closed)"`), the old key's `GameObj` instance is removed from `@@inv`/`@@contents` by `remove_inv_item` but its `@@index` entry is not removed — it just becomes unreachable from any live registry until `prune_index!`'s 900-second TTL sweep evicts it. This is consistent with how the rest of the registry/index system already handles objects that leave a registry (nothing special-cases it elsewhere either), so it's not a regression this stack introduces, just a pre-existing characteristic of the index design that a closed-over-time item now exercises more often. Not worth blocking on — flagging so it's not mistaken for new behavior if someone notices index growth during a session with heavy `INV SEARCH` use.

## Open questions

- `lib/dragonrealms/drinfomon/drparser.rb`'s dash-stripping (`stripped = server_string.strip.sub(/\A-\s*/, '')`, added in #1554) assumes DR always emits a single leading `-` regardless of nesting depth, distinguishing depth only via indentation. The PR body says this was verified against a real capture and I found no counter-example in the test fixtures, but I have no way to confirm from the repo alone that DR *never* emits `--` or similar for deeper nesting. **Settles it:** has this been observed against a real triple-or-deeper-nested container in game, or only the two depths in the fixtures?
- The `InventorySearchStart` regex (`^(?:<roundTime[^>]*/>)?You rummage about your person, looking for`) is claimed in #1556's PR body to have been "verified across all ~90 `inv` verbs" — I can't verify that claim from the repo, only that the two tested phrasings ("looking for pouch...", "looking for armor and shields...") match. **Settles it:** was this an exhaustive check against real game logs for all category verbs, or a sample? If sample, low risk given the shared prefix, but worth knowing before relying on it for every category.

## Coverage notes

Full local clone of `elanthia-online/lich-5`; each PR's own commit diffed against its true stacked parent branch (`merge-base(main, pr-1555)` for #1555, then each predecessor's branch tip for #1554/#1556/#1557) rather than against current `main`, since `main` has advanced with unrelated merges since this stack branched. Read full contents of every touched file (`lib/common/gameobj.rb`, `lib/common/xmlparser.rb`, `lib/dragonrealms/drinfomon/drparser.rb`) at the stack head, plus `lib/stash.rb`/`lib/gemstone/stowlist.rb`/`lib/gemstone/readylist.rb` as blast-radius callers of `GameObj.inv`/`GameObj.containers`. Ran `git log`/`git show` across the stack to confirm ancestry and to compare against `main` for the pre-existing-bug claim in #1556's body. Ran `bundle exec rubocop` (clean) and the full `bundle exec rspec` suite (6367 examples, 0 failures) against the cumulative stack head. Did not have access to real DR game logs beyond what's embedded in the PR bodies/specs, which is the basis for the two open questions above. This is a lich-5 PR, so the `scripts`-repo-only PR-title/changelog conventions don't apply; lich-5's `CHANGELOG.md` is release-please-managed (confirmed via `git log -- CHANGELOG.md`), not hand-edited per PR, so no housekeeping gap there.

PR #1558, opened 2026-09-05
feat(all): GUI add configurable frontend registry

therealatari:
## Summary

- add a dedicated **Frontends** tab to the GTK launcher for detected clients, executable overrides, extra arguments, and custom frontend definitions
- restore available-client radio buttons on Manual Login, with an always-available Custom option using Wrayth compatibility
- retain all configurable frontends and availability status in the Frontends tab and saved-entry editing
- preserve stable frontend identity and declared capabilities through GUI login, CLI login, detachable sessions, and child-process launch
- define each shipped frontend in its own native Ruby catalog file under `lib/common/frontend/`
- store machine-local frontend configuration in a separate, versioned `frontends.yml` rather than credential-bearing `entry.yaml`

## Motivation

The frontend detection changes in 5.20.0 coupled account/character creation to locally detected executables. On systems where no supported frontend was detected, these changes broke the GTK launcher's historical Wrayth default selection and its ability to configure a custom launch frontend.

This change fixes that at the source by separating three concerns:

1. the frontend catalog and protocol capabilities
2. machine-local discovery and launch configuration
3. account/character association with a stable frontend ID

## Catalog architecture

- `lib/common/frontend.rb` owns the registry, validation, deterministic loading, immutable catalog, and compatibility constants.
- Each shipped frontend is a direct Ruby Hash in its own file (`frontend/wrayth.rb`, `frontend/saga.rb`, and so on), using the same native Ruby vocabulary consumed by the registry.
- The complete built-in catalog is validated before registration, then sealed before user settings are applied.
- Wrayth retains the stable internal ID `stormfront` and the `wrayth` alias.
- `lib/common/front-end.rb` remains as a compatibility shim for existing third-party require paths.
- `frontends.yml` is intentionally limited to mutable machine-local overrides and user-created frontend definitions; the GTK editor never writes or executes Ruby source.

### GTK launcher

- The Frontends tab lists built-in and custom definitions with stable IDs and availability status.
- Its catalog/editor layout remains usable in compact saved windows and wraps capability controls as the vocabulary grows.
- Built-ins accept a machine-local executable override and additional CLI arguments.
- Custom definitions accept a label, command, working directory, arguments, and supported protocol capabilities.
- Manual Login shows detected clients as radio buttons plus **Custom**; it does not list unavailable clients.
- Custom automatically checks/reveals the command and working-directory controls. A blank command is rejected before Play authenticates or saves an entry. Its saved/protocol identity is `stormfront` (Wrayth), without requiring the Wrayth executable.
- Account Management and Saved Entry character creation retain registry-backed configuration dropdowns. Saved characters can be reassigned using the inline Frontend dropdown; Escape cancels the edit.
- Frontend settings refresh across launcher tabs without discarding selection or overwriting a saved reassignment during a subsequent Manual Login save.
- Removed the experimental GUI headless mode, listener-port controls, and their persisted fields. Existing headless/detachable CLI flags are unchanged.

### Launching and compatibility

- Existing saved frontend identifiers and aliases continue to resolve normally.
- Explicit per-character Custom Launch values retain precedence over registry-derived commands.
- Registered custom frontends can launch from saved entries and retain their identity through detachable/headless startup.
- The existing no-selection CLI/detachable default remains Profanity.
- Saga keeps its structured environment handoff and now honors configured executable overrides and additional arguments on macOS, Windows, and Linux.
- Avalon and existing Simutronics launch paths retain their platform-specific behavior.
- Windows configured additional arguments travel as an argv list, preserving empty strings, whitespace, quotes, backslashes, and literal shell metacharacters. Legacy Custom commands without separate arguments retain their existing execution semantics. Windows templates with separate arguments must name an executable; implicit shell operators require a wrapper.
- Existing child-launch selectors keep their legacy flags; custom registry IDs use `--frontend=ID`, preserving compatibility with #1520.

### Settings safety

- `frontends.yml` has a versioned, validated schema and bounded scalar/argument inputs.
- Writes are atomic and mode `0600` where supported.
- Unsupported future schema versions are preserved rather than overwritten.
- Built-in identities/capabilities cannot be replaced by user configuration.
- Custom IDs cannot collide with built-ins, aliases, or runtime registrations.
- Registry reloads are synchronized and roll back completely if replacement fails.

## Audit follow-up and #1520 compatibility

- P1: re-read saved entries before Manual Login's whole-collection write, and refresh its cache after frontend edits.
- P2: defer inline selection persistence until `edited`, discard pending changes on Escape, and retain the original entry identity across model refresh.
- P2: validate additional arguments as literal argv values; reject malformed lists instead of trimming/dropping values. The Windows path no longer uses POSIX escaping.
- P2: canonicalize aliases when checking duplicate destination entries.
- GUI-headless mode/port findings are eliminated by removing that experimental feature, not by changing CLI headless behavior.
- Added user configuration/schema examples, command trust-boundary documentation, and regression coverage in [saved-frontend-editing.md](docs/saved-frontend-editing.md).
- Removed unused `LoginHelpers.spawn_login` and its tests as in Doug's patch. Current branch includes upstream main at `94c76c44` and merges cleanly with #1520 at `64d4d15d`.
- Tested the combined #1520/registry session-launcher and login-helper code: 111 examples, 0 failures. This also caught and corrected a semantic test mismatch around legacy frontend flags.
- No optional broad complexity refactor was added.

## Verification

- full suite: 6514 examples, 0 failures (seed 44387)
- RuboCop: 50 changed Ruby files, no offenses
- real GTK under Xvfb: inline commit/cancel, account expansion and credential preservation, detected-client radios, Custom fallback and command-field visibility
- Windows CRT parsing vectors and a real child-process argv round trip on Linux
- `ruby -c lib/main/main.rb` and `git diff --check`: clean
- full-suite run uses a temporary DATA_DIR seeded with the checked-in `spec/fixtures/effect-list.xml`; local socket access is required by existing networking tests

Native Windows/macOS frontend launches have not been verified on this Linux host. Installed player clients and saved player accounts were not changed during these revisions.

## Summary by CodeRabbit

* **New Features**
  * Added a Frontends tab for configuring built-in and custom clients, launch commands, executable paths, working directories, and arguments.
  * Added frontend selection dropdowns for saved characters and manual login, including Custom launch support.
  * Added support for additional frontend clients and improved frontend detection across platforms.
  * Preserved credentials and favorites when changing a saved character’s frontend.

* **Bug Fixes**
  * Improved validation and handling of invalid frontend configurations and launch arguments.
  * Fixed GTK message dialogs so button responses are returned correctly.

* **Documentation**
  * Added guidance for configuring saved frontends, custom launches, command arguments, and local overrides.


therealatari:
Updated the existing branch with the agreed Manual Login UX and audit fixes.

- Manual Login now has detected-client radios plus **Custom**, which reveals the command/directory controls and requires a nonblank command. Custom uses Wrayth compatibility, not a GUI headless mode.
- Removed the GUI headless/port feature; existing CLI headless support remains unchanged.
- Retained the registry tab and saved-entry dropdown editing. Fixed stale cross-tab saves, commit-versus-cancel behavior, literal argument preservation, Windows argv handling, and alias-equivalent duplicate detection.
- Applied the obsolete login-helper removal from Doug's patch. The branch includes current main and merges cleanly with #1520 at `64d4d15d`. A combined-code test caught the legacy frontend flag expectation; existing clients now retain those flags while custom IDs use the registry-safe long form. Combined launcher/helper tests: **111 passing**.

Final local verification: **6,514 tests passing**, **50 changed Ruby files lint-clean**, and the real GTK synthetic-account smoke test passes. Updated the PR description and configuration documentation. No optional broad complexity refactor or installed-player-client changes were made.

Native Windows/macOS frontend launches still need platform verification; the Windows parsing vectors and real argv round trip were exercised on Linux. Ready for another pass.

OSXLich-Doug:
@coderabbitai - review complete PR submission

therealatari:
Addressed both final review findings in `0cce874c`: launch preparation now removes stale `CUSTOMLAUNCH`, `CUSTOMLAUNCHDIR`, and `CUSTOMLAUNCHARGV` fields before appending explicit values; Manual Login now preserves its prior entry cache and continues the authenticated launch if the quick-save reread fails. Added regressions for both paths. Full suite: 6,516 examples, 0 failures; focused RuboCop: 4 files, no offenses. I also updated the Motivation paragraph with Doug’s requested clarification.

OSXLich-Doug:
@coderabbitai - confirm the latest commits address the findings you reported.

OSXLich-Doug (approved):
Good add to the codebase.  Thank you!

mrhoribu (approved):
# PR Review: feat(all): GUI add configurable frontend registry (elanthia-online/lich-5#1558)

**Reviewed:** 49 files, +5194/-1066 at `0cce874c` on 2026-09-07. Based on: full clone (worktree of `pr-1558` against `origin/main` @ `94c76c44`, matching the PR's stated merge base), with `git log`/`git blame` available.
**Verdict:** Approve
**Summary:** Replaces the ad hoc, executable-detection-coupled frontend selection logic with a sealed, thread-safe frontend registry (`Frontend`), a separately-persisted machine-local settings store (`FrontendSettings`/`frontends.yml`), and a shell-injection-conscious launch-argument pipeline (`WindowsCommandLine`, `FrontendLauncher`). The most important thing about this PR is that its riskiest surface — turning user-supplied strings into a spawned process's argv — is handled carefully and is well tested, including a real child-process round trip for the Windows argv path.

## What this PR gets right

- **Registry integrity.** `Frontend.replace_user_configuration!` (`lib/common/frontend.rb:182`) snapshots `@registry`/`@aliases`/`@definitions`/`@user_frontend_ids` before applying a new configuration and rolls back completely on any `StandardError`, so a malformed `frontends.yml` can never leave the catalog half-updated. Built-ins are sealed and deep-frozen (`seal_built_in_catalog!`) before user settings can touch them, and `replace_user_configuration_unlocked!` explicitly refuses to let a custom definition claim a built-in id.
- **Windows argv parsing is correct, not just plausible.** `lib/common/windows_command_line.rb` implements the actual CRT backslash/quote rules (executable-path quoting vs. general-argument escaping, odd/even backslash-run handling before a `"`), and `spec/lib/common/windows_command_line_spec.rb` verifies it against the documented CRT test vectors plus a real spawned child process reading back `ARGV` via JSON — this is exactly the kind of test that would have caught a subtly wrong escaping rule.
- **The `escape_argument`/`render_connection` split correctly keeps `%host%/%port%/%key%` tokens unescaped through `Shellwords.escape` while still round-tripping arguments containing literal `=`.** I hand-verified the `gsub('\\=', '=')` trick against several adversarial inputs (embedded `\=`, multiple `=`, trailing `\\=`) and it round-trips correctly in every case (Shellwords always escapes `=` as exactly the two-character sequence `\=`, and any preceding literal backslash is independently escaped first, so the substring match never crosses an unintended boundary).
- **`frontends.yml` normalization is defense-in-depth, not just validation-on-write.** `FrontendSettings.normalize_*` bounds scalar length (`MAX_SCALAR_BYTES`), argument count (`MAX_ARGUMENTS`), rejects control characters, and re-validates on *load* as well as on save, so a hand-edited or corrupted file degrades to "ignore the bad entry and log" rather than crashing boot.
- **The `update_launch_settings` reassignment path** (`lib/common/gui/account_manager.rb:360`) re-reads `entry.yaml` fresh rather than trusting a UI snapshot, matches on `(char_name, game_code, custom_launch, old_frontend)` to avoid editing the wrong row after a concurrent change, and rejects a reassignment that would collide with another saved entry's frontend — directly addresses the stale-write class of bug the PR description says an earlier audit pass found.

## Minor

- None risen to a level worth blocking on.

## Nits

None found — `rubocop` is clean on all 49 changed Ruby files (see Coverage notes).

## Open questions

- `lib/common/frontend_locator.rb:114-121`: `FrontendLocator#resolve` now checks `definition.dig(:metadata, :configured_executable)` and, when present, always calls `resolve_configured` (a fresh `File.executable?`/stat check) *before* consulting `@cache`/`refresh` — bypassing the discovery cache entirely for any frontend with a configured executable override, regardless of the caller's `refresh: false`. In every call site I found (`frontend_manager_tab.rb:269`, `frontend_selector.rb:77`) this only runs once per tab population/refresh, not on a timer, so it's very unlikely to matter in practice. **Settles it:** is `resolve` ever called from a hot path (e.g., a periodic UI tick) for a frontend that has a configured executable override? If not, this is a non-issue and not worth changing.

## Coverage notes

- **Access:** Full clone via `git worktree` against `origin/main` (matches the PR's own stated merge base `94c76c44`), so `git blame`/`git log -S` were available and used to check for prior context on the newer files (all new, so no pre-existing history to reconcile against).
- **Files read in full:** `lib/common/frontend.rb`, `lib/common/frontend_settings.rb`, `lib/common/frontend_launcher.rb`, `lib/common/windows_command_line.rb` (plus its spec), `lib/common/front-end.rb`, and the diffs for `lib/common/frontend_locator.rb`, `lib/common/gui/account_manager.rb`, `lib/main/main.rb` (launch/spawn call sites), `lich.rbw`/`lib/lich.rb` (require ordering), and `docs/saved-frontend-editing.md` (cross-checked against the implementation, found accurate).
- **Files reviewed at diff-level only, not read in full:** `lib/common/gui/frontend_manager_tab.rb` (588 new lines), `lib/common/gui/account_manager_ui.rb`, `lib/common/gui/frontend_selector.rb`, `lib/common/gui/manual_frontend_selector.rb`, `lib/common/authentication/login_helpers.rb`, `lib/common/session_launcher.rb`, and the individual `lib/common/frontend/*.rb` catalog files — these are lower-risk (GTK view glue and static catalog data) and are covered by substantial new spec files (`frontend_manager_tab_spec.rb` at 445 lines, `account_manager_ui_spec.rb`, `account_frontend_edit_spec.rb`, `saved_frontend_workflow_spec.rb`), which I ran rather than re-deriving by hand.
- **Verification performed directly:**
  - `bundle exec rspec` on the 13 new/changed spec files for this PR: **389 examples, 0 failures**.
  - `bundle exec rubocop` on all 49 changed `.rb` files (Ruby 4.0.5, matching `TargetRubyVersion: 4.0`): **no offenses**.
  - Did **not** run the full 6514-example suite the PR description cites, nor the native GTK/Xvfb smoke test (`spec/native/frontend_inline_smoke.rb`) or Windows-specific manual verification — no Windows host or X server available in this environment. Taking the PR author's reported results for those at face value; the targeted subset above exercises the same code paths under the same Ruby/rubocop versions CI would use.
  - Manually traced the shell-execution path for a "Custom" frontend's launch command (`FrontendLauncher.command` → `with_additional_arguments`/`escape_argument` → `render_connection` → `spawn(launcher_cmd)` in `lib/main/main.rb:509`) to confirm the user-supplied command string is executed through the shell only for the frontend's *own* local, user-authored configuration (not attacker-influenced input) — consistent with the pre-existing "Custom Launch" trust model, not a new exposure.

PR #1559, opened 2026-09-06
fix(gs): combat module - event defs, observer emissions, replay-verified parsing, SQLite recorder

Nisugi:
## Summary

Continues the combat module work from #1523. Built and verified against a replay corpus of recorded live sessions.

- **Observer seam**: `Tracker.on`/`Tracker.off` named subscriptions with emissions in `persist_event`; hits model and def hardening. The `:attack` emit precedes the target-id guard, so recorder-class subscribers see every event while nothing is ever misapplied to a creature.
- **Definition growth**: attack defs split into assaults/spells; flare, sequence, outcome and resolution defs; spell losses, dispel flux, status ticks, positioning; AoE outcomes and hidden status; exchange-proven defs (Pain, burning pair, dispel blink, slump knockdown). Every pattern carries its evidence in comments.
- **Event coverage**: inbound (creature → us) and foreign-target events now emit; targetless attack-born events saved; orphan sink and self-status handling (replay 72.7% → 100.0%).
- **Attribution**: `foreign_caster` (a nearby player's attack we witness) and `unowned` (an actor-less effect tick) are classified and kept off our own damage rollup; a chunk-local foreign-AoE latch lets anonymous swing chains inherit foreign ownership. Adds the paladin `weapon_infusion` proc-cast, both glowbark `plasma_burst` messaging generations, the `:terror` murder-of-crows variant, and the 3p `pulverize` AoE form.
- **SQLite combat recorder**: `Combat::Recorder`, an opt-in library - NOT default functionality. Nothing records unless a script instantiates it and subscribes, and it additionally sits behind the `emit_attacks` tracker setting (off by default). When driven, it persists the observer stream to a per-character 7-table database (sessions/creatures/attacks/resolutions/flares/hits/statuses), with printed roll components stored verbatim and the d100 kept separate.
- **Critranks**: the crit-table edits in this diff are #1547 ("correct messaging drift in 13 crit table patterns") — a standalone PR this branch is based on top of and reviewed there, not new work authored here; once #1547 merges to `main` they drop out of this diff on rebase. (They restore drifted double-space sentence breaks, plus a few trailing `.` → `.?` guards for live lines seen missing their period.)

## Review response (#1559)

Addresses @mrhoribu's review:

- **Major** — the crit lookahead that fills `hit[:crit]` was gated on `track_wounds` alone, but `apply_crit_statuses` consumes it too, so `track_wounds:false + track_statuses:true` silently disabled status tracking. Gate is now `track_wounds || track_statuses`, pinned in both directions.
- **Major** — `PatternGate.rejects?` blanket-disabled its fast path the moment any short-literal pattern landed in `always_scan`, reverting a whole table to full-scan. Fixed to the correct idiom and the spec that had encoded the bug is rewritten to the real contract.
- **Minor** — dropped the dead duplicate `:prone` pattern and the dead spell-deflect copies in `:block` (they live in `:intercept`, which wins first-match); added a mutex over the recorder's public surface (worker-thread `record` vs. consumer-thread `check_idle!`/`close`); made `record_status` transactional like `record_attack`; removed three cross-repo working-note docs swept in by accident.

## Testing

- `bundle exec rspec` — **190 combat examples, 0 failures**; rubocop clean.
- **Replay regression net** (new): a curated 85-blob slice of the replay corpus lands under `spec/fixtures/replay/` (most-common real-feed shape per attack def, plus flare-bearing cases), driven through the real `Processor.parse_events` in CI and asserted against each blob's recorded facts. The previously-external 1000/1000 fidelity claim is now CI-enforced.
- `recorder.rb` gains end-to-end specs (session lifecycle, `:attack` payload unpack, status attribution, idle sessioning, transactional integrity, a threaded `record`/`check_idle!` race with `PRAGMA integrity_check`) — the file previously had zero coverage.
- Recorder verified row-level against raw session logs: aggregate damage lines reconcile exactly across five sessions (count and sum); random per-attack spot checks match resolution numbers, damage and crits verbatim.


mrhoribu (commented):
# PR Review: fix(gs): combat module - event defs, observer emissions, replay-verified parsing, SQLite recorder (elanthia-online/lich-5#1559)

**Reviewed:** 37 files, +6082/-206 at `361be39d` on 2026-09-07. Based on: full local clone (both `main` and the PR head, with history), full RSpec suite run, rubocop run.
**Summary:** A large continuation of the GS4 combat-parsing rewrite (#1523): it grows the regex "def" tables that recognize attack/status/outcome text, adds a pub-sub `Observers`/`Tracker.on`/`off` seam so scripts can subscribe to parsed combat facts, and adds an opt-in SQLite `Combat::Recorder`. The core event-assembly logic is sound and its most load-bearing claims (the `:attack` emit preceding the target-id guard, single-threaded ordered processing) check out against the code. The real problems are a status-tracking gate that silently goes dark under a specific settings combination, a perf-gate helper that degrades to a silent no-op the moment one future pattern triggers it, and a sizeable coverage gap under the newest, largest regex tables — plus several docs files that don't belong in this repo.

## Major

### `apply_crit_statuses` silently does nothing when `track_wounds: false` and `track_statuses: true` — [lib/gemstone/combat/processor.rb:930](lib/gemstone/combat/processor.rb:930), [:1084](lib/gemstone/combat/processor.rb:1084), [:740](lib/gemstone/combat/processor.rb:740)

```ruby
# processor.rb:740
if Tracker.settings[:track_wounds]
  (1..3).each do |offset|
    ...
    hit[:crit] = c.reject { |k, _| k == :regex }
    ...

# processor.rb:930-931
if Tracker.settings[:track_statuses]
  apply_crit_statuses(creature, event)

# processor.rb:1087-1088
event[:hits].each do |hit|
  crit = hit[:crit] or next
```

`apply_crit_statuses` (new in this PR) derives stun/roundtime/position/silenced/slowed/dazed/crippled/limb_favored status changes entirely from `hit[:crit]`. But `hit[:crit]` is only ever populated by the crit lookahead at line 740, which is gated on `Tracker.settings[:track_wounds]` — a *different*, pre-existing setting. Flip `track_wounds` off while leaving the new `track_statuses` on (both are independent, user-facing settings exposed by `Tracker.configure`, and disabling wound tracking for performance while keeping status tracking is a plausible real configuration) and every `hit[:crit]` is `nil`; `crit or next` fires on every hit and `apply_crit_statuses` becomes a permanent no-op with no error, no log line, and no indication anything is wrong. This directly contradicts the design rationale in the surrounding code, which explicitly rejected coupling status visibility to the *UCS* flag for exactly this reason ("gating the state on the UCS flag would mean a non-UCS character sees `silenced?` return false for a genuinely silenced creature, which is worse than having no answer at all") — the PR avoided that coupling and then introduced the identical failure mode via `track_wounds` instead. Under default settings both flags are `true`, so this doesn't fire out of the box, but it's a real, reachable footgun for the settings surface this PR itself documents. Confidence: confirmed by trace.

### `PatternGate.rejects?` degrades to "never reject" — silently, with no test protecting it — [lib/gemstone/combat/defs/pattern_gate.rb:83-85](lib/gemstone/combat/defs/pattern_gate.rb:83)

```ruby
def rejects?(gate, always_scan, line)
  always_scan.empty? && (gate.nil? || !gate.match?(line))
end
```

The intent (per the file's own header) is a ~7us literal-substring pre-check that avoids a ~0.5-1ms full regex-alternation retry on non-matching lines. But `rejects?` only returns `true` — i.e., only ever actually skips work — when `always_scan` (patterns whose literal was too short to gate on) is completely empty. The instant a def module has even one such pattern, `always_scan.empty?` is `false`, the whole expression short-circuits to `false`, and the caller falls through to the full un-gated scan for *every* line, for that *entire* module — not just for the ungated pattern, all of them. This is the wrong idiom: the correct check (used correctly elsewhere in this same PR, e.g. `assaults.rb:100`, `flares.rb:1102`) is `gate&.match?(line) || always_scan.any? { |rx| rx.match?(line) }`. Verified with `ruby -e` against the real `ATTACK_LOOKUP`, status `ALL_LOOKUP`, and `SpellLosses::LOOKUP` constants: all three currently produce an empty `always_scan`, so today `rejects?` happens to degrade to the correct simple case and there's no live bug. But it's a landmine with zero test coverage of its own contract: the next contributor who adds one short-literal pattern to a 400+-entry table (easy to do without noticing) silently reverts that whole table's fast path to full-scan, with nothing surfacing the regression. Given this PR's stated goal is making these tables the fast path, this should be fixed to the correct idiom or at minimum covered by a spec that exercises `rejects?` with a non-empty `always_scan`. Confidence: confirmed by trace and by loading the real constants.

## Minor

### Two dead/duplicate patterns, both introduced in this commit

- [lib/gemstone/combat/defs/statuses.rb:65](lib/gemstone/combat/defs/statuses.rb:65) and [:81](lib/gemstone/combat/defs/statuses.rb:81): the exact same regex `/(?<target>.+?) falls to the ground!/` appears twice in `:prone`'s pattern list (confirmed both lines are new in `361be39d` via `git blame`). The comment on the second copy ("bang-form generic knockdown... the period form is item-drop flavor, not this") frames it as a distinct pattern being added to complement an existing period-form variant — but there is no period-form (`falls to the ground\.`) pattern anywhere in the list, and the second copy is simply unreachable behind the first. Harmless (first-match-wins just never reaches it), but the comment actively misdescribes the surrounding code and should either be corrected or the missing period-form pattern actually added.
- [lib/gemstone/combat/defs/outcomes.rb:104-105](lib/gemstone/combat/defs/outcomes.rb:104) duplicated verbatim at [:206-207](lib/gemstone/combat/defs/outcomes.rb:206): the same two spell-deflected-by-barrier patterns appear in both `:intercept` and `:block`. `:intercept` is checked first in `Outcomes.parse` (first-match-wins), and the `:intercept` category's own doc comment argues this exact line is "categorically NOT `:block`" so hit/block-rate stats stay honest — making the `:block` copies not just dead but contrary to the table's own stated design. No functional effect (intercept always wins), but it's copy-paste cruft in a table meant to be the authoritative outcome taxonomy.

### Cross-thread SQLite access in `Combat::Recorder` has no synchronization — [lib/gemstone/combat/recorder.rb](lib/gemstone/combat/recorder.rb)

`record` is wired as an `Observers.on(:attack, ...)` callback, which fires from whatever thread calls `Observers.emit` — in practice the single `AsyncProcessor` worker thread. But the recorder's own docs say `check_idle!` and `close` are meant to be driven from "a periodic loop" on the *consuming script's* thread. Both paths touch the same `@db` handle with no mutex (contrast with `Observers` itself, which mutex-protects its subscriber lists throughout). A `close` racing an in-flight `record` on the worker thread can hit a closed handle from another thread; a `finish_session`/`UPDATE sessions` from the idle-check thread landing inside `record_attack`'s open transaction would be folded into (and rolled back with) that transaction if it later raises. `record`'s own rescue swallows `SQLite3::Exception` silently, so this would fail quietly rather than crash. This is suspected, not confirmed — it depends on the recorder actually being driven by a script with a separate idle-check loop, which is documented as the intended usage but has no first-party caller in this repo to trace against (see coverage gap below).

### `record_status` is not transactional, unlike `record_attack` — [lib/gemstone/combat/recorder.rb](lib/gemstone/combat/recorder.rb)

`record_status` does an unwrapped creature upsert, then a separate status insert, then possibly a creature update. A failure between the first two leaves an auto-vivified creature row with no corresponding status row — low practical impact, but an inconsistency with the transactional discipline correctly used for `record_attack` (verified against the installed `sqlite3` gem: `Database#transaction` does roll back and re-raise on exception, so the attack-write path is sound).

### Stray, out-of-scope planning documents committed under `docs/` — [docs/COMBAT_DEFS_PROPOSAL11.md](docs/COMBAT_DEFS_PROPOSAL11.md), [docs/COMBAT_DEFS_PROPOSAL12.md](docs/COMBAT_DEFS_PROPOSAL12.md), [docs/SPAWNS_PROPOSAL_2026-09-04.md](docs/SPAWNS_PROPOSAL_2026-09-04.md)

These three files (900+ lines combined) are dated, numbered "round" working notes about *per-creature messaging* content for a different, sibling repository — `SPAWNS_PROPOSAL_2026-09-04.md` opens with `For: lich-5-creature-spawns/lib/gemstone/creatures`. They contain raw diff-against-a-hash notes ("Diffed against the canonical files at `88633e76`") that read as scratch output from the author's own process, not reference documentation for lich-5. This repo's existing `docs/` directory (checked via `git log`) contains only durable material — hardening notes, protocol analysis, a style guide — with no precedent for numbered proposal rounds about another repo's content. `docs/COMBAT_DEFS_ONBOARDING.md`, by contrast, is genuinely useful lich-5 contributor documentation and is fine as-is. Worth asking the author whether the three proposal files were meant to go here or were swept in by accident.

## Test coverage gaps

The full RSpec suite passes (6390 examples, 0 failures) and rubocop is clean across the changed files, but coverage is uneven across the new surface:

- `lib/gemstone/combat/recorder.rb` (489 new lines of SQL/persistence logic — arguably the highest-risk new file in the PR) has **zero** spec coverage; no `recorder_spec.rb` exists, and `grep` for `Combat::Recorder`/`Recorder.new` across the whole worktree finds no instantiation outside the file's own doc comments — there is no first-party caller exercising the wiring end-to-end.
- `lib/gemstone/combat/defs/flares.rb` (365 lines, ~80 flare defs): only one flare has a direct spec.
- `lib/gemstone/combat/defs/outcomes.rb` (411 lines): only one narrow rider is spec'd; the bulk of the file (miss/evade/block/parry/resisted/unaffected) — including the duplicate flagged above — has no spec at all, which is exactly how that duplicate went unnoticed.
- `lib/gemstone/combat/defs/assaults.rb` and `sequences.rb` (both new files, 120 and 77 lines): zero specs.
- `lib/gemstone/combat/defs/statuses.rb`: the large set of new add/remove text patterns has no regex-level spec; the new `creature_base_spec.rb` coverage tests the status-tracking *mechanism* (case canonicalization, duration extension), not whether the new literal game-text patterns match.

None of this is a bug today, but it's the exact gap that let the two dead-pattern duplicates above ship unnoticed, and it means large parts of this PR have no regression net against a future live-wording change.

## Open questions

- `lib/gemstone/critranks/lightning_critical_table.rb:1348` and `plasma_critical_table.rb:1428`/`:1619` change a trailing literal `.` to `.?`. Since none of these patterns anchor on `$`, this is functionally a no-op against well-formed game text and doesn't match the PR's stated "restoring double-space sentence breaks" description for the critrank changes. **Settles it:** was this deliberately handling a variant with no trailing punctuation, or did the double-space-restoration tooling touch lines it shouldn't have?
- Five other critrank edits (`fire_critical_table.rb:163`, `impact_critical_table.rb:2036`, `puncture_critical_table.rb:584`, `unbalance_critical_table.rb:719`, and one more in `plasma`) are wording-variant additions rather than double-space fixes. They look plausible, but unlike the `defs/*.rb` tables — which cite a source log, round number, or sweep count for nearly every non-trivial pattern — these have no cited evidence. **Settles it:** is there a source log for these five, the same way the def-table entries are sourced?
- The PR description's headline verification claim — "100.0% of blobs (1000/1000) reproduce their recorded events" against a replay corpus — can't be checked from this checkout: no corpus, fixture directory, or replay harness script exists in the repo. **Settles it:** is that harness/corpus external and can a sample of its output be attached to the PR, or is it worth landing a slimmed-down version of it as a regression test?
- Does any script (in `dr-scripts`, `scripts`, or elsewhere) actually plan to instantiate `Combat::Recorder`, or is the calling script landing in a follow-up PR? If a caller is imminent, the zero-coverage finding above is temporary rather than a standing gap.

## What this PR gets right

- The inbound/foreign-target/self-attribution guards in `processor.rb` are excellent: each is grounded in a specific, dated, named replay failure, and `processor_inbound_spec.rb` pins each one directly (ogre self-damage, bystander-emote leak, summoned-prop misattribution, ambush self-targeting) — rare to see bug-fix comments this concretely falsifiable, and they hold up under trace.
- The PR's central claim that `Observers.emit(:attack, event)` precedes the target-id guard in `persist_event` (`processor.rb:857-859`) is true, not just asserted — verified directly, along with `Observers`' per-handler exception isolation and mutex-protected subscriber lists, which is solid pub/sub hygiene for a feed meant to run on a shared worker thread.
- The evidence-in-comments discipline in the regex def tables is real: the large majority of non-trivial patterns across `flares.rb`, `spell_losses.rb`, and `statuses.rb` cite a concrete source (log file, round-N sweep count, or a dated ruling), which is what makes a table this size reviewable at all.
- `Creature#add_stun_estimate`/`stun_rounds`/`clear_stun_estimate` is a careful, well-reasoned separation of an advisory crit-table timer from the authoritative `<crtrStatus>` boolean — "extends but never shortens," gated correctly, and exercised by real callers.
- `Combat::Recorder`'s schema is solid: indexes match the documented query patterns (per-session sequence, per-creature hit history, crit-rank distribution), and the storage design (verbatim printed rolls with `d100` kept separate) is deliberate and explained. All queries are properly parameterized — no SQL injection surface anywhere in the file.
- The `AsyncProcessor` single-worker-thread design is the right call and is documented as such: it's what makes the module-level state in `Processor` and mutation of `CreatureInstance` safe without additional locking, and this PR's debug-mode changes (`enable_debug!`/`disable_debug!` forcing `max_threads: 0` and restoring it afterward) correctly preserve that invariant.

## Coverage notes

Full local clone with history (`main` and PR head both present), not diff-only. Ran `bundle exec rspec` for the full suite (6390 examples, 0 failures, including the new/updated combat and creature specs) and `rubocop` against every changed non-doc file (48 files inspected, no offenses) myself. Three focused sub-reviews were run in parallel against non-overlapping scopes — `processor.rb`/`parser.rb`; `recorder.rb`; and `defs/*.rb` + `creature.rb` + `critranks/*` + their specs — each with git-history grounding (`git blame`/`git log -S`) and, for the defs review, direct `ruby -e` verification of the `PatternGate` claims against the real constants rather than trusting a read-through. I independently re-verified the two Major findings and the two dead-pattern Minors by re-reading the exact lines cited. Not independently verified: live-game-text accuracy of any individual regex pattern (no live session or access to the cited log corpus), and the claimed 1000/1000 replay-corpus result (no corpus or harness present in this checkout — see open questions). The recorder's cross-thread finding is marked suspected rather than confirmed because no first-party caller exists to trace the claimed periodic-loop usage against.

mrhoribu (commented):
# PR Re-Review: fix(gs): combat module - event defs, observer emissions, replay-verified parsing, SQLite recorder (elanthia-online/lich-5#1559)

**Reviewed:** update at `c405d708` (three new commits over the previously-reviewed `361be39d`), diffed against that prior review. Full RSpec suite (6503 examples, 0 failures) and rubocop (50 files, no offenses) run against the new head.
**Verdict:** Request changes — one confirmed new bug in the newly-added attribution logic, everything from the original review is resolved.
**Summary:** The author's fix commit (`757f5c96`) correctly resolves every finding from the first review — both Majors, both dead-pattern Minors, the recorder's thread-safety gap, and the stray docs — and a third commit (`c405d708`) lands a genuine replay-fixture regression suite answering the open question about the unverifiable corpus claim. But a second commit landed alongside the fixes (`19086526`, new foreign-caster attribution logic) introduces its own bug: `cast_owner` tracks DoT-cast ownership keyed only by spell name, with no per-target dimension, so a foreign player's DoT tick on one creature can silently inherit "ours" status and get miscounted into the player's own ledger if the player is also running the same DoT type on a different creature in the same processing chunk — the same class of misattribution this commit's own fix was written to prevent.

## Original findings — status

| # | Finding | Status |
|---|---|---|
| Major | `apply_crit_statuses` silently no-ops when `track_wounds:false`/`track_statuses:true` | **Fixed.** [processor.rb:822](lib/gemstone/combat/processor.rb:822) gate changed to `Tracker.settings[:track_wounds] \|\| Tracker.settings[:track_statuses]`. Verified by reading the diff directly. |
| Major | `PatternGate.rejects?` silently disabled the fast-path gate | **Fixed.** [pattern_gate.rb:83-87](lib/gemstone/combat/defs/pattern_gate.rb:83) now `return false if gate&.match?(line)` then `always_scan.none? { |rx| rx.match?(line) }` — the correct idiom. Its spec was rewritten to match. |
| Minor | Duplicate `:prone` pattern in statuses.rb | **Fixed.** Dead duplicate and its misleading comment removed. |
| Minor | Duplicate spell-deflect patterns in `:block` | **Fixed.** Removed from outcomes.rb, with a comment explaining why they belong only in `:intercept`. |
| Minor | Recorder cross-thread `@db` access with no synchronization | **Fixed, and well done.** A single non-reentrant `@mutex` now wraps every public entry point (`record`, `start_session`, `finish_session`, `check_idle!`, `close`), with `_locked` internal variants used for the calls made while the lock is already held — verified there's no re-entrant deadlock (the locked variants never call back through the public wrappers). |
| Minor | `record_status` not transactional | **Fixed.** Wrapped in `@db.transaction do...end`, matching `record_attack`. |
| Minor | Stray planning docs for a different repo committed under `docs/` | **Fixed.** `COMBAT_DEFS_PROPOSAL11.md`, `PROPOSAL12.md`, `SPAWNS_PROPOSAL_2026-09-04.md` all deleted. |
| Coverage gap | `recorder.rb` had zero test coverage | **Fixed.** New `spec/lib/gemstone/combat/recorder_spec.rb` (274 lines). |
| Open question | Unverifiable "1000/1000 replay corpus" claim | **Addressed.** New `spec/lib/gemstone/combat/replay_spec.rb` + 85 curated real-feed fixtures under `spec/fixtures/replay/` drive actual captured game text through the real `Processor.parse_events` and assert against an embedded `# expect:` oracle per blob. Spot-checked several fixtures — they're real captured feed text, not synthetic. This is a genuine regression net, not a rubber stamp. Minor nit: the spec's header comment references `tools/replay_harness.rb` for regenerating fixtures, which isn't part of this repo — a dangling reference to the author's external tooling, harmless but worth a doc tweak. |
| Open question | Critrank `.` → `.?` edits and unsourced wording-variant changes | **Not addressed** — no commits in this update touch `lib/gemstone/critranks/`. Still open; low stakes, doesn't block. |

## New finding (introduced by this update, not previously reviewed)

### `cast_owner` is keyed by spell name only, with no target dimension — foreign DoT ticks can inherit "ours" status when we're running the same DoT elsewhere in the chunk

**[lib/gemstone/combat/processor.rb:639](lib/gemstone/combat/processor.rb:639), [:684-687](lib/gemstone/combat/processor.rb:684), [:709-716](lib/gemstone/combat/processor.rb:709)**

```ruby
# processor.rb:709-716 — ownership recorded per SPELL NAME only
if UNOWNED_TICK_ATTACKS.include?(current_event[:name]) && current_event[:hits].empty?
  cast_owner[current_event[:name]] =
    if current_event[:foreign_caster] then (current_event[:attacker] && current_event[:attacker][:name]) || :foreign
    elsif current_event[:attacker].nil? && line.match?(/\AYou\b/) then :self
    end
end

# processor.rb:684-687 — consulted per SPELL NAME only, no creature id
unowned: (UNOWNED_TICK_ATTACKS.include?(attack[:name]) &&
          attack[:attacker].nil? && !eff_foreign &&
          !line.match?(/\AYou\b/) &&
          cast_owner[attack[:name]] != :self) || nil,
```

`cast_owner` is a flat `{spell_name => owner}` hash. Trigger: in one processing chunk, we cast `pestilence` on creature A ("You exhale a virulent green mist toward A..."), setting `cast_owner[:pestilence] = :self`. Later in the *same* chunk, a different player's pestilence ticks on creature B ("Boils rupture all over B causing 54 points of damage!") — a line that names the victim but no caster, exactly the shape `UNOWNED_TICK_ATTACKS` exists to catch. Its `eff_foreign` is `nil` (the tick line itself names no caster, and no `foreign_latch` was armed for this DoT), and `cast_owner[:pestilence]` is still `:self` from our unrelated cast on A — so `unowned` evaluates to `nil` too. The event has no `foreign_caster` or `unowned` flag, carries a real resolved target (creature B), and falls straight through `persist_event`'s guards, crediting B's 54 damage into our own damage rollup. This is the identical failure mode the commit's own message cites fixing ("Heavenscent's pulverize dumped ~825 swing damage into our web event") — just reached through the new `cast_owner` mechanism instead of the old unconditional one. Group hunts with overlapping AoE DoTs of the same spell family on multiple creatures are a normal, not edge-case, GS4 scenario.

This was independently reproduced with a throwaway spec constructing exactly this two-creature, same-chunk scenario: the tick event's `[:unowned]` came back `nil` instead of the expected `true`. Confirmed by both trace and execution.

**Fix direction:** key `cast_owner` (and the `cast_owner[attack[:name]] != :self` check) by `[spell_name, target_id]` rather than spell name alone, so ownership of a DoT is tracked per victim, not globally per spell.

## Minor (new, in the same commit)

### Pulverize's foreign-attacker capture regex is unusually strict, risking a coverage gap in the exact mechanism it's meant to protect

**[lib/gemstone/combat/defs/attacks.rb:80](lib/gemstone/combat/defs/attacks.rb:80)**

The 3rd-person attacker capture for pulverize uses `(?<attacker>[A-Z][a-z]+)`, unlike the overwhelmingly dominant `(?<attacker>.+?)` convention used everywhere else in this file — including `weapon_infusion`, added in this same commit. If a player name doesn't fit that narrow character class (can't be fully verified from this repo whether GS4 naming rules guarantee it always does), the opener line fails to match, `foreign_latch` never arms, and the anonymous swing/effect lines this whole feature exists to protect fall back to being attributed to "us" by default — silently reproducing the class of bug this commit fixes, via a regex gap rather than a logic gap. Confidence: suspected (the regex narrowness is confirmed; real-world reachability depends on name-format constraints outside this codebase).

## What this update gets right

- Every finding from the first review was actually fixed, not deflected or argued away — including the harder ones (the mutex/re-entrancy work in `recorder.rb` is genuinely correct, not just present).
- The replay-fixture regression suite is a real, substantive response to an open question that could easily have been waved off as "trust the external harness" — landing 85 real fixtures with an expect-oracle and wiring them into CI closes that gap properly.
- The commit messages are unusually precise about *why* each fix works, citing the exact mechanism of the bug being fixed (down to real-feed incident names and dates), which made this re-review faster and more confident.

## Coverage notes

Full local clone with history; ran the complete RSpec suite (6503 examples, 0 failures) and rubocop (50 files, clean) against the new head myself. The `cast_owner`/`foreign_latch` attribution logic (new, not covered by the original review) was reviewed via a dedicated pass that traced the latch lifecycle, the `our_2p` detection, and the `unowned`/`foreign_caster` field consumption by hand and reproduced the finding above with a throwaway spec (not committed). Did not re-verify the new `flares.rb` (`plasma_burst`/chain) or `attacks.rb` (`terror` murder-of-crows) def additions beyond confirming they're pure pattern additions with no attribution-logic involvement — low risk, not exercised further. Did not re-examine the critrank open questions since no commits touched those files in this update.

mrhoribu (commented):
# PR Re-Review, Round 3: fix(gs): combat module - event defs, observer emissions, replay-verified parsing, SQLite recorder (elanthia-online/lich-5#1559)

**Reviewed:** head at `15dc98f6` (two new commits over the previously-reviewed `c405d708`: `247ad509` spawn-tree lineage, `15dc98f6` review-response fixes). Full RSpec suite (6520 examples, 0 failures) and rubocop (26 changed-scope files, no offenses) run against the new head.
**Verdict:** Request changes — one confirmed, reproduced Major bug in the brand-new spawn-tree lineage feature. Everything from round 2 (the `cast_owner` per-spell-only keying bug and the pulverize regex gap) is correctly fixed. The recorder-side rollback-safety rework is solid, real work.

## Round 2 findings — status

| Finding | Status |
|---|---|
| `cast_owner` keyed by spell name only (foreign DoT tick on one creature could inherit "ours" from an unrelated cast on another creature) | **Fixed, verified.** [processor.rb:245-249](lib/gemstone/combat/processor.rb:245) now keys by `[spell_name, target_id \|\| target_name]` via a `cast_owner_key` lambda, applied consistently at both the read site ([:737](lib/gemstone/combat/processor.rb:737)) and write site ([:813](lib/gemstone/combat/processor.rb:813)). My original two-creature repro scenario no longer collides. |
| Pulverize's narrow `[A-Z][a-z]+` attacker capture could fail to arm the foreign latch for names with apostrophes/hyphens | **Fixed.** [attacks.rb:80](lib/gemstone/combat/defs/attacks.rb:80) now uses the file's standard `(?<attacker>.+?)` convention. |

## New finding (introduced by this round, not previously reviewed)

### Multi-target AoE attacks don't actually share spawn-tree lineage — contradicts this round's own stated verification

**[lib/gemstone/combat/processor.rb:471-537](lib/gemstone/combat/processor.rb:471) (target-switch branch) vs. [:676-798](lib/gemstone/combat/processor.rb:676) (attack branch)**

This round adds spawn-tree lineage (`root_ref`/`parent_ref`/`parent_confidence`) so a recorder can compute "correct per-shot spawn-tree damage totals" — its own stated purpose. The commit message claims: *"Verified on a real 460-attack log: 0 NULL roots; the only non-self-roots are genuine multi-target AoE siblings (fire/volley/mount_collapse across distinct creatures)."*

That claim doesn't hold. When a line both switches target and matches a new attack def — the documented shape for AoE-per-target lines like volley's per-arrow hits — the target-switch branch builds a placeholder that correctly *inherits* `root_ref`/`parent_ref` from the prior sibling (`processor.rb:493-495`, `root_ref: current_event && current_event[:root_ref]`). But the attack branch then unconditionally discards that placeholder and builds a fresh `current_event` hash whose lineage is recomputed from scratch, purely from `active_spawn`/`spawn_root` (`processor.rb:786-798`) — it never looks at what the discarded placeholder inherited. For a plain multi-target volley with no `active_spawn` in play, every arrow after the first lands in the `else` branch and becomes its own independent root instead of a child of the first arrow.

This was confirmed by execution, not just by reading: running `Processor.parse_events` against a synthetic two-target volley chunk ("An arrow pierces orc1! ... 20 points of damage!" / "An arrow pierces orc2! ... 15 points of damage!") produced `root_uid=0` and `root_uid=1` — two independent roots, not the shared lineage the feature is built to provide. The net effect is silent: nothing crashes, but `root_attack_id` — the whole point of which is correct per-shot damage rollups — undercounts every multi-target AoE attack by fragmenting it into N single-hit trees. The new spec additions (`processor_inbound_spec.rb:555-629`) cover a lone swing, blink's bracketed cast, and an unbracketed follow-on shot, but never construct the two-different-creature-ids case the commit message claims was verified — which is exactly why this shipped unnoticed.

**Fix direction:** when the attack branch detects it's replacing a same-line switch-artifact (`current_event[:_line] == index`), carry that artifact's `root_ref`/`parent_ref`/`parent_confidence` forward into the new event hash instead of recomputing lineage from `active_spawn`/`spawn_root` — the same pattern already used to carry `:resolutions` forward on the same line (`processor.rb:684-686`).

## Recorder-side rollback safety — solid, verified work

The `in_txn` staging mechanism (`recorder.rb:379-395`) that fixes the dangling-cache-entry risk from round 2 is correctly implemented: `@pending_cache`/`@pending_chunk_rows` are only merged into the live `@creature_cache`/`@chunk_rows` maps *after* the SQLite transaction block returns successfully; an exception mid-transaction skips straight past the publish step, so a rolled-back DB write can never leave the in-memory cache pointing at a row that doesn't actually exist. This is proven by real tests, not mocked around — `recorder_spec.rb:274-335` forces genuine `SQLite3::Exception`s mid-transaction (via `and_wrap_original`, so the actual rollback machinery runs unmodified) and asserts against a second, independent SQLite connection that no dangling reference was left behind. No double-transaction or re-entrancy issue found (`record_attack`/`record_status` are mutually exclusive under one non-reentrant mutex).

Two small notes, neither blocking:
- SQLite's `PRAGMA foreign_keys` is never enabled on the connection, so the original "dangling FK" risk was never a hard failure to begin with — it would have silently written orphaned rows rather than raising. The staging fix still matters for data correctness, just worth knowing it wasn't preventing a crash.
- **Open question:** when a `parent_ref`/`root_ref` can't be resolved to a database row (an event that was rolled back or never emitted), the child silently self-roots rather than surfacing the gap. That fallback is deliberately tested for the *parent-missing* case — but does `processor.rb` guarantee every `root_uid` a child references is always itself emitted under a given recorder config, or can a self-rooted child still fold its damage into the wrong (its own) total in some configuration? Worth a one-line confirmation from the author, especially since the fix for the Major above will touch this exact object-identity machinery.

## What this round gets right

- All four crit-fidelity fixes named in the commit message (the `crit_rank`/`wound_rank` field split — recorder-only, confirmed processor.rb never touches `crit_rank`; the `emit_attacks` addition to the crit-lookahead gate, verified to have no unintended side effect on wound/status *application*, which stays independently gated; flare-hit crits now correctly emitting statuses attributed to the flare's own target; and targetless/inbound attacks with reactive flares no longer dropping the flare loop entirely) all hold up under hand-trace against constructed boundary cases.
- The recorder's transactional-cache-staging fix is genuinely good engineering, proven by tests that force real rollbacks rather than asserting against mocks.
- `spawn_root`/`foreign_latch`/`cast_owner` are all confirmed still correctly chunk-local with no cross-chunk leakage.
- The stated conservatism about not guessing lineage for ambiguous mirror/afterimage echoes is honest — the only path that stamps non-self lineage is the game-declared `active_spawn` bracket case, there's no hidden positional-guess path elsewhere.

## Coverage notes

Full local clone with history; ran the complete RSpec suite (6520 examples, 0 failures) and rubocop (26 files touched by this round, clean) against the new head myself. Two focused sub-reviews ran in parallel on non-overlapping scope: `processor.rb`'s crit-fidelity fixes and the new spawn-tree lineage logic (which reproduced the Major finding above by direct execution against a synthetic chunk); and `recorder.rb`'s rollback-staging rework plus its spec (which independently verified against the actual `sqlite3` gem's transaction semantics, not assumptions). `cast_owner` and pulverize fixes were spot-checked directly by me rather than re-litigated at length, since they were the clearly-scoped round-2 asks and check out exactly as the commit message describes.

Nisugi:
Thanks — the AoE-lineage Major is fixed in e007d80c. The switch-artifact's inherited `root_ref`/`parent_ref`/`parent_confidence` are now captured before the attack branch rebuilds the event and carried forward (same pattern as `:resolutions`), instead of being recomputed from `active_spawn`/`spawn_root`. Verified on the real 460-attack log: children (root != id) went 13 → 56 — the extra 43 were exactly the AoE per-target siblings that had been self-rooting. Added the two-different-creature-ids spec that was missing.

On the open question (can a self-rooted child fold damage into the wrong total under some config): no. A child's `root_uid` is set as `uids[root_ref] || i` where `root_ref` is either the event itself or `spawn_root` — and `spawn_root` is always an own attack emitted *earlier in the same chunk*. So `root_uid` is always ≤ the child's own uid and references an event emitted before it, which is therefore already in `@chunk_rows` when the child records. The only way `chunk_row` returns nil is a **rolled-back** root, and in that case self-rooting is the correct/safe outcome — the intended parent row genuinely doesn't exist, so folding the child's damage into its own (now-independent) tree is accurate, not a leak into an unrelated total. `emit_attacks` off can't produce the gap either: with it off `record` never fires, so there are no children to orphan.

Good catch on `PRAGMA foreign_keys` — you're right it was never enabled, so the pre-fix risk was silent orphan rows rather than a raise. The staging fix stands on data-correctness grounds regardless.

mrhoribu (commented):
# PR Re-Review, Round 4: fix(gs): combat module (elanthia-online/lich-5#1559)

**Reviewed:** head at `e007d80c` (one new commit over the previously-reviewed `15dc98f6`). Combat spec suite (208 examples, 0 failures) and rubocop (both touched files, clean) run against the new head.
**Verdict:** Request changes — the round-3 bug is correctly fixed, but the fix introduces an adjacent, unguarded gap of the same shape it just closed.
**Summary:** The fix for round 3's multi-target-AoE lineage bug is correct and well-tested for the case it targets. But the new code path it adds — carrying a switch-artifact's inherited lineage forward — has no check for whether the *new* event replacing that artifact is itself foreign/inbound/unowned, so a nearby player's own attack landing on a different creature in the same chunk as one of our attacks can get silently grafted onto our spawn tree, corrupting `root_attack_id` rollups the same way the just-fixed bug did, just in the opposite direction (over-counting instead of under-counting).

## Round 3 finding — status

**Multi-target AoE attacks fragmenting into independent spawn-tree roots: fixed for the reported case.** [processor.rb:684-699](lib/gemstone/combat/processor.rb:684) now captures the switch-artifact placeholder's `root_ref`/`parent_ref`/`parent_confidence` before it's discarded, and [:799-807](lib/gemstone/combat/processor.rb:799) carries it forward onto the replacing attack event — exactly the fix direction I suggested last round, and it mirrors the existing pattern already used for `:resolutions` a few lines above. The new spec ([processor_inbound_spec.rb:629-651](spec/lib/gemstone/combat/processor_inbound_spec.rb:629)) builds the exact two-different-creature-ids case (fire at an orc, then a troll) and asserts `root_uid` equality across both — a real behavioral check, not a rubber stamp. I re-ran it directly: 208/208 green, rubocop clean.

## New finding (introduced by this fix)

### The lineage-carrying branch has no exclusion for foreign/inbound/unowned events, unlike its sibling branches — a nearby player's attack can get grafted onto our spawn tree

**[lib/gemstone/combat/processor.rb:799-807](lib/gemstone/combat/processor.rb:799)**

```ruby
if switch_artifact_lineage
  current_event[:root_ref] = switch_artifact_lineage[:root_ref]
  current_event[:parent_ref] = switch_artifact_lineage[:parent_ref]
  current_event[:parent_confidence] = switch_artifact_lineage[:parent_confidence]
elsif active_spawn && spawn_root &&
      !(current_event[:inbound] || current_event[:foreign_target] ||
        current_event[:foreign_caster] || current_event[:unowned] || current_event[:_orphan])
  ...
else
  spawn_root = current_event unless current_event[:inbound] || current_event[:foreign_target] ||
                                    current_event[:foreign_caster] || current_event[:unowned] || current_event[:_orphan]
  current_event[:root_ref] = current_event
end
```

Both the `elsif` and `else` branches explicitly guard against grafting lineage onto (or self-rooting as a linkable node for) an `inbound`/`foreign_target`/`foreign_caster`/`unowned`/`_orphan` event — the file's own stated rule is "inbound/foreign/orphan events are their own root regardless" ([:798](lib/gemstone/combat/processor.rb:798)). The new `if switch_artifact_lineage` branch has no such guard: it grafts the captured lineage unconditionally, regardless of what the new `current_event` turns out to be.

This is reachable with ordinary group-combat text, not a contrived input. `foreign_caster` is set in `parser.rb:110-113` purely from the attacker being a nearby player (a negative-id link or a bare name), independent of whether that same line also functions as a target switch. So: we have an open attack on creature A (current_event is ours, not foreign); a nearby player's own third-person attack lands on a *different* creature B in the same chunk (`"<Heavenscent> swings a warhammer at <troll>!"`). That line both (a) switches target from A to B — the switch guard at `processor.rb:471-473` only checks the *old* current_event's foreign status, which is fine (ours) — building a placeholder that inherits our lineage, and (b) matches an attack def whose attacker is a player, so `attack[:foreign_caster]` comes back true for the new event replacing that placeholder. I reproduced this directly against `Processor.parse_events`: the resulting foreign event carried `foreign_caster: true` and `root_ref` pointing at our own attack on A.

`persist_event`'s existing `return if event[:foreign_caster]` guard still keeps this event off our creature-application ledger, so no damage gets mis-credited to a creature. But `recorder.rb` resolves `root_ref`/`parent_ref` object references to database row ids independently of ledger application — this event's `parent_attack_id`/`root_attack_id` in storage would point into our own attack tree, meaning a foreign player's attack shows up as a child of our root in the stored spawn-tree analytics. This is the mirror image of the bug just fixed: instead of one real AoE fragmenting into too many roots, an unrelated foreign attack now merges into one root it doesn't belong to. Multi-player group hunts routinely interleave our attack lines with other players' third-person combat text against other creatures in the same room within a single chunk — this is ordinary play, not an edge case.

The new spec added in this commit only builds the same-owner two-creature case; nothing in the 208-example suite constructs a foreign/inbound/unowned event landing in this branch.

**Fix direction:** add the same exclusion used by the other two branches: only take the `switch_artifact_lineage` path when the new `current_event` is not itself `inbound`/`foreign_target`/`foreign_caster`/`unowned`/`_orphan` — e.g. `if switch_artifact_lineage && !(current_event[:inbound] || current_event[:foreign_target] || current_event[:foreign_caster] || current_event[:unowned] || current_event[:_orphan])`, falling through to the existing `elsif`/`else` self-rooting logic otherwise.

Confidence: confirmed by trace and by independent execution against a constructed repro chunk.

## What this round gets right

- The fix for the originally-reported bug is exactly the right shape and is properly tested with a real two-different-creature assertion, not a superficial change.
- The commit message is admirably honest about its own prior mistake ("My earlier commit's 'verified on a real log' claim was wrong: the spec never built the two-different-creature-ids case, so it shipped unnoticed") rather than glossing over it.
- Scope discipline: this commit touches only the two files needed for the fix, with no incidental refactoring or drive-by changes to review.

## Coverage notes

Full local clone with history. Ran `bundle exec rspec spec/lib/gemstone/combat/` (208/208) and rubocop against both changed files myself (clean). The new foreign-graft finding was verified two ways: independently by me re-reading `processor.rb:684-699`/`:799-820` and `parser.rb:96-113` directly, and by a dedicated verification pass that additionally reproduced it by executing `Processor.parse_events` against a constructed chunk mixing our own attack with a nearby player's third-person attack on a different creature (throwaway script, not committed). Did not re-examine `recorder.rb` or `defs/*.rb` in this round since neither changed in this commit.

mrhoribu (approved):
# PR Re-Review, Round 5: fix(gs): combat module (elanthia-online/lich-5#1559)

**Reviewed:** head at `25df66d7` (two new commits over the previously-reviewed `e007d80c`: `9db15a94` fixes the round-4 finding, `25df66d7` is an unrelated small fix for UCS positioning persistence). Combat spec suite (211 examples, 0 failures) and rubocop (26 files, clean) run against the new head.
**Verdict:** Approve. No outstanding findings above Nit.
**Summary:** The round-4 gap is fixed correctly and defensively — not just patched for the reported case but restructured so the same class of bug can't recur. The second commit is a small, well-scoped, well-tested fix for an unrelated pre-existing issue (UCS positioning tiers being silently dropped by the recorder) with no interaction with the spawn-tree/attribution logic from prior rounds. Five rounds of review have now converged: every confirmed finding has been fixed and verified, and no new issues surfaced in this round's changes.

## Round 4 finding — status

**Foreign attacks grafting onto our spawn tree via the switch-artifact lineage branch: fixed, and fixed well.** [processor.rb:799-822](lib/gemstone/combat/processor.rb:799) now computes a single `not_ours` flag (`current_event[:inbound] || current_event[:foreign_target] || current_event[:foreign_caster] || current_event[:unowned] || current_event[:_orphan]`) once, and gates all three lineage branches on it (`if switch_artifact_lineage && !not_ours`, `elsif active_spawn && spawn_root && !not_ours`, `else ... unless not_ours`). This is a better fix than the minimal patch I suggested — hoisting the exclusion into one shared flag structurally prevents the exact failure mode that caused this bug in the first place (the commit message is candid about it: "I added a third branch to a guarded conditional and omitted the guard"). The new spec ([processor_inbound_spec.rb:654-681](spec/lib/gemstone/combat/processor_inbound_spec.rb:654)) reconstructs precisely the scenario from my round-4 report — our attack on one creature plus a nearby player's attack on a different creature in the same chunk — and asserts the foreign attack self-roots (`root_uid == its own uid`, `parent_uid` nil) rather than becoming our child. I re-verified this is a real assertion, not a rubber stamp, and confirmed the guard logic directly by re-reading the diff.

## New commit this round (25df66d7) — reviewed, no issues found

This commit is unrelated to the attribution/lineage work from prior rounds — it fixes UCS positioning-tier persistence. Before this fix, `recorder.rb`'s `:ucs` handling only stored a numeric `value` (`data[:value].is_a?(Numeric) ? data[:value] : nil`), but UCS position events (`:position`/`:position_inbound`) carry a string (`'decent'`/`'good'`/`'excellent'`), so every stored `'position'`/`'position_inbound'` status row had a `nil` value column — silently useless for any query needing the tier. The fix adds a `POSITION_TIERS` ordinal map (`decent: 1, good: 2, excellent: 3`) in `defs/ucs.rb`, threads a new `tier:` field through `Observers.emit(:ucs, ...)`, and has the recorder's new `ucs_value` helper prefer the numeric `tier` when present, falling back to the old numeric-only check otherwise (preserving existing behavior for `:tierup`/`:smite_on`/`:smite_off`, which correctly still persist `nil` for their string/non-numeric values per the new spec).

I checked the one thing worth checking here: `apply_ucs_to_target`'s `case` statement previously had no branch for `:position_inbound` at all — but since the `Observers.emit(:ucs, ...)` call sits *after* the `case`, unconditionally, that event was already being emitted before this fix; the new explicit `when :position_inbound` branch only adds a debug-log line and a comment, it doesn't change what reaches Observers. So there's no behavioral regression risk from "handling it explicitly now" — it was already flowing through, just undocumented. The actual bug (and fix) is entirely on the recorder-persistence side, and it's well-tested: [recorder_spec.rb:404-415](spec/lib/gemstone/combat/recorder_spec.rb:404) asserts against real stored rows that `position`/`position_inbound` now persist `2`/`3` instead of `nil`, while `tierup` (no tier) still correctly persists `nil`.

No interaction with the spawn-tree/foreign-caster/cast_owner logic reviewed in prior rounds — this commit doesn't touch `active_spawn`, `foreign_latch`, `cast_owner`, or the lineage branches at all.

## What this round gets right

- The round-4 fix doesn't just patch the reported case — it restructures the code so the three lineage branches share one exclusion flag, closing off the exact mechanism that let the bug ship in the first place (a fourth branch added later without the guard would need to actively ignore `not_ours` to reintroduce this).
- Both commits include tests that reconstruct the actual reported scenario (not just a simplified proxy for it) and assert on the specific behavior that was broken.
- The UCS fix is appropriately scoped — it doesn't touch unrelated code, and it correctly preserves existing behavior for the UCS kinds that were never broken.

## Coverage notes

Full local clone with history. Ran the complete combat spec suite (`bundle exec rspec spec/lib/gemstone/combat/`: 211 examples, 0 failures) and rubocop against all files touched since the last review (26 files, no offenses) myself. Verified both commits by reading their full diffs directly rather than relying on the commit messages, including re-tracing the `not_ours` guard logic against the round-4 repro scenario and confirming the UCS recorder fix's `ucs_value` fallback preserves non-position UCS kinds' existing null-persistence behavior. Did not re-run the full non-combat RSpec suite or the author's external replay-corpus harness for this round, since neither of these two commits touches code paths outside `lib/gemstone/combat/`.

PR #1561, opened 2026-09-07
fix(all): normalize colons in Lich::Util.normalize_lookup

mrhoribu:
## Summary
- `Lich::Util.normalize_lookup` normalized underscores to spaces on the lookup value but did nothing about colons in effect hash keys (e.g. `"Covert Art: Escape Artist"`), so `PSMS.available?("covert_art_escape_artist")` never matched that cooldown entry and reported the maneuver as unavailable/on cooldown even when it wasn't.
- Both the lookup value and the effect registry's hash keys are now normalized the same way (colons and underscores both collapsed to spaces, then squeezed/stripped) before comparing, so punctuation-free PSM feat names correctly match punctuated effect keys.
- Merged the previously-identical `String`/`Symbol` branches and updated the method's doc comment to describe the new normalization and why both sides need it.

## Test plan
- [x] Added specs in `spec/lib/util/util_spec.rb` covering: a punctuated effect key matched by an underscored PSM-style lookup, a plain punctuation-free match, `Symbol` lookup values, a non-matching lookup, the `Integer`/`active?` delegation path, and the `RuntimeError` for unsupported types.
- [x] `bundle exec rspec spec/lib/util/util_spec.rb` — 17 examples, 0 failures
- [x] `bundle exec rspec` (full suite) — 6341 examples, 0 failures
- [x] `rubocop lib/util/util.rb spec/lib/util/util_spec.rb` — no offenses

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved lookup matching for names containing colons, underscores, or repeated whitespace.
  - String and Symbol lookups now behave consistently.
  - Unmatched and unsupported lookup values continue to return clear results or errors as appropriate.

- **Tests**
  - Added coverage for punctuation, whitespace normalization, Symbol lookups, unmatched values, numeric lookups, and unsupported types.

### DIFF
--- a/lib/util/util.rb
+++ b/lib/util/util.rb
@@ -12,9 +12,13 @@ module Lich
     # Normalizes and performs a lookup for an effect based on the provided value.
     #
     # Depending on the type of `val`, this method will:
-    # - For String: Check if the normalized string matches any key in the effect's hash (case-insensitive, underscores replaced with spaces).
+    # - For String: Check if the normalized string matches any key in the effect's hash (case-insensitive, underscores/colons replaced with spaces).
     # - For Integer: Check if the effect is active for the given integer value.
-    # - For Symbol: Check if the normalized symbol matches any key in the effect's hash (case-insensitive, underscores replaced with spaces).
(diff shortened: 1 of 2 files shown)


OSXLich-Doug (approved):
Good catch to normalize colon as well.

PR #1563, opened 2026-09-07
fix(dr): make INV LIST refresh atomic and warn on interruption (05 of 05)

MahtraDR:
**Part 05 of 05 — DR inventory-model series. Depends on #1557** (stack order: #1555 → #1554 → #1556 → #1557 → this). Review/merge the stack first; this PR builds on #1554's `INV LIST` path and #1556's staging mechanism, so its diff will only be minimal once the stack has merged.

Two related robustness fixes surfaced by an adversarial audit of the series, both in the shared `GameObj` inventory model.

## 1. Atomic `INV LIST` refresh + interruption notice

The COMPLETE `INV LIST` path added in the series cleared `GameObj.inv` and every container **up front**, then repopulated. An interrupted or truncated listing — a `<prompt>` arriving before the closing `<output class=""/>`, the exact case the existing safety valve at the prompt was added for — left the model wiped and only half-filled, **silently**, until the next successful scrape. Nothing here is persisted, so this is a torn/partial read of the in-memory model rather than true data loss — but a consumer reading `GameObj` between the interruption and the next refresh sees an incomplete inventory with no signal that the listing didn't finish.

Route `INV LIST` through the staged-refresh mechanism `GameObj` already provides (the same "leave the published snapshot in place until a clean commit" pattern documented on `begin_*`/`commit_*`), instead of clear-then-fill:

- **`GameObj.begin_all_containers` / `commit_all_containers_full`** — a full-replacement container refresh. While open, `new_inv` stages every container placement instead of touching live `@@contents`; commit swaps the whole hash in one reference assignment, so containers absent from the listing are dropped (matching the old clear-then-fill semantics), and an interrupted listing that never commits leaves the previous model intact.
- **`GameObj.discard_inv_refresh`** — drops only the inventory staging buffers (worn + full container) without aborting an unrelated in-flight room/familiar refresh.
- **`remove_inv_item`** now also scrubs the staging buffers, so an item picked into a hand mid-listing is not resurrected in its old container when the staged list commits.
- **drparser** — begin staging on the `INV LIST` header, commit on the clean `<output class=""/>` terminator, and on an interrupted `<prompt>` discard staging (keeping the previous model) and emit a `Lich::Messaging` warn: `"DRParser: 'inv list' did not finish; keeping previous inventory. Re-run 'inv list' to refresh."`

## 2. `GameObj.containers` dup-on-read

`GameObj.containers` duplicated only the outer hash, handing callers the **live inner arrays**. That was safe pre-stack — container arrays were only ever replaced wholesale, never mutated in place — but not once this series began mutating them in place: `remove_inv_item`'s `reject!` (hand pickup / upsert) and `new_inv`'s `push`. A script iterating `GameObj.containers[id]` across a hand pickup would have its array shifted under it and **silently skip an item**.

Fixed by returning duplicated inner arrays (`transform_values(&:dup)`), matching the dup-on-read contract every other registry accessor and instance `#contents` already honor. The `GameObj` elements are shared, not copied. The two in-tree callers (`stash.rb`, `readylist.rb`) are one-shot per-command reads with no write-through dependency, so cost and behavior are unaffected.

## GemStone safety

Byte-for-byte unaffected. The full-refresh mode is only ever entered from the DR `INV LIST` path, so `@@staging_all_containers` stays `false` for GS and `new_inv` behaves exactly as before; `remove_inv_item` is only reached from DR-gated paths. The `containers` accessor change is a read-side snapshot with no behavioral effect on any writer.

## Known limitation

An `XMLParser#reset` resync landing *mid*-`INV LIST` is a double-fault edge: `discard_staged_refreshes` now also clears the full-refresh flag, so subsequent item lines (before the next header) would route to the live model. This remains bounded by the existing prompt safety valve and is no worse than the prior clear-then-fill behavior; not staged around here.

## Testing

- Updated the one test asserting the old `clear_*` behavior; added 4 real-`GameObj` INV LIST scenario tests (model stays visible mid-stream; atomic replace + absent-container drop on a clean finish; previous model kept + warn emitted on interruption; picked-up item not resurrected on commit).
- Added a `GameObj.containers` regression test that drives a real `reject!` mid-iteration and asserts no item is skipped (verified to fail on the old shallow dup).
- Full suite: **5772 examples, 0 failures**. Rubocop clean.

### DIFF
--- a/lib/common/gameobj.rb
+++ b/lib/common/gameobj.rb
@@ -126,6 +126,13 @@ module Lich
       @@staging_fam_pcs       = nil
       @@staging_contents      = {}
 
+      # True while a full-replacement container refresh (DR +INV LIST+) is open.
+      # While set, +new_inv+ routes every container placement into
+      # +@@staging_contents+ (auto-vivifying per-container buffers) instead of the
+      # live +@@contents+, so an interrupted listing never mutates the published
+      # model. See {.begin_all_containers}.
+      @@staging_all_containers = false
+
       # ---------------------------------------------------------------------------
       # Instance interface
       # ---------------------------------------------------------------------------
@@ -356,7 +363,13 @@ module Lich
       # @return [GameObj]
       def self.new_inv(id, noun, name, container = nil, before = nil, after = nil)
         if container
-          target = @@staging_contents[container] || (@@contents[container] ||= [])
+          target = if @@staging_all_containers
+                     # Full INV LIST refresh: stage every container so nothing
+                     # touches the live model until the listing commits cleanly.
+                     @@staging_contents[container] ||= []
+                   else
+                     @@staging_contents[container] || (@@contents[container] ||= [])
+                   end
           find_or_create(target, id, noun, name, before, after)
         else
           find_or_create(@@staging_inv || @@inv, id, noun, name, before, after)
@@ -420,6 +433,11 @@ module Lich
         @@index_mutex.synchronize do
           @@inv.reject! { |obj| obj.id == str_id }
           @@contents.each_value { |list| list.reject! { |obj| obj.id == str_id } }
+          # Scrub any in-flight staging buffers too: a hand pickup during a full
+          # INV LIST refresh must not be resurrected in its old container when the
+          # staged listing commits.
+          @@staging_inv&.reject! { |obj| obj.id == str_id }
+          @@staging_contents.each_value { |list| list.reject! { |obj| obj.id == str_id } }
         end
         nil
       end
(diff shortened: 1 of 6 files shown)


mrhoribu (approved):
Reviewed the full stack (#1555 → #1554 → #1556 → #1557 → #1563) across several rounds as the author addressed feedback.

The one substantive concern raised on this PR — that `INV LIST`'s new atomic staged refresh (`begin_all_containers`/`commit_all_containers_full`/`discard_inv_refresh`) shared `@@staging_contents` with the pre-existing per-container `clearContainer` refresh path, which could silently drop or prematurely-publish an unrelated in-flight container refresh — is resolved by `e8173dd9`. It gives the full refresh its own dedicated `@@staging_all_contents` buffer, fully isolated from per-container staging, with `remove_inv_item`, `clear_all_containers`, `discard_staged_refreshes`, and the TTL-prune live-object set all updated consistently.

I reproduced the original issue directly against the code (not just read it) before and after this fix — both scenarios (interrupted-listing discard, and completing-listing premature-publish) now behave correctly, and the PR's own new specs cover the same cases plus one I hadn't (a per-container refresh already open before `INV LIST` starts).

rubocop clean; targeted specs (gameobj/xmlparser/drparser/inventory) green at 410 examples, 0 failures.

Two minor, non-blocking open questions remain for the author to confirm at their convenience, not merge-blocking:
- The dash-stripping in `drparser.rb` assumes DR always emits a single leading `-` regardless of nesting depth — confirmed against 3+-deep nesting in game, or only the fixture depths?
- `InventorySearchStart`'s "verified across all ~90 inv verbs" — exhaustive log check, or a representative sample?

Approving.

PR #10, opened 2021-08-31
Update script jinx for jinxp

OSXLich-Doug:
Creating test entry jinx for proposed changes to jinxp

This supports addition of a lib dir for jinx, with corresponding required changes in Lich.

Note, Lich rev not bumped here - may want to bump it if testing successful.

PR #1564, opened 2026-09-07
fix(dr): refresh before_name/after_name on a name-preserving container move

MahtraDR:
## What

`find_or_create` / `index_or_create` only backfilled `before_name`/`after_name` when the stored value was `nil`. After a **name-preserving container move** (same `id|noun|name`, e.g. `#10` from `#20` to `#30`) the identity index reused the instance but kept its **stale** `before_name`. A consumer building a GET from `before_name` would target the item's old container; the fields also compose `full_name`, which feeds `type`/`sellable` classification.

## How

Refresh `before_name`/`after_name` whenever a **differing non-nil** observation arrives; a `nil` observation never blanks a known value. Both `find_or_create` and `index_or_create` are updated, and their stale doc summaries corrected.

Because `find_or_create` returns the shared, still-published instance, refreshing it **during an open staged refresh** (a `clearContainer`/container fill, `begin_inv`, or a full `INV LIST` refresh) would publish an uncommitted value and could not be rolled back on discard. So a staged refresh **defers** the metadata change into `@@pending_metadata` and applies it at the matching `commit_*`; non-staged observations still refresh immediately.

The pending map is scoped **per staging buffer** (outer hash keyed by the buffer Array's identity via `compare_by_identity`, inner by GameObj identity), which is what makes it correct:

- two concurrently-open refreshes of the same object can't clobber each other's pending values; a commit publishes only its own buffer's changes;
- `before`/`after` merge **independently** (a before-only then after-only observation keeps both), seeded from the published value so an observation returning to the published value supersedes an earlier intermediate one;
- **every** discard path drops exactly its buffer's entries — `abort_inv`, `abort_container`, `delete_container`, `clear_all_containers`, a `begin_inv` / `begin_container` / `begin_all_containers` restart, `discard_inv_refresh`, and `discard_staged_refreshes` — so a rolled-back refresh can't leak onto the shared instance and the map never grows unbounded.

The deferral integrates with the full `INV LIST` staging path (`begin_all_containers` / `commit_all_containers_full` / `discard_inv_refresh`) added by #1563.

## Scope / history

The nil-only backfill dates to #1234 and reproduces on `main`; independent of the DR inventory-model series. An earlier revision of this PR shipped the deferral with correctness gaps (cross-buffer leakage, whole-pair overwrite, missed discard paths); a follow-up review then showed that simply removing it reintroduced the premature-publication/rollback regression vs. `main`. This revision keeps the deferral but makes it correct and buffer-scoped, and closes the gaps.

## Testing

New regression specs cover: staged isolation until commit; rollback on every discard path; per-buffer scoping; independent field merge; return-to-published superseding an intermediate; both `begin_inv`/`begin_container` restart leaks; and the full `INV LIST` path (defer → `commit_all_containers_full`, and `discard_inv_refresh`/restart rollback). Full suite green (7042 examples). Rubocop clean (pinned 1.86.1).

### DIFF
--- a/lib/common/gameobj.rb
+++ b/lib/common/gameobj.rb
@@ -529,8 +529,9 @@ module Lich
       # based garbage collection.
       #
       # When a matching entry is found, +before_name+ and +after_name+ are
-      # backfilled if they were previously +nil+ and the incoming values are
-      # non-nil. Existing non-nil values are never overwritten.
+      # refreshed from a differing non-nil incoming value; a nil observation
+      # never clobbers a known value. (Hand slots pass no before/after, so in
+      # practice this path only ever backfills.)
       #
       # @example Replace a bare GameObj.new call
       #   # Before:
@@ -1368,8 +1369,8 @@ module Lich
         # and it is re-added to the cleared registry - no allocation required.
         #
         # When a duplicate is found, +before_name+ and +after_name+ are
-        # backfilled if they were previously +nil+ and the incoming values are
-        # non-nil. Existing non-nil values are never overwritten.
+        # refreshed from a differing non-nil observation; a nil observation
+        # never clobbers a known value.
         #
         # @param registry [Array<GameObj>]  the target registry array
         # @param id       [Integer, String]
--- a/lib/common/gameobj.rb
+++ b/lib/common/gameobj.rb
@@ -542,8 +542,8 @@ module Lich
       # @param id     [Integer, String]
       # @param noun   [String, nil]
       # @param name   [String, nil]
-      # @param before [String, nil]   backfills +before_name+ if previously unset
-      # @param after  [String, nil]   backfills +after_name+ if previously unset
+      # @param before [String, nil]   sets +before_name+ from a differing non-nil observation
+      # @param after  [String, nil]   sets +after_name+ from a differing non-nil observation
       # @return [GameObj]
       # @api private Internal identity-index plumbing. No compatibility guarantee.
       def self.index_or_create(id, noun, name, before = nil, after = nil)
@@ -555,8 +555,13 @@ module Lich
           if (entry = @@index[key])
             existing, _ts        = entry
             @@index[key]         = [existing, now]
-            existing.before_name = before if existing.before_name.nil? && !before.nil?
-            existing.after_name  = after  if existing.after_name.nil?  && !after.nil?
+            # Refresh command metadata whenever a differing non-nil observation
+            # arrives -- e.g. a name-preserving move updates +before_name+ from
+            # "get #id in #old" to "get #id in #new". A nil observation never
+            # clobbers a known value (a sighting from a source that carries no
+            # command, such as the hand slot, must not blank it).
+            existing.before_name = before if !before.nil? && existing.before_name != before
+            existing.after_name  = after  if !after.nil?  && existing.after_name  != after
             existing
           else
             new_obj      = GameObj.new(id, noun, name, before, after)
@@ -1370,8 +1375,8 @@ module Lich
         # @param id       [Integer, String]
         # @param noun     [String, nil]
         # @param name     [String, nil]
-        # @param before   [String, nil]   backfills +before_name+ if previously unset
-        # @param after    [String, nil]   backfills +after_name+ if previously unset
+        # @param before   [String, nil]   sets +before_name+ from a differing non-nil observation
+        # @param after    [String, nil]   sets +after_name+ from a differing non-nil observation
         # @return [GameObj]
         def find_or_create(registry, id, noun, name, before = nil, after = nil)
           str_id = id.is_a?(Integer) ? id.to_s : id
@@ -1382,8 +1387,13 @@ module Lich
             if (entry = @@index[key])
               existing, _ts        = entry
               @@index[key]         = [existing, now] # refresh last-seen timestamp
-              existing.before_name = before if existing.before_name.nil? && !before.nil?
-              existing.after_name  = after  if existing.after_name.nil?  && !after.nil?
+              # Refresh command metadata whenever a differing non-nil observation
+              # arrives -- e.g. a name-preserving move updates +before_name+ from
+              # "get #id in #old" to "get #id in #new". A nil observation never
+              # clobbers a known value (a sighting from a source that carries no
+              # command must not blank it).
+              existing.before_name = before if !before.nil? && existing.before_name != before
+              existing.after_name  = after  if !after.nil?  && existing.after_name  != after
               registry.push(existing) unless registry.include?(existing)
               existing
             else
--- a/lib/common/gameobj.rb
+++ b/lib/common/gameobj.rb
@@ -139,6 +139,24 @@ module Lich
       # model. See {.begin_all_containers}.
       @@staging_all_containers = false
 
+      # Deferred +before_name+/+after_name+ refreshes for objects observed while a
+      # staged registry refresh is open. +find_or_create+ returns the shared,
+      # still-published instance, so mutating its metadata mid-refresh would leak
+      # an uncommitted value to readers and could not be rolled back if the refresh
+      # is discarded. Instead the change is recorded here and applied by the
+      # matching +commit_*+ (see {.apply_pending_metadata}), or dropped by every
+      # discard path (+abort_*+, {.delete_container}, {.clear_all_containers}, an
+      # +begin_inv+/+begin_container+ restart, {.discard_inv_refresh}, and
+      # {.discard_staged_refreshes}).
+      #
+      # Scoped PER staging buffer: the outer hash is keyed by the buffer Array's
+      # identity (+compare_by_identity+ -- two empty Arrays are +==+, so content
+      # keying would collide), the inner by GameObj identity. Buffer scoping is
+      # what keeps two concurrently-open refreshes of the same object from
+      # overwriting each other's pending values and lets a discard drop exactly
+      # its own buffer's entries and no others.
+      @@pending_metadata = {}.compare_by_identity
+
       # ---------------------------------------------------------------------------
       # Instance interface
       # ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


mrhoribu (dismissed):
**Reviewed:** 2 files, +23/-8 vs `origin/main` at `4592f734` on 2026-09-08. Based on: full local clone (worktree checkout of the updated PR branch), including history for all three commits, plus an independent local CodeRabbit CLI review (`coderabbit review --committed --base origin/main --agent`).

**Summary:** The author dropped the entire `@@pending_metadata` deferral mechanism from the prior revision and reverted to the original narrow fix (refresh `before_name`/`after_name` on any differing non-nil observation, never on nil). This resolves the previous review's Blocker outright — that machinery, and its abort-path leak, no longer exists. What's left is small, correct in the common case, and well tested, but it reintroduces (in a wider-exposure form) a narrower publish-before-commit hazard that pre-dates this PR and that the abandoned second commit was itself trying to fix. Both my manual review and an independent local CodeRabbit pass converged on this same defect.

## Previously reported Blocker — resolved

The prior revision's `abort_inv`/`abort_container` not clearing `@@pending_metadata` (leaking a discarded refresh's staged metadata onto an unrelated, already-published object) no longer applies: `grep -n "pending_metadata\|staged:" lib/common/gameobj.rb` on the current head returns nothing. The whole deferral mechanism, the `staged:` keyword arg, and the two extra `commit_*` hooks are gone. The PR body's "Scope / history" section explains why — a follow-up review found several correctness defects in that mechanism ("deferred entries leaking across `delete_container` / `begin_container` restart / `clear_all_containers` / abort / cross-container commits, plus whole-pair overwrite and stale-value comparison") and the author chose to drop the isolation guarantee rather than keep patching it, since it wasn't provided pre-PR either. That's a reasonable trade given no identified consumer of `before_name`/`after_name` currently exists — see the Major issue below for why it isn't fully free, though.

## Major

### Immediate (non-deferred) metadata refresh still lets a staged, uncommitted container refresh premature-publish onto — and permanently corrupt on abort — the still-live registry's shared instance

```ruby
# lib/common/gameobj.rb:1509-1520 (find_or_create)
if (entry = @@index[key])
  existing, _ts        = entry
  @@index[key]         = [existing, now] # refresh last-seen timestamp
  existing.before_name = before if !before.nil? && existing.before_name != before
  existing.after_name  = after  if !after.nil?  && existing.after_name  != after
  registry.push(existing) unless registry.include?(existing)
  existing
```

`existing` comes from the shared `@@index`, independent of which `registry` array (live or staging) the caller is writing into. When a container refresh is open (`begin_container`), the object identity-index still returns the *same* `GameObj` instance that the currently-published `@@contents[container_id]` list already holds a reference to — so mutating `existing.before_name` here mutates the value every existing reader of the *not-yet-committed* published container sees too, immediately, mid-refresh. Reproduced directly against this branch:

```ruby
GameObj.new_inv('10', 'gem', 'ruby', '20', 'get #10 in #20')     # published

GameObj.begin_container('20')
GameObj.new_inv('10', 'gem', 'ruby', '20', 'get #10 in #30')     # differing, staged fill

GameObj.containers['20'].first.before_name
# => "get #10 in #30"   -- leaked onto the still-published view before commit_container('20')
```

Worse, if the refresh is then aborted (`abort_container`/`abort_inv` — the rollback path `lib/common/inventory.rb:1594-1595` uses on a partial-fill failure), the mutation is **not** rolled back, because it was never staged separately from the shared instance in the first place:

```ruby
GameObj.abort_container('20')
GameObj.containers['20'].first.before_name
# => "get #10 in #30"   -- still corrupted; the aborted refresh's value survives permanently
```

This is not brand-new architecture — `main` already had the identical shape of hazard for the *nil → non-nil* backfill case (a first-ever observation could premature-publish the same way). What this PR changes is the trigger condition: refreshing on any *differing* non-nil value (not just filling a previously-unset one) means the hazard now fires on every subsequent move, not just an item's first sighting — which is exactly the recurring scenario this PR's own bug report is about (`#10` moving `#20` → `#30`). The abandoned second commit's message named this precisely: "the refresh mutated the published object immediately: readers saw the new metadata before commit, and discarding the staging buffer could not roll it back (P2 - publication/rollback defect)." Reverting the deferral removed that commit's own new defects but restored — and widened the exposure of — the defect it was written to fix.

An independent local CodeRabbit CLI review of this same diff (`coderabbit review --committed --base origin/main --agent`) flagged the identical issue on its own, unprompted:

> `lib/common/gameobj.rb` around lines 1518-1519: Update `find_or_create` so re-seeing an existing object does not mutate the instance held by the published registry during a staged refresh; stage `before_name` and `after_name` changes on the refresh copy or defer reconciliation until commit, preserving the published metadata if the refresh aborts. Add a regression test covering an existing item re-seen during `begin_container` or `begin_inv` followed by abort.

The `@@staging_contents` docblock's stated contract (`lib/common/gameobj.rb:98-113`) — "a reader never observes an empty or half-filled registry mid-stream... it sees the previous complete snapshot until commit" — is violated specifically for `before_name`/`after_name` on a shared, previously-published instance.

I'm requesting changes rather than approving because the abort case is a step worse than what was previously accepted on `main`: it's not just "briefly visible before an already-scheduled commit," it's "permanently wrong after an *explicit rollback*," which is a stronger violation of `abort_inv`/`abort_container`'s documented "leaving the previously published ... visible" contract (`lib/common/gameobj.rb:766-876` doc comments) than existed before this PR. I don't think this needs the full deferral machinery back — just enough isolation that an in-flight staged refresh's metadata write doesn't touch the shared instance until (and unless) its own commit runs.

**Confidence:** confirmed (reproduced against the current PR head; corroborated independently by CodeRabbit).

## Local CodeRabbit review

Ran `coderabbit review --committed --base origin/main --agent` from the PR worktree. It surfaced exactly **one** finding, `major` severity, on `lib/common/gameobj.rb:1518-1519` — the same issue as above, independently arrived at. No other findings (style, tests, docs, or otherwise) were raised.

## What this PR gets right

- Correctly and narrowly reverts the riskier staging/deferral machinery once its own defects surfaced, rather than continuing to patch a design with multiple leak paths — a good call, and the "Scope / history" section of the PR body is an unusually candid writeup of why.
- The core semantic fix (differing non-nil refreshes, nil never clobbers) is correct and applied consistently to both `find_or_create` and `index_or_create`, with docs updated to match in the same commit.
- Regression tests for the two behaviors that matter (`before_name` refreshes on a name-preserving move; a nil observation doesn't blank a known value) are present, focused, and would fail against the old code.
- Full `gameobj_spec.rb` suite green (124 examples on this branch) and Rubocop clean.

## Open question

Is there truly no code path (in lich-5, dr-scripts, or scripts) that reads `before_name`/`after_name` off an object obtained from a container mid-refresh (i.e., between a `begin_container`/`begin_inv` and its matching `commit_*`)? The PR body asserts "no known executing consumer," which is plausible, but the abort case above means a wrong value can now persist *after* an explicit rollback, not just transiently — worth a final confirmation before merging, since that's a stronger claim than "nobody polls mid-stream." If something does consume it today, or will via the in-flight DR inventory-model series (#1555-#1557) mentioned in the PR body, this should be revisited before that lands.

## Coverage notes

Reviewed via a full local worktree checkout of the updated PR branch (`4592f734`) diffed against `origin/main`, plus `git log -p` over all three commits (including the reverted one, for context). Ran the full `spec/lib/common/gameobj_spec.rb` suite (124 examples, 0 failures) and `rubocop` on both changed files (0 offenses). Independently reproduced the Major issue above by loading `lib/common/gameobj.rb` standalone and driving the class API directly (both the mid-refresh premature-publish and the abort-doesn't-roll-back cases). Ran `coderabbit review --committed --base origin/main --agent` as a second, independent automated pass — its single finding corroborates the issue above with no additional findings. Did not review `lib/common/inventory.rb`/`xmlparser.rb` beyond the specific call sites relevant to grounding this issue — both are unchanged by this PR.

mrhoribu (approved):
**Reviewed:** 3 files, +302/-26 vs `origin/main` at `b962d7bc` on 2026-09-08. Based on: full local clone (worktree checkout of the updated PR branch), including history for all four commits, plus an independent local CodeRabbit CLI review (`coderabbit review --committed --base origin/main --agent`).

**Summary:** This revision brings back the metadata-deferral mechanism from two revisions ago, but scoped per staging buffer (outer hash keyed by the buffer Array's identity via `compare_by_identity`, inner by GameObj identity) instead of per object. That's the fix for the Major issue in my last review: I reproduced the mid-refresh premature-publish and the abort-doesn't-roll-back cases against this new revision and both are now correctly handled — a staged refresh's metadata change no longer leaks onto the still-published shared instance, and every discard path (`abort_inv`, `abort_container`, `delete_container`, `clear_all_containers`, a `begin_inv`/`begin_container`/`begin_all_containers` restart, `discard_inv_refresh`, `discard_staged_refreshes`) correctly drops its own buffer's pending entries rather than orphaning or misapplying them. An independent local CodeRabbit pass over this same diff found zero findings, corroborating that read.

## Previously reported Major — resolved

Reproduced against the current head:

```ruby
GameObj.new_inv('10', 'gem', 'ruby', '20', 'get #10 in #20')     # published

GameObj.begin_container('20')
GameObj.new_inv('10', 'gem', 'ruby', '20', 'get #10 in #30')     # differing, staged
GameObj.containers['20'].first.before_name
# => "get #10 in #20"   -- unchanged mid-refresh, as it should be

GameObj.commit_container('20')
GameObj.containers['20'].first.before_name
# => "get #10 in #30"   -- now correctly published atomically with the commit

# and the abort case:
GameObj.begin_container('20')
GameObj.new_inv('10', 'gem', 'ruby', '20', 'get #10 in #99')
GameObj.abort_container('20')
GameObj.containers['20'].first.before_name
# => "get #10 in #30"   -- correctly rolled back to the last committed value, not "get #10 in #99"
```

The buffer-identity keying (`@@pending_metadata = {}.compare_by_identity`, and the inner per-object hash also `compare_by_identity`) is the right fix for the cross-buffer collision risk in the earlier per-object-only design: two concurrently open, content-equal-but-distinct staging buffers (e.g. two containers each starting from a fresh `[]`) can no longer alias each other in the pending map, since a plain `Hash` would treat two empty arrays as the same key. I checked this against the "keeps each open container's deferral scoped to its own buffer" spec (`spec/lib/common/gameobj_spec.rb:1340`) and traced it manually — correct.

I also checked the specific gap classes the PR body claims to have closed, by reading (not just running) each discard path: `begin_container`/`begin_inv`/`begin_all_containers` restart all call `drop_pending_metadata` on the buffer they're about to replace before allocating a new one; `delete_container` and `clear_all_containers` do the same for the buffers they discard; `discard_inv_refresh` drops both `@@staging_inv`'s and every `@@staging_all_contents` buffer's entries without touching unrelated `@@staging_contents` (per-container) buffers, matching its documented scope. All of that lines up with the code, not just the specs asserting it.

## Minor / pre-existing (not introduced by this PR)

`delete_container` (`lib/common/gameobj.rb:762`) only removes a container's key from `@@staging_contents` (the per-container staging hash) and `@@contents` (published) — it never touches `@@staging_all_contents` (the full `INV LIST` refresh's dedicated buffer, from #1563). If a container is deleted mid-`INV LIST` after already being staged into `@@staging_all_contents`, `commit_all_containers_full`'s later wholesale `@@contents = @@staging_all_contents` swap would resurrect it. I confirmed this shape already exists verbatim on `origin/main` before this PR (`delete_container` there is a two-line method with the same gap) — this PR only adds the `drop_pending_metadata` call to it, it doesn't touch the `@@staging_all_contents` omission. Flagging for awareness, not as a blocker on this PR.

## What this PR gets right

- Correctly diagnosed and fixed the exact issue from the last review round, with a design (per-buffer identity keying) that directly addresses the specific collision/leak modes named in the PR body's own retrospective on the previous attempt.
- Test coverage is unusually thorough for this kind of state-machine code: isolation until commit, rollback on every discard path individually (parameterized across `abort_container`/`discard_staged_refreshes`/`clear_all_containers`), per-buffer scoping with two concurrently open containers, independent before/after field merging, the "return to published value supersedes an intermediate" edge case, both restart-leak paths, and the full `INV LIST` integration (`begin_all_containers`/`commit_all_containers_full`/`discard_inv_refresh`). Read through each one against the implementation rather than trusting the assertions — they test what they claim to.
- Full `gameobj_spec.rb` suite green (138 examples on this branch) and Rubocop clean.
- The PR body's "Scope / history" section continues to be a candid, accurate account of what changed and why across revisions — makes this much easier to re-review incrementally.

## Coverage notes

Reviewed via a full local worktree checkout of the updated PR branch (`b962d7bc`) diffed against `origin/main`, plus `git log -p` / `git show --stat` over the new commit. Ran the full `spec/lib/common/gameobj_spec.rb` suite (138 examples, 0 failures) and `rubocop` on both changed files (0 offenses). Independently reproduced the fix for the previously-reported Major issue by loading `lib/common/gameobj.rb` standalone and driving the class API directly. Ran `coderabbit review --committed --base origin/main --agent` as a second, independent automated pass — 0 findings this time (down from 1 major finding on the prior revision). Did not re-verify integration with `lib/dragonrealms/drinfomon/drparser.rb`/`lib/common/inventory.rb` end-to-end (no changes to those files in this PR, and their call sites' signatures are unchanged), beyond confirming the pre-existing `delete_container` gap noted above.

PR #1565, opened 2026-09-07
fix(dr): do not re-add a held item as worn inventory on inv scrape

MahtraDR:
**Depends on #1557 → #1563** (branched off the inventory-model series head; the INV scrape/parse path this touches was reworked by that stack). Merge the stack first.

## What

An `INV SEARCH` / `INV LIST` line for an item in a hand looks like:

```
<d cmd='get #8761784'>a seagull feather quill</d> is in your right hand.
```

The parser only reads the `<d>` element's `cmd`, which for a held item is just `get #8761784` — no `in #container`. So it routed the item to **worn inventory** (`container = nil` → `GameObj.inv`), with no awareness of the "is in your right hand" tail. That recreates the **held/worn duplicate** that #1557's `<right>`/`<left>` reconciliation removes on pickup: pick up an item, then `inv search` for it, and the model shows it as simultaneously in-hand *and* worn.

## How

Detect the `"... is in your right/left hand"` trailing prose — still present on the full line after the `</d>` element — and skip the worn/container store, dropping any stale placement via `remove_inv_item`. The item's real placement is the hand slot, tracked separately by the `<right>`/`<left>` stream.

## Scope / attribution

Pre-existing behavior of the INV scrape (reproduces both before and after the inventory-model series) — the reverse direction of #1557's pickup reconciliation. DR-only path; GemStone's inventory stream is untouched.

## Testing

Repurposed the old "trailing location prose" test (which had codified the buggy hand behavior) to a realistic nested-container line, and added tests asserting a held item is **not** stored as worn for both right and left hand (verified to fail on the old code). Full DR parser suite green; rubocop clean (pinned 1.86.1).

### DIFF
--- a/lib/dragonrealms/drinfomon/drparser.rb
+++ b/lib/dragonrealms/drinfomon/drparser.rb
@@ -239,6 +239,16 @@ module Lich
               id = id_match[:itemID].strip.delete('#').to_s
               noun = nil # This isn't exposed in the DR XML stream
               name = item_name
+              # An item reported "... is in your right/left hand" is held, not worn
+              # or in a container: its placement is the hand slot, tracked
+              # separately by the <right>/<left> stream. Storing it here (cmd has no
+              # "in #container", so container=nil -> worn inv) would recreate the
+              # held/worn duplicate that hand reconciliation removes on pickup. Drop
+              # any stale placement and skip the worn/container store.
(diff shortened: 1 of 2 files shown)


OSXLich-Doug (approved):
Looks good!

PR #1566, opened 2026-09-07
fix(dr): evict orphaned identity-index entries on a name-changing upsert

MahtraDR:
**Depends on #1557 → #1563** (branched off the inventory-model series head; `upsert_inv`/`remove_inv_item` are introduced by that stack). Merge the stack first.

## What

A name-changing `upsert_inv` (the review's own example: `"soft gem pouch"` → `"soft gem pouch (closed)"`) keys a **new** `id|noun|newname` entry in the shared identity index. The prior `id|noun|oldname` instance was just pulled from every registry by `remove_inv_item`, but `remove_inv_item` deliberately leaves `@@index` alone — so the old key lingers, unreachable from any live registry, until the 15-minute TTL prune (`prune_index!`) sweeps it. Under heavy `INV SEARCH` of frequently-renamed items the index accumulates dead orphans between sweeps.

## How

On the **cold** scrape path (`upsert_inv`), evict any index entry for the id whose instance is no longer live, right after the re-add:

```ruby
live = live_registry_objects
@@index_mutex.synchronize do
  @@index.delete_if { |_key, (indexed, _ts)| indexed.id == str_id && !live.include?(indexed) }
end
```

Guarded by **object identity** via `live_registry_objects` (computed outside the lock, matching `sweep_stale!`), so a same-id variant still held anywhere — the just-re-added instance, or a hand slot — survives.

The **hot** hand-pickup path (`remove_inv_item`, called on every DR `<right>`/`<left>` update) is deliberately left untouched to avoid an O(index) scan on every hand event; the TTL prune remains the backstop there. This targets exactly the review's scenario (name-changing INV SEARCH re-observe).

## Scope / attribution

Pre-existing index-lifecycle characteristic (how the index already treats any object leaving a registry), flagged in review as non-blocking — this makes the common scrape case self-clean immediately rather than waiting for the sweep. Not a regression introduced by the stack.

## Testing

Regression tests: the old-name entry is evicted immediately after a name-changing upsert (verified to fail on the old code, where it lingered), and a live same-id variant held in a hand slot is kept. Rubocop clean (pinned 1.86.1).

### DIFF
--- a/lib/common/gameobj.rb
+++ b/lib/common/gameobj.rb
@@ -395,7 +395,23 @@ module Lich
       def self.upsert_inv(id, noun, name, container = nil, before = nil, after = nil)
         str_id = id.is_a?(Integer) ? id.to_s : id
         remove_inv_item(str_id)
-        new_inv(str_id, noun, name, container, before, after)
+        obj = new_inv(str_id, noun, name, container, before, after)
+
+        # A name change keys a NEW index entry ("id|noun|newname"); the prior
+        # "id|noun|oldname" instance was just pulled from every registry by
+        # remove_inv_item, but its index key lingers until the TTL prune. On the
+        # cold scrape path, evict any index entry for this id whose instance is no
+        # longer live so a frequently-renamed item (e.g. a pouch toggling
+        # "(closed)") can't accumulate orphaned variants between sweeps. Guarded by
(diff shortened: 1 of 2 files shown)


OSXLich-Doug (approved):
Prior defect, thanks for closing!

PR #1569, opened 2026-09-08
fix(all): fall back to cleartext SGE auth (7900) when TLS port (7910) is unreachable

MahtraDR:
## Problem

`EAccess.socket` connects only to the **TLS** SGE listener on `eaccess.play.net:7910`. When that port is unreachable, Lich can't authenticate — even though the *same* SGE protocol is answering on the legacy **cleartext** listener `:7900`.

Observed from a live host (SGE now resolves to an AWS address, `3.22.54.28`):

```
$ nc -vz eaccess.play.net 7910   # TLS port
exit=124            # SYN silently dropped — connect times out, no RST

$ nc -vz eaccess.play.net 7900   # cleartext port
Ncat: Connected to 3.22.54.28:7900.
exit=0
```

Because 7910's SYNs are dropped (rather than refused), Lich hangs on the OS connect timeout (~75s) and then fails. The cleartext port is open the whole time — it's exactly what the ecosystem's cleartext reward tools (e.g. `rewards.cr`, `simu-rewards`) use, which is why they keep working while Lich doesn't.

> Note: this reachability asymmetry was confirmed from one network; whether it's a global Simu change or a routing/firewall issue on a given host, the resilient behavior is the same.

## Fix — prefer TLS, fall back to cleartext

- Split `socket()` into `secure_socket` (TLS 7910) and `cleartext_socket` (7900). `socket()` tries TLS first and drops to cleartext **only on a connection-establishment failure**.
- Bound the TLS TCP connect with `Socket.tcp(connect_timeout:)` so a dropped SYN fails over in ~5s instead of hanging ~75s (also applied to `download_pem`).
- **Security preserved:** only `SocketError` / `SystemCallError` / `IO::TimeoutError` trigger the fallback. `OpenSSL::SSL::SSLError` is **not** downgraded — a reachable-but-untrusted 7910 still hard-fails rather than silently dropping to cleartext.
- Guard the post-connect `verify_pem` to the TLS socket; the cleartext `Socket` has no peer cert.

## Why cleartext is acceptable here

The `K`/`A` password hash is applied **identically on both ports**, so the cleartext path never transmits the raw password — it only forgoes transport encryption of an already-obscured exchange. This is the same tradeoff Simutronics themselves make by running the 7900 listener.

## Tests

`spec/lib/common/authentication/eaccess_spec.rb` adds coverage for: TLS-preferred success, cleartext fallback on unreachable TLS (`ETIMEDOUT` and `SocketError`), and **no** downgrade on a TLS cert failure. Full auth suite: `383 examples, 0 failures`; RuboCop clean.

### DIFF
--- a/lib/common/authentication/eaccess.rb
+++ b/lib/common/authentication/eaccess.rb
@@ -25,6 +25,30 @@ module Lich
         # When sent via the L command, the game server starts the character creation flow.
         NEW_CHARACTER_CODE = "0"
 
+        # SGE authentication endpoints. Simutronics exposes the same protocol on
+        # a TLS listener (7910) and a legacy cleartext listener (7900). The K/A
+        # password hash is applied identically on both, so the cleartext path
+        # never transmits the raw password -- it only forgoes transport
+        # encryption of an already-obscured exchange. We prefer TLS and fall back
+        # to cleartext when the TLS port is unreachable (e.g. 7910 firewalled or
+        # its SYNs dropped while 7900 stays open, as after the SGE move to AWS).
+        HOST = "eaccess.play.net"
+        TLS_PORT = 7910
+        CLEARTEXT_PORT = 7900
+
+        # Bounds the TCP connect to the TLS endpoint so a silently-dropped SYN on
+        # 7910 fails over to cleartext in seconds instead of hanging on the OS
+        # connect timeout (~75s).
+        TLS_CONNECT_TIMEOUT = 5
+
+        # Failures to *establish* the TLS connection that warrant a cleartext
+        # retry. OpenSSL::SSL::SSLError is deliberately excluded: a
+        # reachable-but-untrusted 7910 is a security signal we refuse on, not one
+        # we silently downgrade around.
+        CONNECT_ERRORS = [SocketError, SystemCallError]
+        CONNECT_ERRORS << IO::TimeoutError if defined?(IO::TimeoutError)
+        CONNECT_ERRORS.freeze
+
         # @api private
         def self.pem
           @pem ||= File.join(DATA_DIR, "simu.pem")
@@ -36,11 +60,12 @@ module Lich
         end
 
         # @api private
-        def self.download_pem(hostname = "eaccess.play.net", port = 7910)
+        def self.download_pem(hostname = HOST, port = TLS_PORT)
           # Create an OpenSSL context
           ctx = OpenSSL::SSL::SSLContext.new
-          # Get remote TCP socket
-          sock = TCPSocket.new(hostname, port)
+          # Get remote TCP socket with a bounded connect so an unreachable TLS
+          # port fails fast instead of hanging on the OS connect timeout.
+          sock = Socket.tcp(hostname, port, connect_timeout: TLS_CONNECT_TIMEOUT)
           # pass that socket to OpenSSL
           ssl = OpenSSL::SSL::SSLSocket.new(sock, ctx)
           # establish connection, if possible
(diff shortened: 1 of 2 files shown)


MahtraDR:
Tested and works during an eaccess tls outage of DR as a whole. I have mixed feelings about this too.

therealatari (commented):
Local adversarial review of `c982093c` (agent-assisted, with reproductions checked locally).

I support having a practical outage fallback; these suggestions are about making this implementation predictable, not redesigning Simutronics authentication.

## Standards

- **P2 — Describe the credential exposure accurately.** [eaccess.rb:28–34](https://github.com/elanthia-online/lich-5/blob/c982093c18180a32542096f86c76e1c502eabdf7/lib/common/authentication/eaccess.rb#L28-L34) calls K/A a password hash. The existing operation is reversible XOR obfuscation, with its challenge sent on the same connection. An observer of the cleartext exchange can recover the password; “no raw password transmitted” should not imply password confidentiality. This is already explained in [the repository's protocol analysis](https://github.com/elanthia-online/lich-5/blob/c982093c18180a32542096f86c76e1c502eabdf7/docs/eaccess-protocol-analysis.md). Please correct the new comments and PR description to state that tradeoff plainly. This follows the YARD guide's accuracy-over-completeness rule.
- **P2 — The documented fallback boundary is broader in code.** The new documentation promises connection-establishment-only fallback, but the rescue covers more than that. See the independently reproduced Spec finding below; this is the same issue, not an additional bug.
- **P3 / nonblocking — Private-method documentation.** The new `secure_socket(hostname, port)` is missing typed `@param` and `@return` tags required by [docs/YARD-STYLE-GUIDE.md's private-method tier](https://github.com/elanthia-online/lich-5/blob/c982093c18180a32542096f86c76e1c502eabdf7/docs/YARD-STYLE-GUIDE.md). No broader abstraction/refactoring request.

## Spec

- **P2 — Restrict fallback to the intended connection phase.** [The rescue at eaccess.rb:103–107](https://github.com/elanthia-online/lich-5/blob/c982093c18180a32542096f86c76e1c502eabdf7/lib/common/authentication/eaccess.rb#L103-L107) surrounds all of `secure_socket`, not just TCP establishment. I reproduced a successful mocked TLS handshake followed by `File.write(pem, ...)` raising `Errno::EACCES`; this selected the cleartext fallback even though the problem was local certificate storage. Separately, a real localhost TCP connection reset during `SSLSocket#connect` raised `Errno::ECONNRESET` (a `SystemCallError`, not an `OpenSSL::SSL::SSLError`). That also qualifies for fallback despite TCP already being established. Narrow the rescue to the intended TCP connection attempts, including the certificate-bootstrap connection, and let certificate-file and handshake errors propagate. The new tests stub all of `secure_socket`, so they cannot detect this distinction. Add negative tests for those phases.

This does **not** demonstrate bypass of an explicit certificate-verification rejection: `OpenSSL::SSL::SSLError` remains excluded, as intended. TLS-first ordering and the five-second TCP connect settings are present.

## Optional product recommendation

Consider making cleartext fallback explicitly opt-in, with a user-visible warning. That is a recommendation, **not a claim that this PR failed its stated automatic-fallback requirement**. It would let players deliberately accept the legacy transport tradeoff during an outage without changing Simu's protocol.

## Verification

- Ruby 4.0.5: existing `spec/lib/common/authentication` suite — **383 examples, 0 failures**.
- Four additional local characterization checks passed: PEM-write failure selects fallback; current API has no caller opt-in; synthetic K/A bytes are reversible; explicit SSL rejection does not select fallback. These characterize current behavior, not approval of every behavior.
- Real localhost TCP/TLS-reset reproduction: `tcp_established=true handshake_error=Errno::ECONNRESET is_system_call_error=true is_ssl_error=false`.
- No real account credentials or game-server login used. Existing certificate-trust/cleanup design is not being attributed to this PR.

Summary: Standards — 3 findings, highest P2; Spec — 1 finding, P2 (the fallback-boundary issue overlaps the Standards accuracy finding).

therealatari:
Additional live field-test result from the September 8 outage, for context:

I tested the fallback locally against GemStone IV while the normal EAccess path was unavailable. The cleartext EAccess endpoint on port 7900 accepted a TCP connection, but it did not send the initial usable K challenge / complete the EAccess protocol handshake during our probes. Consequently, switching from TLS to cleartext did not produce a login and this PR path still could not launch the character in that outage condition.

At the same time, Play.net website authentication and Play in Browser remained functional. The website issued a one-time game ticket through its web launch path, and feeding that website-issued ticket to the existing Lich SAL launch path successfully authenticated to GSIV. That confirms the game and website ticket path were available while the EAccess protocol path used here was not.

This is a point-in-time GSIV outage observation, not a claim that the cleartext fallback can never work. MahtraDR reports that it did work during a DR-wide TLS outage, so the behavior may depend on which Simutronics service or game endpoint is impaired. The practical limitation is that TLS-versus-cleartext fallback only helps when port 7900 is actually serving the legacy EAccess handshake; a TCP accept alone is insufficient.

MahtraDR:
Closing in favor of #1570

PR #1572, opened 2026-09-08
fix(all): Script.log respects --log-dir/LOG_DIR override; use File.join for paths in script.rb

### LINKED ISSUE #1519: Script.log ignores --log-dir/LOG_DIR, always writes to LICH_DIR/logs
`--log-dir=PATH` (alias `--logs=PATH`) is documented as redirecting where Lich's log output goes, and most of the plumbing for it is correct — but the one place that actually writes `.log` content ignores it entirely.

## What correctly respects the override

- `lich.rbw` (top-of-file ARGV pre-scan, ~lines 19-20): recognizes `--logs=` / `--log-dir=` and sets the `LOG_DIR` constant.
- `lib/constants.rb:7`: `LOG_DIR ||= File.join(LICH_DIR, "logs")` — correct default fallback.
- `lib/init.rb` (~lines 614-619): creates the `LOG_DIR` directory if it doesn't exist.
- `lib/common/session_launcher.rb` (`OPTIONAL_PATH_FLAGS` array, ~line 17): correctly forwards `--logs=` to spawned child Lich processes.

## What's broken

The only code that actually writes a `.log` file is the `Script.log` API surfaced to `.lic` scripts — the `@@elevated_log` proc in `lib/common/script.rb`, around line 258-276:

```ruby
@@elevated_log = proc { |data|
  if (script = Script.current)
    if script.name =~ /\\|\//
      nil
    else
      begin
        Dir.mkdir("#{LICH_DIR}/logs") unless File.exist?("#{LICH_DIR}/logs")
        File.open("#{LICH_DIR}/logs/#{script.name}.log", 'a') { |f| f.puts data }
        true
      rescue
        respond "--- Lich: error: Script.log: #{$!}"
        false
      end
    end
  else
    respond '--- error: Script.log: unable to identify calling script'
    false
  end
}
```

Line 265 hardcodes `LICH_DIR/logs` and never reads the `LOG_DIR` constant. So passing `--log-dir=/custom/path` creates an empty directory at the custom path (via `init.rb`'s mkdir), but every `Script.log` call from every running script still always writes to the default `LICH_DIR/logs`, silently ignoring the override.

This is a confusing, silent failure — a user who passes `--log-dir` expecting script logs to land in their custom directory sees the directory get created (so it looks like the flag "worked"), but finds it empty, with the actual logs sitting in the default location where they didn't intend them.

## Not to be confused with `debug-*.log`

This is a separate, unrelated logging mechanism from Lich's own `debug-*.log` files, which are written via `lib/init.rb`'s `debug_filename = "#{TEMP_DIR}/debug-...log"` and are governed by `--temp`/`TEMP_DIR`, not `LOG_DIR`. `--log-dir` only ever affects `Script.log` output from `.lic` scripts — and currently doesn't even do that correctly.

## Suggested fix

Change the `@@elevated_log` proc in `lib/common/script.rb` to build its path from the `LOG_DIR` constant instead of hardcoding `LICH_DIR/logs`, mirroring how other directory constants (`TEMP_DIR`, `DATA_DIR`, `SCRIPT_DIR`) are correctly referenced elsewhere in the codebase rather than hardcoded.

Found in 5.19.1, confirmed still present on current `main`. Not a version-specific regression — looks like the hardcoded path was just never updated when `LOG_DIR` was introduced as an override.

mrhoribu:
## Summary
- `Script.log` (the `@@elevated_log` proc in `lib/common/script.rb`) hardcoded `LICH_DIR/logs` instead of using the `LOG_DIR` constant, so `--log-dir`/`--logs`/`LOG_DIR` silently had no effect on where script logs were written — the custom directory would be created (empty) by `init.rb`, while actual log content kept going to the default `LICH_DIR/logs`.
- Changed the two hardcoded path references to use `LOG_DIR`, matching how other overridable directory constants (`TEMP_DIR`, `DATA_DIR`, `SCRIPT_DIR`) are referenced elsewhere in the codebase.
- While in this file, also replaced the remaining `"#{DIR}/#{name}"`-style string-interpolated path construction with `File.join` throughout `script.rb` (script loading, `Script.exists?`, `Script.db`/`Script.open_file`, `__find_script_file`, `__header_lines_for`) for consistency and to avoid manual separator handling. No behavior change - `File.join` collapses the same paths (verified for the one case with an existing leading slash).

Fixes #1519

## Test plan
- [x] `bundle exec rubocop lib/common/script.rb` - no offenses
- [x] `ruby -c lib/common/script.rb` - syntax OK
- [x] `bundle exec rspec spec/lib/common/script_lifecycle_spec.rb spec/lib/common/script_buffer_spec.rb` - 165 examples, 0 failures

### DIFF
--- a/lib/common/script.rb
+++ b/lib/common/script.rb
@@ -142,9 +142,9 @@ module Lich
           begin
             if file_name =~ /\.(?:cmd|wiz)(?:\.gz)?$/i
               trusted = false
-              script_obj = WizardScript.new("#{SCRIPT_DIR}/#{file_name}", script_args, false)
+              script_obj = WizardScript.new(File.join(SCRIPT_DIR, file_name), script_args, false)
             else
-              script_obj = Script.new(:file => "#{SCRIPT_DIR}/#{file_name}", :args => script_args, :quiet => options[:quiet], :publish => false)
+              script_obj = Script.new(:file => File.join(SCRIPT_DIR, file_name), :args => script_args, :quiet => options[:quiet], :publish => false)
               trusted = script_obj.labels.length <= 1
             end
             if trusted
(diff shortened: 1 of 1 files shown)


MahtraDR (approved):
## PR Review

**Reviewed:** 1 file, +19/-19 at `6303974` on 2026-09-09. Full clone + PR branch checked out; history, specs, and a CodeRabbit pass all consulted.
**Verdict:** Approve (one minor coverage note)
**Summary:** Fixes #1519 by making the `@@elevated_log` proc build its path from `LOG_DIR` instead of hardcoding `LICH_DIR/logs`, so `--log-dir`/`--logs` finally redirects `Script.log` output. Bundled with a mechanical `"#{DIR}/#{name}"` → `File.join(DIR, name)` refactor across `script.rb`. Both the fix and the refactor check out as correct and behavior-preserving; the only thing worth flagging is that the fixed path has no regression test.

### Minor

**1. The fixed behavior has no regression test** — [`lib/common/script.rb:263-265`](https://github.com/elanthia-online/lich-5/blob/6303974c8c84ae4fc8bdcaa7633bd125dfcf6b4b/lib/common/script.rb#L263-L265)

```ruby
Dir.mkdir(LOG_DIR) unless File.exist?(LOG_DIR)
File.open(File.join(LOG_DIR, "#{script.name}.log"), 'a') { |f| f.puts data }
```

The bug in #1519 is exactly the kind that regresses silently: a hardcoded path drifts from the override constant and nobody notices until logs land in the wrong place. The test plan runs `script_lifecycle_spec` and `script_buffer_spec` (I re-ran both: 165 examples, 0 failures), but grepping `spec/` for `Script.log`, `LOG_DIR`, and `elevated_log` returns **no** spec exercising this path — so the same regression could return unnoticed. Not a defect in the change; an absent test for the path the PR exists to fix.

It's admittedly awkward to test (the writer is an elevated `proc` gated on `Script.current`), but the seam is reachable: stub `Script.current` to return a double whose `name` is slash-free, point `LOG_DIR` at a `Dir.mktmpdir`, invoke the proc, and assert the file lands under `LOG_DIR` and not under `LICH_DIR/logs`. That directly asserts the invariant #1519 is about.

### What this PR gets right

- **The fix is correct and complete for the issue's scope.** [`LOG_DIR ||= File.join(LICH_DIR, "logs")`](https://github.com/elanthia-online/lich-5/blob/6303974c8c84ae4fc8bdcaa7633bd125dfcf6b4b/lib/constants.rb#L7) means the default path is byte-identical to the old hardcoded `LICH_DIR/logs`, so unoverridden installs are unaffected. When `--log-dir=`/`--logs=` is passed, [`lich.rbw:19-20`](https://github.com/elanthia-online/lich-5/blob/6303974c8c84ae4fc8bdcaa7633bd125dfcf6b4b/lich.rbw#L19-L20) sets `LOG_DIR` before `constants.rb`'s `||=` runs, and [`init.rb:551-559`](https://github.com/elanthia-online/lich-5/blob/6303974c8c84ae4fc8bdcaa7633bd125dfcf6b4b/lib/init.rb#L551-L559) creates it — so the proc now writes where the user asked. Both the `mkdir` and the `File.open` were updated consistently.
- **`Script.log` is genuinely the only writer of script `.log` files.** The other `/logs`-ish hit ([`dependency_recovery.rb:153`](https://github.com/elanthia-online/lich-5/blob/6303974c8c84ae4fc8bdcaa7633bd125dfcf6b4b/lib/dependency_recovery.rb#L153)) uses `temp_dir` and is unrelated, matching the issue's own "not to be confused with debug-*.log" note.
- **The `File.join` refactor really is behavior-preserving.** The one genuinely risky site is where `file_name` from `__find_script_file` can carry a leading `/custom/...` (at [`script.rb:145/147`](https://github.com/elanthia-online/lich-5/blob/6303974c8c84ae4fc8bdcaa7633bd125dfcf6b4b/lib/common/script.rb#L145-L147) and [`script.rb:476`](https://github.com/elanthia-online/lich-5/blob/6303974c8c84ae4fc8bdcaa7633bd125dfcf6b4b/lib/common/script.rb#L476)): `"#{SCRIPT_DIR}/#{file_name}"` produced a `//custom` double-slash while `File.join` collapses it to `/custom`. I executed the actual path shapes through both forms **and** through the `@name`/`@custom` derivation regex at [`script.rb:1732-1733`](https://github.com/elanthia-online/lich-5/blob/6303974c8c84ae4fc8bdcaa7633bd125dfcf6b4b/lib/common/script.rb#L1732-L1733) — `expand_path` equal, and `@name`/`@custom` identical in every case (top-level, custom root, custom subdir). The `nil` foot-gun of `File.join` is avoided: the only site that could receive `nil` ([`script.rb:476`](https://github.com/elanthia-online/lich-5/blob/6303974c8c84ae4fc8bdcaa7633bd125dfcf6b4b/lib/common/script.rb#L476)) is guarded by `file_name && ...`.
- **The resolver change is covered.** `script_version_gate_spec.rb` (38 examples, 0 failures on this branch) includes a case literally named *"does not double the /custom/ prefix when listing the custom root"* — exactly the property the `File.join` swap could have broken.
- Clean commit hygiene: fix and refactor are separate, clearly-messaged commits.

### Coverage notes

- **CI** is green on the PR over a green base. I additionally re-ran `script_lifecycle_spec`, `script_buffer_spec`, and `script_version_gate_spec` locally (all pass) and `rubocop lib/common/script.rb` (no offenses).
- **Independence:** a CodeRabbit pass against `main` returned **0 findings** on `script.rb`. The central "no behavior change" claim was checked by *execution* (a script comparing both path-construction forms and the name-derivation regex across all realistic inputs), not by citation.
- **What I did not do:** stand up a full Lich process with a real `--log-dir=` override to observe an end-to-end `.log` landing in the custom directory; I verified each link of that chain statically (arg parse → constant → init mkdir → proc) instead. The regression test in Minor #1 would settle it definitively.
- **Pre-existing, out of scope:** `Dir.mkdir(LOG_DIR)` is non-recursive, so an overridden `--log-dir` at a not-yet-existing *nested* path would fail — but `init.rb` already uses the identical non-recursive `Dir.mkdir(LOG_DIR)` and `exit`s on failure, so `Script.log` is never reached in that scenario. The PR faithfully mirrors the existing pattern.

_Reviewed with an adversarial PR-review pass (grounding + falsification-by-execution + an independent CodeRabbit pass)._

PR #1573, opened 2026-09-08
test(all): fix spec stub race breaking FFI loads, gate WindowsCredentialManager to Windows (fixes #1542)

### LINKED ISSUE #1542: Spec load order: no-op install_gem_requirements stub in login_helpers_spec breaks FFI for later spec files
## Summary

Running these two spec files together fails to load, before any example runs:

```sh
bundle exec rspec spec/lib/common/authentication/login_helpers_spec.rb \
                  spec/lib/common/session_launcher_spec.rb
```

```
An error occurred while loading ./spec/lib/common/session_launcher_spec.rb.
Failure/Error: extend FFI::Library

NameError:
  uninitialized constant Lich::Common::GUI::WindowsCredentialManager::FFI
# ./lib/common/gui/windows_credential_manager.rb:16:in '<module:WindowsCredentialManager>'
...
# ./lib/common/gui_login.rb:5:in '<top (required)>'
# ./spec/login_spec_helper.rb:127:in '<top (required)>'
# ./spec/lib/common/session_launcher_spec.rb:1:in '<top (required)>'
```

`0 examples, 0 failures, 1 error occurred outside of examples`.

## Scope

- Order-dependent: `spec/lib/common/session_launcher_spec.rb` on its own passes.
- **The full suite is unaffected** — `bundle exec rspec` is green (6292 examples, 0
  failures), so CI never sees this. It only bites when running a subset by hand,
  which is the normal loop while working on either file.
- Not a regression from any particular change; reproduces on current `main` content.
  I hit it on a branch, then confirmed it at the branch point with nothing applied.

## Root cause

Two competing stubs of `Lich::Util.install_gem_requirements`, each guarded so
whichever loads first wins and the other backs off:

- `spec/lib/common/authentication/login_helpers_spec.rb:19-25` defines a **no-op**
  version, guarded `unless defined?(Lich::Util)`.
- `spec/login_spec_helper.rb:34-40` defines a version that actually does
  `require 'os'` and `require 'ffi'`, guarded
  `unless respond_to?(:install_gem_requirements)`.

When `login_helpers_spec.rb` loads first, the no-op wins, so `login_spec_helper.rb`
declines to replace it. `login_spec_helper.rb:127` then requires `gui_login.rb`,
which pulls in `lib/common/gui/windows_credential_manager.rb`. That file's line 3
calls `Lich::Util.install_gem_requirements({ 'ffi' => true })` — now a no-op — so
`ffi` is never required, and `extend FFI::Library` on line 16 raises.

Verified directly: after loading `login_helpers_spec.rb` alone,
`Lich::Util.respond_to?(:install_gem_requirements)` is `true` while neither
`lib/util/util.rb` nor `ffi` is in `$LOADED_FEATURES`.

## Possible fixes

1. Have `spec/login_spec_helper.rb` do its `require 'ffi'` / `require 'os'`
   unconditionally rather than only as a side effect of defining the stub, so it
   no longer matters who defined the method first.
2. Or drop the ad-hoc stub in `login_helpers_spec.rb` and let the shared helper own it.
3. Independently of the spec plumbing: `windows_credential_manager.rb` does
   `extend FFI::Library` at module-body level on every platform, but only calls
   `ffi_lib` inside `if OS.windows?`. Guarding the `extend` the same way would make
   the file inert off-Windows and remove the load-order dependency entirely.

Happy to put up a PR for whichever of these you'd prefer.

mrhoribu:
## Summary

Closes #1542.

An earlier version of this PR only gated the FFI-dependent code in `windows_credential_manager.rb` behind `OS.windows?`, without touching the actual competing spec stubs the issue describes. An independent audit of that version found it only masked the bug's non-Windows manifestation: on an actual Windows load, the guard still reached `Lich::Util.install_gem_requirements`, still hit whichever no-op stub won the race, and `extend FFI::Library` still raised the identical `NameError`. I verified this myself by forcing `OS.windows?` true and running the exact two files from the issue against that version - same crash, same line. This revision fixes the actual stub race, not just where it was visible.

**Root cause:** [`spec/login_spec_helper.rb`](spec/login_spec_helper.rb) only required `'os'`/`'ffi'` as a side effect of defining its `install_gem_requirements` stub, guarded so a different file's no-op definition (`login_helpers_spec.rb`) could win the race first and block it from ever registering - after which any FFI-dependent production file loaded later got a no-op instead of the real gem-install/require behavior.

**Fix (two parts):**
1. `spec/login_spec_helper.rb`: require `'os'`/`'ffi'` unconditionally, decoupled from which `install_gem_requirements` definition wins. Verified: forcing `OS.windows?` true on the exact reported two-file order now gets past the `install_gem_requirements`/FFI race entirely (it only fails afterward on a real Win32 DLL load, which needs actual Windows hardware to test further - expected and unrelated to this bug).
2. `windows_credential_manager.rb`: additionally gate the FFI-dependent block (`install_gem_requirements`, `extend FFI::Library`, `CredentialStruct`, `attach_function`) behind `OS.windows?`, so this module also stops paying for FFI initialization it never uses on non-Windows platforms. This is a real improvement, not the fix for #1542 itself - correcting my own earlier framing, which overstated it as "the root-cause fix" and "removing the load-order dependency entirely." Two other things I mischaracterized in that framing: `attach_function` was already gated behind `OS.windows?` before this PR (only `extend FFI::Library`, the `install_gem_requirements` call, and `CredentialStruct`'s definition were unconditional); and the guard still depends on the `OS` constant already being loaded - it isn't a fully standalone/self-contained loader, just inert with respect to FFI off Windows.

## Changes

- `spec/login_spec_helper.rb`: require `os`/`ffi` unconditionally instead of only inside the (guarded) `install_gem_requirements` stub.
- `lib/common/gui/windows_credential_manager.rb`: guard the FFI-dependent block under `OS.windows?`.
- `spec/lib/common/gui/windows_credential_manager_spec.rb`:
  - This spec file could not run in isolation even on `main` (fails with a different error, for the same underlying reason: nothing in its own require chain guarantees `os`/a working `install_gem_requirements`). Added `require 'os'` plus a local guarded no-op `install_gem_requirements` stub (safe here since `os`/`ffi` are already required unconditionally by this file itself) so it's self-contained regardless of load order, on any platform.
  - The `.available? on Windows platform` example previously only passed because `CredentialStruct` was unconditionally defined regardless of platform; stubbing `OS.windows?` after load can't retroactively define Windows-only FFI structures on a non-Windows CI runner. Gated that example behind `if: OS.windows?`, and added a new test that doubles `ffi_lib`/`attach_function` (the only calls that need real Windows) to keep verifying the full install/extend/bind sequence on the Windows branch without touching real Win32 DLLs.
  - Added a regression test asserting this module never calls `Lich::Util.install_gem_requirements` off Windows, with `FFI` hidden so the assertion can't be masked by this spec file's own `require 'ffi'`.
  - Both module-load examples above use `load` to re-execute the production file. The Windows one's fake `attach_function` double defines singleton methods (`CredReadW`, etc.) on whatever module it runs against - which was the real, shared `WindowsCredentialManager` constant, leaking fake bindings into every later example in this file. `stub_const`s the constant to a disposable `Module.new` before each `load` so the reload lands on a throwaway module instead; verified the leak (`respond_to?(:CredReadW)` true after the suite) on the prior commit and confirmed it's gone now.

## Test plan

- [x] Reported order (`login_helpers_spec.rb` then `session_launcher_spec.rb`) on `main`: fails with the reported `NameError`. Passes on this branch (111 examples).
- [x] Same reported order with `OS.windows?` forced true (via a preload script) and real DLL access blocked with FFI doubles: fails identically on the previous (guard-only) revision of this PR; passes on this branch (111 examples) - this is the actual regression check for the stub race itself, not just its non-Windows symptom.
- [x] `windows_credential_manager_spec.rb` in isolation: fails on `main` (`NoMethodError`). Passes on this branch, both for real `OS.windows?` (29/30 examples) and forced-true (gets past the stub race, only fails afterward on a real DLL load that needs actual Windows hardware).
- [x] New regression/coverage tests verified against the pre-fix production file: the off-Windows one fails as expected; the on-Windows-with-doubles one is unaffected either way (it checks a property that was never broken, restoring coverage rather than proving the fix).
- [x] Post-suite check: `WindowsCredentialManager.respond_to?(:CredReadW)` is `false` after the full manager spec runs (was `true` before this fix, confirming the singleton-method leak and its resolution).
- [x] Full suite: `bundle exec rspec` - 7027 examples, 0 failures.
- [x] `rubocop -A` on all changed files - no offenses.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved platform handling for Windows credential management.
  * Non-Windows platforms now avoid loading Windows-specific credential components, preventing related compatibility issues.
  * Windows credential API integration continues to load and operate correctly on supported systems.

### DIFF
--- a/lib/common/gui/windows_credential_manager.rb
+++ b/lib/common/gui/windows_credential_manager.rb
@@ -1,7 +1,5 @@
 # frozen_string_literal: true
 
-Lich::Util.install_gem_requirements({ 'ffi' => true })
-
 module Lich
   module Common
     module GUI
@@ -13,10 +11,13 @@ module Lich
       # Uses DPAPI (Data Protection API) via Credential Manager for encryption
       # No PowerShell subprocess calls - direct Win32 API access
       module WindowsCredentialManager
-        extend FFI::Library
-
-        # Load advapi32.dll and kernel32.dll for Credential Manager functions (Windows only)
+        # Everything FFI-related is Windows-only, so it's guarded together: this keeps
+        # the module inert (no ffi/os requires, no FFI constant references) on load for
+        # every other platform instead of depending on ffi/os already being loaded by
+        # whatever required this file first.
         if OS.windows?
+          Lich::Util.install_gem_requirements({ 'ffi' => true })
+          extend FFI::Library
           ffi_lib 'advapi32', 'kernel32'
         end
 
@@ -36,26 +37,25 @@ module Lich
         # Max credential size (512KB)
         CRED_MAX_CREDENTIAL_BLOB_SIZE = 512 * 1024
 
-        # Credential structure for Win32 API
-        class CredentialStruct < FFI::Struct
-          layout(
-            :flags, :uint32,
-            :type, :uint32,
-            :target_name, :pointer,
-            :comment, :pointer,
-            :last_written, :uint64,
-            :credential_blob_size, :uint32,
-            :credential_blob, :pointer,
-            :persist, :uint32,
-            :attribute_count, :uint32,
-            :attributes, :pointer,
-            :target_alias, :pointer,
-            :user_name, :pointer
-          )
-        end
-
-        # FFI function definitions (Windows only)
+        # Credential structure and FFI function definitions (Windows only)
         if OS.windows?
+          class CredentialStruct < FFI::Struct
+            layout(
+              :flags, :uint32,
+              :type, :uint32,
+              :target_name, :pointer,
+              :comment, :pointer,
+              :last_written, :uint64,
+              :credential_blob_size, :uint32,
+              :credential_blob, :pointer,
+              :persist, :uint32,
+              :attribute_count, :uint32,
+              :attributes, :pointer,
+              :target_alias, :pointer,
+              :user_name, :pointer
+            )
+          end
+
           attach_function :CredReadW, [:pointer, :uint32, :uint32, :pointer], :bool
           attach_function :CredWriteW, [:pointer, :uint32], :bool
           attach_function :CredDeleteW, [:pointer, :uint32, :uint32], :bool
--- a/spec/lib/common/gui/windows_credential_manager_spec.rb
+++ b/spec/lib/common/gui/windows_credential_manager_spec.rb
@@ -3,6 +3,17 @@
 require_relative '../../../spec_helper'
 require 'os'
 require 'ffi'
+
+# spec_helper.rb's own Lich::Util stub doesn't define install_gem_requirements, and
+# this file may load without spec/login_spec_helper.rb (which does) ever having run.
+# `os`/`ffi` are already required unconditionally above, so this only needs to satisfy
+# the call - it isn't standing in for the real gem-install behavior.
+module Lich
+  module Util
+    def self.install_gem_requirements(*); true; end unless respond_to?(:install_gem_requirements)
+  end
+end
+
 require 'common/gui/windows_credential_manager'
 
 RSpec.describe Lich::Common::GUI::WindowsCredentialManager do
@@ -56,6 +67,35 @@ RSpec.describe Lich::Common::GUI::WindowsCredentialManager do
       expect(Lich::Util).not_to receive(:install_gem_requirements)
 
       load File.join(LIB_DIR, 'common', 'gui', 'windows_credential_manager.rb')
+    end
+  end
+
+  describe 'module load on Windows' do
+    # Real ffi_lib/attach_function calls try to dlopen advapi32/kernel32, which only
+    # exist on an actual Windows machine. Double them so this exercises the Windows
+    # branch's call sequence - installer, DLL names, function bindings - without
+    # needing real Windows, instead of skipping this branch on CI entirely.
+    it 'installs ffi, loads the Windows DLLs, and binds the expected Credential Manager functions' do
+      ffi_lib_calls = []
+      attach_function_calls = []
+      original_ffi_lib = FFI::Library.instance_method(:ffi_lib)
+      original_attach_function = FFI::Library.instance_method(:attach_function)
+      FFI::Library.define_method(:ffi_lib) { |*libs| ffi_lib_calls << libs }
+      FFI::Library.define_method(:attach_function) do |name, *_args|
+        attach_function_calls << name
+        define_singleton_method(name) { |*_a| nil }
+      end
+
+      allow(OS).to receive(:windows?).and_return(true)
+      expect(Lich::Util).to receive(:install_gem_requirements).with({ 'ffi' => true }).and_call_original
+
+      load File.join(LIB_DIR, 'common', 'gui', 'windows_credential_manager.rb')
+
+      expect(ffi_lib_calls).to eq([%w[advapi32 kernel32]])
+      expect(attach_function_calls).to eq(%i[CredReadW CredWriteW CredDeleteW CredFree GetLastError])
+    ensure
+      FFI::Library.define_method(:ffi_lib, original_ffi_lib)
+      FFI::Library.define_method(:attach_function, original_attach_function)
     end
   end
 
--- a/spec/lib/common/gui/windows_credential_manager_spec.rb
+++ b/spec/lib/common/gui/windows_credential_manager_spec.rb
@@ -58,6 +58,10 @@ RSpec.describe Lich::Common::GUI::WindowsCredentialManager do
     # never touch Lich::Util.install_gem_requirements off Windows, regardless of
     # whether a real or stubbed version of that method exists.
     it 'never calls Lich::Util.install_gem_requirements off Windows' do
+      # Reload into a disposable module rather than the real WindowsCredentialManager
+      # constant, so this doesn't leave load-time state (or, in the Windows test below,
+      # fake singleton bindings) on the module every other example in this file shares.
+      stub_const('Lich::Common::GUI::WindowsCredentialManager', Module.new)
       allow(OS).to receive(:windows?).and_return(false)
       # This file's own `require 'ffi'` above already defines FFI, which would mask
       # a regression where extend FFI::Library/CredentialStruct escaped the guard
@@ -76,6 +80,12 @@ RSpec.describe Lich::Common::GUI::WindowsCredentialManager do
     # branch's call sequence - installer, DLL names, function bindings - without
     # needing real Windows, instead of skipping this branch on CI entirely.
     it 'installs ffi, loads the Windows DLLs, and binds the expected Credential Manager functions' do
+      # The fake attach_function below defines singleton methods (CredReadW, etc.) on
+      # whatever module it's called on. Reload into a disposable module instead of the
+      # real WindowsCredentialManager constant so those fakes don't leak into it and
+      # get exercised by later examples in this file instead of the real bindings.
+      stub_const('Lich::Common::GUI::WindowsCredentialManager', Module.new)
+
       ffi_lib_calls = []
       attach_function_calls = []
       original_ffi_lib = FFI::Library.instance_method(:ffi_lib)
--- a/spec/lib/common/gui/windows_credential_manager_spec.rb
+++ b/spec/lib/common/gui/windows_credential_manager_spec.rb
@@ -48,6 +48,11 @@ RSpec.describe Lich::Common::GUI::WindowsCredentialManager do
     # whether a real or stubbed version of that method exists.
     it 'never calls Lich::Util.install_gem_requirements off Windows' do
       allow(OS).to receive(:windows?).and_return(false)
+      # This file's own `require 'ffi'` above already defines FFI, which would mask
+      # a regression where extend FFI::Library/CredentialStruct escaped the guard
+      # without calling install_gem_requirements. Hide it so a reload only succeeds
+      # if the guard keeps every FFI reference out of the non-Windows load path.
+      hide_const('FFI')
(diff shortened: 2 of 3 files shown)


mrhoribu:
@coderabbitai review in full again

Nisugi:
## Review

Verified by building worktrees for this branch and `origin/main` and running the specs on both. Full suite shows an identical 60 pre-existing Windows-environment failures on each branch, so **this PR introduces no regression**. Note the example-count gap (7028 here vs 7708 on main) is branch drift — the head is 43 commits behind main, so the "7027 examples, 0 failures" in the test plan was measured against a stale base.

Three findings, one substantive.

### 1. The stub race is only half fixed — and the guard hoist relocated it

`spec/lib/common/gui/master_password_manager_spec.rb` requires the same production file but did not get the `require 'os'` / `require 'ffi'` treatment that `login_spec_helper.rb` and `windows_credential_manager_spec.rb` did. Reproduced on this branch:

```
$ rspec spec/lib/util/textstripper_spec.rb spec/lib/common/gui/master_password_manager_spec.rb --order defined
NameError: uninitialized constant Lich::Common::GUI::WindowsCredentialManager::OS
  at lib/common/gui/windows_credential_manager.rb:18
```

`textstripper_spec.rb` defines its own no-op `install_gem_requirements` and wins the same `unless respond_to?` race the PR description identifies. `os` is therefore never required — and the `if OS.windows?` guard that this PR *hoisted to line 18* now dereferences an undefined `OS`.

On `main` the same file combination failed on `extend FFI::Library`. So the failure mode moved rather than being eliminated, and hoisting the guard earlier in the file is what moved it. The PR body has already walked back one round of "root-cause fix" framing after an audit; this suggests the corrected framing is still overstated, since the underlying race (`unless respond_to?`-guarded stubs competing across spec files) is untouched and still reachable through a different entry point.

Adding the same two requires to `master_password_manager_spec.rb` makes that combination pass (55 examples, 0 failures, verified). That said, patching each spec file as it turns up is treating symptoms one at a time — it may be worth deciding whether the competing-stub pattern itself should be centralized in `spec_helper.rb` so no future spec can re-open this.

### 2. `.and_call_original` can reach the real gem installer

`spec/lib/common/gui/windows_credential_manager_spec.rb:100` calls `.and_call_original` on `install_gem_requirements`. Because every spec-side stub is `unless respond_to?`-guarded, the real `lib/util/util.rb` implementation wins whenever `spec/lib/util/util_spec.rb` has loaded first. This currently passes only because ffi happens to already be installed; in an environment without the gem, a unit test would attempt a live RubyGems install and raise `Please install the failed gems: ffi manually to continue.`

Asserting on the call without `.and_call_original` would keep the coverage without the live-install path.

### 3. Global `FFI::Library` monkeypatch is broader than it needs to be

`spec/lib/common/gui/windows_credential_manager_spec.rb:93` redefines `ffi_lib`/`attach_function` on the process-global `FFI::Library` class. Any lazily-required FFI binding loaded elsewhere during that window would silently get no-op fakes.

The `ensure` block restores both methods correctly on normal and exceptional exit, so this is exposure rather than an observed failure — flagging it because it is the same class of problem as the singleton-method leak this PR already fixed with `stub_const`, and scoping the doubles to a throwaway module would close it the same way.

---

Findings 2 and 3 are test hygiene and wouldn't block on their own. Finding 1 is worth resolving before merge, or at least worth adjusting the PR description so the fix isn't recorded as more complete than it is.

mrhoribu:
@Nisugi thanks for the thorough re-audit — verified and addressed all three findings in [`f3bfc91c`](https://github.com/elanthia-online/lich-5/pull/1573/commits/f3bfc91c):

**1. Stub race (substantive finding)** — reproduced your exact repro and confirmed it went further than `master_password_manager_spec.rb` alone: `textstripper_spec.rb`'s stub is *unguarded* (clobbers even a working implementation), and `master_password_manager_spec.rb` still failed in the reverse order and in isolation, for the same reason `windows_credential_manager_spec.rb` needed a local fix earlier — `spec_helper.rb`'s own `Lich::Util` stub never defined `install_gem_requirements` at all. Rather than patch a fourth file one at a time, fixed it at the actual choke point: `spec_helper.rb` (required first by nearly every spec) now defines a real, safe `install_gem_requirements` that just requires whatever gem was asked for, plus an unconditional `require 'os'`. Verified against your repro, its reverse order, and `master_password_manager_spec.rb` standalone — all pass now. The per-file fixes stay in place as defense in depth for files that don't require `spec_helper.rb` first (`login_helpers_spec.rb`, by design).

**2. `.and_call_original` risk** — agreed, dropped it in favor of a bare mock expectation; the test doesn't need the real call to pass, and this removes the live-install exposure you flagged.

**3. Global `FFI::Library` monkeypatch** — went further than narrowing scope: replaced it entirely with singleton methods defined directly on the disposable module being reloaded into. A module's own singleton methods take priority over ones gained via `extend`, so this shadows `ffi_lib`/`attach_function` for that module only — nothing touches `FFI::Library` process-wide anymore, and there's nothing left to restore in an `ensure`.

Full suite and rubocop stayed clean throughout (verified after each change, not just at the end).

_🤖 Posted by [Claude Code](https://claude.com/claude-code) on behalf of @mrhoribu_

MahtraDR:
## Review — `f3bfc91` · Approve

**Reviewed:** 5 files, +133/-29 at `f3bfc91`, from a full worktree checkout with the suite executed on Ruby 4.0.6.
**Verdict:** Approve — nothing above a foot-gun survived verification.

Fixes the spec load-order race where competing `install_gem_requirements` stubs could clobber the real gem-require behavior and crash FFI-dependent loads, and makes `WindowsCredentialManager` platform-inert off Windows by gating the whole FFI block behind `OS.windows?`. The fix is grounded, the regression test genuinely fails-before/passes-after, and the full suite + rubocop are clean.

### Minor / foot-gun (non-blocking)

**`spec/spec_helper.rb:1895` — stub requires *every* gem regardless of `should_require`, with no rescue.**

```ruby
def install_gem_requirements(gems_to_install, **_kwargs)
  gems_to_install.each_key { |gem_name| require gem_name }
  true
end
```

The real impl (`lib/util/util.rb:308`) only does `require gem_name if should_require`, and the hash value is meaningful — `{ 'kramdown' => false }` in `textstripper.rb` means *install but do not require*. This stub ignores the value, unconditionally `require`s the key, and uses the key as the require path. Two latent foot-guns, neither triggered today (every current key — `ffi`/`os`/`kramdown` — is bundled and gem-name == require-path, which is why the suite passes):

- A future `install_gem_requirements({ 'some_optional_gem' => false })` for a gem not in the bundle will raise `LoadError` at spec-load time (no `rescue`), where the old no-op stubs passed silently. The real impl only *installs* such a gem.
- A gem whose require path differs from its name (e.g. `ruby-openai` → `require 'openai'`) fails here even though the real impl handles it.

It's a stub, so matching the real semantics isn't required — just consider a `rescue LoadError` or a one-line comment so the next person adding an optional-gem requirement doesn't hit a confusing spec-load crash.

### What this gets right

- **The regression test earns its keep.** Swapping the pre-fix (`6f12688`) production file back in makes the `module load (regression for #1542)` example fail with `expected: 0 times … received: 1 time` on `install_gem_requirements`, and pass on the fixed file — a real fail-before/pass-after guard, not a tautology.
- **Honest framing.** The body/commits correct the earlier "root-cause fix / removes load-order dependency entirely" over-claim, note `attach_function` was already gated, and note the guard still needs `OS` pre-loaded. The "module load on Windows" test is correctly described as *restoring* coverage, not proving the fix.
- **Safe off Windows.** Every public method returns early via `available?`/`OS.windows?` before touching `CredentialStruct`/`FFI`; the sole consumer `master_password_manager.rb` requires `os` before `require_relative 'windows_credential_manager'`, so the load-time `OS.windows?` reference stays satisfied — same precondition the code already had.
- **Correct isolation technique.** Reloading into a `stub_const`'d `Module.new` with `ffi_lib`/`attach_function` as the module's *own* singleton methods (shadowing the `extend FFI::Library` versions) avoids both the shared-constant singleton-method leak and a global `FFI::Library` monkeypatch. `hide_const('FFI')` in the off-Windows test prevents the file's own `require 'ffi'` from masking a guard escape.
- **Right altitude.** Fixing it at `spec_helper.rb` (required first by nearly every spec) means the per-file guarded stubs normally never win the race, with the per-file fixes as defense-in-depth for the deliberately-isolated `login_helpers_spec.rb`.

### Coverage / how this was checked

- CI green on the PR (Rubocop + both Ruby test jobs + title). Reproduced locally: full `bundle exec rspec` = **7027 examples, 0 failures**; rubocop on all 5 changed files = no offenses.
- Independent reviewer: **CodeRabbit** (`--agent --base main`) ran, **0 findings** across all 5 files.
- Falsification-by-execution: verified the `#1542` regression test fails against the pre-fix file / passes against the fix; ran the `textstripper_spec + master_password_manager_spec --order defined` repro (55 ex, 0 fail), `master_password_manager_spec` alone (42), and `windows_credential_manager_spec` (30) / `textstripper_spec` (13) / `login_helpers_spec` (91) in isolation.
- **Not covered:** exhaustive RSpec ordering permutations (ran the default randomized full suite once + the specific named orders); real Windows FFI binding (`ffi_lib`/`CredReadW` on actual Windows hardware) — the doubles verify the call sequence, not the real DLL load. Both are inherent CI limits, honestly disclosed in the body.

_Reviewed at `f3bfc91cb37430c70d0aa76f84eed87fd4127157`._

PR #1574, opened 2026-09-08
fix(dr): unpause the exact scripts safe_pause_list paused, by object

MahtraDR:
## Problem

`DRC.safe_pause_list` returned a list of script *names*, and `DRC.safe_unpause_list` restored by re-scanning live `Script.running` and matching those names:

```ruby
Script.running.find_all { |s| s.paused? && !s.no_pause_all && names.include?(s.name) }.each(&:unpause)
```

The unpause set is re-derived from live state at unpause time, not from what was actually paused. Any script that leaves `Script.running` between pause and unpause — gone hidden, or mid start/teardown — or whose `no_pause_all` flag flips, silently drops out of the rescan and is **stranded paused forever**. Because pausing is cooperative, that script parks at its next checkpoint and never wakes. The log also actively misleads: the `Unpausing [...]` message prints the captured names, so it claims to unpause a script the code never touches.

**Real incident:** `almanac`'s `safe_pause_list` paused `t2` while `t2` was mid-`wait_for_script_to_complete` launching a child (`autopick`). By unpause time `t2` was no longer in the rescanned set, so `safe_unpause_list` skipped it even though the message listed it. `t2` stayed paused for the rest of the session (no further output for ~21 min).

## Fix

`safe_pause_list` captures and returns the exact **Script objects** it paused. `safe_unpause_list` unpauses those objects directly (`unpause if paused?`), regardless of their current `Script.running` visibility or flags — undoing exactly the pause it issued. The `Pausing`/`Unpausing` log messages map the same objects to names, so message and action can no longer diverge.

No caller inspects the returned list's contents (all ~15 dr-scripts callers use it only as a truthiness gate and pass it back), so the names→objects contract swap is safe.

## Scope / non-goals

Targeted follow-up to #1525. Like that hotfix, it does **not** remove the fragile pausing-as-coordination model or the multiple non-interoperating pause mechanisms — the lease-based command broker remains the structural fix. This removes one silent-strand failure mode now.

## Testing

- New regression spec in `spec/lib/dragonrealms/commons/common_spec.rb`: a paused script that leaves `Script.running` is still unpaused. Verified **red on the old code** (unpause called 0 times) and **green on the new**.
- Updated the existing safe_pause/unpause specs to the object contract.
- Full suite green (7147 examples, 0 failures) and `rubocop` clean.

### DIFF
--- a/lib/dragonrealms/commons/common.rb
+++ b/lib/dragonrealms/commons/common.rb
@@ -983,23 +983,35 @@ module Lich
         @safe_pause_prev_ignore_pause = @safe_pause_holder&.ignore_pause
         @safe_pause_holder&.ignore_pause = true
 
-        paused_script_list = []
+        # Capture the exact Script objects we pause -- not their names. The
+        # matching safe_unpause_list restores from this list directly instead of
+        # re-scanning Script.running by name. A name + live-rescan restore
+        # silently drops any script that has left Script.running between pause
+        # and unpause (gone hidden, or mid start/teardown) or whose pause flags
+        # changed, stranding it paused forever while the log still claims it was
+        # unpaused. Holding the objects makes the restore undo exactly what the
+        # pause did.
+        paused_scripts = []
         Script.running.find_all { |s| !s.paused? && !s.no_pause_all && s.name != Script.self.name }.each do |s|
           s.pause
-          paused_script_list << s.name
+          paused_scripts << s
         end
-        Lich::Messaging.msg("plain", "DRC: Pausing #{paused_script_list} to run #{Script.self.name}")
-        return paused_script_list
+        Lich::Messaging.msg("plain", "DRC: Pausing #{paused_scripts.map(&:name)} to run #{Script.self.name}")
+        return paused_scripts
       end
 
-      def safe_unpause_list(scripts_to_unpause)
+      def safe_unpause_list(paused_scripts)
         return false unless $safe_pause_lock.owned?
 
-        if scripts_to_unpause.empty?
+        if paused_scripts.empty?
           Lich::Messaging.msg("plain", "DRC: #{Script.self.name} has finished.")
         else
-          Lich::Messaging.msg("plain", "DRC: Unpausing #{scripts_to_unpause}, #{Script.self.name} has finished.")
-          Script.running.find_all { |s| s.paused? && !s.no_pause_all && scripts_to_unpause.include?(s.name) }.each(&:unpause)
+          Lich::Messaging.msg("plain", "DRC: Unpausing #{paused_scripts.map(&:name)}, #{Script.self.name} has finished.")
+          # Unpause exactly the objects we paused, regardless of their current
+          # Script.running visibility or flags. Skip any a peer already unpaused
+          # so we neither emit spurious "is not paused" noise nor fight another
+          # coordinator that has since taken ownership of the pause.
+          paused_scripts.each { |s| s.unpause if s.paused? }
         end
         # Restore the holder's pre-lock pause immunity before releasing the lock,
         # so we never leave a script permanently unpausable.
--- a/spec/lib/dragonrealms/commons/common_spec.rb
+++ b/spec/lib/dragonrealms/commons/common_spec.rb
@@ -1042,9 +1042,9 @@ RSpec.describe Lich::DragonRealms::DRC do
         expect(described_class.safe_pause_list).to be false
       end
 
-      it 'pauses scripts and returns their names' do
+      it 'pauses scripts and returns the paused Script objects' do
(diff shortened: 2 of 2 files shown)


mrhoribu:
# PR Review: fix(dr): unpause the exact scripts safe_pause_list paused, by object (elanthia-online/lich-5#1574)

**Reviewed:** 2 files, +44/-14 at `949626ef` on 2026-09-08. Based on: full local clone (lich-5, plus sibling dr-scripts/scripts repos already checked out alongside it) + isolated worktree build/test run.
**Verdict:** Approve
**Summary:** Changes `DRC.safe_pause_list`/`safe_unpause_list` to capture and restore the exact `Script` objects paused, instead of names re-resolved against a live `Script.running` rescan at unpause time — closing a real stranded-pause bug (documented incident: `almanac`/`t2`/`autopick`, ~21 min stuck) where a script that left `Script.running` between pause and unpause was silently never restored. The fix is correct, narrowly scoped, and thoroughly tested.

## What this PR gets right

- **The fix matches the mechanism, not just the symptom.** I traced `Script#pause`/`unpause`/`paused?` in `lib/common/script.rb:2786-2804` — they're plain ivar reads/writes (`@paused`), entirely independent of `@@running` registry membership. `Script.running` (`lib/common/script.rb:1561-1567`) additionally filters out hidden scripts. So holding the actual object and calling `.unpause` on it works correctly regardless of whether the script has gone hidden or been pruned from the registry mid-window — exactly the failure mode described in the PR body, and I confirmed the object-based approach structurally cannot reproduce it.
- **The "no caller inspects contents" claim is verified, not assumed.** I grepped every `.lic` caller in dr-scripts and scripts (`tessera`, `buff-watcher`, `wild-monitor`, `almanac`, `wand-watcher`, `tendme`, `empathylink`, `tarantula`, `smartlisten`, `heroic-tattoo`, `walkingastro`, `ranger-companion`, `tome`, `burgle` — 14 files) for any use of the returned value besides `until (x = DRC.safe_pause_list)` / passing it straight to `safe_unpause_list`. None inspect names, iterate, or `.include?` against it. The `scripts` repo has zero references. The names→objects contract swap is safe.
- **Dropping the `no_pause_all` recheck at unpause time is deliberate and consistent with the fix's own logic**, not an oversight: the old code re-filtered by `!s.no_pause_all` at unpause, which could itself strand a script if that flag flipped mid-window — the same class of bug this PR fixes for registry visibility. The new code unconditionally undoes exactly what it did, which is the right call and is called out explicitly in the code comment (`lib/dragonrealms/commons/common.rb:1010-1013`).
- **The regression test is real, not decorative.** `unpauses a paused script even after it leaves Script.running` (`spec/lib/dragonrealms/commons/common_spec.rb:1071-1080`) stubs `Script.running` to return `[]` between pause and unpause — reproducing the exact incident shape — and I ran it locally (`bundle exec rspec ... -e safe_pause_list -e safe_unpause_list`): 11 examples, 0 failures. `rubocop` is clean on both changed files. CI (rubocop, both test jobs, PR-title check) is green.
- Object identity also incidentally fixes a latent adjacent bug: two distinct running scripts sharing a name could never be conflated by this path anymore, since matching is now by reference rather than by name string.
- Comments are unusually good here — they explain *why* (the cooperative-pause/registry-visibility mechanism), not what, and will save the next reader from re-deriving the incident from scratch.

## Coverage notes

- Full local clone used for lich-5, plus the sibling dr-scripts/scripts checkouts already present in this environment, so caller impact was verified directly rather than inferred from the PR description.
- Built an isolated `git worktree` for the PR head commit (949626ef) to avoid disturbing unrelated uncommitted work already present in the main lich-5 checkout; ran `bundle exec rspec` (11/11 passing) and `bundle exec rubocop` (clean) there directly, and removed the worktree afterward.
- Ran `coderabbit review --agent` against the same worktree (base `origin/main` at merge-base `3c655903`): 0 findings, corroborating this review.
- Not independently re-verified: the live-incident claim in the PR body (almanac/t2/autopick, ~21 min stall) — taken as given since it's the author's own operational report and the regression test reproduces the mechanism described.
- Noted but out of scope for this review: `dr-scripts/spec/empathylink_spec.rb` mocks `DRC.safe_pause_list` to return string names (`['combat-trainer', 'hunting-buddy']`) rather than objects — stale relative to the new contract. It doesn't matter functionally (the mock never touches real `safe_pause_list`, and empathylink.lic only passes the value through), and it lives in a separate repo, so it's not this PR's responsibility to fix, just worth a heads-up if/when dr-scripts specs get touched next.

mrhoribu (approved):
Approving per the review posted above — https://github.com/elanthia-online/lich-5/pull/1574#issuecomment-5593243104

PR #1575, opened 2026-09-09
feat(all): add opt-in bounded script execution guards

therealatari:
## Summary

- add opt-in cooperative execution guards to an exact `Script` instance
- gate game writes, downstream/upstream reads, pause polling, and explicit guarded waits through one caller-supplied policy
- support guarded named-child startup plus an optional prohibition on descendant script launches
- make cancellation sticky across worker threads, cleanup, rescued interruptions, and script exit
- preserve existing behavior when no guard is installed

## Contract

`Script#with_execution_guard(policy)` installs a scoped policy. `Script.start`, `run`, `run_child`, and `start_child` can install the same policy before a child worker begins. A frozen String represents an impending game write; `nil` represents a checkpoint. Only literal `true` permits continuation.

The API exposes protocol markers for instance guards, guarded startup, and descendant-launch restriction. Guarded code can call `check_execution_guard!` and `execution_sleep`. The exact first cancellation reason wins, and cleanup removes only the owning guard.

This is cooperative control, not a sandbox or preemptive timeout. It does not intercept arbitrary Ruby, direct socket access, or blocking code without checkpoints. Cancellation cannot undo a command already sent. Callers must review the helpers they place inside the scope.

## Compatibility

- scripts without an active guard retain existing sends, waits, reads, pause behavior, and child launches
- children do not inherit a parent guard implicitly
- `allow_script_starts:` defaults true and requires an installed policy when false
- existing helper arguments and return behavior remain intact
- guarded writes intentionally perform their final guard check before displaying and sending; this can change echo/game-text interleaving only for scripts that opt into a guard

## Review scope

The branch's merge of `main` can make GitHub's aggregate diff attribute unrelated upstream files to this branch. The commits authored for this PR are scoped to:

- `lib/common/script.rb`
- `lib/common/script_execution_guard.rb`
- `lib/games.rb`
- `lib/global_defs.rb`
- `docs/script-execution-guard.md`
- the corresponding execution-guard regression specs

Files such as `lib/common/move.rb`, `lib/common/events.rb`, `lib/common/hook_registry.rb`, and `lib/gemstone/combat/**` arrived from upstream and are not authored changes in this PR.

## Verification

- focused execution-guard, lifecycle, write, movement, and global-helper suite: **339 examples, 0 failures**
- hosted full suite at current head: **7,860 examples, 0 failures**
- hosted RuboCop and syntax checks: green
- regressions cover cancellation during buffer reads, policy callback failures, worker ownership, nested/overlapping guards, cleanup, child adoption/startup, forbidden descendants, thread-priority restoration, `dothis`/`dothistimeout`, and unguarded compatibility
- `git diff --check`: clean

The complete public contract and limitations are in `docs/script-execution-guard.md`.

### DIFF
--- a/docs/script-execution-guard.md
+++ b/docs/script-execution-guard.md
@@ -74,7 +74,10 @@ child.join
 raise 'go2 failed' unless child.completed_successfully?
 ```
 
-`Script::START_EXECUTION_GUARD_PROTOCOL == 1` identifies support. `Script.start`,
+`Script::EXECUTION_GUARD_PROTOCOL == 1` identifies the instance guard API
+(`with_execution_guard`, `execution_guard_active?`, `check_execution_guard!`,
+and `execution_sleep`). `Script::START_EXECUTION_GUARD_PROTOCOL == 1`
+separately identifies guarded named-script startup support. `Script.start`,
 `Script.run`, and `Script.run_child` accept the same options hash; `run_child`
 also accepts `execution_guard: policy` alongside its `timeout:` keyword.
 Omitting the option (or passing `nil`) keeps ordinary launch behavior.
--- a/docs/script-execution-guard.md
+++ b/docs/script-execution-guard.md
@@ -1,7 +1,9 @@
 # Script execution guards
 
-For the motivating casting example, existing retry behavior, and verification
-boundaries, see [Spell casting in bounded test sequences](spell-cast-testing-use-case.md).
+This API lets a supervising script place cooperative command, time, and
+cancellation boundaries around existing helpers without changing their normal
+unguarded behavior. Spell-specific integration is intentionally documented and
+reviewed separately.
 
 `Script#with_execution_guard(policy)` installs an optional cooperative policy
 for that script until its block exits. `policy` must be a `Proc` accepting one
--- a/docs/script-execution-guard.md
+++ b/docs/script-execution-guard.md
@@ -15,6 +15,11 @@ argument and returning **literal `true`** to permit continuation:
   `Game._puts` passes the raw command supplied by its caller without adding or
   stripping a prefix.
 
+`Game.puts` retains its script pause checkpoint. Raw `Game._puts` resolves the
+calling script without waiting for unpause, including when its caller holds a
+shared lock such as `Inventory.refresh`. An installed guard still checks every
+raw write under the socket mutex immediately before transport.
+
 For example, this policy bounds elapsed time and attempted writes for one thread:
 
 ```ruby
--- a/docs/script-execution-guard.md
+++ b/docs/script-execution-guard.md
@@ -1,0 +1,147 @@
+# Script execution guards
+
+For the motivating casting example, existing retry behavior, and verification
+boundaries, see [Spell casting in bounded test sequences](spell-cast-testing-use-case.md).
+
+`Script#with_execution_guard(policy)` installs an optional cooperative policy
+for that script until its block exits. `policy` must be a `Proc` accepting one
+argument and returning **literal `true`** to permit continuation:
+
+- `nil` means a checkpoint, such as a downstream read or guarded wait.
+- A frozen `String` means an impending game socket write. It is a detached copy
+  of the exact wire command. `Game.puts` has already added `$cmd_prefix`;
+  `Game._puts` passes the raw command supplied by its caller without adding or
+  stripping a prefix.
+
+For example, this policy bounds elapsed time and attempted writes for one thread:
+
+```ruby
+deadline = Process.clock_gettime(Process::CLOCK_MONOTONIC) + 10.0
+writes = 0
+policy = lambda do |command|
+  within_time = Process.clock_gettime(Process::CLOCK_MONOTONIC) < deadline
+  within_budget = command.nil? || (writes += 1) <= 3
+  within_time && within_budget
+end
+
+begin
+  Script.current.with_execution_guard(policy) do |_guard|
+    fput 'look'
+  end
+rescue Lich::Common::ScriptExecutionGuard::Interrupted => error
+  echo "Stopped: #{error.reason}"
+end
+```
+
+There is no built-in command budget or deadline: the policy supplies those
+decisions. Retries reach the socket check separately; `Game.puts` delegating to
+`Game._puts` is checked once. A permitted write attempt does not prove that the
+server received or executed it. Policies should inspect local observations,
+return promptly, and never send commands. Recursive sends cancel the scope with
+`:reentrant_command`. Reading `Script.current` from the callback is supported.
+
+The block receives a cancellation handle. `guard.cancel!(:manual_hold)` latches
+cancellation; the next checkpoint raises `Interrupted`. Cancellation is shared
+across threads, and the first reason wins. Reasons must be short lowercase
+identifiers; invalid labels become `:cancelled`. Returning anything except `true`
+latches `:checkpoint_rejected` or `:command_rejected`. Callback exceptions become
+`:callback_error`; their original text and cause are not exposed by the
+interruption.
+
+The scope checks both entry and successful block completion. Rescuing an
+interruption inside the block does not clear cancellation. Cleanup closes the
+handle and removes the policy even when the block raises. An overlapping or
+nested guard on the same script raises `ArgumentError`.
+
+The policy belongs to the Script instance. Workers attributed to that script
+share it; separately started child scripts do not inherit it. The owner must
+join its workers before leaving the scope. If workers share policy counters or
+other mutable observations, synchronize that state inside the policy.
+
+## Guarding a named child from startup through cleanup
+
+For an existing script that cannot install its own block scope, pass the policy
+in the normal launch options:
+
+```ruby
+child = Script.start_child('go2', '223 --disable-confirm',
+                           { quiet: true, execution_guard: policy })
+raise 'go2 did not start' unless child
+# Retain this exact object for supervision; a same-name script is not its owner.
+child.join
+raise 'go2 failed' unless child.completed_successfully?
+```
+
+`Script::START_EXECUTION_GUARD_PROTOCOL == 1` identifies support. `Script.start`,
+`Script.run`, and `Script.run_child` accept the same options hash; `run_child`
+also accepts `execution_guard: policy` alongside its `timeout:` keyword.
+Omitting the option (or passing `nil`) keeps ordinary launch behavior.
+
+The policy is installed before the worker starts, after which the normal child
+adoption and startup gate ensure that its first checkpoint belongs to the exact
+child. The callback is evaluated by that child's worker, not by the launching
+parent. A refused initial checkpoint prevents its script body from running and
+is recorded in `exit_error`. Policies must therefore support startup observations
+before the caller has received the returned handle. A caller using a short
+startup handshake must bound that wait and permit cancellation.
+
+Unlike a block scope, this guard stays active during the script body's `ensure`
+blocks and before-dying callbacks, including external-kill cleanup. A cancelled
+policy does not permit cleanup commands; owners must provide a separately
+authorized recovery operation when needed. The guard closes only after cleanup
+and owned workers have terminated, or after an abandoned startup is discarded.
+Rescuing an interruption or calling `exit` cannot turn a cancelled run into a
+successful result. Guarded children still do **not** implicitly guard further
+child scripts; callers must explicitly supervise any such descendants. This
+option supplies no routing rules, automatic return trip, or preemptive timeout.
+
+### Prohibiting descendant launches explicitly
+
+`Script::SCRIPT_START_RESTRICTION_PROTOCOL == 1` identifies support for
+`allow_script_starts: false`. Supply it with the launch policy when an existing
+helper must not create unguarded scripts:
+
+```ruby
+child = Script.start_child('go2', '223 --disable-confirm',
+                           { quiet: true, execution_guard: policy,
+                             allow_script_starts: false })
+```
+
+The same opt-in applies to block scopes, for example read-only map admission:
+
+```ruby
+Script.current.with_execution_guard(->(wire) { wire.nil? }, allow_script_starts: false) do
+  Map.dijkstra(origin_id, destination_id)
+end
+```
+
+Native named, anonymous, and exec-script startup admission checks the actual
+calling worker's guard before reserving a script name, adopting a child, or
+creating a worker. `force: true` and explicitly detached parent options do not
+bypass the check. A prohibited launch latches `:script_start_rejected`, including
+attempts from policy callbacks or teardown callbacks; subsequent commands remain
+cancelled. `run_child` also accepts `allow_script_starts:` as a keyword. A launch
+restriction requires a policy, and the value must be boolean. The default is
+`true`, preserving existing script-start behavior for ordinary guards and scripts.
+
+This is deliberately not a restriction on all Ruby effects or script lifecycle
+operations: it does not intercept direct socket access, arbitrary constructors,
+or separately requested kills. It prevents native script launches from escaping
+a reviewed helper's execution scope, rather than automatically extending that
+scope to unknown descendants.
+
+Checks currently cover the game write seam, downstream `gets`/`gets?`, upstream
+and unique-script reads (`upstream_gets`, `unique_gets`, and their `?` polling
+forms), script pause checkpoints, and helpers routed through
+`Script.execution_sleep`.
+`script.check_execution_guard!` provides an explicit checkpoint, and
+`script.execution_sleep(seconds)` provides a cooperative finite wait. Guarded
+downstream reads, paused-script waits and execution sleeps poll at intervals of
+at most 0.05 seconds, subject to scheduling and callback duration. Code without
+an active guard retains its ordinary behavior.
+
+This is not a sandbox or preemptive timeout. Arbitrary Ruby loops, direct socket
+access, unrelated scripts, and blocking calls without checkpoints are outside
+its coverage. Cancellation cannot undo commands already sent or interrupt an
+operating-system socket write already in progress. A caller must review the
+paths it invokes rather than infer complete coverage from installing a guard.
--- a/lib/common/script.rb
+++ b/lib/common/script.rb
@@ -1079,6 +1079,15 @@ module Lich
       end
       private_class_method :__resolve_current
 
+      # Resolves the calling script for core seams that must not add a pause
+      # checkpoint, including raw writes made while holding a caller's lock.
+      #
+      # @return [Script, nil] the script bound to Thread.current, or nil
+      # @api private
+      def Script.current_without_pause
+        __resolve_current
+      end
+
       # Returns the script bound to the calling thread.
       #
       # Blocks the calling thread while that script is paused (unless it has
--- a/lib/common/script.rb
+++ b/lib/common/script.rb
@@ -9,6 +9,7 @@
 
 require 'weakref'
 require_relative 'script_death'
+require_relative 'script_execution_guard'
 
 module Lich
   module Common
@@ -38,6 +39,12 @@ module Lich
       RAW_THREAD_GROUP_ENCLOSED = ThreadGroup.instance_method(:enclosed?)
       JOIN_WAIT_INTERVAL = 0.05
       CLEANUP_SCRIPT_THREAD_KEY = :lich_cleanup_script
+      EXECUTION_GUARD_MUTEX_INITIALIZER = Mutex.new
+      EXECUTION_GUARD_POLL_INTERVAL = 0.05
+      # Named-script starts accept a policy installed before the worker runs.
+      START_EXECUTION_GUARD_PROTOCOL = 1
+      # Guards can explicitly prohibit native script starts by their workers.
+      SCRIPT_START_RESTRICTION_PROTOCOL = 1
 
       # Nested lock order:
       #   startup -> registry -> per-script lifecycle -> child relationship
@@ -118,6 +125,18 @@ module Lich
           script_args = (options[:args] || String.new)
         end
 
+        options ||= {}
+        allow_script_starts = options.fetch(:allow_script_starts, true)
+        unless allow_script_starts.equal?(true) || allow_script_starts.equal?(false)
+          raise ArgumentError, 'allow_script_starts must be true or false'
+        end
+        if !allow_script_starts && options[:execution_guard].nil?
+          raise ArgumentError, 'restricting script starts requires an execution guard'
+        end
+        unless options[:execution_guard].nil?
+          startup_guard = ScriptExecutionGuard.new(options[:execution_guard], :allow_script_starts => allow_script_starts)
+        end
+
         # fixme: look in wizard script directory
         # Resolve via the shared resolver so script discovery stays identical
         # everywhere (custom/ root, custom/<subdir>/, then SCRIPT_DIR root).
@@ -167,6 +186,7 @@ module Lich
           launcher_thread = Thread.current
           gate_release_failed = false
           begin
+            script_obj.__send__(:__install_startup_execution_guard, startup_guard) if startup_guard
             new_thread = Thread.new {
               begin
                 next unless Script.__send__(:__await_thread_gate, start_gate, launcher_thread, false)
@@ -207,6 +227,12 @@ module Lich
                 else
                   respond '--- error: out of cheese'
                 end
+              rescue ScriptExecutionGuard::Interrupted => error
+                # The initial Script.current pause checkpoint runs before the
+                # ordinary script-body error handler. Preserve its denial as a
+                # failed child result, not an unreported thread exception.
+                script_obj.__send__(:__record_exit_error, error)
+                Script.__send__(:__report_script_error, script_obj, error, :untrusted => !trusted)
               ensure
                 script_obj.kill if script_obj.running? && !script_obj.stopping?
               end
@@ -737,6 +763,7 @@ module Lich
       end
 
       def Script.__begin_start(reservation, name = nil, force: true)
+        __resolve_current&.__send__(:__check_script_start_allowed!)
         normalized_name = name&.downcase
         @@startup_mutex.synchronize do
           __reap_abandoned_startups_locked
@@ -760,6 +787,7 @@ module Lich
       private_class_method :__begin_start
 
       def Script.__begin_library_start(reservation, name, deadline: nil)
+        __resolve_current&.__send__(:__check_script_start_allowed!)
         normalized_name = name.downcase
         @@startup_mutex.synchronize do
           __reap_abandoned_startups_locked
@@ -971,6 +999,8 @@ module Lich
       # Starts a named script and registers it as a child of the current script.
       # Outside a script worker, this behaves like {Script.start}.
       #
+      # @param args [Array] {.start} arguments, including optional
+      #   +execution_guard: Proc+ in the existing options hash
       # @return [Script, nil] the started script, or nil when startup fails
       def Script.start_child(*args)
         parent = Script.current
@@ -987,14 +1017,29 @@ module Lich
       # return value, and a timed-out child is torn down before the raise, so
       # no unsupervised child ever survives this call.
       #
+      # @param args [Array] arguments forwarded to {.start_child}
       # @param timeout [Numeric, nil] maximum seconds to wait, or nil to wait indefinitely
+      # @param execution_guard [Proc, nil] optional lifetime policy, as in {.start}
+      # @param allow_script_starts [Boolean, nil] override the policy's script-start
+      #   permission; nil retains the options hash value (default true)
       # @return [Script] the terminated child
       # @raise [ArgumentError] when the timeout is negative, before the child starts
       # @raise [StartError] when the child cannot be started
       # @raise [TimeoutError] when the child outlives the timeout, after its teardown
-      def Script.run_child(*args, timeout: nil)
+      def Script.run_child(*args, timeout: nil, execution_guard: nil, allow_script_starts: nil)
         raise ArgumentError, 'timeout must be non-negative' if timeout && timeout.negative?
 
+        overrides = {}
+        overrides[:execution_guard] = execution_guard unless execution_guard.nil?
+        overrides[:allow_script_starts] = allow_script_starts unless allow_script_starts.nil?
+        unless overrides.empty?
+          args = args.dup
+          if args.last.is_a?(Hash)
+            args[-1] = args.last.merge(overrides)
+          else
+            args << overrides
+          end
+        end
         child = Script.start_child(*args)
         unless child
           name = args.first.is_a?(Hash) ? args.first[:name] : args.first
@@ -1048,11 +1093,29 @@ module Lich
         script
       end
 
+      # Native helpers use this sleep checkpoint so opted-in scripts can unwind
+      # a long roundtime/retry wait. Ordinary scripts keep the original sleep
+      # behavior; resolving the caller here does not add a pause checkpoint.
+      #
+      # @param seconds [Numeric] requested duration
+      # @return [Numeric] sleep result
+      # @raise [ArgumentError] if the duration is invalid for the active sleep mode
+      # @raise [ScriptExecutionGuard::Interrupted] if the caller's policy cancels
+      def Script.execution_sleep(seconds)
+        script = __resolve_current
+        script ? script.execution_sleep(seconds) : Kernel.sleep(seconds)
+      end
+
       # Starts a script, blocking first if the calling script is paused.
       #
       # @param args [Array] arguments forwarded to the underlying script
       #   start machinery (script name, params, flags -- see callers for the
       #   accepted shapes)
+      # @option args [Proc, nil] :execution_guard optional observation-only
+      #   policy installed before execution, retained through teardown callbacks,
+      #   and removed after all owned workers stop. No child inheritance.
+      # @option args [Boolean] :allow_script_starts (true) false forbids new
+      #   native script launches from this guarded script, including during cleanup
       # @return [Script, nil] the started script, or nil on failure
       # @note Blocks the calling thread while it is itself paused (unless
       #   exempt via +ignore_pause+) before starting anything; a no-op wait
@@ -2403,7 +2466,11 @@ module Lich
       private :__record_exit_error
 
       def __record_successful_exit
+        guard = execution_guard_mutex.synchronize { @startup_execution_guard }
+        guard&.checkpoint!
         lifecycle_mutex.synchronize { @completed_successfully = true unless @exit_error }
+      rescue ScriptExecutionGuard::Interrupted => error
+        __record_exit_error(error)
       end
       private :__record_successful_exit
 
@@ -2597,6 +2664,7 @@ module Lich
       private :__while_running
 
       def __discard_startup
+        __close_startup_execution_guard
         parent = nil
         Script.__send__(:__registry_synchronize) do
           lifecycle_mutex.synchronize do
@@ -2721,6 +2789,7 @@ module Lich
           CHILD_RELATIONSHIP_MUTEX.synchronize { parent = @parent_script }
           parent&.unregister_child(self)
           CHILD_RELATIONSHIP_MUTEX.synchronize { @parent_script = nil }
+          __close_startup_execution_guard
           lifecycle_mutex.synchronize do
             @die_with = @at_exit_procs = nil
             @cleanup_complete = true
@@ -2811,9 +2880,150 @@ module Lich
       # one implementation instead of inlining the sleep loop.
       #
       # @return [void]
+      # @raise [ScriptExecutionGuard::Interrupted] if an installed policy cancels
       def wait_while_paused!
-        sleep 0.2 while paused? and not ignore_pause
+        unless execution_guard_active?
+          sleep 0.2 while paused? and not ignore_pause
+          return
+        end
+        # Guard callbacks may read Script.current. Do not recursively enter the
+        # pause loop while collecting those observations; sends still pass the
+        # guard at the socket and recursive sends are rejected there.
+        guard = execution_guard_mutex.synchronize { @execution_guard }
+        return if guard&.checking?
+
+        check_execution_guard!
+        while paused? && !ignore_pause
+          Kernel.sleep(EXECUTION_GUARD_POLL_INTERVAL)
+          check_execution_guard!
+        end
+        check_execution_guard!
       end
+
+      # Attach a named script's launch policy before releasing its worker gate.
+      # @param guard [ScriptExecutionGuard] already validated policy
+      # @return [void]
+      # @api private
+      def __install_startup_execution_guard(guard)
+        execution_guard_mutex.synchronize do
+          raise ArgumentError, 'script already has an execution guard' if @execution_guard
+
+          @execution_guard = @startup_execution_guard = guard
+        end
+      end
+      private :__install_startup_execution_guard
+
+      # Dispose the launch policy only after cleanup and worker termination (or
+      # rejected startup), never when the body merely returns or is interrupted.
+      # @return [void]
+      # @api private
+      def __close_startup_execution_guard
+        execution_guard_mutex.synchronize do
+          guard = @startup_execution_guard
+          return unless guard
+
+          guard.close!
+          @execution_guard = nil if @execution_guard.equal?(guard)
+          @startup_execution_guard = nil
+        end
+      end
+      private :__close_startup_execution_guard
+
+      # Called at native startup admission before reserving a name or creating
+      # workers. Read the actual caller, not a requested child's parent option.
+      # @return [true] when no active policy prohibits script launches
+      # @raise [ScriptExecutionGuard::Interrupted] when explicitly prohibited
+      # @api private
+      def __check_script_start_allowed!
+        guard = execution_guard_mutex.synchronize { @execution_guard }
+        guard ? guard.check_script_start! : true
+      end
+      private :__check_script_start_allowed!
+
+      # Install an opt-in execution policy for this script for the duration of
+      # the block. The policy receives nil at cooperative checkpoints, or the
+      # immutable raw command (including the client prefix) before a socket write.
+      # Only literal true permits continuation. A denial stays cancelled even if
+      # a helper rescues the interruption; leaving the scope checks it again.
+      #
+      # The scope applies to this script's worker threads, not child scripts.
+      # Owners must join their workers before leaving it. This is cooperative
+      # control, not a sandbox for arbitrary Ruby or a way to undo sent commands.
+      #
+      # @param callback [Proc] observation-only policy, never a command sender
+      # @param allow_script_starts [Boolean] false prohibits native script launches
+      #   by this script while the scope is installed; default true is unchanged
+      # @yieldparam guard [ScriptExecutionGuard] cancellation handle
+      # @return [Object] the block's result when the guard permits completion
+      # @raise [ArgumentError] if callback is not a Proc, the block is missing,
+      #   or another guard is already installed
+      # @raise [ScriptExecutionGuard::Interrupted] if the policy cancels execution
+      def with_execution_guard(callback, allow_script_starts: true)
+        raise ArgumentError, 'execution guard requires a block' unless block_given?
+
+        guard = ScriptExecutionGuard.new(callback, :allow_script_starts => allow_script_starts)
+        execution_guard_mutex.synchronize do
+          raise ArgumentError, 'script already has an execution guard' if @execution_guard
+
+          @execution_guard = guard
+        end
+        begin
+          guard.checkpoint!
+          result = yield guard
+          guard.checkpoint!
+          result
+        ensure
+          guard.close!
+          execution_guard_mutex.synchronize { @execution_guard = nil if @execution_guard.equal?(guard) }
+        end
+      end
+
+      # Check for an installed policy without evaluating or renewing it.
+      # @return [Boolean] whether this script currently has a scoped policy
+      def execution_guard_active?
+        execution_guard_mutex.synchronize { !@execution_guard.nil? }
+      end
+
+      # Evaluate the installed policy, preserving any previous cancellation.
+      # @param command [String, nil] exact wire command, or nil for a checkpoint
+      # @return [true] if no guard is installed or its policy permits continuation
+      # @raise [ScriptExecutionGuard::Interrupted] if the installed policy denies
+      def check_execution_guard!(command: nil)
+        guard = execution_guard_mutex.synchronize { @execution_guard }
+        guard ? guard.checkpoint!(command: command) : true
+      end
+
+      # Sleep cooperatively while a policy is installed. Native helpers use this
+      # instead of one long sleep so cancellation can unwind their ensure blocks.
+      #
+      # @param seconds [Numeric] finite nonnegative duration in seconds
+      # @return [Numeric] requested duration; ordinary scripts retain Kernel.sleep
+      # @raise [ArgumentError] if the duration is invalid for the active sleep mode
+      # @raise [ScriptExecutionGuard::Interrupted] if the installed policy cancels
+      def execution_sleep(seconds)
+        return Kernel.sleep(seconds) unless execution_guard_active?
+
+        unless seconds.is_a?(Numeric) && seconds.finite? && seconds >= 0
+          raise ArgumentError, 'guarded sleep requires a finite nonnegative duration'
+        end
+        deadline = Process.clock_gettime(Process::CLOCK_MONOTONIC) + seconds
+        loop do
+          check_execution_guard!
+          remaining = deadline - Process.clock_gettime(Process::CLOCK_MONOTONIC)
+          break if remaining <= 0
+
+          Kernel.sleep([remaining, EXECUTION_GUARD_POLL_INTERVAL].min)
+        end
+        seconds
+      end
+
+      # Obtain this script's lazily initialized policy lock.
+      # @return [Mutex] the per-script execution guard lock
+      # @api private
+      def execution_guard_mutex
+        @execution_guard_mutex || EXECUTION_GUARD_MUTEX_INITIALIZER.synchronize { @execution_guard_mutex ||= Mutex.new }
+      end
+      private :execution_guard_mutex
 
       def get_next_label
         if !@jump_label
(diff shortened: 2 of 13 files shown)


therealatari:
Addressed the current review feedback in `6da6b914`:

- recheck guarded `gets?` after the buffer read, with a regression that reproduces cancellation during `try_shift`
- remove back-to-back policy evaluations from pause and downstream polling loops
- make the native wait priority test raise while the helper still owns the priority
- require `ostruct` explicitly in the isolated game-write spec
- document the remaining new public guard lifecycle method

Verification: the complete execution-guard-focused set passes with 255 examples and 0 failures; current GitHub RSpec, RuboCop, and syntax checks are green.

mrhoribu (changes requested):
# PR Review: feat(all): add opt-in bounded script execution guards (elanthia-online/lich-5#1575)

**Reviewed:** 11 files, +1868/-54 at `6da6b914` on 2026-09-10. Based on: full clone (fresh checkout of `main`, PR head fetched via `pull/1575/head`), all changed files read in full at head, plus surrounding non-changed call sites for blast-radius tracing.
**Verdict:** Request changes
**Summary:** A carefully designed, cooperative (non-preemptive) cancellation/execution-guard system layered onto `Script`, threaded through pause waits, downstream/upstream/unique reads, sleeps, and the game-socket write seam, with an opt-in script-start restriction. The design itself is sound and thoroughly tested (255 new examples, full 7241-example suite green, rubocop clean). One change made unconditionally to `Game._puts` — not gated behind guard activation — adds a new pause-blocking dependency to a low-level write path that existing core code (`Inventory.refresh`) relies on being pause-independent while holding a module-wide serializing mutex, contradicting the PR's own "no change for unguarded scripts" compatibility claim.

## Major

### 1. `Game._puts` now blocks on pause for every caller, not just guarded ones — stalls `Inventory.refresh` for *all* scripts if one gets paused mid-exchange — [lib/games.rb:638-666](lib/games.rb#L638)

```ruby
def _puts(str)
  script = Script.current
  # A guard callback must not reenter the socket. Let Script's guard
  # recursion check reject it before Ruby attempts to relock the mutex.
  if @mutex.owned? && script.respond_to?(:check_execution_guard!) &&
     script.respond_to?(:execution_guard_active?) && script.execution_guard_active?
    script.check_execution_guard!(command: str.to_s.dup.freeze)
  end
  @mutex.synchronize do
    ...
```

Before this PR, `_puts` never touched `Script.current` at all — it was the deliberately "raw" write path, used specifically where callers wanted to avoid the script/pause-awareness baggage that `Game.puts` already carried. Now `_puts` calls `Script.current`, which (via `Script.__resolve_current` → `script&.wait_while_paused!`, [lib/common/script.rb:1090-1093](lib/common/script.rb#L1090)) unconditionally loop-sleeps until the calling script is unpaused — *regardless of whether that script has an execution guard installed*. The guard-specific logic that follows (`execution_guard_active?`, `check_execution_guard!`) is correctly gated, but the `Script.current` resolution itself is not.

This directly contradicts the PR's own compatibility promise: "scripts without an active guard retain existing sends, waits, reads, pause behavior, and child launches" (PR description, Compatibility section).

Concrete blast radius: `Lich::Common::Inventory.refresh` ([lib/common/inventory.rb:880-907](lib/common/inventory.rb#L880)) acquires a **module-level** `@refresh_mutex` that, per its own comment, "deliberately serializes `.refresh` callers so only one continuation exchange is ever live at a time" ([lib/common/inventory.rb:802-810](lib/common/inventory.rb#L802)) — i.e. every script in the process shares this one mutex. It then calls `Game._puts("_inventory manager #{initial}")` *while still holding `@refresh_mutex`*, and its own doc comment notes "MUST run on a script thread." If any other thread (the CLI, another script) pauses that script while it's inside this critical section — an ordinary, expected operation (`Script#pause`, [lib/common/script.rb:2857](lib/common/script.rb#L2857), is public and callable on any script object from anywhere) — `_puts` now blocks indefinitely in `wait_while_paused!` while still holding `@refresh_mutex`. Every other script that subsequently calls `Inventory.refresh` blocks on that mutex too, until someone manually unpauses the first script. Previously, this scenario was impossible: `_puts` fired the write immediately and released the mutex regardless of pause state.

The same unconditional-`Script.current` exposure applies to every other direct `_puts` caller that exists specifically to bypass `Game.puts`'s script-awareness (e.g. `lib/common/frontend.rb:586-596`, `lib/common/xmlparser.rb:871/940-966`, `lib/common/orderly_shutdown.rb:184`, `lib/common/gameloader.rb:75`) — most of those run off a script's own thread only incidentally or not at all, so `Inventory.refresh` is the clearest concrete trigger, but it is not the only exposed call site.

**Confirmed** by direct code trace (mechanism is deterministic — no guard needed to reproduce, just an external pause during the critical section) plus the diff showing `_puts` had no `Script.current` reference on `main`. Fix: resolve the calling script without invoking its pause-wait (e.g. an internal, pause-free resolver analogous to `Script.__resolve_current`, exposed for this seam) and only apply pause-awareness when a guard is actually active on that script — matching the "unguarded scripts see no behavior change" contract the PR itself states.

## What this PR gets right

- The core `ScriptExecutionGuard#checkpoint!` state machine ([lib/common/script_execution_guard.rb](lib/common/script_execution_guard.rb)) correctly handles: sticky first-cancellation-wins across threads, recursive-checkpoint short-circuiting for nil (non-send) checkpoints so a policy callback can safely read `Script.current`, and reentrant-command rejection for a misbehaving policy that tries to send from inside its own callback.
- The reentrant-send handling in `_puts` is genuinely clever: it pre-checks `@mutex.owned?` *before* attempting `@mutex.synchronize` again, converting what would otherwise be a Ruby `ThreadError` (non-reentrant `Mutex#synchronize` deadlock) into the guard's proper `:reentrant_command` `Interrupted` — verified by the `'rejects callback reentry without a socket write or recursive mutex failure'` spec and traced through the code.
- Script-start admission is checked uniformly and correctly at every native launch path (named via `__begin_start`, library via `__begin_library_start`, anonymous via `SubScript.start`, exec via `ExecScript.start`) against the *actual calling worker's* guard (`Script.__resolve_current`), not a caller-supplied `parent:`/`force:` option — matching the doc's claim that these can't be used to bypass the restriction.
- `wait_until`/`wait_while` in `lib/global_defs.rb` gained an `ensure` around the previously bare trailing `Thread.current.priority = priosave` — a real, necessary fix, since those loops can now raise `ScriptExecutionGuard::Interrupted` mid-wait and previously would have leaked the thread at priority 0 on any exception exit.
- Test coverage is extensive and genuinely adversarial (concurrent threads sharing a guard, mutex-ownership assertions inside guard callbacks, policy-callback-raises-exception paths, nested/nil-checkpoint recursion) — 255 new examples, all passing, and they exercise real concurrency rather than only the happy path.
- `docs/script-execution-guard.md` is precise about what the guard does *not* do (not a sandbox, doesn't intercept direct socket access or arbitrary Ruby, can't undo an in-flight write) — this kind of explicit scope-limiting documentation is exactly right for a cooperative-cancellation primitive that could otherwise be mistaken for a real sandbox.

## Coverage notes

- Full clone of `elanthia-online/lich-5`; PR head fetched directly (`pull/1575/head`) into an isolated scratch checkout separate from the user's own in-progress working tree, so nothing here touched local uncommitted work.
- Read `lib/common/script.rb`, `lib/common/script_execution_guard.rb`, `lib/games.rb`, and `lib/global_defs.rb` in full diff plus surrounding context; traced call sites for `Script.current`, `_puts`, `__begin_start`/`__begin_library_start`, `__launch_child`, and `Script.run_child`'s option-merging logic against `Script.start`'s shared parsing in `@@elevated_script_start`.
- Ran `bundle exec rubocop` on all four core changed `.rb` files: clean, no offenses.
- Ran the six new/changed guard-specific spec files directly (255 examples) and the full repo suite (`bundle exec rspec`, 7241 examples): all green, no failures, on Ruby 4.0.5.
- CodeRabbit CLI (`coderabbit review --agent --base main --committed`) was attempted for a second-opinion pass but returned only a rate-limit error for this account/org (3 included reviews exhausted); no CodeRabbit findings were available to incorporate.
- Did not independently verify the two doc-only commits' prose against every implementation detail line-by-line beyond spot-checking claims that mattered for grounding findings (e.g. the "checked once" claim, the launch-admission bypass claims) — those spot-checks held up.
- Did not review `lib/gemstone/infomon/parser.rb`'s and `lib/gemstone/group.rb`'s `_puts` call sites for pause-blocking exposure in as much depth as `Inventory.refresh`, since dr-scripts/scripts (the primary consumers per this repo's `CLAUDE.md`) are DR-only and don't exercise the Gemstone code path; the `Inventory.refresh` case alone is sufficient to ground the Major finding.

therealatari:
Thanks for catching the paused-owner path. Addressed in e4a55e1c.

Game._puts now resolves its owner through Script.current_without_pause, using the existing resolver without the pause wait. Guard cancellation/reentry checks remain in place, including the recheck inside the socket mutex. Ordinary Game.puts retains its existing pause behavior.

The regression test uses the real Script and Inventory paths: a paused refresh owner must release the shared refresh mutex rather than block subsequent callers. It failed before the fix and passes afterward. The focused guard/game/inventory suite passes (386 examples); the changed Ruby files pass RuboCop.

I also tested a temporary integration checkout containing all four current PR heads against main b9df96c4: 7,439 examples, zero failures; 1,160 files inspected by RuboCop, no offenses. This was a local integration check, not a merge to main or a live deployment.

mrhoribu (approved):
# PR Review: feat(all): add opt-in bounded script execution guards (elanthia-online/lich-5#1575)

**Reviewed:** 11 files, +1868/-54 at `6da6b914` on 2026-09-10 (original review); re-reviewed follow-up commit `e4a55e1c` ("fix(core): preserve pause-free raw game writes") on 2026-09-12. Based on: full clone (fresh checkout of `main`, PR head fetched via `pull/1575/head`), all changed files read in full at head, plus surrounding non-changed call sites for blast-radius tracing.
**Verdict:** Approve
**Summary:** A carefully designed, cooperative (non-preemptive) cancellation/execution-guard system layered onto `Script`, threaded through pause waits, downstream/upstream/unique reads, sleeps, and the game-socket write seam, with an opt-in script-start restriction. The design is sound and thoroughly tested. The original review's one Major finding — `Game._puts` unconditionally resolving `Script.current` and thereby blocking on pause even for unguarded callers holding `Inventory.refresh`'s shared mutex — is fixed in `e4a55e1c` and independently verified below. No new findings.

## Update (2026-09-12): follow-up commit fixes the Major finding

Commit `e4a55e1c` ("fix(core): preserve pause-free raw game writes") directly addresses [finding #1 below](#1-fixed-in-e4a55e1c-gamesputs-now-blocks-on-pause-for-every-caller-not-just-guarded-ones--stalls-inventoryrefresh-for-all-scripts-if-one-gets-paused-mid-exchange--libgamesrb638-666).

The fix adds `Script.current_without_pause` ([lib/common/script.rb:1082-1088](lib/common/script.rb#L1082)), a public wrapper around the existing private `Script.__resolve_current` that resolves the calling script **without** invoking `wait_while_paused!`, and switches `_puts` ([lib/games.rb:639](lib/games.rb#L639)) to use it instead of the pause-aware `Script.current`. `Game.puts` itself still resolves via `Script.current` (unchanged — it keeps its existing pause checkpoint), so the fix is narrowly scoped to the raw write seam that previously had no pause dependency at all. Guard checks (`execution_guard_active?` / `check_execution_guard!`) are untouched and still run correctly against the pause-free-resolved script — a paused, guarded script's raw write is still checked and can still be rejected by its policy, it just no longer blocks on pause state to get there.

I re-traced the mechanism directly: `_puts`'s early reentrancy check and the in-mutex check both now operate on the pause-free-resolved `script`, so the "checked once, immediately before transport" and "reentrant send caught before the mutex deadlocks" properties from the original design are preserved unchanged — only the resolution's pause side effect was removed.

The accompanying new spec, `spec/lib/game_raw_write_pause_spec.rb`, is a genuine regression test for exactly this scenario, not a decorative one: `'lets a paused inventory refresh finish and release the shared refresh mutex'` builds a real `Script` instance with `@paused = true` (never unpaused during the test), runs `Inventory.refresh` on a worker thread attributed to that script, and asserts the worker completes (`first.join(1)).to equal(first)`) and a second, concurrent `Inventory.refresh` call from an unrelated thread also completes — which is only possible if the first call's `Game._puts` did not block on the owning script's pause state while holding `@refresh_mutex`. Under the pre-fix code this test would have timed out. I ran it directly along with the rest of the guard-focused specs (260 examples) and the full suite (7241 examples): all green, and `rubocop` is clean on every changed file.

I did not find anything else to flag in this follow-up commit — it's a minimal, correctly targeted fix with matching documentation (`docs/script-execution-guard.md:18-21` now states the corrected contract) and test coverage. One very minor observation, not worth blocking on: `Script.current_without_pause` is a genuinely public class method (no `private_class_method`, unlike `__resolve_current`) despite being tagged `@api private` in its comment — a `.lic` script could call it directly. That's a documentation-enforced boundary rather than a language-enforced one, consistent with how this PR already exposes other guard-related internals (`with_execution_guard`, `check_execution_guard!`) as real public API, so it's a Nit at most, not a concern.

---

## Original review (2026-09-11)

### Major finding (now fixed — see update above)

### 1. [FIXED in e4a55e1c] `Game._puts` now blocks on pause for every caller, not just guarded ones — stalls `Inventory.refresh` for *all* scripts if one gets paused mid-exchange — [lib/games.rb:638-666](lib/games.rb#L638)

```ruby
def _puts(str)
  script = Script.current
  # A guard callback must not reenter the socket. Let Script's guard
  # recursion check reject it before Ruby attempts to relock the mutex.
  if @mutex.owned? && script.respond_to?(:check_execution_guard!) &&
     script.respond_to?(:execution_guard_active?) && script.execution_guard_active?
    script.check_execution_guard!(command: str.to_s.dup.freeze)
  end
  @mutex.synchronize do
    ...
```

Before this PR, `_puts` never touched `Script.current` at all — it was the deliberately "raw" write path, used specifically where callers wanted to avoid the script/pause-awareness baggage that `Game.puts` already carried. Now `_puts` calls `Script.current`, which (via `Script.__resolve_current` → `script&.wait_while_paused!`, [lib/common/script.rb:1090-1093](lib/common/script.rb#L1090)) unconditionally loop-sleeps until the calling script is unpaused — *regardless of whether that script has an execution guard installed*. The guard-specific logic that follows (`execution_guard_active?`, `check_execution_guard!`) is correctly gated, but the `Script.current` resolution itself is not.

This directly contradicts the PR's own compatibility promise: "scripts without an active guard retain existing sends, waits, reads, pause behavior, and child launches" (PR description, Compatibility section).

Concrete blast radius: `Lich::Common::Inventory.refresh` ([lib/common/inventory.rb:880-907](lib/common/inventory.rb#L880)) acquires a **module-level** `@refresh_mutex` that, per its own comment, "deliberately serializes `.refresh` callers so only one continuation exchange is ever live at a time" ([lib/common/inventory.rb:802-810](lib/common/inventory.rb#L802)) — i.e. every script in the process shares this one mutex. It then calls `Game._puts("_inventory manager #{initial}")` *while still holding `@refresh_mutex`*, and its own doc comment notes "MUST run on a script thread." If any other thread (the CLI, another script) pauses that script while it's inside this critical section — an ordinary, expected operation (`Script#pause`, [lib/common/script.rb:2857](lib/common/script.rb#L2857), is public and callable on any script object from anywhere) — `_puts` now blocks indefinitely in `wait_while_paused!` while still holding `@refresh_mutex`. Every other script that subsequently calls `Inventory.refresh` blocks on that mutex too, until someone manually unpauses the first script. Previously, this scenario was impossible: `_puts` fired the write immediately and released the mutex regardless of pause state.

The same unconditional-`Script.current` exposure applies to every other direct `_puts` caller that exists specifically to bypass `Game.puts`'s script-awareness (e.g. `lib/common/frontend.rb:586-596`, `lib/common/xmlparser.rb:871/940-966`, `lib/common/orderly_shutdown.rb:184`, `lib/common/gameloader.rb:75`) — most of those run off a script's own thread only incidentally or not at all, so `Inventory.refresh` is the clearest concrete trigger, but it is not the only exposed call site.

**Confirmed** by direct code trace (mechanism is deterministic — no guard needed to reproduce, just an external pause during the critical section) plus the diff showing `_puts` had no `Script.current` reference on `main`. Fix: resolve the calling script without invoking its pause-wait (e.g. an internal, pause-free resolver analogous to `Script.__resolve_current`, exposed for this seam) and only apply pause-awareness when a guard is actually active on that script — matching the "unguarded scripts see no behavior change" contract the PR itself states.

## What this PR gets right

- The core `ScriptExecutionGuard#checkpoint!` state machine ([lib/common/script_execution_guard.rb](lib/common/script_execution_guard.rb)) correctly handles: sticky first-cancellation-wins across threads, recursive-checkpoint short-circuiting for nil (non-send) checkpoints so a policy callback can safely read `Script.current`, and reentrant-command rejection for a misbehaving policy that tries to send from inside its own callback.
- The reentrant-send handling in `_puts` is genuinely clever: it pre-checks `@mutex.owned?` *before* attempting `@mutex.synchronize` again, converting what would otherwise be a Ruby `ThreadError` (non-reentrant `Mutex#synchronize` deadlock) into the guard's proper `:reentrant_command` `Interrupted` — verified by the `'rejects callback reentry without a socket write or recursive mutex failure'` spec and traced through the code.
- Script-start admission is checked uniformly and correctly at every native launch path (named via `__begin_start`, library via `__begin_library_start`, anonymous via `SubScript.start`, exec via `ExecScript.start`) against the *actual calling worker's* guard (`Script.__resolve_current`), not a caller-supplied `parent:`/`force:` option — matching the doc's claim that these can't be used to bypass the restriction.
- `wait_until`/`wait_while` in `lib/global_defs.rb` gained an `ensure` around the previously bare trailing `Thread.current.priority = priosave` — a real, necessary fix, since those loops can now raise `ScriptExecutionGuard::Interrupted` mid-wait and previously would have leaked the thread at priority 0 on any exception exit.
- Test coverage is extensive and genuinely adversarial (concurrent threads sharing a guard, mutex-ownership assertions inside guard callbacks, policy-callback-raises-exception paths, nested/nil-checkpoint recursion) — 255 new examples, all passing, and they exercise real concurrency rather than only the happy path.
- `docs/script-execution-guard.md` is precise about what the guard does *not* do (not a sandbox, doesn't intercept direct socket access or arbitrary Ruby, can't undo an in-flight write) — this kind of explicit scope-limiting documentation is exactly right for a cooperative-cancellation primitive that could otherwise be mistaken for a real sandbox.

## Coverage notes

- Full clone of `elanthia-online/lich-5`; PR head fetched directly (`pull/1575/head`, later re-fetched for `e4a55e1c`) into an isolated scratch checkout separate from the user's own in-progress working tree, so nothing here touched local uncommitted work.
- Read `lib/common/script.rb`, `lib/common/script_execution_guard.rb`, `lib/games.rb`, and `lib/global_defs.rb` in full diff plus surrounding context; traced call sites for `Script.current`, `_puts`, `__begin_start`/`__begin_library_start`, `__launch_child`, and `Script.run_child`'s option-merging logic against `Script.start`'s shared parsing in `@@elevated_script_start`.
- For the follow-up commit: re-read the full diff of `e4a55e1c`, traced `Script.current_without_pause` and its use in `_puts` against the reentrancy/mutex-ownership logic that was already verified in the original pass, and read the new `spec/lib/game_raw_write_pause_spec.rb` in full to confirm it exercises the real scenario rather than a stubbed-out approximation.
- Ran `bundle exec rubocop` on all changed `.rb` files at both the original head and the follow-up head: clean, no offenses, both times.
- Ran the guard-specific spec files directly (260 examples at the follow-up head, including the new `game_raw_write_pause_spec.rb`) and the full repo suite (`bundle exec rspec`) at both heads: all green, no failures, on Ruby 4.0.5.
- CodeRabbit CLI (`coderabbit review --agent --base main --committed`) was attempted for a second-opinion pass but returned only a rate-limit error for this account/org (3 included reviews exhausted); no CodeRabbit findings were available to incorporate at either pass.
- Did not independently verify the two doc-only commits' prose against every implementation detail line-by-line beyond spot-checking claims that mattered for grounding findings (e.g. the "checked once" claim, the launch-admission bypass claims) — those spot-checks held up.
- Did not review `lib/gemstone/infomon/parser.rb`'s and `lib/gemstone/group.rb`'s `_puts` call sites for pause-blocking exposure in as much depth as `Inventory.refresh`, since dr-scripts/scripts (the primary consumers per this repo's `CLAUDE.md`) are DR-only and don't exercise the Gemstone code path; the `Inventory.refresh` case alone was sufficient to ground the original finding, and the fix is general (it changes the resolver used by `_puts` itself, not a special case for `Inventory.refresh`), so it covers those call sites too.

mrhoribu:
Follow-up review (based on the local worktree at `e4a55e1c`, full clone + local CodeRabbit pass). Full writeup below; wrote it up in more detail as a review doc but posting the substance here.

---
# PR Review: feat(all): add opt-in bounded script execution guards (elanthia-online/lich-5#1575)

**Reviewed:** 13 files, +2045/-54 at `e4a55e1c` on 2026-09-15. Based on: full local clone (git worktree at PR head, with history) + `gh` metadata/comments/checks.
**Summary:** This adds an opt-in, cooperative cancellation/timeout primitive (`Script#with_execution_guard` and guarded named-script startup) that scripts can install around game writes, downstream/upstream reads, pause waits, and explicit sleeps, with zero behavior change for the overwhelming majority of scripts that never install one. The design and test coverage are unusually thorough for a change this size, and two rounds of prior CodeRabbit review already caught and fixed the two sharpest correctness bugs (a missing post-read recheck in `gets?`, and `Game._puts` resolving through the pause-waiting `Script.current` instead of a pause-free resolver, which could starve `Inventory.refresh`'s shared mutex). The one thing that survived my own pass is a narrow, concrete coverage gap: `dothis` — the direct, structurally-identical sibling of `dothistimeout`, which this same PR converts — was not wired into the guard, so cancellation latency inside it is unbounded relative to the rest of the feature's stated ≤0.05s polling promise.

## Minor

### 1. `dothis` is not guard-aware even though its sibling `dothistimeout` is — `lib/global_defs.rb:1974`

```ruby
def dothis(action, success_line)
  loop {
    Script.current.clear
    put action
    loop {
      line = get
      if line =~ success_line
        return line
      elsif line =~ /^(\.\.\.w|W)ait ([0-9]+) sec(onds)?\.$/
        if $2.to_i > 1
          sleep($2.to_i - "0.5".to_f)
        else
          sleep 0.3
        end
```

`dothistimeout` (`lib/global_defs.rb:2037`, in this same diff) is the near-exact twin of `dothis` — same branches, same messages, same structure — and every `sleep` in it was converted to `Script.execution_sleep` by this PR. `dothis` was left with five raw `sleep` calls (lines 1984, 1990, 1998, 2007, 2019): a "Wait N seconds"/"type ahead" one-shot sleep of up to `$2.to_i - 0.5` seconds, and three `100.times` polling loops.

The polling loops are not actually a problem — each iteration calls `get?`, which routes through `Script#gets?` and is fully guard-checked before and after the buffer read, so cancellation during those loops is still observed at the loop's own 0.1s cadence. The one-shot sleeps are the real gap: `sleep($2.to_i - "0.5".to_f)`, `sleep 0.3`, and `sleep 1` block with no checkpoint at all. Because the outer `loop` re-enters via `Script.current.clear` next, and `Script.current` itself triggers `wait_while_paused!` (a checkpoint when a guard is active), cancellation set by another thread during one of these sleeps is eventually observed — but only after the full sleep duration elapses, which for the "Wait N seconds" branch can be many seconds, not the ~0.05–0.3s the rest of this feature targets.

This reads as an accidental omission rather than a deliberate scope boundary: `spec/lib/common/native_wait_execution_guard_spec.rb:16` lists the native helpers this PR intentionally wired up for guard-awareness — `%w[fput dothistimeout get? clear pause waitrt? waitcastrt? wait_until wait_while]` — and `dothistimeout` is on it while `dothis` is absent, with no other apparent reason (e.g. `matchtimeout` and `selectput`, which are also not on the list, don't have a converted sibling sitting right next to them the way `dothis`/`dothistimeout` do). Suggest converting the same five call sites in `dothis` to `Script.execution_sleep`, matching what was already done in `dothistimeout`, and adding `dothis` to the native-helper guard spec list.

## Open questions

- `lib/common/script_execution_guard.rb:71` rescues bare `Exception` (with an explicit `rubocop:disable Lint/RescueException` and a comment that callback failures "must latch, including script-defined exceptions") around the caller-supplied policy callback. That also catches `SystemExit`, `NoMemoryError`, and `ScriptError` from anything the callback transitively calls, translating them into `:callback_error` instead of letting them propagate. If a policy callback (or a helper it calls) ever invokes `Kernel#exit`, that exit is silently absorbed into a cancellation instead of actually exiting. **Settles it:** is silently absorbing `SystemExit`/similar from within a policy callback intended, or was the broad `rescue Exception` aimed only at ordinary `StandardError`-shaped script bugs?

## What this PR gets right

- **True opt-in, verified.** Every guarded helper (`execution_sleep`, `gets`/`gets?`, `upstream_gets`/`upstream_gets?`, `unique_gets`/`unique_gets?`, `wait_while_paused!`, `Game#_puts`/`#puts`) branches on `execution_guard_active?` and falls straight through to the original `Kernel.sleep`/buffer-shift/write path when no guard is installed. I traced each of these and confirmed the unguarded path is byte-for-byte the pre-PR behavior.
- **Consistent post-read recheck.** After the fix in `6da6b914`, every guarded read (`gets`, `gets?`, `upstream_gets`, `upstream_gets?`, `unique_gets`, `unique_gets?`) rechecks the guard immediately after consuming from the buffer, not just before — so a cancellation that latches mid-read is still observed before the line is handed back to the script.
- **Reentrancy and mutex-ordering are handled correctly.** `ScriptExecutionGuard#checkpoint!` tracks per-thread reentry via `@callbacks`, raises `:reentrant_command` for a recursive send, and returns immediately (without re-invoking the callback) for a recursive checkpoint — so a policy reading `Script.current` from inside its own callback can't deadlock or infinite-loop. `Game#_puts` separately detects the case where the guard callback itself tries to send (which would otherwise hit Ruby's own `ThreadError` for a non-reentrant `Mutex`) via `@mutex.owned?` and raises the friendlier `Interrupted` first.
- **The `Inventory.refresh` fix is real and well-covered.** `Game#_puts` now resolves the calling script via `Script.current_without_pause` instead of `Script.current`, so a raw write made while already holding a shared lock (like `Inventory.refresh`'s mutex) can't get stuck waiting on its own pause state. `spec/lib/game_raw_write_pause_spec.rb` exercises this with a real `Script`/`Inventory` regression, not just a mock.
- **Thorough, well-targeted specs.** 4,600+ lines of new/extended spec coverage across guard lifecycle, cancellation propagation through cleanup/`ensure`/external-kill, thread-priority restoration, and the game-write seam — all of which I ran locally and passed.
- **Documentation is exceptional** (`docs/script-execution-guard.md`): explicit about what is and isn't covered, what "cooperative" means, and the exact cancellation semantics — this is the kind of doc that will actually prevent misuse.

## Coverage notes

- Full local clone: fetched `pull/1575/head` into the existing `elanthia-online/lich-5` checkout and reviewed from a `git worktree` at `e4a55e1c` (HEAD), so history (`git log`/`git show` on the six PR commits) and the full repo were available, not just the diff.
- Ran the focused guard spec files locally (`native_wait_execution_guard_spec.rb`, `script_auxiliary_guard_spec.rb`, `script_execution_guard_spec.rb`, `script_execution_scope_spec.rb`, `game_execution_guard_spec.rb`, `game_raw_write_pause_spec.rb`): 73 examples, 0 failures. Ran `script_lifecycle_spec.rb` (the large pre-existing file this PR extends): 187 examples, 0 failures. Did not re-run the full ~7,000-example suite locally — CI (`Run tests on Ruby`, both jobs) is green, and the author's own PR comments report a 7,246–7,439-example integration run clean.
- Ran `rubocop` on all 12 changed Ruby files (excluding the markdown doc): no offenses.
- Ran a local CodeRabbit pass (`coderabbit review --agent --base main --committed`) against the full worktree: 0 findings. Two prior CodeRabbit review comments existed on this PR (`lib/common/script.rb` `gets?` missing post-read recheck, and a missing `require 'ostruct'` in `spec/lib/game_execution_guard_spec.rb`) — I verified both directly against the code at HEAD and confirmed both are already fixed (in commit `6da6b914`, before the two comments' anchoring commit `e4a55e1c`), so I did not re-report them.
- Did not independently trace every call site across `dr-scripts`/`scripts` that might one day adopt `execution_guard:` — this PR is additive and opt-in, so existing script behavior in those repos is unaffected regardless; blast-radius review here focused on lich-5's own core (`Script`, `Game`, `global_defs.rb`) since that's what every script implicitly runs on.

---

**Note:** this branch currently shows as `CONFLICTING` / merge-dirty against `main` on GitHub — will need a rebase before it can merge.

Nisugi:
Thorough pass over this one. The design is sound and unusually well-documented, and the "no guard installed → byte-identical behavior" property mostly holds. But there are two real bugs in `Game.puts`, plus some design concerns worth resolving. This touches `_puts` — the chokepoint every script's every command flows through — so I've held it to a high bar.

## Blocking

**1. `puts` sends the command *before* displaying it, inverting the ordering — and the echo shows the unguarded string**

In the guarded branch, `_puts wire_command` is called **before** `$_CLIENTBUFFER_.push` and before the `respond` echo. The comment says "Rejected guarded commands must not appear as sent," which is the right intent — but the consequence is that an *accepted* command is written to the socket before the user sees it, reversing the ordering for guarded scripts. And the `respond` echo still prints `str` (unprefixed, un-snapshotted) while the client buffer records `wire_command`. Two different records of the same send.

The unguarded path retains the original order (display, then send), so guarded and unguarded scripts now have observably different echo/send interleaving — which cuts against the compatibility claim, and will show up as reordered output in any script mixing the two.

Suggested shape: do the guard check *without* sending (a `check_execution_guard!(command: wire_command)`), then display, then `_puts` once on the normal path. Preserves ordering and still guarantees a rejected command never displays.

**2. `wire_command` is referenced outside the `if guarded` block that defines it**

```ruby
if guarded
  str = str.to_s.dup.freeze
  wire_command = "#{$cmd_prefix}#{str}".freeze
  _puts wire_command
end

$_CLIENTBUFFER_.push "...#{guarded ? wire_command : "#{$cmd_prefix}#{str}"}\r\n"
```

This works in Ruby — the parser hoists `wire_command` to `nil` because the assignment is lexically earlier — and it's ternary-guarded, so it's not a live crash. But it relies on parse-order local hoisting, and any refactor that moves the assignment into a method or block breaks it silently. Hoist `wire_command = nil` explicitly, or compute it unconditionally.

**3. Double guard check on every guarded write**

`_puts` checks the guard twice on the contended path: once before `@mutex.synchronize` (when `@mutex.owned?`) and once inside. Since `checkpoint!` invokes the user's callback, a guarded write can invoke the policy callback **twice for one command** when the pre-check fires. The callback is documented observation-only, but any policy that counts commands, rate-limits, or debits a budget will double-charge. The comment says checks are "charged once here" — that isn't what the code does.

The pre-check is only reachable when `@mutex.owned?` (re-entrant send from inside a callback), which is exactly the case you want to reject — so the pre-check may be the right place, but the inner check should then be skipped for that path.

## Non-blocking

**4. The `wait_until`/`wait_while` `ensure` change is a genuine fix — worth splitting out.** Adding `ensure Thread.current.priority = priosave` fixes a real pre-existing leak: any exception (including a plain `die`, not just `Interrupted`) previously left the thread at priority 0 forever. This is good and unrelated to execution guards — it'd be worth its own PR so it can merge and backport independently of the guard machinery.

**5. `rescue Exception` swallows all callback context.** In `checkpoint!`, a callback that raises is latched as `:callback_error` with message, backtrace, and cause deliberately discarded. The rationale (don't leak command payloads) is stated, but this makes a buggy policy callback — a typo, a `NoMethodError` — undebuggable. Consider `Lich.log`ging the exception class and backtrace (not the command) while keeping it out of the raised `Interrupted`.

**6. `execution_sleep` turns every sleep in `move`, `fput`, and `dothistimeout` into a 0.05s poll loop for guarded scripts.** A `pause 3600` becomes 72,000 iterations of `clock_gettime` + `check_execution_guard!`, each taking the guard mutex and invoking the user callback. For a long-running guarded script that's a meaningful CPU and callback-volume change. Consider checking the latched `cancelled?` flag (cheap) on each tick and only running the full callback on an interval.

**7. `EXECUTION_GUARD_MUTEX_INITIALIZER` is a single global mutex** used to lazily create every script's per-script mutex. Correct, but it serializes first-touch across all scripts. Simpler: initialize `@execution_guard_mutex` in the constructor and drop the global.

**8. `Script.current_without_pause` is a public class method** that just aliases the private `__resolve_current`. It's documented `@api private` but is publicly callable, and it bypasses the pause checkpoint — a meaningful capability to put on the public surface. Consider `private_class_method` plus `__send__` at the two `games.rb` call sites, matching how the rest of the file handles internals.

**9. `respond_to?` triple-checks repeated in three places.** `script.respond_to?(:check_execution_guard!) && script.respond_to?(:execution_guard_active?) && script.execution_guard_active?` appears twice in `_puts` and again in `puts`, and has already drifted slightly in ordering between them. Extract a single `Script.guarded?(script)` helper so it can't drift further.

## What's good

- The `nil` = checkpoint / frozen `String` = impending write protocol is clean, and "only literal `true` permits continuation" is the right fail-closed default.
- First-cancellation-wins latching with normalized `/\A[a-z][a-z0-9_]{0,63}\z/` reason symbols prevents both reason churn and symbol-table growth from attacker-controlled strings.
- Re-entrancy detection via `@callbacks[Thread.current]` correctly distinguishes "callback is reading `Script.current`" (allowed) from "callback is trying to send" (rejected).
- Guard disposal on both `__discard_startup` and `__finalize_stop`, with `equal?` identity checks so cleanup removes only the owning guard, is careful work.
- 255 focused examples covering the genuinely hard cases (worker ownership, nested guards, cancellation mid-buffer-read).

Items 1–3 are in `Game.puts`/`_puts` and I'd want those fixed before merge, since that's the hot path for every command every script sends. The rest are judgment calls.

Nisugi (changes requested):
# Review of elanthia-online/lich-5 PR #1575

Reviewed 2026-09-15 at `e4a55e1cd906e843b10811a625dc9ab0d2364cf8`, against `073d23a9d75b94029e9d08a7dfbd443335adc65c`.

## Recommendation: request changes

Two reproducible cancellation bugs remain. Both are absent from the existing review comments. The existing focused suite passes, but the additional regression cases in `review_regressions_spec.rb` fail at this head.

### 1. [P2] Preserve cancellation first observed during cleanup in the final result

Location: `lib/common/script.rb:2930-2938` (`__close_startup_execution_guard`), called by `__finalize_stop` at line 2803.

The body marks the script successful through `__record_successful_exit` before the before-dying callbacks run. A callback can subsequently attempt a denied command, rescue `ScriptExecutionGuard::Interrupted`, and return normally. Finalization then closes and removes the cancelled guard without recording its reason in `exit_error`.

Reproduction: start a child with a policy that accepts nil checkpoints and rejects command strings. Its body registers an exit callback which calls `script.check_execution_guard!(command: 'cleanup')` and rescues `Interrupted`. After `child.join`, `child.completed_successfully?` is true and `child.exit_error` is nil. No timing race is needed.

This violates the lifetime contract that rescued cancellation remains a failed run through cleanup. A supervising caller following the documented join/result pattern will incorrectly accept that run as successful. The same omission can lose a cancellation rescued by an owned worker during final shutdown.

Before discarding the lifetime guard, transfer any latched cancellation to lifecycle error state, preserving its original reason. Do this before normal disposal adds `:closed`, and avoid invoking user callbacks while holding the per-script guard mutex. Add coverage for cancellation first observed and rescued during cleanup, not just in the body.

### 2. [P2] Recheck cancellation before returning a downstream timeout

Location: `lib/common/script.rb:3083-3088` (`gets`).

The post-buffer checkpoint only executes when `wait_shift` returns a non-nil line. If cancellation occurs while an empty timed read waits and the deadline expires, `gets` returns nil without checking the guard again. Callers can enter ordinary timeout handling despite an already-latched cancellation. A later checkpoint will still reject game writes, but the read itself has incorrectly returned control.

The new test deterministically cancels at the buffer-return boundary and returns nil. `script.gets(0)` returns rather than raising. Existing post-read tests cover data returns, but omit the empty timeout return.

Check the guard on the terminal timeout path before returning nil. Preserve the single checkpoint per continuing empty polling iteration, so the fix does not restore the duplicate checks removed in the earlier review.

## Assessment of existing review comments

Read the PR description, both inline review comments, the formal reviews including CodeRabbit's nested nitpicks, and all issue discussion comments through the current head.

| Existing concern | Assessment at reviewed head |
| --- | --- |
| Missing `gets?` post-read check | Fixed. Both before/after checks are present, with a regression test. The `gets` timeout issue above is a separate remaining branch. |
| Missing explicit `ostruct` require | Fixed in `game_execution_guard_spec.rb`. |
| Consecutive duplicate pause/downstream polling checks | The cited loops were changed as requested. The downstream timeout exit now needs its own check. |
| Priority restoration test did not raise inside the helper | Fixed. The predicate explicitly calls `owner.check_execution_guard!`, so the test exercises the helper's ensure. |
| Raw `_puts` pause wait can hold the shared inventory refresh mutex | Fixed by `current_without_pause`; the real Script/Inventory regression passes. Ordinary `puts` still uses pause-aware lookup. |
| Guarded send occurs before command echo | Confirmed behavioral change: guarded `_puts` precedes client history and `respond`, allowing server output to overtake the echo. This deserves a deliberate ordering decision/test, but does not contradict the narrower promise to preserve *unguarded* behavior. If changing it, keep one authorization under the socket lock immediately before dispatch. The suggested separate preflight followed by normal `_puts` would evaluate command policy twice and introduce a gap between authorization and dispatch. |
| Echo uses an unsnapshotted string | Incorrect at this head: `str` is copied and frozen before `_puts`. Omitting the wire prefix from the visible echo is existing behavior. |
| `wire_command` is assigned only inside the conditional | No runtime bug: the local is valid Ruby and the ternary only reads it when the guarded branch executed. Explicit initialization would be stylistic. |
| Contended writes invoke command policy twice | Incorrect: `Mutex#owned?` means owned by the current thread, not another contending thread. A recursive command is rejected by the guard before entering `synchronize`; an ordinary write invokes command policy once. Nil pause checkpoints are separate observations. |
| `dothis` retains raw sleeps | Confirmed. A wait-30 response sleeps about 29.5 seconds without checkpoints. Converting it like `dothistimeout` would improve coverage. The documentation limits the 0.05-second guarantee to guarded reads, pause waits, and execution sleeps; it explicitly excludes other blocking calls, so this is a coverage improvement rather than a demonstrated breach of that guarantee. |
| Broad callback exception rescue and lost debugging context | Intentional in code, documentation and a non-StandardError test. `SystemExit` is included. Whether to expose sanitized diagnostics is an API decision; automatically logging backtraces can undermine the stated information-suppression contract. |
| Polling callback volume | Twenty polls per second is the designed cooperative deadline mechanism. Checking only the cancellation flag would stop deadline policies from being evaluated during sleeps unless an independent timer were added. No performance regression was measured. |
| Lazy initializer mutex, public internal resolver, repeated capability checks, splitting priority fix | Maintenance/style choices; no concrete correctness problem established. The initializer lock is only used on first access. The API explicitly provides cooperative control, not a sandbox. |

## Validation and limits

- Ruby 4.0.3 on Windows; isolated PR worktree.
- All seven focused guard/lifecycle spec files: **260 examples, 0 failures**.
- RuboCop on all 12 changed Ruby files: **0 offenses**.
- Two added regression examples: **2 failures**, reproducing the findings above.
- Full PR suite: **7,246 examples, 59 failures, 1 pending**. Baseline comparison recorded separately below.
- GitHub currently reports successful Ruby test and RuboCop checks, but the branch is `CONFLICTING` / `DIRTY` against current main.
- No live game-server exercise or performance benchmark was performed. This review does not validate helpers outside the documented checkpoint coverage.

Save the following regression file at the repository root as `review_regressions_spec.rb`, then run:

```powershell
bundle exec rspec review_regressions_spec.rb --example 'review regression' --format documentation
```

Sources: [PR and formal reviews](https://github.com/elanthia-online/lich-5/pull/1575), [follow-up review](https://github.com/elanthia-online/lich-5/pull/1575#issuecomment-5682939451), [latest concerns](https://github.com/elanthia-online/lich-5/pull/1575#issuecomment-5683034324).

Baseline verification: the unchanged PR base at 073d23a9 produced **7,147 examples, 59 failures, 1 pending** with the same seed (59226). Comparing failed example identifiers shows exactly the same 59 failures on base and PR head. These full-suite failures predate this PR in this environment; the added guard tests introduce no additional full-suite failures.


<details>
<summary>Regression test source (two failing examples)</summary>

```ruby
require_relative 'spec/lib/common/script_lifecycle_spec'
require_relative 'spec/lib/common/script_execution_scope_spec'
root = RSpec.world.example_groups.find { |g| g.description == 'Lich::Common::Script lifecycle extensions' }
root.children.find { |g| g.description == 'named-script lifetime execution guards' }.class_exec do
  it 'review regression: preserves cancellation rescued during cleanup' do
    guarded_fixture(<<~RUBY) do
      script.at_exit_procs << proc {
        begin
          script.check_execution_guard!(command: 'cleanup')
        rescue Lich::Common::ScriptExecutionGuard::Interrupted
        end
      }
    RUBY
      child = script_class.start_child('guarded', execution_guard: ->(command) { command.nil? })
      expect(child.join(2)).to equal(child)
      expect(child).not_to be_completed_successfully
      expect(child.exit_error.reason).to eq(:command_rejected)
    end
  end
end
RSpec.world.example_groups.find { |g| g.description == 'Lich::Common::Script execution guard scope' }.class_exec do
  it 'review regression: observes cancellation during an empty timed read' do
    expect do
      script.with_execution_guard(->(_) { true }) do |guard|
        allow(buffer).to receive(:wait_shift) do |_timeout|
          guard.cancel!(:manual_hold)
          nil
        end
        expect { script.gets(0) }.to raise_error(interrupted)
      end
    end.to raise_error(interrupted)
  end
end

```

</details>

Nisugi:
Corrections to my review above —  Retracting:

**Item 3 (double guard check) — withdrawn entirely.** I misread `@mutex.owned?` as "is the mutex held" rather than "is it held by the current thread." The pre-check therefore only fires on same-thread reentry, which means the caller is already inside `checkpoint!`; that call finds `@callbacks[Thread.current]` set with a non-nil command, latches `:reentrant_command`, and raises before ever reaching `@mutex.synchronize`. The two checks are mutually exclusive by construction — either the pre-check raises, or it was never entered. Nothing double-charges. My premise was wrong.

**Item 1, the "unsnapshotted echo" half — withdrawn.** `str = str.to_s.dup.freeze` rebinds `str` itself, so the later `respond` reads the frozen snapshot. A callback cannot alter what gets echoed. I missed the rebinding.

**Item 2 (conditional `wire_command`) — downgraded out of blocking.** It's valid Ruby, local hoisting is defined behavior rather than an accident, and the ternary makes the `nil` case unreachable. I said as much and then filed it as blocking anyway, which was inconsistent. It's a style preference at most.

What I'd still flag, much more narrowly than I originally put it: on the guarded path `_puts` runs before the client-buffer push and the echo, while on the unguarded path it runs after both. So installing a guard flips send-then-display to display-then-send. Invisible within a single script; it would only show as interleaving against other scripts' output or against game text arriving in response. Quite possibly deliberate given the "rejected commands must not appear as sent" goal — but it is a guarded/unguarded behavior difference, and the Compatibility section reads as though there isn't one. A sentence in the docs would cover it.

That leaves nothing blocking from me. Items 4–9 stand as written, and 4 (the `ensure` priority-restore fix) still seems worth splitting into its own PR on its own merits. Apologies for the noise on 1–3.

therealatari:
Addressed the latest review findings and reconciled this branch with current `origin/main`.

- Guarded `Script#gets(timeout)` now rechecks cancellation at the terminal empty-read boundary before returning `nil`.
- Named-script finalization now records an already-latched, sanitized guard interruption as `exit_error` before ordinary guard disposal. It preserves the first reason and does not invoke policy callbacks during teardown.
- Added direct regressions for cancellation first observed/rescued during cleanup and cancellation latched during an empty timed read.
- Closed Horibu’s remaining `dothis` gap: its one-shot and polling waits now use cooperative `Script.execution_sleep`, with guarded and unguarded behavior covered.
- Resolved the upstream merge by preserving the new bounded `fput` and extracted `Lich::Common::Move` implementation; the latter’s waits are now guard-aware too.

Verification on the merged head:
- 339 execution-guard/lifecycle/game-write/movement/global-helper examples, 0 failures
- 7,858-example full Lich run: only 8 sandbox-denied socket/interface examples failed with `EPERM`; rerunning that exact 13-example networking subset outside the sandbox passed 13/13
- RuboCop clean on all touched guard, movement, global helper, and regression files
- `git diff --check` clean

Latest head: `6e6af636`.

therealatari:
Follow-up on the hosted RSpec failure: the exact seed exposed an order-dependent test leak in spec/lib/common/hook_priority_spec.rb. Its suppressor and survivor registrations remained in the real downstream-hook registry after the example, so the later sentinel integration specs had their chunks suppressed before SynchronizedSocket.write was reached. Commit 98b91b7d clears all five hook registry maps in teardown. Verification: the focused hook/games replay passes (193 examples), RuboCop is clean, and the exact hosted seed now passes all 7,860 examples with 0 failures.

therealatari:
@Nisugi both requested cancellation fixes are now implemented at the current head: cleanup-latched interruption is transferred into lifecycle failure state before guard disposal, and an empty timed downstream read rechecks the guard before returning nil. Both supplied regression shapes are covered. The dothis wait-30 gap is also closed. The exact previously failing hosted seed now passes all 7,860 examples, and every replacement check is green. When convenient, this is ready for your re-review.

Nisugi:
Took another full pass now that the branch has moved (currently at `98b91b7d`).

## Scoping note first

`gh pr diff` is misleading on this branch right now. The merge of `origin/main` at `0c7b87f4` plus main having advanced since means the two-dot diff attributes files to this PR that no PR commit touches — `lib/common/move.rb` (458 lines), `lib/common/events.rb`, `hook_registry.rb`, and the whole `lib/gemstone/combat/**` block. I checked each of the nine PR commits individually; none of them touch `move.rb`. The authored surface is still `script.rb`, `script_execution_guard.rb`, `games.rb`, `global_defs.rb`, and specs.

Anyone landing on this cold will read it as ~2,900 lines across 24 files rather than the much smaller real change. A rebase, or a line in the description saying which files are actually in scope, would help the next reviewer a lot.

## Changes since my last pass

**`6e6af636` — `dothis` waits guard-aware.** Completes the coverage; `dothis` was the one retry loop still on bare `sleep` after `dothistimeout` had been converted. Mechanical and correct.

**`d57879db` — late cancellation preserved.** Two genuine fixes here:

- `gets` now calls `check_execution_guard!` on the timeout-expiry path before returning `nil`. Previously a guard cancelled during that final poll window returned a clean `nil` and the script carried on as though it had merely timed out — the cancellation was silently dropped. That's the kind of gap only a deliberate test finds.
- `__close_startup_execution_guard` now reads `guard.interruption` *before* `close!`, and records it via `__record_exit_error`. This matters because `close!` latches `:closed` through `cancel!`'s `@reason ||=`, so without capturing first, a script cancelled for a real reason would have exited looking like an ordinary close. Ordering is right.

**New `interruption` method** is observation-only, never invokes the callback, builds a fresh `Interrupted` under the mutex — consistent with the class's existing "never leak payload" discipline.

**`98b91b7d`** is spec state isolation only.

## On the three items I retracted

All three stay retracted, and re-reading `checkpoint!` confirms the reentrancy argument in detail: `@callbacks.key?(thread)` with a non-nil command sets `:reentrant_command` and calls `interrupt!` *inside* the `@mutex.synchronize` block, so it raises before `entered` is ever assigned. The `_puts` pre-check and the inner check genuinely cannot both charge the callback.

## Still open from the first review

All non-blocking, unchanged in the code:

- **5** — `rescue Exception` discards callback class and backtrace, making a buggy policy callback hard to debug. A `Lich.log` of class + backtrace (not the command) would keep the payload-hygiene property while restoring debuggability.
- **6** — `execution_sleep` invokes the full policy callback every 50ms; a cheap `cancelled?` check per tick with the callback on a longer interval would cut the callback volume on long waits.
- **7** — global `EXECUTION_GUARD_MUTEX_INITIALIZER` serializes first-touch across all scripts; constructor init would drop it.
- **8** — `Script.current_without_pause` is publicly callable and bypasses the pause checkpoint.
- **9** — the triplicated `respond_to?` chain has already drifted in ordering between its three sites.
- **4** — the `wait_until`/`wait_while` `ensure` priority restore is still bundled here. It's a real pre-existing leak fix (any `die` left the thread at priority 0 permanently) that would merge on its own merits today; I'd still suggest splitting it.

## One new small thing

`cancel!` silently collapses any reason not matching `/\A[a-z][a-z0-9_]{0,63}\z/` to `:cancelled`. Right fail-safe, but a caller passing `:Timeout` or `:"rate-limited"` gets `:cancelled` with no indication their reason was dropped, which will be annoying to debug. A log line on normalization mismatch would cost nothing.

## Verdict

The two fixes since my last pass address real and subtle cancellation-loss bugs, and both are correctly ordered. Nothing blocking from me. The one thing I'd genuinely like before merge is the rebase or scope note — the inflated diff is a real obstacle to further review.

mrhoribu (approved):
# PR Review (follow-up): feat(all): add opt-in bounded script execution guards (elanthia-online/lich-5#1575)

**Reviewed:** 16 files, three-dot diff (merge-base `2e3438b1`) +2123/-60, at `98b91b7d` on 2026-09-16. Based on: full local clone (git worktree at head, with history).
**Verdict:** Approve
**Summary:** Since my last pass the author rebased onto current `main` (resolving the merge-conflict I flagged), fixed the `dothis` guard-coverage gap I reported, and — prompted by a second reviewer's (Nisugi's) findings — landed two more genuine fixes: a guard cancellation that could go unobserved at the exact moment `gets(timeout)`'s wait expires, and a cancellation that could be silently dropped during named-script cleanup instead of surfacing as `exit_error`. I independently verified both new fixes, re-ran the full guard/game/move/hook spec surface and RuboCop locally (all green), and ran a fresh local CodeRabbit pass. Nothing blocking remains from me.

## Previously reported, now fixed

- **`dothis` guard coverage** (`lib/global_defs.rb:1890`, my prior Minor #1): commit `6e6af636` converts all five `sleep` call sites to `Script.execution_sleep`, matching its sibling `dothistimeout`. Verified directly against the file at head — confirmed fixed.
- **Merge conflicts / rebase** (my prior note): `gh pr view --json mergeable,mergeStateStatus` now reports `MERGEABLE` / `CLEAN`. The branch was rebased onto current `main` in `0c7b87f4`.

## New fixes since my last pass, verified

### `gets(timeout)` could silently drop a cancellation at timeout expiry — fixed in `d57879db`

`lib/common/script.rb:3091` now runs `check_execution_guard!` on the timeout-expiry return path, not just the line-returned path:

```ruby
if deadline && Process.clock_gettime(Process::CLOCK_MONOTONIC) >= deadline
  check_execution_guard!
  return nil
end
```

Before this, a cancellation that latched during the final poll window before timeout would be swallowed — the caller got back an ordinary `nil` (indistinguishable from "nothing arrived in time") instead of `Interrupted`. I traced the full method; every return path (line delivered, or timeout) now rechecks the guard, consistent with the pattern already used in `gets?`/`upstream_gets`/`upstream_gets?`.

### Cancellation during named-script cleanup could go unrecorded — fixed in `d57879db`

`ScriptExecutionGuard#interruption` (`lib/common/script_execution_guard.rb:119`) is a new observation-only accessor that returns a fresh `Interrupted` for the currently-latched reason without touching the callback. `__close_startup_execution_guard` (`lib/common/script.rb:2932`) now reads it **before** calling `guard.close!`, and records it via `__record_exit_error` afterward:

```ruby
interruption = execution_guard_mutex.synchronize do
  guard = @startup_execution_guard
  return unless guard
  existing_interruption = guard.interruption
  guard.close!
  ...
  existing_interruption
end
__record_exit_error(interruption) if interruption
```

The before/after ordering here is load-bearing, not cosmetic: `close!` itself calls `cancel!(:closed)`, and `cancel!`'s `@reason ||= normalized` means it never overwrites an existing reason — but if `interruption` were read *after* `close!`, an ordinary, never-cancelled script would also show a `:closed` interruption (since `close!` always latches something), and every guarded script's exit would spuriously get an `exit_error`. Reading first correctly distinguishes "was already cancelled for a real reason" (non-nil) from "closing normally" (nil). I confirmed `__record_exit_error` is `@exit_error ||= error` (first-write-wins), so this can't clobber an exit_error already recorded by another path either. This closes a real gap: a script whose own cleanup code rescues an `Interrupted` internally (swallowing it) would previously finish looking like an ordinary, successful completion.

### Also verified as sound

- `0c7b87f4` (merge) and the accompanying `Script.execution_sleep` conversions re-applied to `lib/common/move.rb`: `move()` was extracted to `Lich::Common::Move` by an unrelated PR that landed on `main` in the interim; the rebase correctly re-applied the same six sleep-site conversions to the new location. Confirmed via `git diff origin/main...HEAD -- lib/common/move.rb` — line-for-line the same substitutions as before, just relocated.
- `98b91b7d`: adds an `after` block to `spec/lib/common/hook_priority_spec.rb` clearing five hook-registry class variables. This matches the order-dependent leak the author diagnosed from a hosted CI failure (a suppressor/survivor registration surviving into later specs); the fix is minimal and targeted.

## On Nisugi's still-open items

I read the full review thread. Items 5–8 (bare `rescue Exception` losing callback class/backtrace, per-tick callback cost in `execution_sleep`, the global `EXECUTION_GUARD_MUTEX_INITIALIZER`, and `current_without_pause` being publicly callable) are all real observations and all correctly labeled non-blocking — I'd weigh them the same way. I flagged the `rescue Exception` one myself last round as an open question for the same reason.

One correction: **item 9** claims the triplicated `script.respond_to?(:check_execution_guard!) && script.respond_to?(:execution_guard_active?) && script.execution_guard_active?` check "has already drifted slightly in ordering" between its three call sites in `lib/games.rb`. I diffed all three occurrences (`_puts`'s pre-check at line 656, `_puts`'s in-mutex check at line 663, and `puts` at line 691) character-for-character at the current head and they're identical in ordering. Whether this was true of an earlier revision or a misreading, it doesn't hold now — worth not acting on as written. The underlying suggestion (extract a single `Script.guarded?(script)` helper) is still reasonable on its own DRY merits, just not because of drift that's currently observable.

## Coverage notes

- Fetched `pull/1575/head` fresh and rebuilt the worktree at the new head (`98b91b7d`); confirmed the real PR-authored diff via three-dot (`git diff origin/main...HEAD`), since `gh pr diff`'s two-dot output currently over-attributes unrelated files (`lib/common/events.rb`, `lib/common/hook_registry.rb`, `lib/gemstone/combat/**`) that no PR commit touches — this is `main` having moved since the merge, exactly as Nisugi flagged, and it cost me a few minutes to confirm was noise rather than scope creep worth reporting.
- Ran the full guard/game/move/hook-priority spec surface locally: 322 examples (`native_wait_execution_guard_spec.rb`, `script_auxiliary_guard_spec.rb`, `script_execution_guard_spec.rb`, `script_execution_scope_spec.rb`, `game_execution_guard_spec.rb`, `game_raw_write_pause_spec.rb`, `move_spec.rb`, `hook_priority_spec.rb`) plus `script_lifecycle_spec.rb` (188 examples) — 0 failures across both runs.
- RuboCop on all 15 changed Ruby files: no offenses.
- CI (`gh pr checks`): all green at the current head.
- Ran a fresh local CodeRabbit pass against the new head: one new finding, in `spec/lib/game_execution_guard_spec.rb:50` — its `after` block doesn't restore `Script.current` after the `before` block sets it. I checked this against `spec/spec_helper.rb:67-70`, which has a suite-wide `config.before { Script.current = nil ... }` that resets it before *every* example regardless of file — so this doesn't actually leak between examples, and I'm not carrying it forward as a finding.

therealatari:
@Nisugi The two cancellation defects from your requested-changes review are fixed at `98b91b7d`, and your follow-up re-review found nothing blocking. I have also added an explicit authored-file scope note to the PR body so the merged-upstream files are not mistaken for this PR's changes.

Could you replace the stale `CHANGES_REQUESTED` review with an approval when convenient? The hosted full suite is **7,860 examples, 0 failures**, and the remaining checks are green.

Nisugi (approved):

PR #1576, opened 2026-09-09
feat(gs): add trustworthy combat observation provenance

therealatari:
## Summary

- attach immutable connection/game/character/room/sequence/time provenance to native combat observations
- preserve initiating provenance through async processing and held-cast completion without rereading mutable current-room state
- add per-processing-invocation batch metadata so consumers can distinguish complete native emission sets
- snapshot transient attack-observer demand once per invocation
- preserve provenance across complete standalone room-object/player roster refreshes while continuing to reject partial, nested, mixed-room, and navigation-bearing fragments

## Why

Bounded supervised controllers and Recorder consumers need to know whether a native combat event came from the exact live session, room epoch, and ingress window they own. Current mutable room state is insufficient for delayed or asynchronous events, and a root callback alone cannot prove that a multi-event parser batch is complete.

This adds observation metadata only. It does not enable the tracker, send commands, persist configuration, infer missing provenance, or claim causal ownership. Legacy/direct/replay input without a validated source remains `nil`.

## Contract

- `Combat::Tracker.observation_context` returns the current immutable binding only when available
- native ingress captures a monotonic timestamp immediately after socket read on the exact parser thread
- `event[:source]` carries validated scalar provenance through the async worker
- `event[:observation_batch]` identifies index and size within one complete `Processor.process` invocation
- consumer subscription affects attack demand beginning at invocation boundaries, never halfway through a batch

## Verification

- focused combat/provenance suite previously passed 245 examples with 0 failures
- roster-refresh production-hook regression passes for complete standalone refreshes and preserves fail-closed behavior for transition-bearing fragments
- all changed Ruby files pass RuboCop and hosted CI is green

See `docs/combat-observation-provenance.md` for the full consumer obligations and limitations.


therealatari:
Self-review hardening is now pushed in `36bee9da`: socket provenance uses the timestamp captured immediately after the read (separate from queue timing), transient attack demand is snapshotted once per processing invocation, and the batch counter is reset by the spec harness. Focused suite: 245 examples, 0 failures; 9 changed Ruby files pass RuboCop; GitHub CI is green.

mrhoribu (approved):
# PR Review: feat(combat): add trustworthy observation provenance (elanthia-online/lich-5#1576)

**Reviewed:** 10 files, +623/-31 at `0933d075` on 2026-09-11. Based on: full local clone (worktree checkout of the PR branch, merge base `073d23a9`), full test suite run, RuboCop run.
**Summary:** Adds an optional, validated `source:` provenance hash (connection/game/character/room-epoch/sequence/receipt-time) that flows from the socket reader through the async worker into parsed combat events, plus a per-`Processor.process` batch id/index/size and a one-time snapshot of transient `:attack` observer demand. The design is unusually well-grounded against its own hardest edge cases (mid-string room transitions, split tags, held-cast supersession, reconnects) and is backed by 249 new/updated examples, all passing. No blocking or major issues found.

## Nits

### 1. Redundant ternaries around calls that already default `source: nil`

```ruby
# lib/gemstone/combat/tracker.rb:245, :247, :454
source ? @async_processor.process_async(chunk, source: source) : @async_processor.process_async(chunk)
source ? Processor.process(chunk, source: source) : Processor.process(chunk)
source ? process(chunk, source: source) : process(chunk) unless chunk.empty?
# lib/gemstone/combat/async_processor.rb:97
source ? Processor.process(chunk, source: source) : Processor.process(chunk)
```

`process_async`, `Processor.process`, and `Tracker#process` all declare `source: nil` as the default, so `X(chunk, source: source)` and `X(chunk)` are behaviorally identical when `source` is `nil` — the ternary in each of these four spots is dead branching. Collapsing to the unconditional form (`@async_processor.process_async(chunk, source: source)`, etc.) removes four branches with no behavior change. Non-blocking.

## Open questions

- `lib/gemstone/combat/tracker.rb:428` (`next server_string if room_roster_refresh?(server_string)`) drops a matched roster-refresh fragment from `@buffer` entirely, not just from source consideration — a scope slightly wider than the PR's own "this adds observation metadata only" framing. Functionally this looks safe (component lines are already ignored inside `parse_events`, `Tracker.buffer` isn't consumed by anything in dr-scripts/scripts, and the behavior is covered by 7 dedicated specs in `tracker_observation_spec.rb`), so this reads as a deliberate, well-tested design choice rather than scope creep that needs undoing. **Settles it:** was excluding roster refreshes from the buffer (vs. just from source-binding) an intentional, reviewed part of "preserve provenance across roster refreshes," or a side effect worth calling out separately in the PR description? One line either way closes this.

## What this PR gets right

- **Fail-closed by construction.** `observation_source`, `observation_binding`/`observation_context`, and `ingest_source` all validate positively (type + range + non-empty checks) and return `nil` on any doubt, rather than trying to sanitize or coerce. Combined with `NilClass#method_missing` semantics in this codebase (see `lib/common/class_exts/nilclass.rb`), a `nil` source degrades safely through the rest of the pipeline.
- **The `SOURCE_TRANSITION` heuristic is grounded in the actual dispatch order**, not assumed: `Game#process_server_string` (`lib/games.rb:1013,1109`) really does run `process_xml_data` (which can advance `XMLData.room_count` mid-string) before `process_downstream_hooks`, which is exactly the scenario the regex-based guard exists to catch. I traced several of the adversarial regex cases (nested component, split tag, trailing-content-after-close) by hand against `ROOM_ROSTER_REFRESH`/`SOURCE_TRANSITION` and they behave as documented and as tested.
- **Batch/attack-demand snapshotting is actually snapshotted.** `attack_events_requested?` is computed once per `Processor.process` call and threaded explicitly through every `event_savable?`/`persist_event`/`parse_events` call site (verified by grep — no call site relies on the live-lookup default), so a subscriber joining or leaving mid-batch can't produce a half-gated batch, matching the documented contract.
- **Concurrency story is explicit and consistent with prior design.** `@observation_batch_id`, `@held_cast`, `@deferred_emits` are module-level state whose safety depends on serialized `Processor.process` calls; that invariant is upheld by the single-threaded `AsyncProcessor` worker (pre-existing from #1523) and is honestly documented ("Direct callers of `Processor.process` must serialize complete invocations") rather than silently assumed.
- **Backward compatibility verified, not assumed.** Every new parameter (`source:`, `ingress_monotonic_at:`, `include_attack_events:`) has a default that reproduces the exact legacy call shape; I grepped dr-scripts, scripts, and ProfanityFE for any use of the touched internals (`enqueue_server_string`, `Combat::Tracker` process/observation APIs, `Processor`/`AsyncProcessor`) and found only `.enabled?`/`.stats` usage in `scripts`, neither of which changed shape.
- **Test coverage matches the hard cases, not just the happy path**: mixed-room chunks, buffer-overflow truncation, wrong-thread hook execution, missing ingress time, partial/nested/transition-bearing roster fragments, held-cast supersession with a changed or missing later binding, and mutation-after-enqueue for the async copy are all exercised.

## Coverage notes

- Full local clone via `git worktree` off `origin/main` merge-base `073d23a9`; diff, blame, and full file contents were all available (not diff-only).
- Ran the entire spec suite (`bundle exec rspec`): 7169 examples, 0 failures, including the 5 changed/added spec files (249 examples) in isolation.
- Ran RuboCop on all 9 changed Ruby files (project's own `.rubocop.yml`, via `bundle exec rubocop`): no offenses.
- CodeRabbit CLI is installed and authenticated for this account, but the run hit the account's usage-based review rate limit ("Rate limit exceeded... 3 included reviews currently available... not linked to an assigned seat") and returned no findings before erroring out. No CodeRabbit pass was incorporated into this review.
- Cross-repo impact checked by grep only (dr-scripts, scripts, ProfanityFE checkouts available locally); no dynamic-dispatch or metaprogramming-based call sites were found that a static grep could miss, but that's a known limitation of the method.
- Did not test against a live game connection — all verification is via the existing spec suite (which stubs `XMLData`, `Game`, `DB_Store`, etc.) and manual regex/logic tracing, not a running Lich session.

therealatari:
Thanks — addressed the cleanup and merged current main in a4fdcfba.

The four forwarding calls now always pass source: source; the callees already accept source: nil and existing positional calls remain supported.

Dropping a strictly recognized standalone roster refresh from the combat buffer is intentional. XML parsing has already applied the roster update, and the hook still returns the original fragment for downstream delivery. The combat processor already skips component lines because their creature links can create phantom targets. Excluding them from the tracker buffer also prevents roster-only links from satisfying its creature gate or displacing combat text at the buffer limit. Mixed, nested, split, and transition-bearing fragments still invalidate provenance. This is now documented and covered by regression tests.

Current main holds pre-flares across chunks, so the merge preserves their original observation receipt and rejects missing or changed bindings rather than assigning a later chunk's provenance. Upstream parser and registry behavior is retained.

Standalone full suite: 7,316 examples, zero failures. RuboCop: 1,153 files clean. A separate temporary integration checkout with all four updated PR heads also passes 7,439 examples and full RuboCop.

mrhoribu (approved):
Re-reviewed after the merge with `main` (a4fdcfba). That merge pulled in a large batch of independent combat-module work that landed on `main` in the interim (held pre-flares, Holy Weapon proc-release children, new inbound/outcome-only event shapes), which meant the provenance feature had to be re-threaded through a substantially restructured `processor.rb`. I checked that re-threading directly:

- Every event-creation site in the merged `parse_events` still carries a `source:` key consistent with the documented rules — grepped and checked all of them, including the shapes introduced by the concurrent upstream work.
- The new `combined_observation_source` helper generalizes the held-cast binding check to also cover the new `@held_pre_flares` state ("an attack that claims them takes the oldest contributing receipt, or nil when any contributing binding is absent or different"), and it traces correctly against both call sites.
- `event_savable?`/`persist_event` call sites still pass `include_attack_events:` explicitly everywhere, so the "snapshot demand once per invocation" contract survived the restructuring.

Both items from the previous pass are resolved: the redundant ternaries are gone, and the roster-refresh buffer-exclusion rationale is now spelled out in the docs and covered by a dedicated test (`'keeps roster refreshes out of the creature gate and buffer limit'`).

Ran a CodeRabbit pass this time (rate-limited on the previous review). It raised one `major` finding that `COMBAT_RELEVANT_PATTERN`/`combat_relevant?` might drop the new outcome-only lines the provenance spec covers. I don't think it holds: that filter predates this PR and isn't touched by it, both cited example lines already contain `<pushBold/>` (itself one of the filter's match literals), the gate is chunk-level so a single line without a bold link still rides through on any other combat-relevant line in the same chunk, and there's already a purpose-built bypass for creature-less lines (`Definitions::Attacks.attackerless_line?`) that predates this PR too.

Full suite passes (7316 examples, 0 failures) and RuboCop is clean on all 9 changed files.

Approving.

PR #1577, opened 2026-09-09
feat(all): add static-only map route selection

therealatari:
## Summary

- add an opt-in `static_only:` keyword to class- and room-level map Dijkstra routing
- skip executable/dynamic edges without evaluating them when static routing is requested
- accept only plain String `wayto` values and finite, nonnegative real Numeric weights in that mode
- preserve existing routing and dynamic weight evaluation when the keyword is omitted or false

## Motivation

Some callers need to inspect whether an ordinary static route exists before granting movement authority. The existing pathfinder may evaluate `StringProc` weights while selecting a route. `static_only: true` provides a side-effect-free selection subset without changing map data, route preferences, or the default behavior used by existing callers.

This is route selection only. It does not execute movement, validate command text, prove a destination safe, or sandbox arbitrary map extensions. A route that requires a dynamic edge is deliberately reported as unreachable in static mode even when normal routing could traverse it.

## Compatibility

- `Map.dijkstra(source, destination)` and `room.dijkstra(destination)` retain their current behavior
- `static_only: false` is equivalent to omitting the keyword
- existing positional instance overrides remain compatible with default class dispatch
- return values remain the existing predecessor/distance hashes

## Verification

Shared GS/DR map examples cover:

- skipping executable weights before invocation
- selecting a longer static alternative without changing default preference
- reading a selected static weight once
- rejecting nonplain commands and invalid/nonfinite weights
- preserving legacy positional instance overrides

Hosted Ruby tests, RuboCop, syntax checks, and PR-title validation are green.

### DIFF
--- a/docs/static-map-routing.md
+++ b/docs/static-map-routing.md
@@ -1,0 +1,19 @@
+# Optional static map routing
+
+`Map.dijkstra(source, destination, static_only: true)` and
+`room.dijkstra(destination, static_only: true)` reuse the existing pathfinder but
+skip edges unless `wayto` is a plain String and `timeto` is a finite, nonnegative
+real Numeric. StringProc weights are skipped before evaluation. The existing
+`dijkstra_hashes` aliases accept the same keyword.
+
+Omitting the keyword, or passing `false`, preserves normal route selection and
+dynamic weight evaluation. Existing positional instance overrides remain
+compatible with default class dispatch. No preference weights or map entries
+are changed. Return values remain the existing predecessor/distance hashes.
+
+Static routing can find a longer ordinary route or leave the destination
+unreachable when dynamic edges are necessary. It does not execute movement,
+validate the meaning of a String command, prove a destination safe, or guarantee
+that mutable map data remains unchanged after planning. Callers must validate
+commands and current room transitions and enforce their own execution budget.
+This restriction is not a sandbox for arbitrary Ruby objects or map extensions.
--- a/docs/static-map-routing.md
+++ b/docs/static-map-routing.md
@@ -7,9 +7,14 @@ real Numeric. StringProc weights are skipped before evaluation. The existing
 `dijkstra_hashes` aliases accept the same keyword.
 
 Omitting the keyword, or passing `false`, preserves normal route selection and
-dynamic weight evaluation. Existing positional instance overrides remain
-compatible with default class dispatch. No preference weights or map entries
-are changed. Return values remain the existing predecessor/distance hashes.
+dynamic weight evaluation. In both cases, class dispatch passes only the
+positional destination, so existing positional-only `Room#dijkstra` overrides
+remain compatible. Explicitly passing `static_only: true` forwards that keyword
+to the instance method. Custom overrides must accept and honor the keyword to
+support static routing; an override accepting only the positional destination
+raises `ArgumentError` when called with this opt-in. No preference weights or
+map entries are changed. Return values remain the existing predecessor/distance
+hashes.
 
 Static routing can find a longer ordinary route or leave the destination
 unreachable when dynamic edges are necessary. It does not execute movement,
--- a/lib/common/map/map_base.rb
+++ b/lib/common/map/map_base.rb
@@ -523,12 +523,13 @@ module Lich
         # Class-level dijkstra dispatcher
         # @param source [Integer, String, Object] room, room id or lookup string
         # @param destination [Integer, Array, nil] Target room(s) or nil for full graph
+        # @param static_only [Boolean] skip executable/dynamic edges without evaluating them
         # @return [Array<Hash>, nil] see Room#dijkstra
-        def dijkstra(source, destination = nil)
+        def dijkstra(source, destination = nil, static_only: false)
           if source.is_a?(self)
-            source.dijkstra(destination)
+            static_only ? source.dijkstra(destination, static_only: true) : source.dijkstra(destination)
           elsif (room = self[source])
-            room.dijkstra(destination)
+            static_only ? room.dijkstra(destination, static_only: true) : room.dijkstra(destination)
           else
             echo 'Map.dijkstra: error: invalid source room'
             nil
@@ -769,9 +770,12 @@ module Lich
         # rooms actually reached rather than the highest id plus one, and
         # iteration yields pairs rather than slots.
         # @param destination [Integer, Array, nil] Target room(s) or nil for full graph
+        # @param static_only [Boolean] use only plain String wayto edges with
+        #   finite, nonnegative real Numeric weights; never evaluate StringProc
+        #   weights in this mode. This is route selection, not command validation.
         # @return [Array<Hash>, nil] [previous, distances] keyed by room id, or
         #   nil when the search failed
-        def dijkstra(destination = nil)
+        def dijkstra(destination = nil, static_only: false)
           self.class.load unless self.class.loaded?
           source = @id
           visited = {}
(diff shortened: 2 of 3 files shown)


mrhoribu (approved):
# PR Review: feat(all): add static-only map route selection (elanthia-online/lich-5#1577)

**Reviewed:** 3 files, +94/-5 at `ed85e9f1` on 2026-09-11. Based on: full clone (PR ref fetched into a local worktree of the real `elanthia-online/lich-5` checkout, full spec suite and RuboCop run locally, git history mined for context, downstream callers in `dr-scripts` and `scripts` checked against the new signature).
**Verdict:** Approve
**Summary:** Adds an opt-in `static_only:` keyword to `Map.dijkstra`/`Room#dijkstra` (and their `dijkstra_hashes` aliases) that skips `StringProc`/dynamic edges without ever evaluating them, restricting the route to plain-`String` `wayto` values with finite nonnegative-real `Numeric` weights. The default (keyword omitted or `false`) path is untouched. It's a small, precisely scoped, thoroughly tested change with no default-path or existing-caller regressions.

## Minor

None.

## Nits

### 1. Duplicated ternary in `Map.dijkstra`'s two dispatch branches — `lib/common/map/map_base.rb:528-537`

```ruby
def dijkstra(source, destination = nil, static_only: false)
  if source.is_a?(self)
    static_only ? source.dijkstra(destination, static_only: true) : source.dijkstra(destination)
  elsif (room = self[source])
    static_only ? room.dijkstra(destination, static_only: true) : room.dijkstra(destination)
  else
```

The `static_only ? …dijkstra(destination, static_only: true) : …dijkstra(destination)` ternary is repeated verbatim for the two ways a room gets resolved. It reads fine as-is and every branch is covered by tests, so this is purely a "could be one line shorter" observation:

```ruby
room = source.is_a?(self) ? source : self[source]
return (echo 'Map.dijkstra: error: invalid source room') || nil unless room
static_only ? room.dijkstra(destination, static_only: true) : room.dijkstra(destination)
```
Non-blocking.

## Open questions

- `lib/common/map/map_base.rb:530,532` forward `static_only: true` as an actual keyword argument to `source.dijkstra`/`room.dijkstra`, but the `static_only: false`/omitted path deliberately calls with a single positional argument specifically so a legacy override such as `define_singleton_method(:dijkstra) { |destination| … }` keeps working (this is exactly what the "keeps legacy positional instance overrides compatible" test at `spec/lib/common/map/game_map_shared_examples.rb:462-467` checks — but only for `static_only: false`). Calling such an override with `static_only: true` would raise `ArgumentError: wrong number of arguments` (verified interactively: a one-arg `def dijkstra(destination)` raises when called as `dijkstra(x, static_only: true)`). I found no current caller in lich-5, dr-scripts, or scripts that overrides `dijkstra`/`dijkstra_hashes` on a room instance or subclass — `map_dr.rb`/`map_gs.rb`'s `Room < Map` classes don't override it either — so nothing is broken today. **Settles it:** is a positional-only `dijkstra` override on a `Room` instance/subclass a pattern this codebase (or a downstream script repo) is expected to support going forward? If yes, it's worth a line in `docs/static-map-routing.md` noting that `static_only: true` requires the target to accept the keyword; if no, this is already fine as-is.

## What this PR gets right

- **Correctness of the numeric gate**: `weight.is_a?(Numeric) && weight.real? && weight.finite? && weight >= 0` correctly rejects `Complex` (via `real?`, checked before any `>=` comparison would raise on a non-orderable type), `Float::NAN`/`Float::INFINITY` (via `finite?`), negative weights, and non-Numeric values like `'0.5'` or `true` — all verified against actual Ruby semantics (`Integer#finite?`/`#real?` are `true`, `Complex#real?` is `false`) and matched by the test at `spec/lib/common/map/game_map_shared_examples.rb:448-460`, including the `weight == 0` edge case (0 is truthy in Ruby, so the shared `next unless edge_weight` guard doesn't accidentally drop a zero-cost static edge).
- **`instance_of?(String)` rather than `is_a?(String)`** correctly excludes String subclasses (`lib/common/map/map_base.rb:820`), and is exercised by the `Class.new(String).new('north')` case in the same test.
- **Single-read weight capture**: `weight = room.timeto[adj_room]` is read once and reused, rather than re-read after validation — this closes a TOCTOU-shaped gap where a mutable/computed `timeto` accessor could swap in a `StringProc` between the check and the use. Explicitly covered by `spec/lib/common/map/game_map_shared_examples.rb:439-446`, which stubs `timeto[]` to return a valid weight then a raising `StringProc` on a second call and asserts only one read happens.
- **No blast radius**: confirmed empirically — no caller in dr-scripts or scripts passes a third positional argument or overrides `dijkstra`/`dijkstra_hashes`, and `dijkstra_hashes` is a plain `alias`, so it picks up the new keyword automatically without a second implementation to keep in sync.
- **Docs match implementation**: `docs/static-map-routing.md` accurately describes the gate (plain String + finite nonnegative real Numeric), the aliasing, and explicitly disclaims that this is route selection only, not command validation or a sandbox — which matches what the code actually does.

## Coverage notes

- Full PR diff, full file contents at head, and complete repo history were available via a local worktree fetched with `git fetch origin pull/1577/head` against the existing `elanthia-online/lich-5` checkout (no changes made to that checkout's working tree or branch).
- Ran `bundle exec rspec spec/lib/common/map/` at the PR head commit: 675 examples, 0 failures.
- Ran `rubocop` on both changed Ruby files: no offenses (matches the PR's green CI `Run Rubocop on Ruby` check).
- Verified Numeric predicate semantics (`finite?`, `real?` on Integer/Complex) by executing them directly rather than assuming.
- Checked every `.dijkstra(`/`dijkstra_hashes` call site in `dr-scripts` and `scripts` (both currently checked out) for third-positional-argument usage or overrides that the new keyword could break — none found.
- Attempted a local CodeRabbit pass (`coderabbit review --agent --base-commit <merge-base> --committed`) as a second opinion; it hit this account's free-tier limit ("You've used all 3 included reviews currently available", 35-minute retry window) before returning any findings. Given the diff is small (94/-5 across 3 files) and already received a thorough manual pass plus a full local test/lint run, I did not wait out the rate limit — nothing here turns on a second automated opinion.
- Did not review `.github/workflows/lint_pr.yaml`'s scope allowlist as a blocker: the PR title's `all` scope is on the allowed list, so no CI/title issue exists despite the commit message using scope `map` (a name not on that list) — commit-message scope isn't what's validated.

therealatari:
Addressed the cleanup in 1a333fc4.

The class-level dijkstra entry point now uses a single source-to-room resolution path, retaining the existing invalid-source behavior. Shared GemStone/DragonRealms tests cover both room objects and IDs.

The documentation now explicitly distinguishes legacy dispatch (keyword omitted or false) from static_only: true. Opting into static-only routing requires a custom override that accepts and honors the keyword; a positional-only override raises ArgumentError rather than silently losing the routing restriction.

Verification: 683 map examples, zero failures; changed Ruby files pass RuboCop. A temporary integration checkout with all four updated PR heads passes 7,439 examples and full RuboCop (1,160 files).

mrhoribu (approved):
Re-reviewed after the latest commit (1a333fc).

The new commit directly addresses the one open question from my earlier pass: `docs/static-map-routing.md` now documents that `static_only: true` forwards the keyword to the instance method, and that a positional-only `Room#dijkstra` override will raise `ArgumentError` unless it accepts the keyword. It's backed by two new tests confirming that behavior (`ArgumentError` for a non-keyword-aware override, correct forwarding for a keyword-aware one), plus two more covering the "invalid source room" echo path for both static and non-static dispatch. It also collapses the duplicated ternary in `Map.dijkstra` into a single `room = source.is_a?(self) ? source : self[source]` lookup.

Verified against the updated head in a local worktree: full map spec suite passes (683/683, up from 675 — the expected +8 for four new shared examples × two games), and RuboCop is clean on both changed Ruby files. The refactor is behaviorally identical on existing paths.

No blockers, no majors, nothing outstanding. Approving.

PR #1578, opened 2026-09-10
feat(gs): Lich::Gemstone::Stance, a shared stance setter

Nisugi:
## Summary

Lich can read stance (`Char.stance`, `Char.percent_stance`) but has never been able to set one, so `Spell#cast`, bigshot, ebounty, eloot and ecleanse each carry their own copy of the same send-and-confirm sequence. This adds that sequence once, as `Lich::Gemstone::Stance`, and makes `Spell#cast` use it.

## What it does

```ruby
Stance.change('defensive')              # send, wait for the game to confirm -> true / false
Stance.change(:gua)                     # prefixes and symbols
Stance.change(80)                       # "cman stance 80" when Stance Perfection is known, else the band (guarded)
Stance.change('guarded', wait: false)   # fire and forget through fput
Stance.change('offensive', force: true) # send even if already there
Stance.safest                           # 'guarded' during cast roundtime, else 'defensive'
Stance.at?('def')                       # already there?
Stance::CONFIRM                         # every line the game answers a stance change with
```

- No-op when already in the target stance, waits out roundtime, returns `true` on confirmation and `false` when the game refuses (unable to change, cast roundtime), the character is dead, or nothing confirms within the timeout.
- An unrecognised stance raises `ArgumentError`. That is a caller bug, not a game state.
- `Spell#cast` restores its after-cast stance through `Stance.change` instead of three inline `dothistimeout` calls. Behaviour is unchanged except the confirm regex now also accepts "You fall back into a ... stance", which the old inline regex missed.

## Files

- `lib/gemstone/stance.rb` (new), required from the Gemstone branch of `gameloader.rb`
- `lib/common/spell.rb`: after-cast stance block replaced by a private `restore_stance_after_cast`
- `spec/lib/gemstone/stance_spec.rb`: 26 examples

## Testing

- `bundle exec rspec spec/lib/gemstone/stance_spec.rb` green, rubocop clean on the touched files.
- In game: `Stance.change` through each name, prefix, and percent form; a stance spell cast returns to the after-stance through the new path; `Stance.change(:def)` during cast roundtime returns `false` with "Cast Roundtime in effect" rather than hanging.

## Why now

First of two small lib additions that let the e-scripts drop their private copies of common helpers (a shared `wield` is the other, in a separate PR). Independent of that PR.

### DIFF
--- a/lib/common/gameloader.rb
+++ b/lib/common/gameloader.rb
@@ -28,6 +28,7 @@ module Lich
         require File.join(LIB_DIR, 'attributes', 'enhancive.rb')
         require File.join(LIB_DIR, 'gemstone', 'society.rb')
         require File.join(LIB_DIR, 'gemstone', 'infomon', 'status.rb')
+        require File.join(LIB_DIR, 'gemstone', 'stance.rb')
         require File.join(LIB_DIR, 'gemstone', 'experience.rb')
         require File.join(LIB_DIR, 'attributes', 'spellsong.rb')
         require File.join(LIB_DIR, 'gemstone', 'infomon', 'activespell.rb')
--- a/lib/common/spell.rb
+++ b/lib/common/spell.rb
@@ -740,19 +740,7 @@ module Lich
                 cast_result = dothistimeout cast_cmd, 5, merged_results_regex
               end
               if ((@stance && force_stance != false) || force_stance == true)
-                if @@after_stance
-                  if Char.stance !~ /#{@@after_stance}/
-                    waitrt?
-                    dothistimeout "stance #{@@after_stance}", 3, /^You (?:are now in|move into) an? \w+ stance|^You are unable to change your stance\.$/
-                  end
-                elsif Char.stance !~ /^guarded$|^defensive$/
-                  waitrt?
-                  if checkcastrt > 0
-                    dothistimeout 'stance guarded', 3, /^You (?:are now in|move into) an? \w+ stance|^You are unable to change your stance\.$/
-                  else
-                    dothistimeout 'stance defensive', 3, /^You (?:are now in|move into) an? \w+ stance|^You are unable to change your stance\.$/
-                  end
-                end
+                restore_stance_after_cast
               end
               if cast_result =~ /^Cast at what\?$|^Be at peace my child, there is no need for spells of war in here\.$|^Provoking a GameMaster is not such a good idea\.$/
                 dothistimeout 'release', 5, /^You feel the magic of your spell rush away from you\.$|^You don't have a prepared spell to release!$/
@@ -773,6 +761,22 @@ module Lich
           @@cast_lock.delete(script)
         end
       end
+
+      # After a stance spell is cast, put the character back into the stance
+      # they asked for (Spell.after_stance), or the safest one the game allows.
+      # Stance.change no-ops when already there and waits out roundtime itself.
+      #
+      # @return [void]
+      def restore_stance_after_cast
+        if @@after_stance
+          Lich::Gemstone::Stance.change(@@after_stance)
+        elsif Char.stance !~ /^guarded$|^defensive$/
+          Lich::Gemstone::Stance.change(Lich::Gemstone::Stance.safest)
+        end
+      rescue ArgumentError => e
+        echo "cast: #{e.message}"
+      end
+      private :restore_stance_after_cast
 
       def force_cast(target = nil, arg_options = nil, results_of_interest = nil, force_stance: nil)
         unless arg_options.nil? || arg_options.empty?
--- a/lib/common/spell.rb
+++ b/lib/common/spell.rb
@@ -768,8 +768,13 @@ module Lich
       #
       # @return [void]
       def restore_stance_after_cast
-        if @@after_stance
-          Lich::Gemstone::Stance.change(@@after_stance)
+        # after_stance can be a blank string when a script captured Char.stance
+        # before the first pbarStance update arrived. Treat that as "no
+        # preference" rather than letting it raise, which is what the old
+        # inline regex check did.
+        after = @@after_stance.to_s.strip
+        if !after.empty?
+          Lich::Gemstone::Stance.change(after)
         elsif Char.stance !~ /^guarded$|^defensive$/
           Lich::Gemstone::Stance.change(Lich::Gemstone::Stance.safest)
         end
--- a/lib/gemstone/stance.rb
+++ b/lib/gemstone/stance.rb
@@ -1,0 +1,153 @@
+# frozen_string_literal: true
+
+# Stance: read and set the character's combat stance.
+#
+# Lich has always been able to read stance (Char.stance, Char.percent_stance)
+# but every script that needs to set one carries its own copy of the same
+# send-and-confirm sequence. Spell#cast had one, bigshot has one, ebounty,
+# eloot and ecleanse each have one. This module is that sequence, once.
+#
+#   Stance.change('defensive')            # send, wait for the game to confirm
+#   Stance.change(:off)                   # prefixes and symbols are fine
+#   Stance.change(80)                     # perfect stance when Stance Perfection is known,
+#                                         # otherwise the band it falls in (guarded)
+#   Stance.change('guarded', wait: false) # fire and forget through fput
+#   Stance.change('offensive', force: true) # send even if already offensive
+#   Stance.safest                         # 'guarded' during cast roundtime, else 'defensive'
+#
+# change returns true when the character is in the requested stance (or already
+# was), false when the game refused (cast roundtime, unable to change, dead) or
+# no confirmation arrived within the timeout. An unrecognised stance raises
+# ArgumentError; that is a caller bug, not a game state.
+module Lich
+  module Gemstone
+    module Stance
+      NAMES = %w[offensive advance forward neutral guarded defensive].freeze
+
+      # Percent-of-defense band each named stance occupies. The game reports
+      # stance_value as a 0..100 number; checkstance has always used these
+      # ranges to decide which name a value belongs to.
+      BANDS = {
+        'offensive' => (0..0),
+        'advance'   => (1..20),
+        'forward'   => (21..40),
+        'neutral'   => (41..60),
+        'guarded'   => (61..80),
+        'defensive' => (81..100),
+      }.freeze
+
+      # Every line the game can answer a stance change with. Shared with
+      # Spell#cast, which used to carry its own copy.
+      CONFIRM = Regexp.union(
+        /^You (?:are now in|move into) an? \w+ stance/,
+        /^You fall back into an? \w+ stance/,
+        /^You are unable to change your stance\./,
+        /^Cast Roundtime in effect/,
+      ).freeze
+
+      REFUSED = /^You are unable to change your stance\.|^Cast Roundtime in effect/.freeze
+
+      DEFAULT_TIMEOUT = 3
+
+      # @return [String] current stance name as the game reports it
+      def self.current
+        Char.stance
+      end
+
+      # @return [Integer] current stance as a 0..100 percentage
+      def self.value
+        Char.percent_stance
+      end
+
+      # @return [Boolean] whether Stance Perfection (cman stance N) is trained
+      def self.perfection?
+        return false unless defined?(Lich::Gemstone::CMan)
+        Lich::Gemstone::CMan.known?('stance_perfection')
+      rescue StandardError
+        false
+      end
+
+      # The stance to fall back to when the caller just wants to be safe:
+      # defensive, unless a cast roundtime is running, in which case the game
+      # will not allow defensive and guarded is the best available.
+      #
+      # @return [String]
+      def self.safest
+        checkcastrt > 0 ? 'guarded' : 'defensive'
+      end
+
+      # Resolve anything a script might hand us into [name, percent].
+      #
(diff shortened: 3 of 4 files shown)


mrhoribu (commented):
# PR Review: feat(gs): Lich::Gemstone::Stance, a shared stance setter (elanthia-online/lich-5#1578)

**Reviewed:** 4 files, +372/-13, at `b429c44e` on 2026-09-11. Based on: full local worktree (branch checked out via `git fetch origin pull/1578/head`), so call-site and blast-radius checks are grounded against the real `lich-5` tree, not just the diff.
**Verdict:** Request changes
**Summary:** Adds `Lich::Gemstone::Stance` as a shared send-and-confirm stance setter and switches `Spell#cast`'s after-cast restore to use it. The refactor of `Spell#cast` is a faithful, well-tested translation of the old inline logic — but the new `Stance.normalize` prefix matcher has a real bug: it only checks that the first three characters of the input match a stance name, not that the whole input is a prefix of it, so garbage strings like `"guardian"` or `"advertised"` silently resolve to a stance instead of raising `ArgumentError` as documented and intended. A second, lower-severity issue (caught by a CodeRabbit pass and independently verified) means the percent-target no-op check never fires for characters without Stance Perfection.

## Major

### 1. `Stance.normalize` accepts any string whose first three letters happen to match a stance, not just genuine prefixes — `lib/gemstone/stance.rb:95`

```ruby
name = NAMES.find { |n| n.start_with?(stripped[0, 3]) } if stripped.length >= 3
raise ArgumentError, "Stance: unknown stance #{target.inspect}" if name.nil?
```

The check compares `name`'s first three characters against `stripped`'s first three characters, but never checks that the rest of `stripped` agrees with `name`. Anything ≥3 chars long that happens to start with `off`/`adv`/`for`/`neu`/`gua`/`def` matches, regardless of what follows. Verified by loading the module straight from the PR's `lib/gemstone/stance.rb` and calling it directly:

```
Stance.normalize("guardian")    => ["guarded", nil]     # should raise ArgumentError
Stance.normalize("advertised")  => ["advance", nil]
Stance.normalize("foreclosure") => ["forward", nil]
Stance.normalize("definitely")  => ["defensive", nil]
Stance.normalize("neuralgic")   => ["neutral", nil]
```

This directly contradicts the PR's own stated contract ("An unrecognised stance raises `ArgumentError`; that is a caller bug, not a game state") and the docstring on `normalize` ("a stance name, a prefix ... @raise ArgumentError when target is not a stance"). None of the 26 spec examples catch it — the invalid-input tests (`'sideways'`, `''`, `nil`) all happen to fail on the length/prefix check for the wrong reason (no 3-letter overlap, or too short), so the missing-suffix-validation path is untested.

For this PR's only production call site (`Spell#restore_stance_after_cast`), the input is always `Char.stance` or a hardcoded literal, so it isn't reachable today. But the PR's own description frames `Stance` as the shared module that `bigshot`, `ebounty`, `eloot`, and `ecleanse` are meant to adopt next, several of which take stance-ish config/user input — once one of them passes an unvalidated string through `Stance.change`, this turns "reject bad input" into "silently do the wrong thing to the character's stance."

Fix: require the whole stripped input, not just its first three characters, to be a prefix of the candidate name:

```ruby
name = NAMES.find { |n| n.start_with?(stripped) } if stripped.length >= 3
```

Confirmed by tracing the code and re-running it in isolation against the actual `lib/gemstone/stance.rb` shipped in this PR.

## Minor

### 2. `at?`'s no-op check compares exact percent even when only band-level control exists — `lib/gemstone/stance.rb:103-107, 121`

```ruby
def self.at?(target)
  name, percent = normalize(target)
  return value == percent unless percent.nil?
  current.to_s.downcase == name
end
```

For a percent target, `at?` always compares the exact `value == percent`, regardless of whether `perfection?` is true. Without Stance Perfection, `change`'s own `command` branch (`stance.rb:123`) falls back to `"stance #{name}"` — i.e. only the *band* is controllable, not the exact percent — so this is the one case where `at?` should be comparing `current.to_s.downcase == name` (the same band-level check it already uses for a name target three lines down), not the exact percent. As written, a non-Perfection character sitting at, say, 70% (already in the `guarded` band) asking for `Stance.change(80)` will never short-circuit on `at?` and will resend `"stance guarded"` every time, since `70 == 80` is always false — even though `"stance guarded"` is the only command this path can ever issue and the character is already exactly where that command would put them. That contradicts the module's own doc ("No-op when already in the target stance") and the PR body's own example (`Stance.change(80) # ... else the band (guarded)`), and it isn't exercised by the test suite — the one existing "in the band but not at the percent" test (`stance_spec.rb:398-404`) stubs `perfection?` to `true`, so it never touches this branch.

Traced through `Stance.change` and confirmed the exact-percent comparison is unconditional at `stance.rb:106`; the fix mirrors what `CMan.known?`'s own `perfection?` flag is already there to select — compare by band when `perfection?` is false, by exact percent when it's true. (Surfaced by a CodeRabbit pass over this branch; verified independently against the source rather than taken on its word.)

**Open question this raises:** the worst case — the game responding to a redundant `"stance guarded"` with something other than the standard confirm text, causing `dothistimeout` to time out and `change` to return `false` for a character that was already safe — is plausible but not something I could verify without a live session; `ecleanse.lic`'s own `change_stance` (which loops resending the same "stance X" command until `percent_stance` matches exactly) suggests a repeated same-band command is at minimum harmless and does get re-confirmed, but the interaction with a strict `DEFAULT_TIMEOUT` of 3 seconds in this new module hasn't been observed in-game. **Settles it:** does resending `"stance guarded"` while already in the guarded band reliably re-trigger a `CONFIRM`-matching line, or can it ever go unconfirmed?

### 3. `CONFIRM` and `REFUSED` duplicate the same two regex literals — `lib/gemstone/stance.rb:107-114`

```ruby
CONFIRM = Regexp.union(
  /^You (?:are now in|move into) an? \w+ stance/,
  /^You fall back into an? \w+ stance/,
  /^You are unable to change your stance\./,
  /^Cast Roundtime in effect/,
).freeze

REFUSED = /^You are unable to change your stance\.|^Cast Roundtime in effect/.freeze
```

The two refusal patterns are typed out twice. `change` relies on both being kept in sync: `dothistimeout` matches against `CONFIRM` to decide whether to return at all, and the result is then checked against `REFUSED` to decide `true` vs `false`. If a future edit adjusts the wording in one constant and not the other (e.g. the game's refusal text changes and only `CONFIRM` gets updated so `dothistimeout` still returns), `change` would return `true` on a line that's actually a refusal. Low risk today since both are static and freshly written, but worth deriving `REFUSED` from a shared sub-list (or vice versa) so they can't drift.

## Nits

### 4. `Stance.change` normalizes the target twice — `lib/gemstone/stance.rb:184-187`

```ruby
def self.change(target, wait: true, force: false, timeout: DEFAULT_TIMEOUT)
  name, percent = normalize(target)
  return false if Status.dead?
  return true if !force && at?(target)
```

`normalize(target)` runs here, and `at?(target)` immediately re-runs it. Harmless (any invalid target already raised on the first call), just a redundant re-parse on every non-forced call. Could pass the already-computed `name`/`percent` into an internal helper instead of calling public `at?` a second time, but not worth doing unless this is touched again.

## Open questions

- `lib/gemstone/stance.rb:120` — `Char.stance` (`XMLData.stance_text`) defaults to `String.new` (empty string) at `lib/common/xmlparser.rb:120` before the game ever sends a `pbarStance` update. If `bigshot.lic`'s `Spell.after_stance = Char.stance` (`bigshot.lic:4928`) were ever captured at that point, `restore_stance_after_cast` would now raise-and-echo `"cast: Stance: unknown stance \"\""` on every subsequent cast, where the old regex-based check (`if @@after_stance` with an empty string, then `Char.stance !~ //`) silently no-op'd instead. In practice `bigshot.lic` only touches `after_stance` deep in an established hunting loop, long after the stance bar has been populated, so this looks unreachable in real play. **Settles it:** is there any path (reconnect mid-session, a differently-timed script) where `Spell.after_stance` could be set before the first `pbarStance` XML arrives? If not, this is a non-issue; if so, `normalize` should treat blank/empty as "unknown, no-op" rather than "raise."

## What this PR gets right

- The `Spell#cast` refactor is a clean, behavior-preserving extraction — `restore_stance_after_cast` mirrors the old `if @@after_stance / elsif` structure exactly, and the one intentional behavior change (accepting "You fall back into a ... stance" in the confirm regex) is called out explicitly in the PR body rather than smuggled in.
- `Stance.rb` is required in `gameloader.rb` right after `infomon/status.rb`, which is the correct load-order dependency (`Stance.change` calls `Status.dead?`) — checked and confirmed in `lib/common/gameloader.rb:29-31`.
- Good defensive design in `perfection?` (`rescue StandardError => false`) so a `CMan` lookup failure can't take down a stance change that doesn't actually need Stance Perfection.
- Solid test coverage of the documented behaviors (forced sends, fire-and-forget via `fput`, dead-character short-circuit, custom timeout, percent-vs-band interaction) — 26 examples, all currently green.

## Coverage notes

- Full local worktree checked out at the PR head SHA (`b429c44e`); all four changed files read in full, plus the surrounding call graph (`lib/attributes/char.rb`, `lib/gemstone/infomon/status.rb`, `lib/gemstone/psms/cman.rb`, `lib/gemstone/psms.rb`, `lib/common/xmlparser.rb`, `lib/global_defs.rb`) and the only real-world caller of `Spell.after_stance=` across the `dr-scripts`/`scripts` repos (`bigshot.lic`).
- Ran `bundle exec rspec spec/lib/gemstone/stance_spec.rb` locally: 26 examples, 0 failures, matching the PR's CI. Confirmed the CI's own rubocop/test runs (both green in `checks.txt`).
- Ran `rubocop` on all four changed files: the new/changed lines are clean; the offenses rubocop reports in `lib/common/spell.rb` are all in pre-existing one-line method defs (`spell.rb:898-902`) untouched by this diff, not introduced by this PR.
- Reproduced the `Stance.normalize` bug by loading the shipped module directly and calling it in isolation (see Major # 1) — `confirmed`, not inferred from reading alone.
- Ran `coderabbit review --agent --base-commit <merge-base> --committed` against the worktree (authenticated as `mrhoribu`). It returned one finding (the `at?` exact-vs-band percent comparison, folded into Minor # 2 above after independent verification against the source — GitHub's own CodeRabbit check shows "Review skipped: automatic reviews are disabled" for the PR itself, so this was a local run, not the PR's check).
- Did not test in a live game session (per the PR's own "Testing" section, that was done by the author). The behavioral claims about `Char.stance`'s initial empty-string state and `pbarStance` timing are grounded in `lib/common/xmlparser.rb` source, not runtime observation.

Nisugi:
Thanks @mrhoribu, that prefix bug is a real one and the reproduction made it easy to confirm. All four items are addressed in 6bcb7e02.

**Major 1, `normalize` prefix matching.** Fixed exactly as you suggested: the whole stripped input must now be a prefix of the candidate name, not just its first three characters. I verified against the shipped module rather than only through the specs, and all five of your examples plus `"offer"` now raise `ArgumentError`, while genuine prefixes of any length (`def`, `defe`, `defens`, `guard`) still resolve.

You were right that the existing invalid-input examples passed for the wrong reason. Added two: one asserting the six look-alikes raise, one asserting prefixes longer than three characters still work, so the suffix path is no longer untested.

**Minor 2, `CONFIRM` / `REFUSED` duplication.** Split the patterns into `ACCEPTED` and `DECLINED` arrays and derived both constants from them, so the set `dothistimeout` waits on and the set `change` treats as a refusal cannot drift. Added a spec that walks `DECLINED` and asserts each pattern matches both constants, and that acceptance lines match `CONFIRM` only.

**Nit 3, double normalize.** Added a private `at_normalized?` taking already-parsed values. Public `at?` delegates to it and `change` reuses the `name`/`percent` it already computed.

**Open question, blank `Char.stance`.** Your tracing is correct, and I couldn't rule out the timing window as firmly as I'd like, so I closed the path rather than argue it's unreachable. `restore_stance_after_cast` now treats a blank `after_stance` as "no preference" and falls through to the safest-stance branch, which is what the old inline regex check effectively did. That leaves `normalize`'s "unknown stance raises" contract intact for real callers instead of loosening it for a case that's really about uninitialized state.

Stance spec is 30 examples, 0 failures. Rubocop clean on all three touched files. The full suite shows 56 failures, but the same 56 fail on the unmodified commit here (frontend locator and session launcher, platform-specific), so they're pre-existing in my environment and untouched by this change.

mrhoribu (approved):
Re-reviewed after the `6bcb7e02` follow-up commit. The blocking issue is resolved — `Stance.normalize` now requires the whole input to be a genuine prefix of a stance name, not just its first three characters, so garbage strings like `"guardian"` or `"advertised"` correctly raise `ArgumentError` instead of silently resolving to a stance. Verified by re-running the original repro directly against the fixed code; new specs cover it too (30 examples, all green).

The `CONFIRM`/`REFUSED` duplication is also fixed, and better than asked — both are now derived from a shared `DECLINED` array so they can't drift apart, with tests locking that in.

One minor item is still open and wasn't touched by this commit: `at?`'s no-op check (`lib/gemstone/stance.rb:106`) compares exact percent even for characters without Stance Perfection, so `Stance.change(80)` will keep resending `"stance guarded"` when the character is already in that band (only band-level control is possible without Perfection, so the exact-percent comparison never no-ops there). Low impact — redundant command, not a crash or wrong return value — worth a fast-follow but not blocking.

Nice fixes, thanks for the quick turnaround.

PR #1579, opened 2026-09-10
feat(gs): Lich::Stash.wield and Stash.hands put named items into hands

Nisugi:
## Summary

`stash_hands` / `equip_hands` remember and restore whatever was held. This is the other half: get a specific item into a hand, named by a profile string, an id, a GameObj, or a ready-list slot, and reconcile both hands to a wanted state without dropping anything. Every command is sent by `#id`, so an ambiguous noun never reaches the game.

## What it does

```ruby
Lich::Stash.find_items('rod')                 # every distinct candidate, best first, nothing sent
Lich::Stash.find_item('vultite broadsword')  # the best one, raises only if nothing matches
Lich::Stash.wield('broadsword')              # into whichever hand the game picks
Lich::Stash.wield(:weapon, hand: :left)      # ready-list slot, specific hand (swaps if needed)
Lich::Stash.hands(right: 'broadsword', left: :shield)
Lich::Stash.hands(right: nil)                # empty the right hand, leave left alone (:keep is the default)
```

- **Resolution.** `find_items` matches the way `find_container` does (case-insensitive, words may be non-adjacent) and ranks by specificity (whole name, whole words inside the name, noun only) then location (hands, ready list, worn, containers), which is the order the game resolves a bare noun. Identical names collapse to one. When the GameObj registries miss, it refreshes `Lich::Common::Inventory` and matches the full tree, so an item in a sack nobody has looked in still resolves. `wield` echoes one line naming what it picked when the name meant more than one kind.
- **Containers.** Before the `get`, `wield` walks the item's parent chain in the inventory snapshot and opens each container whose current flag says closed, outermost first, each confirmed. Locked stops it before any `get`. If the item doesn't arrive (usually a container closed since Lich last looked), it refreshes once, reopens, and retries the `get` once before raising.
- **Hands.** The target hand is emptied through `stash_hands` first, worn items are `remove`d and the rest `get`, arrival is confirmed by id, and a `swap` follows if the game put it in the other hand. `hands` resolves every name before touching anything and does one `swap` when the two wanted items are in each other's hands.

The existing `Stash.container` look-and-open path is untouched; `add_to_bag` still uses it.

## Files

- `lib/stash.rb`: `find_item`, `find_items`, `wield`, `hands`, `open_path_to`, `open_container`, and the small readers they're built from (`hand_holding`, `in_hand?`, `empty_hand?`, `worn?`, `container_holding`)
- `spec/lib/stash_wield_spec.rb`: 37 examples with fakes for ReadyList and the Inventory snapshot

## Testing

- `bundle exec rspec spec/lib/stash_wield_spec.rb` green, rubocop clean on both files.
- In game: wield by name, id, and ready slot; from an open sack (no `open` sent), a closed sack (`open` then `get`), a sack never looked in (one `_inventory manager` then `get`), and a sack closed after Lich saw inside it (`get` misses, refresh, `open`, `get`); `hands` reconciles and empties; a name matching several kinds echoes the pick.

## Why now

Second of two small lib additions that let the e-scripts drop their private copies of common helpers (bigshot `cmd_wield`, eherbs and eloot hand helpers). Independent of the Stance PR.

### DIFF
--- a/lib/stash.rb
+++ b/lib/stash.rb
@@ -453,6 +453,15 @@ module Lich
       wanted = { right: right, left: left }
       resolved = wanted.transform_values { |want| want == :keep || want.nil? ? want : find_item(want) }
 
+      # A wanted item sitting in the hand the caller asked to keep can only be
+      # moved by a swap, which would change that hand. Refuse before touching anything.
+      HANDS.each do |hand|
+        other = hand == :right ? :left : :right
+        item = resolved[hand]
+        next unless item.is_a?(GameObj) && wanted[other] == :keep && hand_holding(item) == other
+        raise ArgumentError, "hands: #{item.name} is in the #{other} hand, which was asked to be kept"
+      end
+
       # Both wanted items are present but in each other's hands: one swap.
       if resolved[:right].is_a?(GameObj) && resolved[:left].is_a?(GameObj) &&
          hand_holding(resolved[:right]) == :left && hand_holding(resolved[:left]) == :right
--- a/lib/stash.rb
+++ b/lib/stash.rb
@@ -493,17 +493,22 @@ module Lich
 
       # Empty first, so a wanted item can land in a freed hand. A hand asked to
       # be emptied that is holding the item wanted in the OTHER hand needs a
-      # swap, not a stash: stashing would drag the item into a container only
-      # for the wield below to fetch it straight back out.
+      # swap, not a stash: stashing would drag the wanted item into a container
+      # only for the wield below to fetch it straight back out. If the other
+      # hand is occupied by something unwanted, stash THAT and then swap, so
+      # the item the caller asked for never goes into a container.
       HANDS.each do |hand|
         next unless resolved[hand].nil? && wanted.key?(hand) && !(wanted[hand] == :keep)
         next if empty_hand?(hand)
         other = hand == :right ? :left : :right
         other_item = resolved[other]
-        if other_item.is_a?(GameObj) && hand_holding(other_item) == hand && empty_hand?(other)
-          waitrt?
-          dothistimeout 'swap', 3, /^You don't have anything to swap!|^You swap/
-          next if empty_hand?(hand)
+        if other_item.is_a?(GameObj) && hand_holding(other_item) == hand
+          free_hand(other) unless empty_hand?(other)
+          if empty_hand?(other)
+            waitrt?
+            dothistimeout 'swap', 3, /^You don't have anything to swap!|^You swap/
+            next if empty_hand?(hand)
+          end
         end
         free_hand(hand)
       end
--- a/lib/stash.rb
+++ b/lib/stash.rb
@@ -281,10 +281,9 @@ module Lich
     #   ready-list slot (:weapon, :shield, ...); or a name matched the way
     #   find_container matches, case-insensitive and with words allowed to be
     #   non-adjacent ("vultite broadsword" matches "vultite hand-forged broadsword")
-    # @param loud_fail [Boolean] raise when not found or ambiguous
-    # @return [GameObj, nil]
-    # @raise [RuntimeError] when loud_fail and the item is missing, or a name
-    #   matches more than one item (that is a profile bug the user should see)
+    # @param loud_fail [Boolean] raise when nothing matches
+    # @return [GameObj, nil] the best candidate from find_items for a name
+    # @raise [RuntimeError] when loud_fail and the item is missing
     def self.find_item(param, loud_fail: true)
       return param if param.is_a?(GameObj)
 
@@ -292,17 +291,28 @@ module Lich
               when Integer then GameObj[param.to_s]
               when Symbol then find_ready_item(param)
               when String
-                if param =~ /\A\d+\z/
-                  GameObj[param]
-                else
-                  matches = known_items.select { |obj| name_matches?(obj, param) }.uniq(&:id)
-                  matches = inventory_matches(param) if matches.empty?
-                  fail "Item[name: #{param}] matches #{matches.size} items: #{matches.map(&:name).join(', ')}" if matches.size > 1 && loud_fail
-                  matches.first
-                end
+                param =~ /\A\d+\z/ ? GameObj[param] : find_items(param).first
               end
       fail "could not find Item[#{param.inspect}]" if found.nil? && loud_fail
       found
+    end
+
+    # Every distinct item a name could mean, best first, without sending
+    # anything. Ordered by how specific the match is (whole name, then part
+    # of the name, then noun only) and then by where the item is (hands,
+    # ready list, worn, containers), which is the order the game itself
+    # resolves a bare noun in. Items with identical names are interchangeable
+    # and collapse to the first.
+    #
+    # @param name [String]
+    # @return [Array<GameObj>] empty when nothing matches
+    def self.find_items(name)
+      ranked = known_items_ranked.select { |obj, _loc| name_matches?(obj, name) }
+      ranked = inventory_matches(name).map { |obj| [obj, 4] } if ranked.empty?
+      ranked.sort_by.with_index { |(obj, loc), i| [match_specificity(obj, name), loc, i] }
+                    .map(&:first)
+            .uniq(&:id)
+            .uniq(&:name)
     end
 
     # The full inventory tree, refreshed on request through the game's
@@ -380,7 +390,14 @@ module Lich
     # @raise [RuntimeError] when the item cannot be found or never arrives
     def self.wield(param, hand: nil)
       fail "wield: hand must be :right, :left or nil, got #{hand.inspect}" unless hand.nil? || HANDS.include?(hand)
-      item = find_item(param)
+      if param.is_a?(String) && param !~ /\A\d+\z/
+        candidates = find_items(param)
+        fail "could not find Item[#{param.inspect}]" if candidates.empty?
+        item = candidates.first
+        echo "wield: #{param} -> #{item.name} (of #{candidates.size} kinds: #{candidates.map(&:name).join(', ')})" if candidates.size > 1
+      else
+        item = find_item(param)
+      end
 
       holding = hand_holding(item)
       if holding
@@ -503,6 +520,28 @@ module Lich
       ReadyList.ready_list[slot]
     end
 
+    # known_items with a location rank: 0 hands, 1 ready list, 2 worn, 3 in a container.
+    #
+    # @return [Array<Array(GameObj, Integer)>]
+    def self.known_items_ranked
+      hands = [GameObj.right_hand, GameObj.left_hand].compact.reject { |obj| obj.id.nil? }
+      ready = ReadyList.checked? ? ReadyList.ready_list.values.compact : []
+      hands.map { |obj| [obj, 0] } +
+        ready.map { |obj| [obj, 1] } +
+        GameObj.inv.to_a.map { |obj| [obj, 2] } +
+        GameObj.containers.values.flatten.map { |obj| [obj, 3] }
+    end
+
+    # 0 when the name is the whole item name, 1 when it is a run of whole
+    # words inside it, 2 when only the noun (or a loose match) hit.
+    def self.match_specificity(obj, name)
+      wanted = name.strip.downcase
+      actual = obj.name.to_s.downcase
+      return 0 if actual == wanted
+      return 1 if actual =~ /(?:\A|\s)#{Regexp.escape(wanted)}(?:\s|\z)/ && wanted.include?(' ')
+      2
+    end
+
     # Every item Lich currently knows the character has: hands, worn, and the
     # contents of containers that have been looked in.
     #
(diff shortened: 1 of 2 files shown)


mrhoribu (commented):
# PR Review: feat(gs): Lich::Stash.wield and Stash.hands put named items into hands (elanthia-online/lich-5#1579)

**Reviewed:** 2 files, +691/-0 at `ad65eb8` on 2026-09-11. Based on: full fresh clone of `elanthia-online/lich-5`, PR branch checked out.
**Verdict:** Request changes
**Summary:** Adds `Lich::Stash.find_item(s)`, `.wield`, and `.hands` — a name/id/ready-slot resolver plus commands that put a specific item into a hand or reconcile both hands to a wanted state, everything targeted by `#id` as advertised. The resolution/ranking/container-opening logic is careful and well-grounded against the real `GameObj`/`Lich::Common::Inventory`/`ReadyList` contracts, but both `wield` and `hands` reuse the legacy `stash_hands` to free a hand, and that silently pushes onto the same global, cross-script restore stacks the pre-existing `empty_hand`/`fill_hand` script API drains — corrupting that shared state every time a hand needs clearing.

## Blockers

### 1. `wield`/`hands` corrupt the shared `$fill_right_hand_actions` / `$fill_left_hand_actions` restore stacks — `lib/stash.rb:413,478`

```ruby
if hand
  stash_hands(**{ hand => true }) unless empty_hand?(hand)
```

`stash_hands` (unchanged, `lib/stash.rb:123-262`) isn't a private helper — it's the exact method `lib/global_defs.rb:1934/1937/1944/1949` exposes to every `.lic` script as `empty_hand`/`empty_right_hand`/`empty_left_hand`, and it unconditionally records a restore action onto a global, process-wide stack before returning: `$fill_left_hand_actions.push(actions) if left` / `$fill_right_hand_actions.push(actions) if right` (`lib/stash.rb:260-261`). Those exact globals are drained by `equip_hands` (`lib/stash.rb:574-598`), which is what `fill_hand`/`fill_right_hand`/`fill_left_hand` (`lib/global_defs.rb:1957-1969`) call.

Every time `wield` or `hands` needs to free an occupied hand — any call where the target hand isn't already empty — it invokes the real `stash_hands`, leaving behind a "how to get the item I just displaced back" action on this shared stack, with nothing to ever consume it (the displaced item is deliberately abandoned in a container; `wield`/`hands` never intend to restore it). Lich runs every `.lic` script as a `Thread` inside one process (`lib/common/script.rb:170`), so `$fill_right_hand_actions` is literally shared, LIFO, mutable state across every script running concurrently for that character, not just within the caller of `wield`.

Concretely: script A calls `empty_right_hand` to free a hand for some manual action, intending to later call `fill_right_hand` to get its own item back. If script B (or the same script, elsewhere) calls `Lich::Stash.wield(...)` in between and that call needs to clear the right hand, it pushes its own leftover restore action on top of A's. A's later `fill_right_hand` now pops and executes B's leftover action instead of its own, fetching the wrong item into A's right hand, while A's real restore action stays buried on the stack, waiting to fire the next time *anything* calls `fill_right_hand`/`equip_hands()` with no way to know it doesn't belong to them.

This is invisible to the new test suite because the spec stubs `stash_hands` at the top level for every example (`spec/lib/stash_wield_spec.rb:85`, `allow(described_class).to receive(:stash_hands)`), so the real push never executes during tests — the interaction with the pre-existing global mechanism can't surface from inside this PR's own spec file.

**Fix direction:** give `wield`/`hands` a way to free a hand that doesn't touch the public restore stacks — a private variant of `stash_hands` that skips the `$fill_*_actions.push`, or pop-and-discard the entry `stash_hands` just pushed for that hand immediately after the call.

Confidence: confirmed (traced against the unchanged `stash_hands`/`equip_hands` implementation and their real call sites in `lib/global_defs.rb`).

## Major

### 2. `Stash.hands` stashes-then-refetches instead of a single swap when the hand being emptied already holds the wanted item — `lib/stash.rb:452-489`

```ruby
HANDS.each do |hand|
  next unless resolved[hand].nil? && wanted.key?(hand) && !(wanted[hand] == :keep)
  stash_hands(**{ hand => true }) unless empty_hand?(hand)
end

HANDS.each do |hand|
  item = resolved[hand]
  next unless item.is_a?(GameObj)
  wield(item, hand: hand)
end
```

The "swap when both wanted items are in each other's hands" optimization (`lib/stash.rb:469-473`) only fires when *both* `resolved[:right]` and `resolved[:left]` are `GameObj`s. It misses the equally ordinary case of `hands(right: some_item, left: nil)` where `some_item` is *currently sitting in the left hand*. There, the "empty" loop above unconditionally calls the real `stash_hands(left: true)` on the very item the caller wants moved into the right hand — physically stowing it in a container — and only then does the `wield` step fetch it back out. That costs two extra game commands (`_drag` into a bag, then `get` back out) where a single `swap` would put the item directly where it belongs, the same way `wield`'s own already-implemented fast path (`lib/stash.rb:402-409`) already handles this for a bare `wield(item, hand:)` call.

Worse, `stash_hands` → `add_to_bag` → `try_or_fail` (`lib/stash.rb:35-40`) raises a bare `RuntimeError` if the `_drag` doesn't resolve within its timeout. A call that should always trivially succeed via one `swap` can instead raise a confusing `Error[command: _drag ...]` from deep inside `stash_hands`, for a request that never needed to touch a container.

I verified this by running the real code (stubbing only `fput`/`dothistimeout`/`stash_hands` as observation points, the same pattern the PR's own spec harness uses) against `hands(right: sword, left: nil)` with `sword` starting in the left hand: `stash_hands` is invoked with `{left: true}` before `wield` ever runs, confirming the item is routed through stash-then-refetch rather than a swap.

Notably, the analogous case with the default `:keep` — `hands(right: sword)` with sword already in the left hand — *is* explicitly guarded: it raises `ArgumentError` rather than swap an item out of a hand asked to be kept (`lib/stash.rb:461-466`, added specifically for this in commit `0c420f07`). The `nil` case doesn't need a guard (emptying the hand is exactly what's wanted), but it does need the same swap treatment `wield` already gives a bare call, not the stash-and-refetch path.

Confidence: confirmed (traced and reproduced against the real `hands`/`wield`/`stash_hands` implementation).

## Minor

### 3. `name_matches?` doesn't escape its regex, unlike the sibling `match_specificity` added in the same PR — `lib/stash.rb:568-569`

```ruby
def self.name_matches?(obj, param)
  obj.name =~ %r[#{param.strip}]i || obj.name =~ %r[#{param.sub(' ', ' .*')}]i
end
```

This deliberately mirrors `find_container`'s long-standing unescaped pattern (`lib/stash.rb:16`, present since 2024) — the doc comment says as much ("Same matching find_container uses"). But `match_specificity`, added in this same PR a few lines above (`lib/stash.rb:549-555`), does call `Regexp.escape(wanted)`. A profile string or item name containing regex metacharacters (an unbalanced paren, a stray `+` or `*`) can raise `RegexpError` out of `name_matches?` in a way `match_specificity` is already immune to. This isn't a regression the PR introduces on its own — the pattern is copied from existing code — but the PR makes it reachable from every `wield`/`hands` call rather than only from container lookups, and it sits inconsistently next to the escaping the PR's own new code does two methods away.

Confidence: confirmed (pattern present in code; not exercised by any spec with regex-special characters).

## What this PR gets right

- The resolution/ranking logic (`find_items`, `match_specificity`, `known_items_ranked`) is careful and well-tested: 39 passing examples cover location ranking, specificity ranking, and the identical-name collapse.
- Every real production API this PR depends on — `GameObj.right_hand`/`left_hand`/`inv`/`containers`/`[]`, `Lich::Common::Inventory`'s `refresh`/`current`/`Snapshot#[]`/`#all`, `Item#parent_item`/`#closed?`/`#locked?`/`#in_room?`/`#at_feet?`, `ReadyList.ready_list`/`.valid?`/`.checked?`/`.check` — was checked against the actual implementation and matches exactly; the spec's fakes (`InvItem`, `FakeInventory`) use the real method names and semantics rather than inventing an API.
- Empty-hand handling (`empty_hand?`, `hand_holding`) correctly accounts for both a literal `nil` hand (pre-first-XML-tag) and the live "Empty" `GameObj` the parser reports once running, matching `lib/common/gameobj.rb`'s actual `new_right_hand`/`new_left_hand` contract rather than just the test fixtures.
- The `hands(right: sword)` case where `sword` is in a hand the caller asked to `:keep` is explicitly and correctly refused with `ArgumentError` before any command is sent (`lib/stash.rb:461-466`) — good defensive design; it just wasn't extended to the `nil` case (Major #2).
- Container discovery reuses the `Lich::Common::Inventory` snapshot correctly (`open_path_to`, `inventory_matches`) rather than duplicating the older, more limited `Stash.container` look-and-open path, which the PR leaves untouched as stated in its description.
- YARD-style docs on every new public method, consistent with the convention already used in `lib/common/inventory.rb`/`gameobj.rb`.

## Coverage notes

- Reviewed from a full fresh clone of `elanthia-online/lich-5`, PR branch checked out via `gh pr checkout` (the pre-existing local `~/lich-5` checkout had unrelated uncommitted work on a different branch and was left untouched).
- `bundle exec rspec spec/lib/stash_wield_spec.rb`: 39 examples, 0 failures. `bundle exec rubocop lib/stash.rb spec/lib/stash_wield_spec.rb`: no offenses.
- Ran a CodeRabbit CLI pass (`coderabbit review --agent --base main`); both of its findings were investigated — the regex-escaping one is Minor #3 above, and its restore-stack finding is the basis for Blocker #1, expanded here with the concrete cross-script mechanism and the production call sites (`lib/global_defs.rb`) its own output didn't cite.
- Blast radius: this is new API surface with zero existing callers inside lich-5 itself (no other file references `Lich::Stash::(wield|hands|find_item)`); the PR description states dr-scripts' `bigshot`/`eherbs`/`eloot` are the intended future consumers, which live in a separate repo not reviewed here.
- Not tested against a live game session (not available from this environment); the container-opening and retry-on-miss paths (`open_path_to`, the refresh-and-retry-once in `wield`) are verified against the spec's fakes and the real `Lich::Common::Inventory` source, not against live server responses.

Nisugi:
Thanks for the review, all three are fixed, plus a fourth the second finding pointed near. Pushed as `c5464b0c` and `6de4b212`.

**1. Restore-stack corruption (blocker).** Confirmed, and the cross-script mechanism you traced is exactly right. Added a private `free_hand(hand)` that records the stack depth, calls `stash_hands`, then pops only the entries that call added. Entries another script already pushed are left alone. All three call sites in `wield` and `hands` now go through it. Three specs cover this, including one asserting a pre-existing entry survives a `wield` that frees a hand.

**2. Stash-then-refetch (major).** Confirmed. `hands` now swaps instead of stashing when the hand being emptied holds the item wanted in the other hand.

Tracing this turned up a second case your report didn't name. Your repro had the other hand empty; with **both** hands full the code stashed the hand holding the *wanted* item, then fetched it back out:

```
stash left    <- the wanted item, into a container
stash right   <- the unwanted item
get #101      <- the wanted item, back out
```

Now it stashes only what is actually in the way, then swaps:

```
stash right   <- only the unwanted item
swap
```

Two commands instead of three, and the item the caller asked for never enters a container. Same end state.

**3. Unescaped regex (minor).** Fixed. `name_matches?` escapes both patterns now, matching what `match_specificity` already did. The loose non-adjacent-words pattern is otherwise byte-identical to before: I split on a literal single space rather than `split(' ')`, which is awk-style in Ruby and would have collapsed repeated spaces and changed matching for names with double spaces.

mrhoribu (approved):
Re-reviewed after the two follow-up commits (`c5464b0c`, `6de4b212`) — both issues from my earlier pass are resolved:

**Global restore-stack corruption** (`wield`/`hands` calling `stash_hands`, which pushes onto the process-wide `$fill_right_hand_actions`/`$fill_left_hand_actions` stacks shared with the public `empty_hand`/`fill_hand` script API): fixed via the new `free_hand(hand)` helper (`lib/stash.rb:391-401`), which records the stack depth before stashing and pops back down to it afterward — discarding only the entries it added, leaving any pre-existing entries from other scripts untouched. Verified against the exact mechanism, and the new "hand-restore stacks" spec block covers both the cleanup and the non-interference case directly.

**`hands` stashing the wanted item instead of swapping it** (`hands(right: X, left: nil)` when `X` was already in the left hand): fixed in the reworked empty-loop (`lib/stash.rb:491-516`) — it now detects when the hand being emptied holds the item wanted elsewhere and does a `swap` (freeing the other hand first if it holds something unwanted) instead of stashing the wanted item into a container and fetching it back. Traced both branches by hand against the two new tests and they match.

The regex-escaping inconsistency in `name_matches?` is also fixed, with escaping applied per-segment so the original wildcard-between-words matching behavior is preserved exactly, plus new tests for metacharacter-bearing names.

Ran the full suite myself: 47/47 examples pass, rubocop clean, and CI is green on `6de4b21`. One non-blocking nit: `free_hand` isn't added to the `private_class_method` list alongside the other internal-only helpers — purely cosmetic.

Approving.

PR #1581, opened 2026-09-10
feat(gs): Lich::Gemstone::Bank, Currency refresh, and WEALTH parser lines

Nisugi:
## Summary

Moves the bank handling that eloot, eherbs and ebounty each duplicate into core, and gives Currency a way to refresh on demand.

**`Lich::Gemstone::Bank`** (new, `lib/gemstone/bank.rb`)
- `here?`, `pinefar?`, `f2p?`
- `account` parses the per-town BANK ACCOUNT listing into `{ balance:, max:, banks:, total: }`, with `balance` being this town's bank matched by room location; nil where the bank refuses access.
- `deposit(amount = :all)`, `deposit_note`, `withdraw(amount, note: false)`, `note_value`, `note_in_hand`, `notes_in(container)`.
- Pinefar depository wording (give/ask banker, `wait_for_banker`) and the free-to-play cap path (`deposit_f2p` fills to the cap and takes notes; `withdraw_f2p` deposits stowed notes to cover a shortfall).
- All replies are sent through `dothistimeout` against one `Pattern` set, so a reworded game line is a one-place fix.

**`Lich::Gemstone::Currency`**
- `silver(refresh: false)`, `refresh(all: false)` sending plain `wealth` (QUIET is now the default; LOUD is the option) or `wealth all`.
- `silver_total(refresh:)`, `notes(refresh:)`, `refresh_notes` sending `wealth notes`.

**Infomon parser**
- `WealthSilverTotal` ("You are carrying a total of N silver.") to `currency.silver_total`
- `WealthNotes` ("Total note value: N") to `currency.notes`

Both are anchored patterns in the fast-path list. Requests go through `Lich::Util.issue_command` quietly; Infomon parses before downstream hooks, so hiding the response from the front end does not hide it from Infomon.

## Test plan

- [x] `spec/lib/gemstone/bank_spec.rb`, 30 examples
- [x] `spec/lib/gemstone/currency_spec.rb`, 6 examples
- [x] `spec/lib/gemstone/infomon_spec.rb` extended for the two new lines
- [x] rubocop clean on touched files
- [x] In game: `Currency.refresh`, `silver_total`, `notes`, `Bank.account`, deposit, withdraw, note withdraw/read/deposit. Pinefar and F2P paths are spec-only.

### DIFF
--- a/lib/common/gameloader.rb
+++ b/lib/common/gameloader.rb
@@ -34,6 +34,7 @@ module Lich
         require File.join(LIB_DIR, 'gemstone', 'psms.rb')
         require File.join(LIB_DIR, 'attributes', 'char.rb')
         require File.join(LIB_DIR, 'gemstone', 'currency.rb')
+        require File.join(LIB_DIR, 'gemstone', 'bank.rb')
         # require File.join(LIB_DIR, 'gemstone', 'character', 'disk.rb') # dup
         require File.join(LIB_DIR, 'gemstone', 'group.rb')
         require File.join(LIB_DIR, 'gemstone', 'critranks')
--- a/lib/gemstone/bank.rb
+++ b/lib/gemstone/bank.rb
@@ -278,22 +278,31 @@ module Lich
         taken = 0
         want = amount
         refused = false
-        notes_in(stow).each do
-          if bal.positive?
-            got = withdraw_silver(bal)
-            if got.nil?
-              refused = true
-              break
-            end
+
+        # Drain what the account already holds before touching any notes: that
+        # has to happen even when the stow container is empty.
+        if bal.positive?
+          got = withdraw_silver(bal)
+          if got.nil?
+            refused = true
+          else
             taken += got
             want -= got
-            bal = 0
           end
+        end
+
+        until refused || want <= 0
           note = notes_in(stow).first
-          break if note.nil?
+          if note.nil?
+            refused = true
+            break
+          end
           fput "get ##{note.id}"
           value = deposit_note(note)
-          break if value.nil?
+          if value.nil?
+            refused = true
+            break
+          end
           got = withdraw_silver([value, want].min)
           if got.nil?
             refused = true
@@ -301,8 +310,8 @@ module Lich
           end
           taken += got
           want -= got
-          break if want <= 0
         end
+
         Currency.refresh
         refused && taken.zero? ? nil : taken
       end
--- a/lib/gemstone/bank.rb
+++ b/lib/gemstone/bank.rb
@@ -211,26 +211,38 @@ module Lich
       # @param amount [Integer, :all]
       # @param stow [GameObj, nil] where notes go; the STOW default container
       # @return [Integer, nil] silver put away in total, nil when the bank refused
+      #   before anything was deposited
       def self.deposit_f2p(amount = :all, stow: StowList.default)
         total = 0
+        refused = false
+        remaining = amount == :all ? nil : amount
         loop do
           info = account
-          return nil if info.nil?
+          if info.nil?
+            refused = true
+            break
+          end
           carried = Currency.silver(refresh: true).to_i
-          carried = [carried, amount].min unless amount == :all
+          carried = [carried, remaining].min if remaining
           break unless carried.positive?
 
           max = info[:max]
           if max.nil? || info[:balance] + carried < max
-            dothistimeout("deposit #{carried}", 3, Pattern::DEPOSIT)
-            total += carried
+            got = deposited(dothistimeout("deposit #{carried}", 3, Pattern::DEPOSIT))
+            refused = got.nil?
+            total += got.to_i
             break
           end
 
           room = max - info[:balance]
           if room.positive?
-            dothistimeout("deposit #{room}", 3, Pattern::DEPOSIT)
-            total += room
+            got = deposited(dothistimeout("deposit #{room}", 3, Pattern::DEPOSIT))
+            if got.nil?
+              refused = true
+              break
+            end
+            total += got
+            remaining -= got if remaining
           end
           carried = Currency.silver(refresh: true).to_i
           note_size = carried >= F2P_NOTE_BUFFER ? max : max - (F2P_NOTE_BUFFER - carried)
@@ -240,7 +252,7 @@ module Lich
           Lich::Stash.add_to_bag(stow, note)
         end
         Currency.refresh
-        total
+        refused && total.zero? ? nil : total
       end
 
       # Free-to-play withdrawal: the balance may be short of +amount+ while notes
@@ -254,18 +266,23 @@ module Lich
         return nil if info.nil?
         bal = info[:balance]
         if bal >= amount
-          dothistimeout("withdraw #{amount} silver", 3, Pattern::WITHDRAW)
+          got = withdraw_silver(amount)
           Currency.refresh
-          return amount
+          return got
         end
 
         taken = 0
         want = amount
+        refused = false
         notes_in(stow).each do
           if bal.positive?
-            dothistimeout("withdraw #{bal} silver", 3, Pattern::WITHDRAW)
-            taken += bal
-            want -= bal
+            got = withdraw_silver(bal)
+            if got.nil?
+              refused = true
+              break
+            end
+            taken += got
+            want -= got
             bal = 0
           end
           note = notes_in(stow).first
@@ -273,14 +290,17 @@ module Lich
           fput "get ##{note.id}"
           value = deposit_note(note)
           break if value.nil?
-          take = [value, want].min
-          dothistimeout("withdraw #{take} silver", 3, Pattern::WITHDRAW)
-          taken += take
-          want -= take
+          got = withdraw_silver([value, want].min)
+          if got.nil?
+            refused = true
+            break
+          end
+          taken += got
+          want -= got
           break if want <= 0
         end
         Currency.refresh
-        taken
+        refused && taken.zero? ? nil : taken
       end
 
       # The Pinefar banker wanders; give him a moment to be at the counter.
@@ -295,6 +315,17 @@ module Lich
         end
         true
       end
+
+      # WITHDRAW N SILVER, confirmed.
+      #
+      # @api private
+      # @return [Integer, nil] silver handed over, nil when refused or unanswered
+      def self.withdraw_silver(amount)
+        result = dothistimeout("withdraw #{amount} silver", 3, Pattern::WITHDRAW)
+        return nil if result.nil? || result =~ Pattern::WITHDRAW_REFUSED
+        result =~ /(?<silver>[\d,]+) silver/ ? Regexp.last_match[:silver].delete(',').to_i : amount
+      end
+      private_class_method :withdraw_silver
 
       # @api private
       def self.deposited(result)
--- a/lib/gemstone/bank.rb
+++ b/lib/gemstone/bank.rb
@@ -45,9 +45,12 @@ module Lich
         # Replies that mean the withdrawal did not happen.
         WITHDRAW_REFUSED = /seem to have that much|looks at you suspiciously|chuckles at you|taps her quill|purses her lips/.freeze
 
-        ACCOUNT_START   = /You currently have an account|you don't have access/i.freeze
+        ACCOUNT_START   = /You currently have the following amounts on deposit|You currently have an account|you don't have access/i.freeze
+        ACCOUNT_LINE    = /^\s+(?<bank>.+?) Bank: (?<silver>[\d,]+)$/.freeze
+        ACCOUNT_TOTAL   = /^\s+Total: (?<silver>[\d,]+)$/.freeze
         ACCOUNT_BALANCE = /in the amount of (?<silver>[\d,]+) silver/.freeze
         ACCOUNT_MAX     = /a maximum of (?<silver>[\d,]+) silvers/.freeze
+        ACCOUNT_END     = /urchin bank runner uses remaining|a maximum of|you don't have access|<prompt/i.freeze
         NO_ACCESS       = /you don't have access/i.freeze
         NOTE_VALUE      = /has a value of (?<silver>[\d,]+) silver and reads/.freeze
         NOTE_READ       = /Hold in right hand to use|has a value of/.freeze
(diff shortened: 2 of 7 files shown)


mrhoribu (commented):
# PR Review: feat(gs): Lich::Gemstone::Bank, Currency refresh, and WEALTH parser lines (elanthia-online/lich-5#1581)

**Reviewed:** 7 files changed (+864/-2 per GitHub; `gh pr diff`), at `4bb4f729` on 2026-09-11. Based on: full clone of the PR branch plus `main`, merge-base `073d23a9`.
**Verdict:** Request changes
**Summary:** Centralizes GemStone bank verbs (deposit/withdraw, Pinefar depository, F2P cap handling) into a new `Lich::Gemstone::Bank`, and gives `Currency` an on-demand `WEALTH` refresh plus two new parsed fields (`silver_total`, `notes`). The new module is thoroughly unit-tested (30 examples) and the commit history shows several real bugs already caught and fixed in review-by-self before this was opened — but the F2P withdraw path still has a gap the test suite doesn't exercise: when the account balance is short and the stow container holds no notes, `withdraw_f2p` never attempts the withdrawal at all and reports success with `0` instead of failure.

## Major

### 1. `Bank.withdraw_f2p` skips withdrawing the available balance, and reports `0` instead of `nil`, whenever the stow container holds no notes — `lib/gemstone/bank.rb:268-308`

```ruby
taken = 0
want = amount
refused = false
notes_in(stow).each do
  if bal.positive?
    got = withdraw_silver(bal)
    ...
```

`notes_in(stow).each` is being used as the loop driver for the whole "drain balance, then convert notes" sequence — but the `bal.positive?` branch that actually withdraws the account's own balance lives *inside* that `.each` block. If `notes_in(stow)` returns `[]` (no bank notes currently sitting in the stow container — the common case for any F2P character who hasn't already gone through a `deposit_f2p` note-conversion cycle), the block body never runs at all: `bal` is never drained, no command is ever sent to the game, and the method falls straight through to `refused && taken.zero? ? nil : taken` with `refused == false` and `taken == 0`, returning `0`.

Concretely: an F2P character with 2,000 silver in their account, no notes in their pack, calling `Bank.withdraw(8000)` gets back `0` — not the 2,000 they actually had available, and not `nil` to signal the request wasn't satisfied. Since only `nil`/`false` are falsy in Ruby, every natural guard a caller would write (`if amount = Bank.withdraw(8000)`, `unless amount.nil?`) treats this `0` as a successful outcome, silently masking a withdrawal that never happened.

This is easy to miss because the same "0 means nothing to do" convention is correct and intentional for `deposit`/`deposit_f2p` (tested: "reports 0 with nothing to deposit"), where 0 really does mean nothing needed doing. For `withdraw`, the established contract is the opposite: the non-F2P path (`lib/gemstone/bank.rb:223`) and the F2P early-return path (`bal >= amount` at `lib/gemstone/bank.rb:272-276`, via `withdraw_silver`) both map "couldn't get the money" to `nil`. The multi-note F2P loop is the one path that conflates "never attempted" with "successfully withdrew zero."

The same root cause also produces `0` (instead of `nil`) whenever the loop *does* run but exits via `break if note.nil?` (ran out of notes) or `break if value.nil?` (a note failed to deposit) while `taken` is still 0 — those breaks never set `refused = true`, unlike the two `got.nil?` branches that do.

Fix direction: withdraw the available `bal` unconditionally before/outside the notes loop (not gated on `notes_in(stow)` being non-empty), and set `refused = true` on the `note.nil?`/`value.nil?` exits too when nothing has been taken yet.

Confirmed by tracing the code directly; not exercised by any existing spec (all four `withdraw_f2p`/f2p `.withdraw` examples stub at least one note into `sack.contents`, or a `bal >= amount` balance that takes the early return).

## Minor

### 2. `Bank.deposit` at Pinefar always reports `0`, even on a real, large deposit — `lib/gemstone/bank.rb:367-372`, `lib/gemstone/bank.rb:175-189`

```ruby
def self.deposited(result)
  return nil if result.nil? || result =~ Pattern::NO_ACCESS
  return 0 if result =~ /no coins to deposit/
  result =~ /(?<silver>[\d,]+)/ ? Regexp.last_match[:silver].delete(',').to_i : 0
end
```

The Pinefar banker's confirmation line ("Smiling greedily, Hurshal takes your silvers and says...") never echoes a figure, so `deposited` falls through to the `: 0` default. This is exercised directly by the spec (`'gives the banker at Pinefar after he is at the counter'` asserts `deposit` returns `0` after sending `800` silver), so it's a known, tested limitation rather than an oversight — but it contradicts the method's own doc comment, "@return ... silver deposited as the bank reported it" (`lib/gemstone/bank.rb:170-174`): the bank didn't report a refusal, it reported nothing, and the code treats that identically to "nothing to deposit." A caller summing `Bank.deposit` return values to track silver banked in a session (which the doc comment invites) will permanently undercount every Pinefar deposit. Worth a one-line doc caveat at minimum, since the current comment overpromises.

## Nits

### 3. `withdraw` and `withdraw_silver` duplicate the same refusal check and silver-extraction regex — `lib/gemstone/bank.rb:210-227`, `lib/gemstone/bank.rb:359-364`

```ruby
return nil if result.nil? || result =~ Pattern::WITHDRAW_REFUSED
...
result =~ /(?<silver>[\d,]+) silver/ ? Regexp.last_match[:silver].delete(',').to_i : nil
```

This exact pair of lines appears in both `withdraw`'s non-note/non-Pinefar branch and the private `withdraw_silver` helper that the F2P path calls, differing only in whether the outgoing command says `"silvers"` or `"silver"`. Both spellings are individually locked in by tests (`sent == ['withdraw 8000 silvers']` vs `['withdraw 8000 silver']`), so this reads as deliberate rather than accidental, but it's still the same parsing logic maintained in two places — a future change to `WITHDRAW_REFUSED` or the silver regex has two call sites to remember. Non-blocking.

## Open questions

- `lib/gemstone/bank.rb:65` (`Pattern::WITHDRAW = Regexp.union(WITHDRAW_RESULT, DEBT)`) has no `NO_ACCESS` alternative, unlike the deposit path's `Regexp.union(Pattern::DEPOSIT, Pattern::NO_ACCESS)` at `lib/gemstone/bank.rb:185`. If a bank ever refuses a *withdraw* attempt with the same "you don't have access to an account here" line it uses for deposits and `BANK ACCOUNT`, `withdraw_reply`'s `dothistimeout` would sit out its full timeout waiting for a line that will never come, rather than returning immediately. The end result (`nil`) is the same either way, so this only costs latency, not correctness — but it's worth confirming whether that refusal line is actually reachable from a `WITHDRAW` command in-game, or whether the game always answers withdrawals differently even on refusal (in which case there's nothing to fix).
- `lib/gemstone/bank.rb:145` (`local_bank`) reads `Room.current&.location.to_s`. `Lich::Common::Map#location` can hold `false` as a real value (see `lib/common/map/map_gs.rb:401`, which explicitly checks `room.location.nil? || room.location == false || room.location == ''`) for a room whose location hasn't been successfully surveyed yet. `false.to_s` is `"false"`, a non-empty string, so `local_bank` wouldn't short-circuit on the `location.empty?` guard and would instead search for a bank named "false" among `banks.keys` — failing to match (returning `nil`) and causing `account[:balance]` to fall back to `0` even when a real matching bank balance is listed. Given how `Bank.account`/`Bank.balance` are meant to be used (interactively, standing in a real, presumably already-mapped room), is an un-surveyed `location == false` actually reachable at the point a script calls into this module, or does normal mapping/room-entry always resolve it to a string or `nil` first?

## What this PR gets right

- The bank protocol (deposit/withdraw/notes/Pinefar/F2P cap dance) is genuinely gnarly, and the commit history (`08a9151f`, `af25ee59`, `122b7ac0`, `4bb4f729`) shows the author iterating against their own tests to kill real bugs — the debt-notice-as-preface fix in particular (a teller's bill line arriving before the actual withdraw answer) is the kind of thing that's easy to get wrong and easy to miss without in-game testing, which the PR description says was done.
- 30 spec examples for `Bank` alone, covering the account-listing parse, multi-town matching, Pinefar, and several F2P cap-dance passes — and they're not shallow: several specifically pin down "counts only what was confirmed" and "never deposits more than requested across passes" behavior that would be trivial to regress.
- `Currency.silver`/`silver_total`/`notes` keeping the "read cached Infomon value" and "send a command to force a refresh" concerns cleanly separated via a `refresh:` keyword, rather than always sending a command, is a sensible API shape for something scripts will call frequently.
- The private `deposited`/`withdraw_reply`/`withdraw_silver` helpers correctly consolidate the response-parsing regexes into one place, per the PR's stated goal of "a reworded game line is a one-place fix" — the one gap (finding 3, `withdraw`'s own branch not routing through `withdraw_silver`) is minor and doesn't undermine that goal for the patterns themselves.

## Coverage notes

- Full clone of `elanthia-online/lich-5`, PR branch checked out via `gh`/`git fetch pull/1581/head`; diff cross-checked against `gh pr diff` to rule out unrelated `release-please` version/changelog churn between `main` and the PR's merge-base (confirmed only the 7 files GitHub reports are actually part of this PR).
- Ran the full spec suite (`bundle exec rspec`): 7,192 examples, 0 failures, including the new `bank_spec.rb`/`currency_spec.rb` and extended `infomon_spec.rb` (90 examples for the touched files alone).
- Ran `rubocop` on all 7 changed files: no offenses.
- A local CodeRabbit pass (`coderabbit review --agent --base-commit <merge-base> --committed`) was attempted per the review workflow but hit the org's usage-based review rate limit ("You've used all 3 included reviews currently available", ~51 minute cooldown) before producing output — no CodeRabbit findings are included in this review as a result.
- Traced `Room`/`GameObj`/`StowList`/`Lich::Common::Account`/`Lich::Stash.add_to_bag`/`Lich::Util.issue_command` call sites and signatures directly in the checked-out `lib/` to confirm the new module's assumptions about each (module nesting for bare `Room`/`StowList` resolution, `Account.type` returning `"FREE"` for F2P, `location` possibly being `false`, etc.) rather than taking them on faith — findings above cite the specific lines each check was grounded against.
- No manual in-game verification was performed (not available in this environment); the PR description states the author did test deposit/withdraw/note flows in-game, with the explicit exception of the Pinefar and F2P paths, which it flags as "spec-only."

Nisugi:
Thanks for the review — all three findings addressed, plus one thing the review raised as an open question that turned out to be a real bug. Pushed as `82014b8e`, `549969d1`, `01ae5cf6`.

### 1. `withdraw_f2p` skipping the balance drain — confirmed, fixed in `82014b8e`

Correct diagnosis. The drain lived inside `notes_in(stow).each`, so an empty stow container meant the block never ran: no command sent, `0` returned. Hoisted the drain above the loop and drove the loop on what's still wanted:

```ruby
if bal.positive?
  got = withdraw_silver(bal)
  ...
end

until refused || want <= 0
  note = notes_in(stow).first
  ...
end
```

Also took the second half of the finding — `break if note.nil?` and `break if value.nil?` now set `refused`, so a withdrawal that was never attempted reports `nil` rather than `0`. A partial withdrawal still reports what was actually handed over, which the existing "reports only what was actually handed over when a later step fails" example pins.

Two new examples, both of which fail against the previous code.

### 2. Pinefar `deposit` returning 0 — doc caveat added in `01ae5cf6`

Agreed it's a tested limitation rather than an oversight, and agreed the doc comment overpromised. `@return` now says a successful Pinefar deposit reports 0 and that callers shouldn't sum these to track silver banked.

### 3. `withdraw` / `withdraw_silver` duplication — not taken

Leaving this one. The two commands genuinely differ (`silvers` vs `silver`) and both spellings are pinned by tests, so folding them means a word parameter that exists only to satisfy the dedupe. Happy to revisit if you feel strongly.

### Open question 1: is the no-access line reachable from a WITHDRAW? — yes, fixed in `549969d1`

Checked in game. Two cases:

```
>withdraw 10000 silver
The teller flips through the books and then looks up with a suspicious expression
and says, "I'm sorry, Cogruel, you don't seem to have that much in the account."
```

That one was already covered — `seem to have that much` is in both `WITHDRAW_RESULT` and `WITHDRAW_REFUSED`.

```
>withdraw 10000 silver
[Because your account is free, you don't have access to bank accounts in multiple
towns.  If you wish to move your account to here, you may acquire a bank account
transfer form from the SimuCoins Store.]
```

This one was in neither pattern, so `dothistimeout` sat out its full 3s waiting for an answer that had already arrived. Result was still `nil`, so it cost latency rather than correctness — but `withdraw_f2p` calls `withdraw_silver` once per stowed note, paying the timeout every pass. `NO_ACCESS` moved above `WITHDRAW` and added to both unions.

Also folded the teller's non-committal `^The teller makes` reply into `WITHDRAW_REFUSED`. That one is hardening, not a bug fix — it carries no silver figure so it already fell through to `nil`; naming it just makes that intentional rather than incidental.

**Worth flagging separately:** the `replies` helper stubs `dothistimeout` and discards the regex argument, so no behaviour-level example in this file can distinguish a missing pattern from a present one. My first attempt at covering this passed against the unfixed code and proved nothing. The new expectations assert against `Pattern::WITHDRAW` / `Pattern::WITHDRAW_REFUSED` directly. Every withdraw and deposit example in the file has that same blind spot, if you want to take it further than this PR.

### Open question 2: is `location == false` reachable? — treating it as reachable, fixed in `01ae5cf6`

Rather than reason about whether an unsurveyed room can be observed here, I guarded it — the failure is silent and the guard is one line:

```ruby
location = Room.current&.location
return nil unless location.is_a?(String)
return nil if location.empty?
```

Worth noting the old failure mode was worse than "returns nil": `false.to_s` is `"false"`, which passes `empty?` and then matches no bank, so `account[:balance]` quietly fell back to `0` even when the listing did carry a balance for this town. A caller checking `Bank.balance` before a withdraw would see 0 and skip it.

### Verification

Three commits, split per fix, each passing standalone (41 / 46 / 47 bank examples) so the history bisects. Four of the five new examples fail against the previous code.

| | Examples | Failures |
|---|---|---|
| `4bb4f729` (previous head) | 7192 | 60 |
| `01ae5cf6` | 7200 | 56 |

The 4 resolved are the bank ones above; the remaining 56 are pre-existing and unrelated (frontend launcher, wine, session launcher). rubocop clean on both touched files.

In-game coverage is unchanged from the PR description — the two WITHDRAW refusals above were confirmed live, Pinefar and the F2P cap dance remain spec-only.

Nisugi:
Correction on finding 3 — I was wrong to decline it. Fixed in `8939f313`.

I justified leaving the duplication by saying the two commands "genuinely differ (`silvers` vs `silver`) and both spellings are pinned by tests." That premise doesn't hold. Both spellings go to the same verb, at the same teller, in the same situation — the game takes either. And the specs pinning both were written from the implementation, so they recorded the accident rather than any game-side requirement. Citing them as evidence of intent was circular.

So: `withdraw` now sends `silver` (what `withdraw_silver` already used, and what I'd confirmed in game) and routes its coin branch through `withdraw_silver`. The refusal check and the silver-extraction regex now live in one place, which was your point.

```ruby
got = withdraw_silver(amount)
Currency.refresh
got
```

On the parsing side the word is `silvers?` wherever a figure is read, so a reply in either form still matches. Worth noting `ACCOUNT_MAX` was the one pattern that actually *required* the plural — `/a maximum of (?<silver>[\d,]+) silvers/` — and would have missed a singular. That's a latent bug this pass happens to close.

The Pinefar branch keeps its own wording (`ask banker for N silvers` is a different verb) and returns early rather than sharing the extraction, since that banker confirms without echoing a figure — the same limitation noted on `deposit`.

47 bank examples, 98 across bank/currency/infomon, rubocop clean.

mrhoribu (approved):
Reviewed the full bank/currency change plus the follow-up fixes.

One Major finding from the initial pass (`Bank.withdraw_f2p` never attempted to drain the account balance, and reported `0` instead of `nil`, whenever the stow container held no notes) and two open questions (missing `NO_ACCESS` on the withdraw wait pattern; `Room#location` possibly being `false` on an unsurveyed room) are all fixed as of `8939f313`, each with a new spec pinning the fix down. Verified by tracing the fixed code directly, not just reading the commit messages.

- `bundle exec rspec`: 7,200 examples, 0 failures
- `rubocop`: no offenses on any touched file

Looks good to merge.

PR #1582, opened 2026-09-10
fix(gs): Injured cache lives on the class and keys on an injury fingerprint

Nisugi:
## Summary

Two bugs made every `Injured.able_to_*?` predicate unusable:

- **First call raised `NoMethodError`.** The cache ivars and `@cache_mutex` were assigned inside `class << self`, so they landed on the singleton class. The methods read them on `Injured` itself, where the mutex was `nil`, so `get_injury_data` blew up on `nil.synchronize`.
- **Cache never invalidated.** The cache key was the live `XMLData.injuries` hash, which `XMLParser` mutates in place. The stored key and the current key were always the same object, so after the first fill a healthy character kept returning `true` for casting after taking a rank 3 head wound, and a cached rejection survived healing.

## Fix

- Move the constant, cache ivars, and mutex into the class body so they live on `Injured`.
- Key the cache on a frozen `[part, wound, scar]` fingerprint built from `XMLData.injuries`, so in-place parser updates invalidate it.
- Keep the cache rather than removing it: each miss calls `Scars.all_scars`, which toggles the injury mode and costs two `_injury` round trips.
- Document that the predicates call `fix_injury_mode` and can block waiting on the game. They are not instantaneous checks.

Known blind spot, noted in a comment: the fingerprint is built in injury mode 2, where a scar underneath an active wound is not visible, so such a scar changing will not trigger a refetch.

## Test plan

- [x] New `spec/lib/gemstone/injured_spec.rb`: mutex location, in-place injury update flips a cached `true`, healing flips a cached `false`, cache hit vs refetch counted via `all_scars`, rank 1 scar ignored / rank 2 not, Sigil of Determination bypasses rank 2 but not rank 3. All `fix_injury_mode` calls are stubbed so no `_injury` commands are sent.
- [x] `bundle exec rspec spec/lib/gemstone/injured_spec.rb spec/lib/games_spec.rb`: 170 examples, 0 failures
- [x] `bundle exec rubocop` on both files: no offenses
- [ ] In-game: call `Injured.able_to_cast?` before and after taking a head wound and confirm it flips

### DIFF
--- a/lib/gemstone/injured.rb
+++ b/lib/gemstone/injured.rb
@@ -4,25 +4,38 @@ module Lich
   module Gemstone
     # Injured class for checking ability to perform actions based on wounds and scars
     class Injured < Gemstone::CharacterStatus
-      class << self
-        BODY_PART_GROUPS = {
-          eyes: %i[leftEye rightEye],
-          arms: %i[leftArm rightArm],
-          hands: %i[leftHand rightHand],
-          legs: %i[leftLeg rightLeg],
-          feet: %i[leftFoot rightFoot],
-          head_and_nerves: %i[head nsys]
-        }.freeze
+      BODY_PART_GROUPS = {
+        eyes: %i[leftEye rightEye],
+        arms: %i[leftArm rightArm],
+        hands: %i[leftHand rightHand],
+        legs: %i[leftLeg rightLeg],
+        feet: %i[leftFoot rightFoot],
+        head_and_nerves: %i[head nsys]
+      }.freeze
 
-        # Cache variables
-        @injury_cache_key = nil
-        @wounds_cache = nil
-        @scars_cache = nil
-        @cache_mutex = Mutex.new
+      # Cache variables. These must live on Injured itself (the class body),
+      # not inside `class << self`, where they would land on the singleton
+      # class and the methods below would read nil.
+      @injury_cache_key = nil
+      @wounds_cache = nil
+      @scars_cache = nil
+      @cache_mutex = Mutex.new
+
+      class << self
+        # Build an immutable fingerprint of the current injury state.
+        # XMLData.injuries is mutated in place by the parser, so the hash itself
+        # cannot serve as a cache key: it would always compare equal to itself.
+        # Note: this runs in injury mode 2 ('both'), where a scar underneath an
+        # active wound is not visible, so such a scar changing will not invalidate
+        # the cache. That is a narrow case and acceptable given each cache miss
+        # costs two _injury round trips in Scars.all_scars.
+        def injury_fingerprint
+          XMLData.injuries.map { |part, v| [part, v['wound'], v['scar']] }.freeze
+        end
 
         # Get cached or fresh injury data
         def get_injury_data
-          current_key = XMLData.injuries
+          current_key = injury_fingerprint
 
           # Fast path: return cached data if key hasn't changed
           if @injury_cache_key == current_key && @wounds_cache && @scars_cache
@@ -77,6 +90,11 @@ module Lich
         def bypasses_injuries?
           Effects::Buffs.active?("Sigil of Determination")
         end
+
+        # NOTE: every able_to_*? predicate calls fix_injury_mode('both') and, on a
+        # cache miss, Scars.all_scars toggles the injury mode. Each of those can
+        # send an _injury command and block for up to 7.5 seconds waiting on the
+        # game. These are not instantaneous checks; avoid calling them in tight loops.
 
         # Check if character is able to cast spells
         def able_to_cast?
--- a/lib/gemstone/injured.rb
+++ b/lib/gemstone/injured.rb
@@ -92,9 +92,11 @@ module Lich
         end
 
         # NOTE: every able_to_*? predicate calls fix_injury_mode('both') and, on a
-        # cache miss, Scars.all_scars toggles the injury mode. Each of those can
-        # send an _injury command and block for up to 7.5 seconds waiting on the
-        # game. These are not instantaneous checks; avoid calling them in tight loops.
+        # cache miss, Scars.all_scars toggles the injury mode to 'scar' and back.
+        # Each of those sends an _injury command and waits up to 7.5 seconds for the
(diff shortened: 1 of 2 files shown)


mrhoribu (approved):
# PR Review: fix(gs): Injured cache lives on the class and keys on an injury fingerprint (elanthia-online/lich-5#1582)

**Reviewed:** 2 files, +133/-15 at `fc430ce0` on 2026-09-11. Based on: full clone (branch checked out via `gh pr checkout`-equivalent fetch), with `bundle exec rspec` and `bundle exec rubocop` run locally against the PR branch.
**Verdict:** Approve
**Summary:** Fixes two real, independently-verified bugs in `Lich::Gemstone::Injured` — cache ivars landing on the singleton class instead of the class (causing `nil.synchronize` on first call) and a cache key that was a live, mutated-in-place reference to `XMLData.injuries` (so the cache could never invalidate). Both root causes were reproduced standalone with plain Ruby to confirm the mechanism, not just the symptom, and both fixes are correct. No callers exist elsewhere in the repo yet, so blast radius is effectively zero.

## Blockers

None.

## Major

None.

## Minor

None.

## Nits

### 1. "up to 7.5 seconds" undercounts the worst case for a single cache miss — `lib/gemstone/injured.rb:94-97`

```ruby
# NOTE: every able_to_*? predicate calls fix_injury_mode('both') and, on a
# cache miss, Scars.all_scars toggles the injury mode. Each of those can
# send an _injury command and block for up to 7.5 seconds waiting on the
# game. These are not instantaneous checks; avoid calling them in tight loops.
```

`fix_injury_mode` waits `150.times { sleep 0.05; ... }` — 7.5s per call (`lib/games.rb:1249-1264`). `Scars.all_scars` calls it twice (`fix_injury_mode('scar')`, then `fix_injury_mode('both')` in its `ensure`, `lib/gemstone/scars.rb:98-106`), so a single cache-miss call chain can legitimately block up to ~15s if both round trips time out, not 7.5s. The comment says "each... can block for up to 7.5 seconds," which is technically true per-call but reads as if 7.5s is the ceiling for the whole predicate call. Worth a word change (e.g. "up to 15 seconds total") so a reader doesn't budget for the smaller number.

## Open questions

None outstanding — the two grounding questions that mattered (does the ivar-placement bug reproduce, does the cache actually fail to invalidate) were both settled by reproducing the underlying Ruby scoping behavior directly rather than trusting the PR description.

## What this PR gets right

- **Fixes are grounded in actual Ruby semantics, not just described.** Reproduced both bugs in isolation:
  - Ivars assigned inside `class << self` at the class-body level land on the singleton class; methods reading them via `self` (the class itself) see `nil`. Confirmed: `@mutex.synchronize` raises `NoMethodError: undefined method 'synchronize' for nil` under the old structure, and returns cleanly once the ivars move to the class body — exactly as claimed.
  - `Foo::BAR` raises `NameError: uninitialized constant Foo::BAR` when `BAR` is defined inside `class << self`, confirming `BODY_PART_GROUPS` was previously inaccessible as `Injured::BODY_PART_GROUPS` from outside the class (nobody currently does this, but the new placement is strictly more correct).
- **The new cache key is actually sound.** `XMLData.injuries` (`lib/common/xmlparser.rb:137`) is initialized once with a fixed 16-entry key set and only ever has its nested `'wound'`/`'scar'` values mutated in place (`lib/common/xmlparser.rb:809-876`) — keys are never added or removed at runtime. That means `injury_fingerprint`'s `.map` over the hash produces a stable, comparably-ordered array every call, so `==` between two frozen fingerprint arrays is a correct staleness check.
- **The disclosed blind spot is real and correctly scoped.** In injury mode 2 ('both'), a scar under an active wound is not reported by the parser, so the fingerprint can't see it change. This is called out explicitly in the code and PR body rather than silently accepted, and is a reasonable tradeoff given the two-round-trip cost of a full rescan.
- **Test suite exercises the actual production mutation pattern.** The regression tests mutate `XMLData.injuries['head']['wound'] = 3` in place — the same way `XMLParser` does — rather than reassigning `XMLData.injuries` wholesale, so they genuinely exercise the fingerprint-invalidation path the bug was in, not a weaker proxy for it.
- **Double-checked locking around the mutex is implemented correctly**: the pre-lock fast-path check and the post-lock recheck both compare against the same `current_key` local, so a second thread that loses the race to populate the cache doesn't refetch.
- **No blast radius risk.** `Injured` has no callers anywhere else in the repo yet (`grep` across the tree turns up only the file's own internal uses of `BODY_PART_GROUPS`, and the class isn't referenced by dr-scripts/scripts since those are DR-only) — this is a self-contained fix to code nothing outside this file yet depends on.

## Coverage notes

- Access level: full clone, PR branch checked out via `git fetch origin pull/1582/head` in a scratch clone (the local `lich-5` working directory had unrelated uncommitted changes on a different branch, so a separate clone was used instead of checking out in place).
- `bundle exec rspec spec/lib/gemstone/injured_spec.rb spec/lib/games_spec.rb` run locally: 170 examples, 0 failures — matches CI.
- `bundle exec rubocop lib/gemstone/injured.rb spec/lib/gemstone/injured_spec.rb` run locally: no offenses — matches CI.
- All GitHub Actions checks (Rubocop, both test jobs, PR title validation) pass; no prior review comments existed on the PR.
- CodeRabbit CLI is authenticated for this account but returned `Error: Rate limit exceeded` (free plan, no seat assigned) on `coderabbit review --agent --base main --committed`; no CodeRabbit pass was available to cross-check against.
- Not independently verified in-game (the author's own test plan marks this unchecked too) — the fix's correctness rests on the reproduced Ruby scoping behavior and the unit tests, not a live character session.
- Noted but not raised as a finding: `able_to_*?` methods call `fix_injury_mode("both")` *before* acquiring `get_injury_data`'s mutex, while `Scars.all_scars` (inside the mutex) temporarily switches mode to `'scar'` and back. Two threads calling different `able_to_*?` methods concurrently could theoretically interleave `_injury` mode-switch commands outside the mutex's protection. This pattern is unchanged by this PR (the `fix_injury_mode("both")` calls at the top of each predicate predate it) and is out of scope for a cache-correctness fix.

PR #1583, opened 2026-09-10
feat(gs): PSM command and results_regex readers, factored out of use

Nisugi:
## Summary

Each PSM category's `.use` built its command string and result regex inline. This exposes both as readers and has `.use` call them, so the PSM modules stay the one source of truth for usage strings and result patterns while a caller that sends and confirms on its own terms (an engine with its own roundtime and timeout discipline) no longer has to copy the tables.

**Shared, in `PSMS`**
- `PSMS.command(verb, usage, target, forcert_count:)`: verb plus usage, a GameObj or Integer target as `#id`, a String target as given, FORCERT when the count is above zero, a bare command when the verb is nil.
- `PSMS.results_regex(name, *patterns, results_of_interest:)`: the shared failures, the "X what?" and cooldown refusals for that name, the technique's own patterns, and any extras.

**Per category**, `CMan`, `Armor`, `Feat`, `Shield`, `Warcry`, `Weapon`:
- `.command(name, target = "", forcert_count: 0)` returns the exact string `.use` would send (e.g. `cman bullrush #12345`, `guard Dissonance`).
- `.results_regex(name, results_of_interest: nil)` returns the union `.use` waits on. Armor keeps its "not wearing any armor" line; Weapon's assault techniques wait on the assault line rather than a result line and roundtime, as before.

`.use` behaviour is unchanged: same availability checks, RT waits, `dothistimeout` and stun retry, and Weapon's assault loop. Feat's bare `guard` / `protect` verbs are now a named constant.

## Test plan

- [x] `spec/lib/gemstone/psms/command_spec.rb` (new): the two helpers plus one real technique per category against the registry data
- [x] `psms_spec.rb` and `qstrike_spec.rb` still green (84 examples total)
- [x] rubocop clean on touched files

### DIFF
--- a/lib/gemstone/psms.rb
+++ b/lib/gemstone/psms.rb
@@ -218,6 +218,50 @@ module Lich
         /^You can't reach .+!$/,
         / attempting to .+ would be a rather awkward proposition\.$/,
       )
+
+      # The command a technique is sent with, as each category's +use+ sends
+      # it: the verb, the technique's usage word, an optional target, and
+      # FORCERT. Exposed so a script that sends and confirms on its own
+      # terms (an engine with its own roundtime and timeout discipline)
+      # still gets the exact command +use+ would send.
+      #
+      # @param verb [String, nil] "cman", "shield", ... or nil for a bare command (FEAT GUARD)
+      # @param usage [String] the technique's usage word
+      # @param target [String, Integer, GameObj] a GameObj or id is sent as #id; a String as given
+      # @param forcert_count [Integer] more than 0 appends FORCERT
+      # @return [String]
+      #
+      # @example
+      #   PSMS.command("cman", "bullrush", GameObj.targets.first)  # => "cman bullrush #12345"
+      def self.command(verb, usage, target = "", forcert_count: 0)
+        cmd = verb.nil? || verb.to_s.empty? ? usage.to_s : "#{verb} #{usage}"
+        if target.is_a?(GameObj)
+          cmd += " ##{target.id}"
+        elsif target.is_a?(Integer)
+          cmd += " ##{target}"
+        elsif target.to_s != ""
+          cmd += " #{target}"
+        end
+        cmd += " forcert" if forcert_count > 0
+        cmd
+      end
+
+      # Every line that answers a technique command: the shared failures, the
+      # "X what?" and cooldown refusals for this technique, its own result
+      # messaging, and any extra patterns. This is the regex each category's
+      # +use+ waits on, so a caller confirming the command itself matches the
+      # same lines +use+ would.
+      #
+      # @param name [String] the technique name as the caller gave it
+      # @param patterns [Array<Regexp, nil>] the technique's result regex and friends; nils are skipped
+      # @param results_of_interest [Regexp, nil] extra lines the caller wants to see
+      # @return [Regexp]
+      def self.results_regex(name, *patterns, results_of_interest: nil)
+        parts = [FAILURES_REGEXES, /^#{name} what\?$/i, /^#{name} is still in cooldown\./i]
+        parts.concat(patterns.compact)
+        parts << results_of_interest if results_of_interest.is_a?(Regexp)
+        Regexp.union(*parts)
+      end
     end
   end
 end
--- a/lib/gemstone/psms/armor.rb
+++ b/lib/gemstone/psms/armor.rb
@@ -210,36 +210,13 @@ module Lich
       def Armor.use(name, target = "", results_of_interest: nil, forcert_count: 0)
         return unless Armor.available?(name, forcert_count: forcert_count)
 
-        name_normalized = PSMS.name_normal(name)
-        technique = @@armor_techniques.fetch(PSMS.find_name(name_normalized, "Armor")[:long_name])
-        usage = technique[:usage]
-        return if usage.nil?
+        usage_cmd = Armor.command(name, target, forcert_count: forcert_count)
+        return if usage_cmd.nil?
 
-        in_cooldown_regex = /^#{name} is still in cooldown\./i
+        results_regex = Armor.results_regex(name, results_of_interest: results_of_interest)
 
-        results_regex = Regexp.union(
-          PSMS::FAILURES_REGEXES,
-          /^#{name} what\?$/i,
-          in_cooldown_regex,
-          technique[:regex],
-          /^Roundtime: [0-9]+ sec\.$/,
-          /^\w+ [a-z]+ not wearing any armor that you can work with\.$/
-        )
-
-        results_regex = Regexp.union(results_regex, results_of_interest) if results_of_interest.is_a?(Regexp)
-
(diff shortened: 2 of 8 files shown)


mrhoribu (commented):
# PR Review: feat(gs): PSM command and results_regex readers, factored out of use (elanthia-online/lich-5#1583)

**Reviewed:** 8 files, +324/-165, at `d781359c` on 2026-09-11. Based on: full clone (`gh pr checkout`, `main...feat/psm-command`), full spec suite run, rubocop run.
**Summary:** Each PSM category's `.use` inlined its command string and result regex; this PR extracts both into `.command`/`.results_regex` readers (shared logic centralized in `PSMS.command`/`PSMS.results_regex`), so an external caller can build the same command and matcher `.use` would without duplicating the technique tables. The refactor is faithful line-for-line to the six categories' prior behavior — verified by diffing each `.use` body against its new `.command`/`.results_regex` pair and by running the full spec suite (644 examples, 0 failures) plus the 11 new examples, all green. One real asymmetry survives in the new public surface: `Warcry.command` skips the input validation every sibling `.command` performs.

## Minor

### 1. `Warcry.command` doesn't validate the technique name the way every sibling `.command` does — `lib/gemstone/psms/warcry.rb:208-210`

```ruby
def Warcry.command(name, target = "", forcert_count: 0)
  PSMS.command("warcry", PSMS.name_normal(name), target, forcert_count: forcert_count)
end
```

`CMan.command`, `Armor.command`, `Shield.command`, and `Feat.command` all resolve the technique first — `@@xxx.fetch(PSMS.find_name(PSMS.name_normal(name), "Category")[:long_name])` — so an unknown name raises `KeyError` immediately (`PSMS.find_name` returns `nil` on no match; `PSMS.name_normal` at `lib/util/util.rb:59-66` does plain string munging with no validation of its own, and `Hash#fetch` on the resulting `nil` key raises since none of the technique hashes carry a `nil` key). `Warcry.command` never resolves against `@@warcries` at all — it hands `PSMS.name_normal(name)` straight to `PSMS.command` as the literal command word — because warcry techniques carry no `:usage` field (`warcry.rb:22-68`), so building the string genuinely doesn't need the table.

The result: `Warcry.command("hollar")` (a typo) silently returns `"warcry hollar"` instead of raising, while `CMan.command("bullrus")` raises `KeyError` for the same class of mistake. This doesn't regress `Warcry.use` — that path still hits the fetch-and-raise inside `Warcry.results_regex` a few lines later (`warcry.rb:182-185`), so a bad name passed to `.use` still crashes, just one call later than before. But it does mean the very use case this PR is written for — "a caller that sends and confirms on its own terms" by calling `.command` directly without going through `.use`'s checks — gets silently wrong output for warcries specifically, while getting a clear crash for every other category. Given the whole point of exposing these readers is to let new callers build on them directly, this inconsistency is likely to surprise the first person who typos a warcry name in that position.

Confidence: confirmed (traced through `PSMS.find_name`, `Hash#fetch`, and the `NilClass` `method_missing` patch that makes `nil[:long_name]` return `nil` rather than raise, at `lib/common/class_exts/nilclass.rb`).

**Fix:** have `Warcry.command` resolve `PSMS.find_name(PSMS.name_normal(name), "Warcry")` (or fetch `@@warcries`) before building the string, matching the other five categories, even though the technique hash isn't otherwise needed for the command word.

## Nits

None — the diff is otherwise a clean, mechanical extraction. Comment placement (e.g. moving the "with forcert we don't want to wait for rt" comment above the `unless forcert_count > 0` guard instead of dangling off the old `else`) is a small readability improvement, not a nit against it.

## What this PR gets right

- **Faithful extraction.** Every one of the six categories' `.command`/`.results_regex` pair was checked line-by-line against the code it replaced, including the two trickier cases: Feat's bare `guard`/`protect` verb (now `BARE_COMMANDS`) and Weapon's assault-technique branch, which never appended FORCERT even in the original code — the second commit (`d781359c`) specifically fixes the extraction to preserve that, and it's covered by a dedicated assertion in the new spec.
- **The `results_of_interest` type check quietly tightens for `Weapon`.** The original `Weapon.use` unioned `results_of_interest` on a bare truthy check (`if results_of_interest`), while every sibling used `if results_of_interest.is_a?(Regexp)` — an existing inconsistency, given `Weapon.use`'s own doc comment already promised `[Regexp, nil]`. Routing Weapon through the shared `PSMS.results_regex` (which does the `.is_a?(Regexp)` check) silently brings it in line with its own documented contract and its siblings. Worth knowing this happened, since it's a narrowing behavior change buried inside an otherwise mechanical refactor, but it fixes a real inconsistency and matches the documented types on both sides.
- **New spec file is well-targeted.** `spec/lib/gemstone/psms/command_spec.rb` exercises the shared helpers directly and one real technique per category against the actual registry data (not mocks), including the FORCERT-omitted-for-assault case. All 11 examples pass, and the full `spec/lib/gemstone/` suite (644 examples) is unaffected.
- **No blast radius beyond the six files.** Grepped the whole checkout for other callers of `CMan.use`/`.command`/`.results_regex` (and the other five categories) — none exist inside lich-5 itself, so this is a pure internal refactor plus new API surface, not a behavior change for any existing internal caller.

## Coverage notes

- Full local clone via `gh pr checkout` against `elanthia-online/lich-5`, diffed `main...feat/psm-command`; read every changed file in full (not just the diff hunks) plus the unrelated call sites via repo-wide grep.
- Ran `bundle exec rspec spec/lib/gemstone/psms/command_spec.rb` (11/11 pass) and the full `spec/lib/gemstone/` directory (644/644 pass).
- Ran `bundle exec rubocop` against all 8 changed files: no offenses (ruby 4.0.5, matching the repo's `TargetRubyVersion: 4.0`).
- CodeRabbit CLI is installed and authenticated but returned `rate_limit` (0 of 3 included reviews left, on-demand not available for this account/org) before producing any findings — no CodeRabbit input in this review.
- Did not attempt to exercise `.use`/`.command` against a live game session (not possible outside the Lich runtime); correctness here rests on the spec suite plus manual trace against the pre-refactor code, not a live in-game confirmation.

mrhoribu (approved):
Reviewed the extraction of PSM `.command`/`.results_regex` readers from `.use` across CMan/Armor/Feat/Shield/Warcry/Weapon.

- Diffed each category's new `.command`/`.results_regex` against the code it replaced — faithful extraction, including the trickier cases (Feat's bare `guard`/`protect`, Weapon's FORCERT-omitted-on-assault path).
- Ran the full `spec/lib/gemstone/` suite (645/645 passing) and `rubocop` on all changed files (clean).
- Flagged one asymmetry in the first pass: `Warcry.command` skipped the input validation every sibling `.command` performs, so a typo'd warcry name would silently build a bogus command instead of raising. Fixed in a56f9dfe (`@@warcries.fetch(...)` added purely for its raising side effect, plus a new spec covering the unknown-name case) — re-verified, no remaining concerns.

PR #1584, opened 2026-09-10
feat(gs): Lich::Gemstone::Fog, the ways home on the spell and society readers

Nisugi:
## Summary

Getting home from the field is something any script may need, not a hunting rule, so it belongs in core. Today the only implementation is bigshot's `fog_return` / `fog_return_spirit` / `fog_return_voln`, which a script has to either run bigshot to get or copy. This is that routine in Lich, on the readers that already exist.

`Lich::Gemstone::Fog`:

- `known?(method)`, `available?(method)`, `available` for `:spirit_guide` (130), `:symbol_of_return` (Voln), `:travelers_song` (1020), `:sigil_of_escape` (Sunfist) and `:familiar_gate` (930), read from `Spell` and the `Societies::OrderOfVoln` / `Societies::GuardiansOfSunfist` readers.
- `return(method, rift:, resting_room:)`: waits out roundtime, pulses mana for a known spell it cannot afford, sends the method, and answers whether the room changed rather than assuming it did. Keeps bigshot's Spirit Guide <-> Symbol of Return fallback and the second cast when the first lands in the Rift with the destination elsewhere.
- Takes bigshot's `fog_return` numbering (1-5) as well as names, so a profile value can be passed straight through. 6 (custom commands) is the script's own business and is not here.

Policy stays with the caller: which method, whether to fog at all, and any custom command list.

Loaded from the GS branch of `gameloader.rb` after `society.rb`.

## Test plan

- [x] `bundle exec rspec spec/lib/gemstone/fog_spec.rb` (13 examples: normalize, the readers, each of the five, both fallbacks, the Rift second cast, the mana pulse, an unknown method)
- [x] rubocop clean
- [x] full suite: the 57 failures are the same frontend, launcher and setup-files examples that fail on a clean `origin/main` worktree here (Windows environment), none in this change
- [ ] In game: `Lich::Gemstone::Fog.available`, then `Lich::Gemstone::Fog.return(:spirit_guide)` from the field

### DIFF
--- a/lib/common/gameloader.rb
+++ b/lib/common/gameloader.rb
@@ -27,6 +27,7 @@ module Lich
         require File.join(LIB_DIR, 'attributes', 'skills.rb')
         require File.join(LIB_DIR, 'attributes', 'enhancive.rb')
         require File.join(LIB_DIR, 'gemstone', 'society.rb')
+        require File.join(LIB_DIR, 'gemstone', 'fog.rb')
         require File.join(LIB_DIR, 'gemstone', 'infomon', 'status.rb')
         require File.join(LIB_DIR, 'gemstone', 'experience.rb')
         require File.join(LIB_DIR, 'attributes', 'spellsong.rb')
--- a/lib/gemstone/fog.rb
+++ b/lib/gemstone/fog.rb
@@ -204,15 +204,24 @@ module Lich
         moved_from?(start)
       end
 
-      # Where we are, as the server reports it: the room counter (which
-      # steps on every move) and the server room id. Both are set for an
-      # unmapped room, where Room.current is nil and a map id would compare
-      # nil to nil and call a real move a failure.
+      # Where we are, as the server reports it: the server room id, which
+      # is set for an unmapped room too (Room.current is nil there, and a
+      # map id would compare nil to nil and call a real move a failure),
+      # with the room counter alongside for the rare stream that carries
+      # no id.
       # @api private
-      def self.here = [XMLData.room_count, XMLData.room_id]
+      def self.here = { id: XMLData.room_id.to_s, count: XMLData.room_count }
 
+      # A move is a different server room id. The counter alone is not
+      # evidence: it steps on every room refresh, moved or not, so it is
+      # consulted only when an id is missing on either side.
       # @api private
-      def self.moved_from?(start) = here != start
+      def self.moved_from?(start)
+        now = here
+        return now[:id] != start[:id] unless now[:id].empty? || start[:id].empty?
+
+        now[:count] != start[:count]
+      end
 
       # The map id, for the Rift check only.
       # @api private
--- a/lib/gemstone/fog.rb
+++ b/lib/gemstone/fog.rb
@@ -1,0 +1,223 @@
+# frozen_string_literal: true
+
+# Fog: the ways home. Spirit Guide (130), Symbol of Return (Voln), Traveler's
+# Song (1020), Sigil of Escape (Sunfist) and Familiar Gate (930) each take a
+# character out of the field to a known room. Every hunting script carries a
+# copy of the same routine (bigshot's fog_return family, and the copies in the
+# scripts that grew out of it); this is that routine once, on the spell and
+# society readers, with the answer confirmed on the room changing rather than
+# assumed.
+#
+#   Lich::Gemstone::Fog.available                        # => [:spirit_guide, :symbol_of_return]
+#   Lich::Gemstone::Fog.return(:spirit_guide)            # => true when the room changed
+#   Lich::Gemstone::Fog.return(2, rift: true, resting_room: 4)   # bigshot's numbering
+#
+# Policy stays with the caller: which method a profile picks, whether to fog at
+# all, and any custom command list are the script's. What each method is, what
+# it costs, how it is sent and how it is confirmed are here.
+module Lich
+  module Gemstone
+    module Fog
+      # The Rift's exit: a fog that lands here needs a second cast to go on
+      # (bigshot fog_return_spirit, fog_return_voln).
+      RIFT_ROOM = 2635
+
+      # Seconds to wait for the room to change after a send. A fog resolves
+      # in one server pulse; a Traveler's Song walk takes longer.
+      CONFIRM_TIMEOUT = 8
+
+      METHODS = %i[spirit_guide symbol_of_return travelers_song sigil_of_escape familiar_gate].freeze
+
+      # bigshot's fog_return setting (1-5; 6 is the profile's own commands)
+      NUMBERS = { 1 => :spirit_guide, 2 => :symbol_of_return, 3 => :travelers_song, 4 => :sigil_of_escape, 5 => :familiar_gate }.freeze
+
+      SPELLS = { spirit_guide: 130, travelers_song: 1020, familiar_gate: 930 }.freeze
+
+      # @param method [Symbol, Integer, String] a METHODS name or bigshot's number
+      # @return [Symbol, nil]
+      def self.normalize(method)
+        return NUMBERS[method] if method.is_a?(Integer)
+        return NUMBERS[method.to_i] if method.to_s =~ /\A\d+\z/
+
+        sym = method.to_s.downcase.to_sym
+        METHODS.include?(sym) ? sym : nil
+      end
+
+      # Does this character have the method at all.
+      def self.known?(method)
+        case normalize(method)
+        when :spirit_guide, :travelers_song, :familiar_gate then spell_known?(SPELLS[normalize(method)])
+        when :symbol_of_return then voln&.known?('return') || false
+        when :sigil_of_escape then sunfist&.known?('escape') || false
+        else false
+        end
+      end
+
+      # Known and affordable right now.
+      def self.available?(method)
+        case normalize(method)
+        when :spirit_guide, :travelers_song, :familiar_gate
+          num = SPELLS[normalize(method)]
+          spell_known?(num) && Spell[num].affordable?
+        when :symbol_of_return then voln&.available?('return') || false
+        when :sigil_of_escape then sunfist&.available?('escape') || false
+        else false
+        end
+      end
+
+      # @return [Array<Symbol>] every method the character can use right now
+      def self.available
+        METHODS.select { |m| available?(m) }
+      end
+
+      # Go home by +method+. Waits out roundtime first, pulses mana when the
+      # spell is known but unaffordable, and answers whether the room changed.
+      #
+      # Spirit Guide and Symbol of Return fall back to each other when the
+      # first does not move us, and cast a second time when the first cast
+      # lands in the Rift and the destination is elsewhere - bigshot's rules,
+      # kept.
+      #
+      # @param method [Symbol, Integer, String]
+      # @param rift [Boolean] the destination is reached through the Rift
+      # @param resting_room [Integer, nil] with +rift+, the room the fog is for
+      # @return [Boolean] whether the room changed
+      def self.return(method, rift: false, resting_room: nil)
+        method = normalize(method)
+        return false if method.nil?
+
+        start = room_id
+        sleep 0.5
+        waitcastrt?
+        waitrt?
+        case method
+        when :spirit_guide then spirit_guide(rift: rift, resting_room: resting_room)
+        when :symbol_of_return then symbol_of_return(rift: rift, resting_room: resting_room)
+        when :travelers_song then travelers_song
+        when :sigil_of_escape then sigil_of_escape
+        when :familiar_gate then familiar_gate
+        end
+        moved_from?(start)
+      end
+
+      # --- the five ------------------------------------------------------------
+
+      # @api private
+      def self.spirit_guide(rift: false, resting_room: nil, from_voln: false)
+        start = room_id
+        pulse_mana(130)
+        if available?(:spirit_guide)
+          cast_and_settle(130)
+          second_cast_from_rift(rift, resting_room, start) { cast_and_settle(130) if available?(:spirit_guide) }
+        end
+        return true if moved_from?(start)
+
+        # 130 did not move us: Symbol of Return, once (fog_return_spirit)
+        symbol_of_return(rift: rift, resting_room: resting_room, from_spirit: true) if !from_voln && known?(:symbol_of_return)
+        moved_from?(start)
+      end
+
+      # @api private
+      def self.symbol_of_return(rift: false, resting_room: nil, from_spirit: false)
+        start = room_id
+        if known?(:symbol_of_return)
+          fput 'symbol of return'
+          wait_for_move(start)
+          second_cast_from_rift(rift, resting_room, start) { fput 'symbol of return'; wait_for_move(RIFT_ROOM) }
+        end
+        return true if moved_from?(start)
+
+        # the symbol did not move us: Spirit Guide, once (fog_return_voln)
+        spirit_guide(rift: rift, resting_room: resting_room, from_voln: true) if !from_spirit && known?(:spirit_guide)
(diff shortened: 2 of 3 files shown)


Nisugi:
Both findings fixed in the follow-up commit:

- **Rift second cast**: `spirit_guide` now pulses mana again before the second cast out of 2635, as bigshot's `fog_return_spirit` does. Spec: the first cast spends the mana, the pulse restores it, the second cast lands in the resting room.
- **Unmapped rooms**: movement is confirmed on the server's room counter and room id (`XMLData.room_count`, `XMLData.room_id`), both of which are set for an unmapped room; `Room.current.id` is kept for the Rift check only. Spec: a fog between two unmapped rooms answers true and does not fire the fallback.

15 examples, 0 failures; rubocop clean.

Nisugi:
Fixed in the follow-up commit. A move is now a change in the server room id (`XMLData.room_id`); the room counter is consulted only when an id is missing on either side, since it steps on every room refresh whether or not the character moved. Specs: a refresh of the same room is not a move and the fallback fog still fires; the counter stands in when there is no id. 17 examples, 0 failures; rubocop clean.

mrhoribu (changes requested):
# PR Review: feat(gs): Lich::Gemstone::Fog, the ways home on the spell and society readers (elanthia-online/lich-5#1584)

**Reviewed:** 3 files, +452/-0 at `2a01c272` on 2026-09-11. Based on: full clone (worktree checkout of the PR head against a freshly-fetched `main`), plus `lib/gemstone/societies/order_of_voln.rb`, `guardians_of_sunfist.rb`, `lib/common/xmlparser.rb`, `lib/common/map/map_gs.rb`, `lib/global_defs.rb`, `lich.rbw`, and `scripts/scripts/bigshot.lic` (the routine this PR ports) all read at head to verify call sites.
**Verdict:** Request changes
**Summary:** This adds `Lich::Gemstone::Fog`, a faithful core port of bigshot's `fog_return` family (Spirit Guide/Symbol of Return/Traveler's Song/Sigil of Escape/Familiar Gate), confirming success by watching the room change instead of assuming it. The five-method dispatch, the society-reader integration, and the Spirit Guide ↔ Symbol of Return fallback/Rift-recast logic are all correctly ported and well tested (17 examples, rubocop clean). Two real problems remain in the move-confirmation primitive the whole module leans on: the "no room id" fallback it documents can't actually occur given `XMLData`'s own contract, and the Traveler's Song path uses the same fixed confirmation timeout the code's own comment says is too short for that spell.

## Major

### 1. `moved_from?`'s "missing id" fallback is unreachable, and the case it was built for doesn't get the fallback — `lib/gemstone/fog.rb:207-224`

```ruby
def self.here = { id: XMLData.room_id.to_s, count: XMLData.room_count }

def self.moved_from?(start)
  now = here
  return now[:id] != start[:id] unless now[:id].empty? || start[:id].empty?

  now[:count] != start[:count]
end
```

The comment above `here` (`fog.rb:207-211`) says the counter is there "for the rare stream that carries no id," and `moved_from?`'s own comment says it's "consulted only when an id is missing on either side." Both describe `XMLData.room_id` as something that can be *missing*. It can't be: `lib/common/xmlparser.rb:144-146` initializes `@room_id = 0` with the comment *"Default 0 (not nil) so a read... is a valid 'no UID' value rather than a NoMethodError"* — 0 is the documented, permanent sentinel for "no UID," not an absence. `XMLData.room_id` is `attr_reader`-only (`xmlparser.rb:13`) and every assignment site converts through `.to_i` or `Digest::MD5...to_i(16)`, so it is always an `Integer`, never `nil`. Consequently `XMLData.room_id.to_s` is never `""`, `now[:id].empty?` is never true, and the counter-based branch in `moved_from?` can never execute in production — it is dead code, and the design goal it exists for ("confirm a move even when the room has no UID," `fog.rb:207-211`) is not actually met by it.

That goal still matters, because it's real: `0` (or, when `Lich::Claim::Lock` is owned — the normal case, since `Lich::Claim.lock` fires on every `<nav>` tag, `xmlparser.rb:401`) an `MD5.hexdigest([title, description, exits]).to_i(16)` hash (`xmlparser.rb:434-436`, `1273-1276`) is what a truly unmapped room's id becomes. Two different unmapped rooms with identical title/description/exit text — the exact ambiguity the `Lich::Claim` fuzzy-matching machinery exists to work around — hash identically, and `moved_from?` would then report "not moved" on a real move. The five fog destinations themselves are all fixed, mapped, real-UID rooms, so this is unlikely to fire on the *destination* side; the risk is on the *start* side (fogging out mid-hunt from an unmapped room) landing in another unmapped room the code can't currently distinguish — narrower than the doc comment implies, but real, and not what "missing" was supposed to cover.

The PR's own spec doesn't catch this because it can't reach it: `spec/lib/gemstone/fog_spec.rb:440-444` exercises the fallback by setting `XMLData.room_id = nil` directly, which is not a state the real `XMLData` ever enters (confirmed above), so the one test written for "no room id" tests a state that cannot occur, while the state that actually needs the fallback (collision between two non-empty ids) has no test.

**Confirmed:** the fallback path is unreachable given `XMLData`'s documented contract, and the accompanying test exercises an impossible state. **Suspected:** how often two visited unmapped rooms actually share identical text in live play — worth confirming with the in-game step the PR's own test plan left unchecked.

### 2. `CONFIRM_TIMEOUT` is fixed at 8s for every method, including the one the code says needs longer — `lib/gemstone/fog.rb:24-26`, `143-146`

```ruby
# Seconds to wait for the room to change after a send. A fog resolves
# in one server pulse; a Traveler's Song walk takes longer.
CONFIRM_TIMEOUT = 8
...
def self.travelers_song
  pulse_mana(1020)
  cast_and_settle(1020) if available?(:travelers_song)
end
```

`travelers_song` routes through `cast_and_settle` (`fog.rb:172-179`), which calls `wait_for_move` (`fog.rb:200-205`), which polls for exactly `CONFIRM_TIMEOUT` (8) seconds — the same constant used for Spirit Guide's instant fog. The comment two lines above the constant says outright that a Traveler's Song walk "takes longer" than the one-pulse fogs, but nothing in the code gives it a longer budget. If the walk legitimately takes more than 8 seconds (which the comment asserts), `Fog.return(:travelers_song)` returns `false` — "did not move" — while the character is still correctly en route, contradicting the documented `@return [Boolean] whether the room changed` contract on `.return` (`fog.rb:83`) for exactly the one method the author flagged as an outlier.

Notably, bigshot's own `fog_return` (`scripts/scripts/bigshot.lic:6478-6481`) never waits for the room to change after casting 1020 at all — it fires the cast and returns, sidestepping this entirely. This PR chose to add a uniform confirm-wait for every method, which is a reasonable improvement in principle, but Traveler's Song needs its own budget (or its own leg, mirroring bigshot's "don't wait" behavior) rather than sharing `CONFIRM_TIMEOUT` with instant fogs.

**Confirmed:** the code contradicts its own stated reasoning — self-contained in the diff, no external dependency needed to see it. **Open question below** on what timeout is actually correct, since that requires live timing.

## Nits

- `known?`/`available?` (`fog.rb:46-53`, `56-65`) each call `normalize(method)` twice — once in the `case`, once more to index `SPELLS`. Harmless (idempotent, cheap), but `case`'s captured value could be reused instead of recomputing it a line later.

## What this PR gets right

- The Spirit Guide ↔ Symbol of Return mutual fallback (`fog.rb:104-140`) correctly bounds itself with `from_voln`/`from_spirit`, matching bigshot's `from_voln`/`from_130` guards (`bigshot.lic:6419-6461`) line for line in intent — traced by hand, no infinite-recursion path exists.
- Using `Societies::OrderOfVoln.known?('return')` / `.available?('return')` instead of bigshot's raw `Spell[9825].known?` is a genuine improvement: verified against `lib/gemstone/societies/order_of_voln.rb:291,401-407,460-463` that this correctly folds in society membership and rank, which a bare spell-known check doesn't.
- `room_id` (map id, for the Rift check only) correctly uses `Room.current&.id` after the module's own comment (`fog.rb:207-209`) calls out the `nil == nil` trap that a naive comparison would fall into — the fix is applied where it's needed even though (per the finding above) an equivalent gap survives one level down in `moved_from?`.
- Bare `Room`, `Spell`, and `Societies::OrderOfVoln`/`GuardiansOfSunfist` constant references all resolve correctly from `Lich::Gemstone::Fog`'s nesting — verified against `lich.rbw:147` (`include Lich::Common`, making `Lich::Common::Room`/`Spell` reachable globally) and `lib/gemstone/society.rb:168` (`Societies` nested directly under `Lich::Gemstone`, resolved lexically). No `NameError` risk despite the mixed namespaces.
- Test suite is genuinely useful: 17 examples cover normalize, both readers, all five methods, both fallback directions, the Rift second-cast (including the mana-pulse-again case), and an unknown-method no-op — verified passing (`bundle exec rspec spec/lib/gemstone/fog_spec.rb`, 17 examples/0 failures) and rubocop-clean on this checkout.
- `gameloader.rb`'s require placement is safe and correctly ordered after `society.rb`.

## Open questions

- `fog.rb:26` — is 8 seconds actually too short for a live Traveler's Song walk to complete? The comment at `fog.rb:24-25` asserts it is longer than a fog's "one server pulse" but the code doesn't act on that. **Settles it:** the unchecked in-game test-plan item (`Lich::Gemstone::Fog.return(:spirit_guide)` from the field) extended to also try `:travelers_song` and timing how long the walk actually takes.
- `fog.rb:207-224` — how often, in practice, do two different rooms a character might fog between (start and destination) lack a UID and share identical title/description/exit text? If this is effectively never in real Voln-outpost/Sunfist-chapterhouse/etc. destinations, the dead fallback path is a documentation/test-accuracy issue rather than a live-correctness one, and can be downgraded accordingly.

## Coverage notes

- Full clone via `git worktree` against the PR head (`pr-1584-review`, commit `2a01c272`) and a freshly fetched `main` (merge-base `073d23a9`) — diff, blame, and full file contents all available, not diff-only.
- Ran the new spec (`bundle exec rspec spec/lib/gemstone/fog_spec.rb`): 17 examples, 0 failures. Ran `rubocop` on both changed Ruby files: no offenses.
- Cross-checked every external symbol the new code calls (`XMLData.room_count`/`.room_id`, `Room.current`/`.id`, `Spell[]`/`.known?`/`.affordable?`/`.cast`, `Societies::OrderOfVoln`/`GuardiansOfSunfist` and their `known?`/`available?`, and the four global script functions `waitrt?`/`waitcastrt?`/`fput`/`dothistimeout`) against their actual definitions on `main` — all resolve and match the assumed signatures, per a dedicated sub-agent pass plus my own follow-up reads of `xmlparser.rb`, `map_gs.rb`, `society.rb`, and `lich.rbw`.
- Compared the ported logic line-by-line against `scripts/scripts/bigshot.lic:6403-6509` (`mana_pulse`, `fog_return_spirit`, `fog_return_voln`, `fog_return`), the routine this PR is explicitly porting.
- Read the PR's existing review-comment history on GitHub: two earlier self-review rounds already fixed a missing second mana-pulse before the Rift recast, and switched the primary move check from `Room.current.id` to `XMLData.room_id` specifically to handle unmapped rooms. Finding 1 above is a residual gap in that same area that neither round caught — it requires knowing `XMLData.room_id`'s default-0/never-nil contract from `xmlparser.rb`, which isn't visible from the diff or from `fog.rb` alone.
- CodeRabbit CLI is authenticated but hit its usage-based rate limit for this org's plan during the review ("Rate limit exceeded... You've used all 3 included reviews currently available", ~39 min cooldown) — no CodeRabbit pass was obtained; this review's findings are from manual investigation only.
- Did not run the in-game verification step the PR's own test plan leaves unchecked; both findings above would benefit from it and neither can be settled from static review alone.

Nisugi:
Both majors and the nit are fixed in `60402a4a`. Thanks for the `xmlparser.rb` dig on the first one — the never-nil contract is not visible from `fog.rb`, and it turns the finding into a better fix than the one I would have written from the diff alone.

**1. The unreachable fallback.** Confirmed, and confirmed for the reason you give: `@room_id = 0` is a permanent sentinel (`xmlparser.rb:144-146`), `attr_reader`-only, every write through `.to_i`, so `.to_s` is never `""` and the counter branch was dead. The test that "covered" it set `XMLData.room_id = nil`, a state the parser never enters, so it was testing an impossibility.

Rather than delete the branch, I rebuilt it on the ambiguity that is actually real — the one you identified as the live risk. A room arriving with no UID gets `MD5(title, description, exits).to_i(16)` (`xmlparser.rb:434-436`), so two unmapped rooms whose text reads the same share an id, and a fog between them read as "did not move". The id stays the primary mark; when it is *unchanged* and above `MAX_ROOM_UID` (an MD5, not a real UID), the room counter decides, since it steps once per room stream. An unchanged real UID is still never a move, so the room-refresh case you and the earlier round both cared about still answers false and still fires the fallback.

Specs now reach the state that can occur instead of the one that can't: a fog between two same-text unmapped rooms confirms and fires no fallback, and a refresh of an unmapped room does not confirm and does fire it.

**2. The shared timeout.** Also yours — the comment asserted the song was the outlier and the code ignored it. `travelers_song` now waits on `TRAVELERS_SONG_TIMEOUT` (60) while the four fogs keep `CONFIRM_TIMEOUT` (8); `wait_for_move`/`cast_and_settle` take the budget as a keyword defaulting to the short one. Two specs assert which budget each leg asks for.

I went with a longer wait rather than bigshot's "don't wait at all" so `.return` keeps meaning the same thing for all five methods — a caller that gets `true` has actually arrived. Happy to switch to the bigshot behaviour if you'd rather the song be explicitly fire-and-forget.

**3. Nit.** `known?`/`available?` now bind `normalize`'s result once and index `SPELLS` with it.

**Open questions.** Both still stand and both are the in-game step:

- 60s for the song is an estimate, not a measurement. It is one constant if the real walk is longer.
- The same-text-unmapped-rooms collision rate in live play is still unmeasured. It no longer gates correctness the way it did — the code now handles the collision instead of documenting a fallback that couldn't run — but I haven't seen it fire in the field.

20 examples, 0 failures; rubocop clean on both files.

mrhoribu (approved):
Re-reviewed at `60402a4a`. Both issues from my earlier pass are fixed, and I verified the fixes against the actual diff rather than just the commit message:

- **Traveler's Song timeout:** `wait_for_move`/`cast_and_settle` now take a `timeout:` keyword, and `travelers_song` passes a new `TRAVELERS_SONG_TIMEOUT = 60` instead of sharing the 8s `CONFIRM_TIMEOUT` used by the instant fogs. New specs assert the right budget goes to the right leg.

- **Unmapped-room move detection:** `moved_from?` now distinguishes a real room UID (small integer) from xmlparser's MD5 stand-in for a room with no UID (a `MAX_ROOM_UID = 1_000_000_000` threshold — real UIDs run in the tens of thousands, e.g. the Rift is `2635`, so the margin is generous), and only falls back to the room counter when the id is an *unchanged* MD5 stand-in. That's exactly the gap I'd flagged: two different unmapped rooms with identical title/description/exit text hashing to the same id. New tests (`arrive_uidless`) cover that case directly, and the old test that asserted on a state `XMLData.room_id` can't actually reach (`= nil`) is gone.

Ran `bundle exec rspec spec/lib/gemstone/fog_spec.rb` (20 examples, 0 failures) and `rubocop` on both changed files — clean.

One narrow corner case left, not blocking: `uidless?(0)` is `false`, so a room still at xmlparser's raw pre-substitution `0` sentinel (rather than the MD5-substituted value) would be treated as a real UID. In practice this window is vanishingly narrow — the MD5 substitution runs on essentially every `<nav>`→`<compass>` cycle, so any room a script is actually standing in when it calls `Fog.return` has almost certainly already been through it. Worth a one-line follow-up if you want to close it out completely, but not a reason to hold this.

Approving.

PR #1585, opened 2026-09-10
fix(gs): infomon - a spell's start message refreshes a refreshable timer

Nisugi:
## Summary

The Infomon parser puts a spell up on its start message only when the spell is not already active (`spell.putup unless spell.active?`). When the game refreshes an effect that is still running, Lich keeps the old expiry. Effects the game lists in the spells window are corrected by ActiveSpell's duration sync, but timers that exist only through their start and end messages have no other source, and the effect list already marks which of those are refreshable (`span='refreshable'`); nobody was asking.

Where it showed: Briar Betrayer (9105) is a two-minute effect the game reports only through the BetrayerPanel, so Lich's timer runs from the RAISE pulse line. A raise inside the live window refreshes it in the game (no end line between, the end line two minutes after the last raise) while Lich's timer expired on the old schedule. On one character's logs since 2026-09-06 that happened 376 times, and 173 of those were followed within 30 seconds by a volley the script's `Spell[9105].active?` gate should have held.

The change: a start message for an active spell calls `putup` when the duration is refreshable, which resets it to the full duration. Anything else keeps the timer it has, exactly as before. Stackable durations are left alone here: those are real spells the window reports, and stacking them from the message as well would count a cast twice.

`data/effect-list.xml` in the scripts repo needs 9105's duration marked `span='refreshable'` to get this behaviour; that is a separate PR there.

## Test plan

- [x] `bundle exec rspec spec/lib/gemstone/infomon/spell_message_refresh_spec.rb` (3 examples: an inactive spell goes up, a refreshable one resets from near-expiry, a non-refreshable one keeps its timer)
- [x] `spec/lib/gemstone/infomon_spec.rb`, `spec/lib/gemstone/infomon/`, `spec/lib/common/spell_spec.rb`: 93 examples, 0 failures
- [x] rubocop clean
- [x] In game with the local effect list marked refreshable: 9105 reads two minutes after every raise, and volley no longer fires inside the window

### DIFF
--- a/lib/common/spell.rb
+++ b/lib/common/spell.rb
@@ -345,6 +345,14 @@ module Lich
         (self.timeleft > 0) and @active
       end
 
+      # The cast-type predicates below (stackable?, refreshable?, multicastable?)
+      # and the duration formulas resolve against the 'self' cast-type unless a
+      # :caster or :target naming someone else is supplied. A bare call therefore
+      # answers for a self-cast only, which is not the same answer for spells
+      # whose target cast-type differs: in the current effect list 14 group buffs
+      # (Bravery 211 and Heroism 215 among them) are stackable when self-cast but
+      # refreshable when cast on someone else. Pass :caster/:target whenever the
+      # spell may have come from another character.
       def stackable?(options = {})
         if options[:caster] and (options[:caster] !~ /^(?:self|#{XMLData.name})$/i)
           if options[:target] and (options[:target].downcase == options[:caster].downcase)
--- a/lib/gemstone/infomon/parser.rb
+++ b/lib/gemstone/infomon/parser.rb
@@ -633,7 +633,10 @@ module Lich
               spell = Spell.list.find do |s|
                 line =~ /^#{s.msgup}$/
               end
-              spell.putup unless spell.active?
+              # A start message while the spell is already up is the game
+              # refreshing it: a refreshable timer resets to its full
+              # duration, anything else keeps the timer it has.
+              spell.putup if !spell.active? || spell.refreshable?
               # add various cooldowns back without affecting parse speed
               Spells.require_cooldown(spell)
               :ok
--- a/spec/lib/gemstone/infomon/spell_message_refresh_spec.rb
+++ b/spec/lib/gemstone/infomon/spell_message_refresh_spec.rb
@@ -1,0 +1,70 @@
+# frozen_string_literal: true
+
+require_relative '../../../spec_helper'
+
+# Load spell data for realistic testing
+load_spell_data
+
+require "common/sharedbuffer"
+require "common/buffer"
+require "games"
+require "gemstone/overwatch"
+require "gemstone/infomon"
+require "attributes/stats"
+require "attributes/resources"
+require "attributes/skills"
+require "attributes/spells"
+require "gemstone/currency"
+require "gemstone/infomon/status"
+require "gemstone/experience"
+require "util/util"
+require "gemstone/psms"
+require "gemstone/psms/ascension"
+
+# Top-level aliases the parser uses at runtime
+Skills = Lich::Gemstone::Skills unless defined?(Skills)
+Spell = Lich::Common::Spell unless defined?(Spell)
+Spells = Lich::Gemstone::Spells unless defined?(Spells)
+
+# A spell's start message while the spell is already up. The game sends
+# it when a refreshable effect is refreshed (a recast of Celerity, another
+# RAISE under Briar Betrayer), and the timer has to follow: before this
+# the parser skipped the message for an active spell, so Lich let its own
+# timer run out while the game's had two minutes to go.
+RSpec.describe Lich::Gemstone::Infomon::Parser, 'a start message for an active spell' do
+  let(:celerity) { Lich::Common::Spell[506] }   # refreshable in the effect list
+  let(:briar)    { Lich::Common::Spell[9105] }  # no span: keeps its timer
+
+  before do
+    allow(Lich::Gemstone::Infomon).to receive(:set)
+    allow(Lich::Gemstone::Spells).to receive(:require_cooldown)
+    celerity.putdown
+    briar.putdown
+  end
+
+  after do
+    celerity.putdown
+    briar.putdown
+  end
+
+  it 'puts an inactive spell up' do
+    described_class.parse('You suddenly start moving light-footedly.')
+    expect(celerity.active?).to be true
+    expect(celerity.timeleft).to be > 0.5
+  end
+
+  it 'resets a refreshable spell to its full duration' do
+    celerity.putup
+    celerity.timeleft = 0.05 # three seconds left on Lich's clock
+    described_class.parse('You suddenly start moving light-footedly.')
+    expect(celerity.timeleft).to be > 0.5
+  end
+
+  it 'leaves a non-refreshable spell on the timer it has' do
+    briar.putup
+    briar.timeleft = 0.05
+    described_class.parse('As you begin to raise your ruic longbow, the briars imbedded in your flesh release their stored blood in a massive pulse of power that you can feel in the core of your very being.  The vines lose all crimson hues, and strength courses through your blood.')
+    expect(briar.active?).to be true
+    expect(briar.timeleft).to be <= 0.05
+  end
+end


mrhoribu (approved):

mrhoribu (approved):
# PR Review: fix(gs): infomon - a spell's start message refreshes a refreshable timer (elanthia-online/lich-5#1585)

**Reviewed:** 2 files (+74/-1) at `184679f6` on 2026-09-11. Based on: full local clone (worktree checkout of the PR branch against `origin/main`), plus the `elanthia-online/scripts` checkout for the live `effect-list.xml` data the fix depends on.
**Summary:** Fixes a real, well-documented bug (`spell.putup unless spell.active?` never refreshed a still-active spell's timer on a repeat start message, so message-only-tracked effects like Briar Betrayer expired on Lich's clock before the game's). The fix is correct for spells whose refreshable/stackable status doesn't depend on who cast it, and is well tested for that case. It's incomplete for a specific, real class of group-cast buffs, for reasons explained below — bounded in practice by a separate correction mechanism, but worth a follow-up.

## Minor

### 1. `refreshable?` is called without cast-type context, so the fix silently doesn't apply to 14 group buffs when received from another caster — `lib/gemstone/infomon/parser.rb:639`

```ruby
spell.putup if !spell.active? || spell.refreshable?
```

`Spell#refreshable?` (`lib/common/spell.rb:372-394`) takes an `options` hash with `:caster`/`:target` and, when given neither (as here), falls through to `@duration['self'][:refreshable]` — it always reports the *self-cast* duration's refreshable flag, never the target-cast one, unless the caller passes options.

The live `effect-list.xml` (`elanthia-online/scripts/scripts/effect-list.xml`, the file `Spell` actually downloads and parses per `lib/common/spell.rb:159`) has 14 spells where the self-cast duration is `stackable` but the target-cast duration is `refreshable`: Spirit Defense (103), Lesser Shroud (120), Spirit Shield (202), Unpresence (204), Purify Air (207), Bravery (211), Heroism (215), Spell Shield (219), Prayer of Protection (303), Benediction (307), Warding Sphere (310), Relieve Burden (314), Blink (1215), and Mantle of Faith (1601). All of these are `availability='group'` or `'all'` — cast by one character onto party members — and their `<message type='start'>` text (e.g. Bravery's `You (?:and your group stand tall and )?feel more confident\.`) is not cast-type-specific, so `parser.rb` has no way at this call site to tell whether the character casting it and the character reading it are the same person.

Concretely: caster casts Bravery on their group. Everyone but the caster experiences it as the *target* cast-type (refreshable=true, per the data). But `spell.refreshable?` here reads the *self* cast-type flag (stackable=true → refreshable=false), so when the caster recasts Bravery on the group while it's still up, recipients' `!spell.active? || spell.refreshable?` evaluates `false || false` = `false` — `putup` is skipped, and Lich's timer keeps running on the stale (pre-refresh) schedule. That's exactly the symptom this PR sets out to fix (Spell[x].active? going false early, per the Briar Betrayer case in the PR description), just for a different, common set of spells.

This is bounded in practice: these 14 are all ordinary circle spells that show up in the game's "Active Spells/Buffs" dialog, and `Lich::Gemstone::ActiveSpell.update_spell_durations` (`lib/gemstone/infomon/activespell.rb:122-164`) independently overwrites `spell.timeleft`/`spell.active` from that window feed whenever an update is requested — which is the mechanism the PR description says already corrects window-tracked effects. So the gap here is a staleness window until the next active-spells sync, not a permanent miss, and it's consistent with how `time_per`, `max_duration`, `stackable?`, and the existing (pre-PR) `spell.putup` call at this same site already default to the self-cast formula when called without options — this isn't a new pattern this PR invented, just a new call site (`refreshable?`) that inherits it.

Confirmed: `Spell#refreshable?`'s self-default and the 14-spell mismatch are verified directly against `lib/common/spell.rb` and the live `effect-list.xml`; that the ambiguity is genuinely unresolvable from `line` alone is confirmed by `@msgup` construction in `Spell#initialize` (`lib/common/spell.rb:83-84`), which has no cast-type dimension at all. That `:caster`/`:target` options are meant to disambiguate this exact case (not dead API surface) is confirmed by `elanthia-online/scripts/scripts/ewaggle.lic:964,985,1015,1440,1478,1503`, which calls `spell.stackable?`/`spell.refreshable?` with `:target =>` specifically to get the right cast-type. `suspected` only on how often the active-spells sync actually catches this in practice (I didn't play-test the multi-caster case) — everything else here is read from source and data.

No test in this PR covers the mismatch case (`spec/lib/gemstone/infomon/spell_message_refresh_spec.rb` only exercises Celerity, where self and target agree, and Briar Betrayer, which has neither). Given the bug is bounded by `ActiveSpell`, I'd treat this as a documented follow-up rather than a blocker — either pass caster/target context through if it becomes available, or note the limitation where `refreshable?` is defined so the next person calling it without options doesn't assume it covers both cast types.

## What this PR gets right

- Correctly scopes the fix to `span='refreshable'` only, deliberately leaving `stackable` effects alone rather than double-counting a recast (`lib/gemstone/infomon/parser.rb:636-639`, matches the PR description's stated reasoning) — this is the right call given `putup`'s stacking behavior in `lib/common/spell.rb:529-536`.
- The new spec (`spec/lib/gemstone/infomon/spell_message_refresh_spec.rb`) is not tautological: I traced through `time_per`/`max_duration`/`putup` by hand for all three cases (inactive→active, active+refreshable→reset, active+non-refreshable→unchanged) and the assertions (`timeleft > 0.5` vs `<= 0.05`) actually discriminate the fixed behavior from the old behavior.
- One-line, minimal diff to the actual defect; no unrelated refactoring.
- Full local run: `bundle exec rspec` — 7150 examples, 0 failures; `rubocop` — clean on both changed files.

## Coverage notes

- Full clone via `gh`, PR branch checked out into an isolated `git worktree` (the user's own `lich-5` checkout had unrelated uncommitted changes on a different branch, left untouched).
- Ran the new spec, the full `gemstone/infomon` + `spell_spec.rb` suite (93 examples), and the entire repo suite (7150 examples) — all green. Ran `rubocop` on both changed files — no offenses.
- Cross-checked the effect-list.xml dependency against the live source of truth in a local `elanthia-online/scripts` checkout (`scripts/scripts/effect-list.xml`), since `lich-5`'s own `data/effect-list.xml` is a bundled fallback/download target, not what's edited for this fix — confirmed the PR description's claim that Briar Betrayer (9105) isn't yet marked `refreshable` there, and that's correctly deferred to a separate PR.
- CodeRabbit CLI is installed and authenticated but returned "Rate limit exceeded" on this run, so no second-model pass was folded in here.
- Did not play-test in-game; the Minor finding's real-world frequency (how often the active-spells sync races the message-based refresh) is unverified beyond static/data analysis.

Nisugi:
Verified the review's finding and it holds up. `refreshable?` (`lib/common/spell.rb:372`) does fall through to `@duration['self'][:refreshable]` when called without options, and re-scanning the live `effect-list.xml` gives exactly the 14 spells listed, same numbers and availability.

One adjustment to the framing: for those 14, the new code is behaviourally identical to the old. A `stackable` self-duration makes `refreshable?` false, so `!spell.active? || spell.refreshable?` collapses back to `putup unless active?`. Nothing regresses — it's a pre-existing gap this PR doesn't close rather than one it opens.

Pushed 9d10518 with a comment where the predicates are defined, noting that a bare call answers for a self-cast only and naming the group-buff mismatch. It sits above `stackable?` to cover `refreshable?` and `multicastable?` too, since all three share the default.

On threading caster/target through the parser — there's no way to do it from the matched text. I checked the three spells whose start message has a group alternation (Bravery 211, Heroism 215, Spell Shield 219) against the wiki, and all three are self-cast-only spells whose group version is triggered by EVOKE, not a targeted CAST. Both alternation branches are caster views:

- **Heroism 215** — `sinks into your skin` is the self-cast message; `you and your group` is the EVOKE caster's. A recipient sees `A brilliant aura surrounds <Name>.`, matching neither.
- **Bravery 211** — a recipient sees `You feel more confident.`, byte-identical to the self-cast message.
- **Spell Shield 219** — same shape: `An opalescent aura surrounds you.` for both.

So there's no cast-type signal to recover at this call site, and no follow-up worth filing against the parser.

Worth noting separately, since the durations are wider apart than "a staleness window" suggests: self-cast is 2 hours base, while the EVOKE versions last 60s (211, 215) and 120s (219). That implies two pre-existing bugs — a received EVOKE of Bravery/Spell Shield being timed on the self formula, and a received Heroism not matching any start message so it isn't tracked at all. Neither is touched or worsened by this PR; flagging them only so they're on record.

Test run: `spec/lib/common/spell_spec.rb` + `spec/lib/gemstone/infomon/` + `infomon_spec.rb` — 93 examples, 0 failures. rubocop clean.

Nisugi:
Correcting myself on one point above. I said the parser has no cast-type context to work with. That's true for most of the 14, but not all of them.

Three of them encode the distinction in the start message already, as an optional group clause:

- Bravery 211 — `You (?:and your group stand tall and )?feel more confident\.`
- Heroism 215 — `A brilliant aura surrounds you and (?:sinks into your skin|your group)\.  ...`
- Spell Shield 219 — `An opalescent aura surrounds you(?: and your group)?\.`

The group phrasing goes to the caster who blanketed the group; a recipient gets the plain self-phrasing. So which alternation branch matched does tell you the cast type, and the parser is discarding it by matching `/^#{s.msgup}$/` as one fused pattern and then calling `refreshable?` with no options.

The other 11 are a single invariant string with no group variant, so a recipient's line really is indistinguishable from the caster's. Those need the schema change.

That makes this partly closable here, not wholly deferred as I claimed. I'd still rather not fold it into this PR — it's a different mechanism (parse the matched branch, derive cast type, pass `:target`) against a different set of spells, and this PR's one-line change is worth keeping reviewable on its own. But it should be its own PR against this repo rather than the scripts-repo issue I described, and I'll scope it that way.

Nisugi:
Retracting my previous comment. I claimed Bravery/Heroism/Spell Shield encode the cast type in the start-message alternation and were therefore fixable in the parser. That was wrong, and I had Heroism's branches backwards.

All three are self-cast-only spells with a group version triggered by EVOKE, not a targeted CAST. The alternation branches are both **caster** views, not caster-vs-recipient:

- **Heroism 215** — `sinks into your skin` is the self-cast message (confirmed against the wiki's Start message). `you and your group` is what the EVOKE caster sees. A recipient of an EVOKE sees `A brilliant aura surrounds <Name>.`, which matches neither branch, so `SpellUpMsgs` never fires and no timer is set at all.
- **Bravery 211** — a recipient sees `You feel more confident.`, byte-identical to the self-cast message. No signal to recover.
- **Spell Shield 219** — same shape as Bravery: `An opalescent aura surrounds you.` for both self-cast and recipient.

So there is no cast-type information in the matched text for any of the three, and the branch-aware parsing I proposed can't work. My earlier comment should be disregarded.

The underlying data mismatch is still real, and the durations make it sharper than "a staleness window": self-cast is 2 hours base, while the EVOKE versions are 60s (211, 215) and 120s (219). Two different bugs sit behind it, neither of which the refresh predicate addresses:

1. A received EVOKE of Bravery/Spell Shield would be timed on the self formula (~2h instead of 60-120s), if it registers as a received cast at all.
2. A received EVOKE of Heroism produces no matching start message, so it isn't tracked.

Both are pre-existing and out of scope here — this PR's one-line change doesn't touch either, and doesn't regress them. I'm dropping the follow-up I promised, since the mechanism I proposed for it doesn't exist. The comment added in 9d10518 documenting the self-cast default still stands and is accurate.

Apologies for the noise in the thread.

PR #1586, opened 2026-09-10
feat(gs): Combat::Messages - non-combat message families as observer events

Nisugi:
## Summary

Widens `Combat::Observers` into a general parser seam: the non-combat lines hunting and cleansing scripts keep private `DownstreamHook` regexes for become observer events, scanned only while something subscribes.

- **`defs/messages.rb`**: eight families (disarm, hazard, ambush, hold, bless, archery, marks, reaction), 40 defs. Every one is a line bigshot's `hunt_monitor`, ecleanse's `set_hooks` or eohunter's watch rules matched privately: a weapon knocked away or wrenched, the itchy curse, the infected wound, the hive traps, the ambusher and the bolt, the snake's coils, the item limit, a bless shrugged off, an arrow stuck, the aim, the bonded return, the 703 haze and 1614 rebuke marks, Swift Justice charges, arcane reflexes, the weapon reaction prompt. Each def is an event name, one pattern (no top-level alternations, so `PatternGate` gates every family), and a payload block; payloads carry `:raw`.
- **`Combat::Messages`**: delivers them through `Observers`. The Tracker's hook chunks on the prompt and drops chunks that name no creature, which is most of these lines, so Messages has its own hook: installed on the first subscription to a message event, removed with the last. A family is scanned only while one of its events has a subscriber; matching runs on one ordered worker thread, never on the game stream (the Observers subscriber contract applies).
- **`Observers.on_change`**: a callback after every `on`/`off`/`clear!`, which is how Messages recomputes the active families.
- `Tracker.on(:disarm_seen) { ... }` subscribes the same way as combat facts and needs the tracker neither enabled nor scanning.

## Cost

Measured on a 250k-line bigshot session log, first 100k lines, `Messages.scan` per line:

| subscribers | per line |
|---|---|
| none | 0.14 µs |
| one family | 0.52 µs |
| all eight | 4.54 µs |

With no subscribers the hook is not installed at all.

## Tests

`spec/lib/gemstone/combat/message_defs_spec.rb`: every family against the game line it was pinned to, every family gated, ordinary lines yield nothing. `messages_spec.rb`: no family, no hook and no scan without a subscriber; a combat subscription activates nothing; `:any` activates all; the worker delivers; the hook goes up with the first subscription and down with the last. The existing combat specs pass unchanged (265 examples).

## Follow-up

eohunter will replace its `watch.rb` rules with these subscriptions; ecleanse and bigshot can do the same one hook at a time. Two of eohunter's remaining rules (UAC positioning tier and followup) are already `:ucs` facts and were not duplicated here.

### DIFF
--- a/lib/gemstone/combat/defs/messages.rb
+++ b/lib/gemstone/combat/defs/messages.rb
@@ -1,0 +1,157 @@
+# frozen_string_literal: true
+
+#
+# Message Definitions - non-combat game lines scripts react to, as
+# observer events.
+#
+# The combat defs answer "what happened to whom in a fight". These answer
+# the other questions a hunting script keeps a DownstreamHook open for:
+# a weapon knocked away, a curse or infection taken, a trap sprung, an
+# ambusher arriving, a bolt, a bless shrugged off, an arrow stuck, a
+# charge counter, a spell mark on a creature. Every one of these was a
+# private regex in bigshot, ecleanse or eohunter; here each is one def in
+# a family, gated by PatternGate like the combat tables, scanned only
+# while something subscribes to the family (see Combat::Messages).
+#
+# A def is an event name, a pattern, and a data block that turns the
+# MatchData into the event payload. Payloads always also carry :raw, the
+# line. Patterns are pinned to the messaging the scripts were matching in
+# the wild (ecleanse set_hooks, bigshot hunt_monitor, eohunter watch
+# rules, 2026-09), one form per def so PatternGate can gate each; a
+# top-level alternation would gate nothing.
+#
+
+require_relative 'pattern_gate'
+
+module Lich
+  module Gemstone
+    module Combat
+      module Definitions
+        module Messages
+          MessageDef = Struct.new(:event, :pattern, :data)
+          Family = Struct.new(:name, :defs, :gate, :always_scan) do
+            # Every event this family can emit.
+            def events = defs.map(&:event).uniq
+
+            def rejects?(line) = PatternGate.rejects?(gate, always_scan, line)
+          end
+
+          LINK = /<a exist="(?<id>[^"]+)" noun="(?<noun>[^"]+)">[^<]+<\/a>/i.freeze
+
+          # Build a family from [event, pattern, data] rows.
+          def self.family(name, rows)
+            defs = rows.map { |event, pattern, data| MessageDef.new(event, pattern, data) }
+            gate, always = PatternGate.build(defs.map(&:pattern))
+            Family.new(name, defs.freeze, gate, always)
+          end
+
+          NONE = ->(_m) { {} }
+
+          FAMILIES = [
+            # ecleanse set_hooks 1618: the line-driven recoveries. The disarm
+            # lines carry the weapon's noun; what hand it was in and where
+            # we stood is the subscriber's to read at the moment.
+            family(:disarm, [
+                     [:disarm_seen, %r{Your <a exist="[^"]+" noun="(?<noun>[^"]+)">[^<]+</a> is knocked from your grasp}, ->(m) { { kind: :recover, noun: m[:noun] } }],
+                     [:disarm_seen, %r{your <a exist="[^"]+" noun="(?<noun>[^"]+)">[^<]+</a> at .+?\.  The weapon rebounds off of the hardened .+? and is wrenched from your hand\.  It slides along the ground and disappears into the shadows!}, ->(m) { { kind: :recover, noun: m[:noun] } }],
+                     [:disarm_seen, %r{^Your <a exist="[^"]+" noun="(?<noun>[^"]+)">[^<]+</a> strikes one of the bony protrusions on <pushBold/>an? <a exist="\d+" noun="[^"]+">[^<]+</a><popBold/> \w+ and it is wrenched out of your grasp!}, ->(m) { { kind: :recover, noun: m[:noun] } }],
+                     [:disarm_seen, %r{^You swing your <a exist="[^"]+" noun="(?<noun>[^"]+)">[^<]+</a> at <pushBold/>(?:an?|the) <a exist="[^"]+" noun="[^"]+">[^<]+</a><popBold/>\.  The weapon strikes one of the bony protrusions on the <pushBold/><a exist="[^"]+" noun="[^"]+">[^<]+</a><popBold/> \w+ and it is wrenched out of your grasp!}, ->(m) { { kind: :recover, noun: m[:noun] } }],
+                     [:disarm_seen, %r{Your <a exist="[^"]+" noun="(?<noun>[^"]+)">[^<]+</a> tears free from your hands and floats}, ->(m) { { kind: :telekinetic_recover, noun: m[:noun] } }],
+                     [:disarm_seen, %r{The webbing entangles your <a exist="[^"]+" noun="(?<noun>[^"]+)">[^<]+</a>, rendering it useless}, ->(m) { { kind: :recover_weapon_webbing, noun: m[:noun] } }],
+                     [:sanctum_transform, %r{Striking with a serpent's unsettling quickness, .*\.  Vile .*, kindling it into an unholy semblance of life\.  The .* form twists and mutates, sprouting scales and cold eyes as it transforms into a <a exist="\d+" noun="(?<noun>[^"]+)">[^<]+</a>!}, ->(m) { { noun: m[:noun] } }]
+                   ]),
+
+            # ecleanse: afflictions and traps that take a trip to clear.
+            family(:hazard, [
+                     [:itchy_curse, /You shiver slightly as an invisible rash covers your body/, NONE],
+                     [:infected_wound, /The flesh around the wound feels hot and cold at the same time, heavy with infection\./, NONE],
+                     [:hive_trap, /You notice a flickering glint in the shadows/, ->(_m) { { kind: :apparatus } }],
+                     [:hive_trap, /The apparatus flickers with deadly radiance/, ->(_m) { { kind: :apparatus } }],
+                     [:hive_trap, /The ground churns violently as flashes of chitin jut from its depths/, ->(_m) { { kind: :ground } }],
+                     [:hive_trap, /The ground underfoot churns violently and huge chitinous mandibles flash as the insectoid monstrosity below goes into a feeding frenzy!/, ->(_m) { { kind: :ground } }],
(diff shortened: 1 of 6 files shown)


mrhoribu (changes requested):
# PR Review: feat(gs): Combat::Messages - non-combat message families as observer events (elanthia-online/lich-5#1586)

**Reviewed:** 6 files, +568/-0 at `0522ac3d` on 2026-09-11. Based on: full clone (worktree checkout of the PR head against `origin/main` merge-base `073d23a9`).
**Verdict:** Request changes
**Summary:** Adds a second `DownstreamHook`-fed event source (`Combat::Messages`, backed by 41 line-defs across 8 families in `defs/messages.rb`) that installs/removes itself as subscribers come and go, delivered through the existing `Combat::Observers` seam. The def/pattern/gate machinery is solid and well-tested; the install/uninstall lifecycle that reacts to subscription changes has a genuine, reproduced race that can leave an active subscriber's hook silently uninstalled.

## Major

### 1. `Messages.refresh!` races itself under concurrent (un)subscription, silently dropping an active subscriber's hook — `lib/gemstone/combat/messages.rb:59-64`

```ruby
def refresh!
  active = families.select { |f| f.events.any? { |e| Observers.any_for?(e) } }
  @mutex.synchronize { @active = active.freeze }
  active.empty? ? uninstall! : install!
  @active
end
```

`@active` is written under `@mutex`, but the resulting `install!`/`uninstall!` call ([`lib/gemstone/combat/messages.rb:111`](lib/gemstone/combat/messages.rb:111) / [`:123`](lib/gemstone/combat/messages.rb:123)) happens outside it, based on a locally-computed `active` snapshot. `refresh!` runs on every `Observers.on`/`off`/`clear!` ([`observers.rb:78-81`](lib/gemstone/combat/observers.rb:78), wired via `Observers.on_change { refresh! }` at [`messages.rb:170`](lib/gemstone/combat/messages.rb:170)), from whatever thread called `on`/`off` — in Lich that's a per-script thread, and it's entirely ordinary for two hunting scripts (e.g. ecleanse and bigshot) to subscribe around the same time at session start.

Two `refresh!` calls with different outcomes (one settling on empty, one on non-empty) can interleave so that the *last* `@active` write is non-empty but the *last* `install!`/`uninstall!` call is `uninstall!` — leaving `@hook == false` while a real subscriber exists. I reproduced this deterministically against the actual production methods (not a synthetic analogue): a `refresh!` that computed `active = []` before a subscription, paused right before its `uninstall!` call, let a second `refresh!` subscribe-and-`install!` run to completion, then resumed:

```
after subscribe:              active=[:hold]  installed?=true
after stale uninstall released: active=[:hold]  installed?=false
```

At that point the family is "active" per `Messages.active_families`, but the `DownstreamHook` is gone, so no game line is ever enqueued again — the subscriber (e.g. something relying on `:rooted`/`:unrooted` to know when it's safe to act) gets nothing, with no error, no log line, indefinitely, until *some other* subscription change happens to trigger a `refresh!` that recomputes a non-empty active set again. For a script depending on `:rooted` for safety, that's a silent, potentially session-long loss of a combat-relevant signal.

Fix: hold `@mutex` across the read-decide-act sequence (compute `active`, decide, and call `install!`/`uninstall!` all inside one `synchronize` block, or make `install!`/`uninstall!` themselves mutex-protected and re-check the *current* `@active` rather than a stale local variable).

**Confidence:** confirmed (reproduced against the actual `refresh!`/`install!`/`uninstall!`/`@active`/`@mutex` code, via a controlled two-thread interleaving in `/tmp/reviewpr-lich-5-1586/race_repro2.rb`).

## Minor

### 2. `LINK` constant is dead code — `lib/gemstone/combat/defs/messages.rb:39`

```ruby
LINK = /<a exist="(?<id>[^"]+)" noun="(?<noun>[^"]+)">[^<]+<\/a>/i.freeze
```

Defined but never referenced anywhere in this file or elsewhere in the repo (`grep -rn "Messages::LINK"` returns nothing). Every family builds its own inline `<a exist=... noun=...>` fragment instead of reusing it. Either wire it in or drop it.

### 3. `messages_spec.rb` globally overrides `Lich.log`/`Lich.logged` without the repo's guard convention — `spec/lib/gemstone/combat/messages_spec.rb:5-9`

```ruby
module Lich
  def self.log(msg); (@logged ||= []) << msg; end

  def self.logged = @logged ||= []
  module Gemstone; module Combat; end; end
end
```

`spec/spec_helper.rb:682-733` documents and applies a fine-grained-guard convention specifically so mocks from different spec files can coexist regardless of load order (`def log(msg); ... end unless respond_to?(:log)`, etc.), because Ruby module reopening is process-global and permanent for the rest of the `rspec` run. This file's `def self.log` has no such guard, so if it happens to load before `spec_helper.rb` establishes its own guarded version, it wins permanently for every other spec in the run. I ran the full suite (`bundle exec rspec`, 7167 examples) at four different seeds (44513, 1, 2, 999) and got 0 failures each time — file-load order is apparently stable enough here that this doesn't currently bite — but it's a latent trap for the next spec file that lands earlier in load order and expects `spec_helper`'s `Lich.log`. Give it the same `unless respond_to?(:log)` guard.

## What this PR gets right

- The def/pattern/gate structure (`MessageDef`, `Family`, `PatternGate`) is clean and mirrors the existing combat-defs shape well — `family.gate`/`always_scan` line up correctly with `PatternGate.build`/`rejects?`, and I verified all 41 defs across the 8 families get a real literal gate (`always_scan: 0` everywhere), matching the PR's "gates every family" claim.
- The regex patterns are grounded in already-battle-tested private script regexes (ecleanse/bigshot/eohunter), not invented from scratch — a much stronger correctness basis than typical new regex additions.
- Subscriber-error isolation (`Observers.emit`, `Observers#changed`) and worker-error isolation (`Messages#run_loop`'s rescue) are both handled, consistent with the existing `Observers` contract.
- `spec/lib/gemstone/combat/message_defs_spec.rb` and `messages_spec.rb` cover the subscription-gate behavior thoroughly (no-subscriber quiescence, per-family activation, `:any`, hook up/down, worker delivery) — this is exactly the kind of state-machine test coverage that's easy to skip.
- Rubocop is clean on all six changed files; `rubocop -A` found nothing to fix.

## Coverage notes

- Access level: full clone via `git worktree`, diffed against `origin/main` (the local `main` branch was stale and briefly gave a misleading multi-hundred-file diff before I re-based the comparison against `origin/main`).
- Ran the full spec suite (`bundle exec rspec`) four times at different seeds: 7167 examples, 0 failures each time. Ran `rubocop` on all six changed files: 0 offenses.
- The race in Finding 1 was reproduced against the real production methods with an injected pause to force the interleaving; I did not find a way to trigger it with plain concurrent load (1000 iterations of unguarded random-jitter concurrent subscribe/unsubscribe produced 0 observed mismatches), so treat the *frequency* in real usage as unquantified even though the mechanism and worst-case outcome are demonstrated, not theoretical.
- No CodeRabbit pass: the `coderabbit` CLI is not available in this environment, so that cross-check was skipped entirely.
- Did not independently verify the 41 regex patterns against real DragonRealms game-log text (no such log was available in this environment) — correctness here rests on the PR's stated provenance (pinned to the exact lines ecleanse/bigshot/eohunter were already matching) plus the unit tests, which are necessarily self-referential (built from the same `link`/`bold` helpers used to construct the patterns).

Nisugi:
Thanks — all three addressed in `4037afd7`.

**1. `refresh!` race.** Fixed as suggested: `@mutex` now covers the whole read-decide-act, so the active-set computation, the `@active` publish, and the `install!`/`uninstall!` call can't be split. Checked the widened critical section for deadlock — `install!`/`uninstall!` only touch the hook registry (plain hash writes in `HookRegistry`) and never re-enter `refresh!` or `ensure_worker`, and `Observers.changed` already invokes callbacks outside its own mutex, so lock ordering stays one-way.

Added a regression spec that forces the interleaving by injecting a pause into the production mutex, stalling the stale `refresh!` in the gap between publishing `@active` and acting on it. It fails on the old code with `active_families == [:hold]` and the hook uninstalled — the exact state you described — and passes on the fix. Worth noting my first attempt at this test passed on the buggy code too; pausing inside `any_for?` is too early, since on the old code those calls happen before the mutex block and the stale `uninstall!` is short-circuited by `return unless @hook`.

**2. `LINK`.** Dropped. Wiring it in would mean rewriting 41 patterns that are only trustworthy because they're the pinned originals, so removing the dead constant seemed like the better trade.

**3. Spec guard.** Added `unless respond_to?(:log)`. Nothing in the file asserts on the captured messages — `Lich.log` only needs to exist for the worker's rescue to call — so the guard is purely additive. Left `logged` unguarded since it's defined nowhere else in the suite.

**On the suite:** combat is 266 examples / 0 failures across seeds 1, 2, 999, 44513. I can't reproduce your clean full-suite run — this Windows box has 57 pre-existing failures at baseline, all in frontend/launcher/wine/GTK/Windows-path specs. I diffed the failing example lists before and after my changes and they're identical, so nothing here introduced them, but flagging it since your environment evidently doesn't hit them. Rubocop clean on all changed files.

Your caveat about the 41 regexes being unverifiable against real game text stands — these are GemStone lines, and the provenance argument is all I have there too.

mrhoribu (approved):
Reviewed the original diff adversarially and found one real issue: `Messages.refresh!` wrote `@active` under `@mutex` but called the resulting `install!`/`uninstall!` outside it, based on a stale local snapshot. Reproduced it deterministically against the actual production methods — two interleaved subscribe/unsubscribe calls could leave a live subscriber's family marked active while the `DownstreamHook` had been silently removed, so it would get no events, no error, until some unrelated later subscription change happened to fix it. Also flagged a dead `LINK` constant in `defs/messages.rb` and a new spec that redefined `Lich.log` globally without the `unless respond_to?(:log)` guard `spec_helper.rb`'s own comments call for.

All three are fixed in 4037afd7:
- `refresh!` now holds `@mutex` across the whole read-decide-act sequence, closing the race. Re-ran my interleaving repro against the fixed code (with the pause injected at the same point the new regression test uses) — no more mismatch, hook state now tracks real subscriber state. Confirmed no reentrancy/deadlock was introduced: `install!`/`uninstall!` don't touch `@mutex` themselves, and `ensure_worker` (which does) is never called from inside `refresh!`'s critical section.
- Dead `LINK` constant removed.
- `Lich.log` mock now guarded, matching convention, with a comment explaining why.

The added regression test (forcing the exact stale-uninstall interleaving via a pausing wrapper around the production mutex) is a good one — it would fail on the old code and passes now.

Full suite: 7168 examples, 0 failures (multiple seeds). Rubocop: clean on all changed files.

PR #1580, opened 2026-09-10
feat(gs): Lich::Gemstone::Mana.pulse, the MANA PULSE verb once

Nisugi:
## Summary

Adds `Lich::Gemstone::Mana.pulse(spell = nil, timeout: 2)`, one home for the MANA PULSE verb that several EO scripts each carry a copy of.

- With no argument it sends `mana pulse` and returns true only when mana was gained.
- With a spell number it pulses only when the spell is known and not currently affordable, so a caller can write `Mana.pulse(1010); cast` without its own bookkeeping.
- Waits out roundtime first; returns false on the "already pulsed" and "nothing to pulse" replies.
- Loaded from the gemstone game loader after `status.rb`.

## Test plan

- [x] `spec/lib/gemstone/mana_spec.rb`, 10 examples
- [x] rubocop clean on touched files
- [x] Exercised in game

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1587, opened 2026-09-10
feat(all): bounded fput with max_resends, interrupt, resend_transient and failure symbols

Nisugi:
## Summary

`fput`'s refusal ladder has no bounds: a `...wait N` or "struggle to stand" loop can resend forever, a transient refusal ends the send with a bare `false` the caller cannot tell from any other failure, and nothing stops a wait when the script wants out. bigshot's `bs_put` and eohunter's send ladder each carry a private copy of the ladder for exactly those reasons. This adds what they added as options on the existing trailing Hash, with every default the old behaviour, so no script changes.

| option | meaning | default |
|---|---|---|
| `max_resends:` | cap on refusal-driven resends (roundtime, stand, stun/web, waitingfor miss) | nil, unbounded |
| `interrupt:` | a callable checked on every wait; true ends the send at once | nil |
| `resend_transient:` | resend a "can't seem" / "don't seem" after a quarter second, under the cap, the way `bs_put` does, instead of giving up | false |
| `failures: :symbol` | `:no_response`, `:too_many_resends`, `:interrupted`, `:dead`, `:refused` instead of `false` | `false` |

With no interrupt the roundtime wait is the single `sleep(N)` it always was; with one it sleeps in tenths so the interrupt lands promptly. `dothistimeout` gains `interrupt:`, checked on every read. `waitrt?` and `waitcastrt?` gain `interrupt:` and `cap:`, sleep in slices while roundtime remains, and answer whether there was roundtime to wait out, as before.

## Tests

The existing 26 fput examples pass unchanged. New: the cap; each failure symbol; the default `false` with the line pushed back; the transient resend under the cap; the interrupt during a roundtime wait and during silence; the single sleep without an interrupt; a waitingfor miss counting as a resend; `waitrt?` plain, interrupted, and capped. 37 examples.

## Follow-up

eohunter's `Actions::Base#send_through_ladder` becomes `fput(cmd, max_resends: 5, timeout: 30, interrupt:, resend_transient: true, failures: :symbol)` and its `settle_rt` becomes `waitrt?(interrupt:, cap: 15)` once this ships, removing the ladder copy the review of that script called out.

### DIFF
--- a/lib/global_defs.rb
+++ b/lib/global_defs.rb
@@ -295,21 +295,35 @@ def checkcastrt
   [0, XMLData.cast_roundtime_end.to_f - Time.now.to_f + XMLData.server_time_offset.to_f].max
 end
 
-def waitrt?
-  sleep checkrt
-  return true if checkrt > 0.0
-  return false if checkrt == 0
+# @param interrupt [#call, nil] checked every tenth of a second; true ends
+#   the wait early
+# @param cap [Numeric, nil] the longest wait allowed, in seconds
+# @return [Boolean] whether there was roundtime to wait out
+def waitrt?(interrupt: nil, cap: nil)
+  had_rt = checkrt > 0.0
+  stop_at = cap ? Time.now + cap : nil
+  while checkrt > 0.0
+    return had_rt if interrupt && interrupt.call
+    return had_rt if stop_at && Time.now >= stop_at
+
+    sleep([checkrt, 0.1].min)
+  end
+  had_rt
 end
 
-def waitcastrt?
-  #  sleep checkcastrt
-  current_castrt = checkcastrt
-  if current_castrt.to_f > 0.0
-    sleep(current_castrt)
-    return true
-  else
-    return false
+# @param interrupt [#call, nil] checked every tenth of a second
+# @param cap [Numeric, nil] the longest wait allowed, in seconds
+# @return [Boolean] whether there was cast roundtime to wait out
+def waitcastrt?(interrupt: nil, cap: nil)
+  had_rt = checkcastrt.to_f > 0.0
+  stop_at = cap ? Time.now + cap : nil
+  while checkcastrt.to_f > 0.0
+    return had_rt if interrupt && interrupt.call
+    return had_rt if stop_at && Time.now >= stop_at
+
+    sleep([checkcastrt.to_f, 0.1].min)
   end
+  had_rt
 end
 
 def checkpoison
@@ -1644,11 +1658,52 @@ def fput(message, *waitingfor)
   unless (script = Script.current) then respond('--- waitfor: Unable to identify calling script.'); return false; end
   waitingfor.flatten!
 
-  # Optional timeout via trailing Hash argument: fput('cmd', 'pattern', timeout: 30)
-  # Default 60s prevents infinite hangs when the game stops responding.
-  # Use timeout: 0 to disable (original behavior).
+  # Options via a trailing Hash argument: fput('cmd', 'pattern', timeout: 30)
+  #   timeout:          seconds with no game response before giving up (60;
+  #                     0 disables, the original behavior)
+  #   max_resends:      how many times a refusal ("...wait 3", "struggle to
+  #                     stand", stunned) may trigger a resend before giving
+  #                     up (nil, the original: unbounded)
+  #   interrupt:        a callable checked on every wait; true ends the
+  #                     send at once (nil: never)
+  #   resend_transient: on a transient refusal that is not a stun or a
+  #                     web (a "can't seem", "don't seem"), resend after a
+  #                     quarter second instead of giving up (false, the
+  #                     original; bigshot's bs_put resends)
+  #   failures:         :false (the original: every failure returns false)
+  #                     or :symbol - :no_response, :too_many_resends,
+  #                     :interrupted, :dead, :refused - so a caller can
+  #                     tell them apart
   options = (waitingfor.pop if waitingfor.last.is_a?(Hash)) || {}
-  timeout = options[:timeout] || options['timeout'] || 60
+  option = ->(key) { options[key] || options[key.to_s] }
+  timeout = option.call(:timeout) || 60
+  max_resends = option.call(:max_resends)
+  interrupt = option.call(:interrupt)
+  resend_transient = option.call(:resend_transient) ? true : false
+  symbols = option.call(:failures) == :symbol
+  fail_with = ->(reason) { symbols ? reason : false }
+  interrupted = -> { interrupt && interrupt.call ? true : false }
+  # With an interrupt, sleep in slices so it lands within a tenth of a
+  # second; without one, the plain sleep of before. True when interrupted.
+  wait = lambda do |seconds|
+    if interrupt.nil?
+      sleep(seconds)
+      return false
+    end
+    slices = (seconds / 0.1).ceil
+    slices.times do
+      return true if interrupted.call
+
+      sleep(0.1)
+    end
+    false
+  end
+  resends = 0
+  # a refusal that asks for a resend: false when the cap allows it
+  over_cap = lambda do
+    resends += 1
+    !max_resends.nil? && resends > max_resends
+  end
 
   clear
   put(message)
@@ -1658,9 +1713,11 @@ def fput(message, *waitingfor)
     string = get?
 
     if string.nil?
+      return fail_with.call(:interrupted) if interrupted.call
+
       if timeout > 0 && (Time.now - timer > timeout)
         echo "fput: No game response for #{timeout}s to '#{message}'"
-        return false
+        return fail_with.call(:no_response)
       end
       pause 0.1
       next
@@ -1669,36 +1726,51 @@ def fput(message, *waitingfor)
     timer = Time.now # Reset timeout on any game response
 
     if string =~ /(?:\.\.\.wait |Wait )(?<wait_time>[0-9]+)/
+      return fail_with.call(:too_many_resends) if over_cap.call
+
       hold_up = Regexp.last_match[:wait_time].to_i
-      sleep(hold_up) unless hold_up.nil?
+      return fail_with.call(:interrupted) if wait.call(hold_up)
+
       clear
       put(message)
       next
     elsif string =~ /^You.+struggle.+stand/
+      return fail_with.call(:too_many_resends) if over_cap.call
+
       clear
-      fput 'stand'
+      stood = fput('stand', options)
+      return stood if symbols && stood.is_a?(Symbol)
+
       next
     elsif string =~ /stunned|can't do that while|cannot seem|^(?!You rummage).*can't seem|don't seem|Sorry, you may only type ahead/
       if dead?
         echo "You're dead...! You can't do that!"
         sleep 1
         script.downstream_buffer.unshift(string)
-        return false
+        return fail_with.call(:dead)
       elsif checkstunned
         while checkstunned
+          return fail_with.call(:interrupted) if interrupted.call
+
           sleep("0.25".to_f)
         end
       elsif checkwebbed
         while checkwebbed
+          return fail_with.call(:interrupted) if interrupted.call
+
           sleep("0.25".to_f)
         end
       elsif string =~ /Sorry, you may only type ahead/
         sleep 1
+      elsif resend_transient
+        sleep 0.25
       else
         sleep 0.1
         script.downstream_buffer.unshift(string)
-        return false
+        return fail_with.call(:refused)
       end
+      return fail_with.call(:too_many_resends) if over_cap.call
+
       clear
       put(message)
       next
@@ -1711,7 +1783,9 @@ def fput(message, *waitingfor)
           script.downstream_buffer.unshift(string)
           return foundit
         end
-        sleep 1
+        return fail_with.call(:too_many_resends) if over_cap.call
+        return fail_with.call(:interrupted) if wait.call(1)
+
         clear
         put(message)
         next
@@ -2026,13 +2100,17 @@ def dothis(action, success_line)
   }
 end
 
-def dothistimeout(action, timeout, success_line)
+# @param interrupt [#call, nil] checked on every read; true ends the wait
+#   at once and returns nil, the way a timeout does
+def dothistimeout(action, timeout, success_line, interrupt: nil)
   end_time = Time.now.to_f + timeout
   line = nil
   loop {
     Script.current.clear
     put action unless action.nil?
     loop {
+      return nil if interrupt && interrupt.call
+
       line = get?
       if line.nil?
         sleep 0.1
--- a/lib/global_defs.rb
+++ b/lib/global_defs.rb
@@ -1690,8 +1690,8 @@ def fput(message, *waitingfor)
   #   max_resends:      how many times a refusal ("...wait 3", "struggle to
   #                     stand", stunned) may trigger a resend before giving
   #                     up (nil, the original: unbounded)
-  #   interrupt:        a callable checked on every wait; true ends the
-  #                     send at once (nil: never)
+  #   interrupt:        a callable checked on every wait and before every
+  #                     resend; true ends the send at once (nil: never)
   #   resend_transient: on a transient refusal that is not a stun or a
   #                     web (a "can't seem", "don't seem"), resend after a
   #                     quarter second instead of giving up (false, the
@@ -1705,6 +1705,9 @@ def fput(message, *waitingfor)
   timeout = option.call(:timeout) || 60
   max_resends = option.call(:max_resends)
   interrupt = option.call(:interrupt)
+  unless interrupt.nil? || interrupt.respond_to?(:call)
+    raise ArgumentError, "fput: interrupt: must respond to call"
+  end
   resend_transient = option.call(:resend_transient) ? true : false
   symbols = option.call(:failures) == :symbol
   fail_with = ->(reason) { symbols ? reason : false }
@@ -1725,6 +1728,9 @@ def fput(message, *waitingfor)
     false
   end
   resends = 0
+  # true after 'stand' went out and before its reply came back; the reply
+  # is not the answer to message, so message goes out again on top of it
+  standing = false
   # a refusal that asks for a resend: false when the cap allows it
   over_cap = lambda do
     resends += 1
@@ -1757,16 +1763,21 @@ def fput(message, *waitingfor)
       hold_up = Regexp.last_match[:wait_time].to_i
       return fail_with.call(:interrupted) if wait.call(hold_up)
 
+      standing = false
       clear
       put(message)
       next
     elsif string =~ /^You.+struggle.+stand/
+      # stand in this frame, under the same cap and interrupt, instead of
+      # a nested fput('stand') that started its own count and could not
+      # be interrupted; a persistent struggle recursed until the stack
+      # gave out
       return fail_with.call(:too_many_resends) if over_cap.call
+      return fail_with.call(:interrupted) if interrupted.call
 
+      standing = true
       clear
-      stood = fput('stand', options)
-      return stood if symbols && stood.is_a?(Symbol)
-
+      put('stand')
       next
     elsif string =~ /stunned|can't do that while|cannot seem|^(?!You rummage).*can't seem|don't seem|Sorry, you may only type ahead/
       if dead?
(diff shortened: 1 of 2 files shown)


Nisugi:
**Merge note: waits on #1575 (bounded script execution guards).**

Clean against `main` alone, but once #1575 merges this conflicts in four regions of `lib/global_defs.rb`: #1575 replaced the ordinary sleeps in `waitrt?`, `waitcastrt?`, `fput` and `dothistimeout` with `Script.execution_sleep`; this PR rewrites the same waits for interruption, caps, resend limits and failure symbols.

**Resolution (rehearsed in a combined checkout, main + #1575 + #1576 + #1592 + #1578 + #1580 + this):**

- Keep this PR's options and control flow; route every wait through `Script.execution_sleep`. That includes the `wait` lambda inside `fput`, which merges *automatically* with bare `sleep` on both of its paths and would sit through a script's cancellation on the two longest waits fput takes (a `...wait N` roundtime, a stun). Resolved form:

  ```ruby
  wait = lambda do |seconds|
    if interrupt.nil?
      Script.execution_sleep(seconds)
      return false
    end
    slices = (seconds / 0.1).ceil
    slices.times do
      return true if interrupted.call

      Script.execution_sleep(0.1)
    end
    false
  end
  ```

- The caller's `interrupt:` and the script's execution guard stay separate mechanisms. A guard cancellation raises `ScriptExecutionGuard::Interrupted` through the waits; it is never converted into `false` or `:interrupted`. `dothistimeout`'s guarded waits from #1575 are unchanged.
- #1575's native guard spec synchronises on a single sleep longer than 20 s, which bounded waits no longer take; it changes to "the first guarded sleep", asserting the same thing (cancel unwinds the helper at once, no extra command). New examples drive the production `fput` for both cancellation mechanisms, the resend cap, and the legacy-vs-bounded contracts.

**Already in this PR (716e0dfd):** `waitrt?`/`waitcastrt?` with no options keep the legacy contract, one sleep then whether roundtime remains, so existing callers see no change; the bounded contract applies only to option-bearing calls.

Combined checkout results: gemstone + guard suites 785 examples, 0 failures; guard spec 17 examples; an adversarial pass confirmed every #1575 checkpoint survives and no rescue swallows the exception. One thing noticed on the way, pre-existing in #1575 rather than this merge: `wait_until`/`wait_while` still `sleep 0.25` between polls (they checkpoint via `wait_while_paused!` at the top of each iteration, so cancellation still unwinds within one poll).

I'll rebase and push the resolved version as soon as #1575 lands.

mrhoribu (approved):
# PR Review: feat(all): bounded fput with max_resends, interrupt, resend_transient and failure symbols (elanthia-online/lich-5#1587)

**Reviewed:** 2 files, +378/-31 at `716e0dfd` on 2026-09-11. Based on: full local clone (history included), all tests run.
**Summary:** Adds bounded-resend, interruptible, and typed-failure options to `fput`, plus matching `interrupt:`/`cap:` options on `waitrt?`, `waitcastrt?`, and `dothistimeout`, all additive and defaulted to the old behavior. The implementation is correct everywhere it was traced and the legacy-contract regression from the first commit was caught and fixed within the PR itself (second commit); the remaining gap is that two of the five new failure paths (`struggle to stand`, and interrupt during the `checkstunned`/`checkwebbed` waits) — the exact scenarios named as the PR's motivation — ship with zero test coverage.

## Minor

### 1. `struggle to stand` and stun/web-interrupt paths are new logic with no test coverage — `lib/global_defs.rb:1763`, `lib/global_defs.rb:1777`

```ruby
elsif string =~ /^You.+struggle.+stand/
  return fail_with.call(:too_many_resends) if over_cap.call

  clear
  stood = fput('stand', options)
  return stood if symbols && stood.is_a?(Symbol)
```

This branch is new behavior, not a passthrough of the old code: it now caps resends via `over_cap`, forwards the caller's full `options` (including `interrupt:` and `failures:`) into the recursive `fput('stand', options)`, and propagates a symbol failure from that recursive call upward. The same is true of the `interrupted.call` checks newly added inside the `checkstunned`/`checkwebbed` `while` loops at lines 1779 and 1785. None of these three additions are exercised by any spec — `grep -n "struggle\|stand"` against `spec/lib/global_defs_spec.rb` turns up only the production code itself, no `it` block. The PR's own "Tests" section lists what the 37 new examples cover (cap, each failure symbol, default `false`, transient resend, interrupt on a roundtime wait and on silence, single sleep, waitingfor-miss-as-resend, `waitrt?` plain/interrupted/capped) and stand/stun/web-interrupt aren't on that list — this isn't a review guess, it's a self-reported gap.

This matters more than a typical missing-test note because "struggle to stand" is the second scenario named in the PR description as the motivating failure mode ("a `...wait N` or 'struggle to stand' loop can resend forever"). I traced this branch by hand against the mirrored spec copy (verified byte-identical to the production function) and it behaves correctly — `over_cap` bounds the resend, `options` propagation means `interrupt:` now actually reaches the nested stand attempt where before this PR it silently wouldn't have, and symbol failures propagate — but "traced correctly by a reviewer" is not the same guarantee as "a test that fails if this regresses." A follow-up PR that touches this branch has nothing to catch a regression here.

**Fix:** add at least one example each for (a) `struggle to stand` respecting `max_resends`/propagating a `:too_many_resends` from the nested stand call, and (b) an `interrupt:` firing while `checkstunned`/`checkwebbed` is `true`.

## Nits

### 2. Two of the five new sleep sites in `fput` don't check `interrupt`, contradicting the stated contract — `lib/global_defs.rb:1790`, `lib/global_defs.rb:1792`

```ruby
elsif string =~ /Sorry, you may only type ahead/
  sleep 1
elsif resend_transient
  sleep 0.25
```

The PR doc comment says `interrupt:` is "a callable checked on every wait," and that's true for the roundtime wait (`wait.call`), the waitingfor-miss wait (`wait.call(1)`), and the stun/web loops (`interrupted.call` each 0.25s slice) — but these two plain `sleep` calls don't check it. In practice the exposure is small (bounded to ≤1s and ≤0.25s respectively, and the interrupt does get honored on the very next loop iteration), so this is cosmetic rather than a real hang risk, but it's worth either fixing for consistency or narrowing the doc comment's "every wait" claim to be accurate.

## What this PR gets right

- The bounded/legacy dual-contract split in `waitrt?`/`waitcastrt?` (`interrupt.nil? && cap.nil?` → old body verbatim, otherwise the new polling body) is exactly right, and I confirmed by grep that every existing caller in this repo (`lib/stash.rb`, `lib/common/spell.rb`, all of `lib/dragonrealms/commons/*`, all of `lib/gemstone/**`, `lib/common/map/map_gs.rb`) calls the bare no-arg form, so none of them can observe the contract-changing bug the second commit (716e0dfd) fixed. That second commit's own message is candid about what it fixes and why — a first-commit self-review catching a real regression before it shipped is exactly what should happen.
- Passing the caller's full `options` Hash into the recursive `fput('stand', options)` call (line 1767) is a real improvement over the pre-PR code (which called bare `fput 'stand'`): it's what makes `interrupt:` actually able to break out of a stuck stand-loop, which is one of the two scenarios the PR is written to fix. (Untested, per Minor #1, but correctly designed.)
- `option.call(:key)` unifies the symbol/string dual-key lookup that the original code only did for `timeout`; extending that same lookup to all five options rather than inventing a second convention is good consistency.
- I hand-traced every new branch of `fput` (roundtime wait+cap, waitingfor-miss+cap, transient resend+cap, interrupt during a roundtime wait in slices, interrupt during silence, dead/stunned/webbed failure symbols) against the actual test expectations and the arithmetic holds in each case — e.g. `max_resends: 2` with three `...wait` responses correctly allows exactly 2 resends (3 total `put` calls) before returning on the third.

## Coverage notes

- Full local clone of `elanthia-online/lich-5`, PR branch `pr-1587` (`716e0dfd`) checked out against `main`; both commits and their messages reviewed, not just the squashed diff.
- Ran `bundle exec rspec spec/lib/global_defs_spec.rb`: 41 examples, 0 failures. Ran `bundle exec rubocop lib/global_defs.rb spec/lib/global_defs_spec.rb`: no offenses.
- Ran the full suite (`bundle exec rspec`): 8 pre-existing failures in `spec/lib/common/setup_files_spec.rb`, confirmed present on `main` with the identical random seed (`--seed 52651`) — a pre-existing test-order-dependent issue unrelated to this PR, not counted against it.
- Verified the spec file's mirrored copies of `fput`, `waitrt?`, and `waitcastrt?` (kept local to the spec file rather than `require`d, per the file's own top-of-file comment explaining why) are byte-for-byte identical to the production functions in `lib/global_defs.rb`, so the 41 passing examples are actually exercising the shipped logic and not a stale copy.
- Checked blast radius for `waitrt?`, `waitcastrt?`, and `dothistimeout` across the whole repo (`lib/gemstone/**`, `lib/dragonrealms/**`, `lib/stash.rb`, `lib/common/spell.rb`, `lib/common/map/map_gs.rb`, `lib/gemstone/group.rb`, `lib/gemstone/infomon/cli.rb`) — every call site uses the pre-existing positional/no-arg form, so none is affected by the new keyword arguments.
- CodeRabbit CLI is installed and authenticated but returned a rate-limit error ("You've used all 3 included reviews currently available... 53 minutes" wait) before producing any findings — no CodeRabbit pass was available for this review.
- Did not review `dr-scripts`/`scripts` for the described follow-up work (migrating `eohunter`'s `send_through_ladder`/`settle_rt`) since that's explicitly future work in a different repo, not part of this diff.

MahtraDR (commented):
# PR Review: feat(all) bounded fput — max_resends, interrupt, resend_transient, failure symbols (elanthia-online/lich-5#1587)

**Reviewed:** 2 files, +378/-31 at `716e0dfd` on 2026-09-13. Based on: full local clone (both commits + history), worktree checkout, all specs run, rubocop, and an independent CodeRabbit CLI pass.
**Verdict:** Approve with comments — one Major worth fixing before `max_resends:`/`interrupt:` are relied on for the stand path.
**Summary:** Adds opt-in bounds to `fput`'s refusal ladder (`max_resends:`, `interrupt:`, `resend_transient:`, `failures: :symbol`) plus `interrupt:` on `dothistimeout`/`waitrt?`/`waitcastrt?`, every default preserving old behavior. The legacy paths are genuinely untouched and every existing caller is unaffected. The one thing to fix: **the "struggle to stand" branch — the exact scenario the PR names as motivation — is bounded by neither `max_resends` nor `interrupt`**, so a character that genuinely cannot stand recurses to a stack overflow. This was proven by execution, not inference.

> This consolidates my own pass with @mrhoribu's review (#5182478326). We independently landed on the same branch as the key risk; the difference is that a runnable test flips its severity from "correct but untested" to "provably broken on its motivating case." Where mrhoribu caught something I missed, it's credited below.

---

## Major

### 1. `max_resends` and `interrupt` both fail to bound a persistent "struggle to stand" — `lib/global_defs.rb:1763`

```ruby
elsif string =~ /^You.+struggle.+stand/
  return fail_with.call(:too_many_resends) if over_cap.call

  clear
  stood = fput('stand', options)
  return stood if symbols && stood.is_a?(Symbol)

  next
```

The other three refusal types (`...wait N`, stun/web, waitingfor-miss) **loop** within one `fput` frame, so they share one `resends` counter and one `interrupt` check and are genuinely bounded. "Struggle to stand" is the odd one out: it **recurses** into a fresh `fput('stand', options)`, and each recursion re-initialises `resends = 0` (line 1727). On a *persistent* struggle the outer frame never reaches its second loop iteration — `stood = fput('stand', options)` never returns because the nested frame is itself recursing — so `over_cap` never accumulates past 1 at any level. There is also no `interrupted.call` in this branch, and the nested frame only checks the interrupt on its `nil`/`wait`/`stun` paths, none of which pure struggle-spam ever reaches. Both new safety valves are inert on this one path.

This contradicts the PR's own contract twice: the summary says the change exists because "a `...wait N` or 'struggle to stand' loop can resend forever" and "nothing stops a wait when the script wants out," and the doc comment (lines 1690–1692) lists "struggle to stand" among what `max_resends` caps. A character who is bound/legged/overloaded emits "You struggle to stand" on every stand attempt — the motivating case — and hits neither bound.

**Confirmed by execution.** Three specs run against the mirrored `fput` in `spec/lib/global_defs_spec.rb` (verified byte-identical to production at this SHA):

```
it 'caps a persistent struggle-to-stand loop (PR claim)'   # get? always "You struggle to stand"
  expect(fput('get sword', max_resends: 3, failures: :symbol)).to eq(:too_many_resends)
  => SystemStackError: stack level too deep          # FAILS

it 'honors interrupt on a persistent struggle-to-stand'    # get? always "You struggle to stand"
  expect(fput('get sword', interrupt: -> { true }, failures: :symbol)).to eq(:interrupted)
  => SystemStackError: stack level too deep          # FAILS

it 'caps a persistent ...wait loop (works)'                # get? always "...wait 2 seconds."
  expect(fput('get sword', max_resends: 3, failures: :symbol)).to eq(:too_many_resends)
  => passes                                          # the loop-based branch IS bounded
```

This is where my read and @mrhoribu's diverge: he hand-traced this branch and reported it correct ("`over_cap` bounds the resend... symbol failures propagate," and that `options` propagation "is what makes `interrupt:` actually able to break out of a stuck stand-loop"), grading it Minor/missing-test. That trace holds for a *single* struggle→stand→success — but the bug lives in the *sequence*, which a single-event trace can't surface and an executed stateful test does. `over_cap` does bound at depth-1; it never engages when the struggle is the persistent case the PR is written to fix.

The infinite recursion itself is pre-existing (old code did `fput 'stand'; next`), so this is **not a regression** — default callers behave exactly as before. But this PR is the change that promises to bound it, and for this branch it does not.

**Fix:** count and interrupt in the same frame instead of recursing, e.g.
```ruby
elsif string =~ /^You.+struggle.+stand/
  return fail_with.call(:too_many_resends) if over_cap.call
  return fail_with.call(:interrupted) if interrupted.call
  clear
  put('stand')            # send stand, then fall through to the loop
  next
```
so the shared `resends`/`interrupt` apply. This changes the send pattern (the loop re-reads rather than nesting a full send), so it wants its own tests against the stand success/refusal sequences — which also closes the coverage gap mrhoribu flagged.

---

## Minor

### 2. Two of the five `fput` sleep sites don't check `interrupt`, contradicting the "every wait" contract — `lib/global_defs.rb:1790`, `lib/global_defs.rb:1792`

*(Caught by @mrhoribu; confirmed.)*

```ruby
elsif string =~ /Sorry, you may only type ahead/
  sleep 1
elsif resend_transient
  sleep 0.25
```

The doc comment says `interrupt:` is "checked on every wait," which holds for the roundtime wait (`wait.call`), the waitingfor-miss wait (`wait.call(1)`), and the stun/web loops (`interrupted.call` per 0.25s slice) — but these two plain `sleep`s don't. Exposure is small (bounded to ≤1s and ≤0.25s, and the interrupt is honored on the next loop iteration), so cosmetic rather than a hang risk — but either check it for consistency or narrow the doc comment's "every wait" claim.

### 3. On give-up, `resend_transient` swallows the last refusal line — `lib/global_defs.rb:1791`

```ruby
elsif resend_transient
  sleep 0.25
else
  sleep 0.1
  script.downstream_buffer.unshift(string)
  return fail_with.call(:refused)
end
return fail_with.call(:too_many_resends) if over_cap.call
```

The default (`else`) path unshifts the refusal back onto `downstream_buffer` so the player sees it. The `resend_transient` path never unshifts — fine while retrying, but when `over_cap` finally trips and it returns `:too_many_resends`, the final "You can't seem to…" line has been consumed and is never surfaced. A script leaning on `resend_transient` shows the player nothing on the eventual give-up. Consider unshifting `string` before that `over_cap` return.

### 4. `interrupt:` on the struggle path is dropped unless `failures: :symbol` — `lib/global_defs.rb:1768`

Independent of Major #1: `return stood if symbols && stood.is_a?(Symbol)`. A caller passing `interrupt:` but not `failures: :symbol` gets a nested `fput('stand', options)` that returns `false` on interrupt; `symbols` is false, the guard is skipped, and the outer frame does `next` — the interrupt is discarded and the send continues. Interrupt honoring shouldn't depend on opting into symbols. (Folds into the Major #1 fix if the branch stops recursing.)

---

## Nits

- **Non-callable `interrupt:` raises rather than degrading** (`lib/global_defs.rb:1711`): any truthy non-callable (`interrupt: true`) reaches `.call` and raises `NoMethodError` mid-send. Documented as `[#call, nil]`, so a foot-gun note — a `respond_to?(:call)` guard would fail faster and clearer.
- **The spec mirrors `fput`/`waitrt?`/`waitcastrt?` by hand** (the file documents why — loading `global_defs.rb` pollutes global methods). Faithfully copied here and byte-identical today, but the tests validate a copy: a future edit to real `fput` not mirrored stays green while shipping broken. Pre-existing pattern, now larger (`spec/lib/global_defs_spec.rb:17`, `:455`).

---

## Open questions

- **How often does a real "struggle to stand" fail to resolve in one attempt?** If `stand` always succeeds (branch always depth-1), Major #1's stack overflow is unreachable and it drops to "the doc overstates the cap." **Settles it:** is there any state (bound, prone-locked, encumbered, certain injuries) where `stand` returns "struggle to stand" indefinitely? One yes/no from a player closes it. (Code path executed; game-frequency is the part I can't run.)

---

## What this PR gets right

- **Legacy paths are genuinely untouched.** `waitrt?`/`waitcastrt?` with no options are the old bodies verbatim (commit 2 exists precisely to restore that after commit 1's rewrite changed the return contract — a first-commit self-review catching a real regression before it ships is exactly right). I confirmed every one of the ~90 `waitrt?`/`waitcastrt?`/`dothistimeout` call sites in `lib/` (`stash.rb`, `common/spell.rb`, all of `dragonrealms/commons/*`, all of `gemstone/**`, `common/map/map_gs.rb`, `gemstone/group.rb`, `gemstone/infomon/cli.rb`) uses the no-option form and so is unaffected.
- **Passing the caller's full `options` into the recursive `fput('stand', options)`** is a real design improvement over the pre-PR bare `fput 'stand'` — it's the right intent (make `interrupt:` reach the nested attempt); it just needs the non-recursive shape from Major #1 to actually hold.
- **The wait-time bounding for the three loop-based branches is correct** — `max_resends: 2` with three `...wait` responses allows exactly 2 resends (3 `put` calls) before returning on the third — and the tests cover the cap, each symbol, the default `false`, the transient resend, the interrupt on a roundtime wait and on silence, the single sleep without an interrupt, and a waitingfor-miss-as-resend.
- **`option.call(:key)` unifies the symbol/string dual-key lookup** the original did only for `timeout`, extended consistently to all five options.

---

## Coverage notes

- **Access:** full local clone; PR branch fetched from `upstream refs/pull/1587/head`, worktree at `716e0dfd`, confirmed still head at post time. Both commits and their messages reviewed.
- **CI:** all green on the PR (Rubocop, two Ruby test jobs, title validation); base `main` green.
- **Executed:** `bundle exec rspec spec/lib/global_defs_spec.rb` → 41 examples, 0 failures; `ruby -c` clean. Per @mrhoribu, the full `bundle exec rspec` has 8 pre-existing failures in `spec/lib/common/setup_files_spec.rb`, present on `main` with the same seed — a pre-existing order-dependent issue unrelated to this PR. Wrote/ran 3 falsifying specs (Major #1's two failures + the passing `...wait` contrast).
- **Independent pass:** ran the CodeRabbit CLI against `main` → **No findings** across the reviewed files. (mrhoribu's run hit the 3-review rate limit and produced none, so this fills that gap.)
- **State/sequence axis:** repeated each of the four refusal types as an infinite stream — this is what surfaced Major #1; interleaved interrupt with the struggle stream; interrupt during silence and during roundtime; the give-up path of `resend_transient`. **Not probed at runtime:** real game wording/persistence of "struggle to stand" (Open Question), non-callable `interrupt:` (reasoned only), live roundtime timing (specs mock `sleep`/`Time.now`).
- **Precision vs recall:** the findings above are real (executed or traced). Not exhaustively fuzzed: all option combinations (`resend_transient:` + `interrupt:` + nested stand). Confidence in the spec mirror rests on it matching source at this SHA (diff-verified), which is itself the Nit above.

## Repo conventions I did not apply

- **Versioning:** `lib/version.rb` carries `# x-release-please-version`, so the version is bumped automatically from the `feat:`/`fix:` commit types — no manual bump expected, and the two commits are correctly typed. I did not flag a missing bump.
- No `CONTRIBUTING.md`/`CLAUDE.md`/PR template beyond the CI checks, which already pass.

MahtraDR (approved):
Re-reviewed at `efe51d5d` (commit "stand in the same frame…") — this clears every finding from my prior review (#5189062565):

- **Major** — the persistent "struggle to stand" recursion is gone: `put('stand')` now runs in-frame under the shared `over_cap`/`interrupt`, with the `standing` flag resending the command after the stand reply. My two falsifying cases (persistent struggle with `max_resends:` and with `interrupt:`) that returned `SystemStackError` before now return `:too_many_resends`/`:interrupted`.
- **Minor** — `type ahead` and `resend_transient` sleeps now honor `interrupt:`; `resend_transient` give-up now surfaces the last refusal to `downstream_buffer`; non-callable `interrupt:` raises a clear `ArgumentError`.

Verified against the new head: CI green, spec suite green (47 examples), an independent CodeRabbit pass came back with no findings, and I re-ran my original falsifying tests plus adversarial probes of the new `standing` state machine (waitingfor interaction, an unrelated line preceding the stand reply, interrupt turning true after `stand`) — all pass. Nice work.

PR #1588, opened 2026-09-10
fix(gs): Group.broken? waits on Lich::Claim::Lock, the constant that exists

Nisugi:
## Summary
- `Group.broken?` waited on `Lich::Gemstone::Claim::Lock`, but Claim is defined at `Lich::Claim` (lib/gemstone/claim.rb), so the call raised `NameError` before the group check ran.
- One-line fix to the constant. Nothing in lib or scripts calls `broken?` today; scripts that do will now get the documented true/false instead of an exception.

## Test plan
- [x] Loaded claim.rb standalone: `Lich::Claim::Lock` resolves, `Lich::Gemstone::Claim` raises NameError
- [x] `bundle exec rubocop lib/gemstone/group.rb` clean

### DIFF
--- a/lib/gemstone/group.rb
+++ b/lib/gemstone/group.rb
@@ -259,7 +259,7 @@ module Lich
       #
       # @return [Boolean] true if group state is inconsistent
       def self.broken?
-        sleep(0.1) while Lich::Gemstone::Claim::Lock.locked?
+        sleep(0.1) while Lich::Claim::Lock.locked?
         if Group.leader?
           return true if (GameObj.pcs.empty? || GameObj.pcs.nil?) && !@@members.empty?
           return false if (GameObj.pcs.empty? || GameObj.pcs.nil?) && @@members.empty?


mrhoribu (approved):

mrhoribu (approved):
# PR Review: fix(gs): Group.broken? waits on Lich::Claim::Lock, the constant that exists (elanthia-online/lich-5#1588)

**Reviewed:** 1 file, +1/-1, at `bf37ff9e` on 2026-09-11. Based on: full local clone (existing checkout at `/home/ryan/lich-5`, PR head checked out in an isolated worktree so the reviewer's own uncommitted local changes were untouched).
**Verdict:** Approve
**Summary:** One-line constant fix — `Group.broken?` referenced `Lich::Gemstone::Claim::Lock`, which has never existed; the real constant is `Lich::Claim::Lock`. The change swaps the module path to match, and every other call site in the codebase already used the correct one.

## Blockers
None.

## Major
None.

## Minor
None.

## Nits
None.

## What this PR gets right

- **Root cause, not symptom.** `lib/gemstone/claim.rb:1-3` defines `module Lich; module Claim` — the `module Gemstone` wrapper is explicitly commented out (`# module Gemstone # test this?`), so `Lich::Gemstone::Claim` has never resolved. The four other live call sites (`lib/common/xmlparser.rb:434,1111,1276` and `lib/gemstone/infomon/xmlparser.rb:580`) all correctly use `Lich::Claim::Lock` / `Lich::Claim`, confirming `group.rb` was the sole outlier rather than there being some ambiguity about which path is correct.
- **Load order guarantees the constant exists here.** `lib/common/gameloader.rb:21` requires `gemstone/claim.rb` before `gemstone/group.rb` (line 38) in `GameLoader.gemstone`, and `Group` is Gemstone-only code, so unlike the `xmlparser.rb` call sites (which are shared between DR and GS and correctly guard with `defined?(Lich::Claim)`), no guard is needed in `group.rb` — the constant is unconditionally loaded before this method can run.
- **Accurately scoped in the description.** The PR body states no lib/script code currently calls `Group.broken?` — verified via repo-wide grep, there are zero callers outside the definition itself. So this fixes a landmine for future callers without being a behavior change anyone is depending on today; the risk of the change is effectively zero.
- **History confirms it was always broken.** `git blame`/`git log -S` show the wrong constant was introduced in `916c7ea7` (PR #850, 2025-04-25) and has been wrong ever since — this isn't undoing a deliberate recent change.

## Coverage notes

- Full local clone available (`elanthia-online/lich-5` checked out with `fork`/`origin` remotes); PR head fetched (`refs/pull/1588/head`) and checked out into an isolated `git worktree` so the review didn't touch the primary checkout's existing uncommitted work.
- Ran `bundle exec rubocop lib/gemstone/group.rb` on the PR worktree: clean, no offenses.
- Ran `bundle exec rspec spec/lib/gemstone/group_load_spec.rb`: 1 example, 0 failures. No spec exists for `Group.broken?` itself (none needed to be added — no behavior test would meaningfully cover a constant-name fix beyond what rubocop/load already confirm, and there are no callers to regression-test).
- Ran a local CodeRabbit pass (`coderabbit review --agent --base main`) against the worktree: 0 findings on the single reviewed file.
- Searched the full repo for `Group.broken?` / `.broken?` callers and for all `Lich::Claim`/`Lich::Gemstone::Claim` references — both confirmed via grep, not assumption.

Issue #42, opened 2021-09-20
Feat:  All login methods to direct FE with .sal calls ('Method 0' login)


OSXLich-Doug:
Category:  Lich ecosystem (general use features)

    Feat: Convert all authorized logins to game to 'Method 0' (directly call FE with .sal)
      Impacts: lich.rbw
      Approach: Break out GUI and launch methods, then remove all other attempts to log in and use 'Method 0' concept
      To-do: break out and test
      Needs: review
      Currntly working: Doug
      Status:
        09/19/2021 - prototype created, needs break out and testing

PR #1589, opened 2026-09-10
feat(gs): society command readers, factored out of use

Nisugi:
## Summary
- Each society reader's `use` built its command inline: the entry's `:usage` verb, else `sign of` / `sigil of` / `symbol of` plus the short name, then the target. This exposes it as `CouncilOfLight.command(name, target)`, `GuardiansOfSunfist.command` and `OrderOfVoln.command`, returning the exact string `use` sends and nil for an unknown ability.
- `Society.command(entry, prefix, target)` holds the shared rule and `use` calls it, so nothing is duplicated. A String target is sent as given, unchanged from before.
- Behavior change worth noting: a `GameObj` or `Integer` target is now appended as `#id`. The society `use` methods never formatted those targets before this PR, so this is new behavior riding along with the refactor, not a pure extraction. It differs from the PSM precedent in #1583, where the `#id` branch was already inline in `.use` and was genuinely just moved. No caller in dr-scripts or `scripts` passes a `GameObj`/`Integer` target here today, and `#id` targeting is the established convention elsewhere (`lib/common/spell.rb`, `lib/stash.rb`), so this reads as a latent-bug fix rather than a regression.
- Same shape as the PSM `command` readers in #1583, so an engine that sends and confirms on its own terms can consume the society tables instead of copying the prefix rule. A `results_regex` reader is not included: the society entries carry no result pattern yet, so `use` stays unverified until those are collected.

## Test plan
- [x] New `spec/lib/gemstone/society_command_spec.rb`: the shared rule, the three readers by short and long name, usage verbs (`signal`, `smite`), GameObj / id / String targets, nil for an unknown ability
- [ ] In game: `;e echo Lich::Gemstone::Societies::OrderOfVoln.command('holiness')` prints `symbol of holiness`, and `use` still sends it

### DIFF
--- a/lib/gemstone/societies/council_of_light.rb
+++ b/lib/gemstone/societies/council_of_light.rb
@@ -323,13 +323,26 @@ module Lich
           end
 
           if available?(sign_name)
-            command = sign[:usage] || "sign of #{sign[:short_name]}"
             waitrt?
             waitcastrt?
-            fput "#{command} #{target}".strip
+            fput Society.command(sign, "sign of", target)
           else
             Lich::Messaging.msg("warn", "You cannot use the #{sign_name} sign right now.")
           end
+        end
+
+        ##
+        # The command {use} sends for a sign, without sending it.
+        #
+        # @param sign_name [String] The short or long name of the sign
+        # @param target [String, Integer, GameObj, nil] Optional target (a GameObj or id as `#id`)
+        # @return [String, nil] e.g. "sign of striking", nil when the sign is unknown
+        #
+        def self.command(sign_name, target = nil)
+          sign = self[sign_name]
+          return nil unless sign
+
+          Society.command(sign, "sign of", target)
         end
 
         ##
--- a/lib/gemstone/societies/guardians_of_sunfist.rb
+++ b/lib/gemstone/societies/guardians_of_sunfist.rb
@@ -286,13 +286,26 @@ module Lich
           end
 
           if available?(sigil_name)
-            command = sigil[:usage] || "sigil of #{sigil[:short_name]}"
             waitrt?
             waitcastrt?
-            fput "#{command} #{target}".strip
+            fput Society.command(sigil, "sigil of", target)
           else
             Lich::Messaging.msg("warn", "You cannot use the #{sigil_name} sigil right now.")
           end
+        end
+
+        ##
+        # The command {use} sends for a sigil, without sending it.
+        #
+        # @param sigil_name [String] The short or long name of the sigil
+        # @param target [String, Integer, GameObj, nil] Optional target (a GameObj or id as `#id`)
+        # @return [String, nil] e.g. "sigil of contact", nil when the sigil is unknown
+        #
+        def self.command(sigil_name, target = nil)
+          sigil = self[sigil_name]
+          return nil unless sigil
+
+          Society.command(sigil, "sigil of", target)
         end
 
         ##
--- a/lib/gemstone/societies/order_of_voln.rb
+++ b/lib/gemstone/societies/order_of_voln.rb
@@ -425,13 +425,26 @@ module Lich
           end
 
           if self.available?(symbol_name)
-            command = symbol[:usage] || "symbol of #{symbol[:short_name]}"
             waitrt?
             waitcastrt?
-            fput "#{command} #{target}".strip
+            fput Society.command(symbol, "symbol of", target)
           else
             Lich::Messaging.msg("warn", "You cannot use the #{symbol_name} symbol right now.")
           end
+        end
+
+        ##
+        # The command {use} sends for a symbol, without sending it.
+        #
+        # @param symbol_name [String] The short or long name of the symbol
+        # @param target [String, Integer, GameObj, nil] Optional target (a GameObj or id as `#id`)
+        # @return [String, nil] e.g. "symbol of holiness", nil when the symbol is unknown
+        #
+        def self.command(symbol_name, target = nil)
+          symbol = self[symbol_name]
+          return nil unless symbol
+
+          Society.command(symbol, "symbol of", target)
         end
 
         ##
--- a/lib/gemstone/society.rb
+++ b/lib/gemstone/society.rb
@@ -134,6 +134,25 @@ module Lich
       end
 
       ##
+      # The command a society ability's +use+ sends, without sending it.
+      #
+      # An entry with a `:usage` string sends that verb ("signal", "smite");
+      # any other sends `<prefix> <short_name>` ("sign of striking"). A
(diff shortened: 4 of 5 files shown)


mrhoribu (approved):
# PR Review: feat(gs): society command readers, factored out of use (elanthia-online/lich-5#1589)

**Reviewed:** 5 files, +110/-6 at `56959c97` on 2026-09-11. Based on: full clone (git worktree checked out to the PR head, `main` at `073d23a9` as merge base).
**Summary:** Extracts the command-string-building logic out of each society's `use` into a shared `Society.command`, and exposes a per-society `.command` reader that returns the exact string `use` sends — mechanically sound, well-tested, rubocop-clean, and the full 7150-example suite still passes. The one thing worth a second look: the refactor silently changes what `use` sends for `GameObj`/`Integer` targets, and the PR frames itself as behavior-preserving the way its cited precedent (#1583) genuinely was — but for Societies it isn't, because that formatting never existed in `use` before.

## Minor

### 1. `use`'s target formatting changed for `GameObj`/`Integer`, unlike the PSM precedent this PR cites — `lib/gemstone/society.rb:148-153`

```ruby
def self.command(entry, prefix, target = nil)
  base = entry[:usage] || "#{prefix} #{entry[:short_name]}"
  target = "##{target.id}" if target.respond_to?(:id) && !target.is_a?(String)
  target = "##{target}" if target.is_a?(Integer)
  "#{base} #{target}".strip
end
```

Before this PR, `OrderOfVoln.use`/`CouncilOfLight.use`/`GuardiansOfSunfist.use` built the command with bare interpolation: `"#{command} #{target}".strip` (e.g. still visible in the diff context at `lib/gemstone/societies/order_of_voln.rb:426` pre-change). For a `GameObj` target that sent `target.to_s`, which is the object's **noun** (`GameObj#to_s` returns `@noun`, [gameobj.rb:197](lib/common/gameobj.rb#L197)) — e.g. `"symbol of holiness orc"`. For an `Integer` target it sent the bare number with no `#` — e.g. `"symbol of holiness 42"`.

After this PR, both now go through `Society.command`, which appends `#id`: a `GameObj` becomes `"symbol of holiness #12345"`, an `Integer` becomes `"symbol of holiness #42"`. That's a real behavior change to a public method's wire output, not just an internal refactor — `git log --follow` on `order_of_voln.rb` (and the same for the other two society files) shows this `#id` conversion never existed in `use` at any point in the file's history (only two commits ever touched it: `0964bbf0` introducing the class, `94800fd8` adding membership checks).

This is worth flagging specifically because the PR explicitly invokes #1583 as precedent ("Same shape as the PSM command readers in #1583") and that precedent genuinely was behavior-preserving: PSM's `Armor.use`/`CMan.use`/etc. already had the identical `if target.is_a?(GameObj) ... ##{target.id} ... elsif target.is_a?(Integer) ... ##{target}` branch inline before #1583 extracted it (visible as removed `-` lines in that PR's diff, e.g. `armor.rb`). For Societies, there was no such branch to extract — this PR is introducing the `#id` convention to `use`, not just relocating it.

Practically, this looks like a latent-bug fix rather than a regression: `#id` targeting is the established convention everywhere else a `GameObj` is turned into a game command in this codebase (`lib/common/spell.rb:659`, `lib/stash.rb:28,44,63,...`), and nothing in dr-scripts or the `scripts` repo currently calls `Societies::*.use` with a `GameObj` or `Integer` target (grepped both trees for `OrderOfVoln.use`/`CouncilOfLight.use`/`GuardiansOfSunfist.use` — no hits), so nothing in the ecosystem currently depends on the old, arguably-broken formatting. There's also no pre-existing spec for `use` (before or after this PR) that would have caught the change either way.

Given no known caller is broken and the new behavior is more consistent with the rest of the codebase, this doesn't rise to Major — but the PR description's "as before" framing understates what changed, and a future git-blame on a targeting complaint will land here without an accurate trail. Worth either calling out the `use` behavior change explicitly in the PR description (the way #1583's description does distinguish what stayed the same), or confirming it was in fact deliberate and just say so.

**Confidence:** confirmed (traced against the diff's own pre-change context, `GameObj#to_s`, and file history; the "no current caller" and "established convention elsewhere" claims are grepped, not assumed).

## Nits

- `Society.command`'s target-type check is duck-typed (`target.respond_to?(:id) && !target.is_a?(String)`) where the cited PSM precedent uses an explicit `target.is_a?(GameObj)` ([PR #1583 diff, `psms.rb`](https://github.com/elanthia-online/lich-5/pull/1583)). Not a bug for any realistic caller — DR/GS4 GameObj is the only object type in this ecosystem that responds to `:id` and gets passed around like this — but it's a style divergence from the PR's own "same shape" claim, and duck-typing would silently swallow, say, a `Room` object if one were ever passed by mistake instead of raising or falling through to the string branch.

## What this PR gets right

- The three per-society `.command` readers are byte-for-byte what `use` now sends (`use` calls the exact same `Society.command(entry, prefix, target)`), so the PR's central claim — reader output matches `use` output — is mechanically guaranteed, not just tested-and-hoped.
- `nil`/String/Integer/GameObj target handling was traced through `NilClass#method_missing` (`lib/common/class_exts/nilclass.rb`) and confirmed correct: `nil.respond_to?(:id)` is `false` (no `respond_to_missing?` override), so a `nil` target still collapses to the base command via `.strip`, matching the old `"#{command} #{target}".strip` behavior for the no-target case.
- Good awareness of RSpec load-order hazards: the second commit's fix (stubbing both `Stats` and `Lich::Gemstone::Stats`) is a real, well-diagnosed fix for cross-spec constant resolution, and the commit message explains why.
- Full suite (7150 examples) and rubocop both clean on the touched files.

## Coverage notes

- Reviewed via a `git worktree` checked out to the PR head commit (`56959c97`) off a full local clone, diffed against `main` at `073d23a9` (the actual merge base). Ran the new spec file, the full `spec/lib/gemstone/` directory, and the entire suite (`bundle exec rspec`) — all green, no interference from the new spec. Ran `rubocop` on all 5 touched files — clean.
- Checked `gh pr view`/`gh pr diff` for #1589 and #1583 (the cited precedent) to verify the "same shape" and "as before" claims directly against #1583's actual diff rather than taking the PR description's word for it.
- Grepped dr-scripts and the `scripts` repo trees for any existing caller of `Societies::{OrderOfVoln,CouncilOfLight,GuardiansOfSunfist}.use` — none found, so the Minor finding above is currently inert in practice.
- CodeRabbit pass **not run**: `coderabbit review --agent` hit a hard rate limit (`0` of `3` included reviews remaining on this account/org, 55-minute cooldown, not eligible for on-demand). Proceeded without it per the review protocol rather than blocking; findings above are entirely manual trace + grounding.

PR #1590, opened 2026-09-10
feat(all): Spell.results_regex, the lines cast waits on, exposed

Nisugi:
## Summary
- `Spell#cast` waits on `@@results_regex`, a private class variable with no reader. A script that sends a cast itself and confirms on its own terms (the eohunter engine, with its own roundtime cap and interrupt) had to copy those lines, and the copy has already drifted from core.
- `Spell.results_regex(results_of_interest: nil)` returns the same union `cast` builds, with the optional extra pattern merged the way `cast` merges it. No change to `cast`.

## Test plan
- [x] `spec/lib/common/spell_spec.rb` gained an example covering the union and the extra-pattern merge; 36 examples green

### DIFF
--- a/lib/common/spell.rb
+++ b/lib/common/spell.rb
@@ -143,6 +143,17 @@ module Lich
         # self # rubocop Lint/Void: self used in void context
       end
 
+      # Every line that answers a cast: the regex {#cast} waits on, exposed
+      # so a caller that sends and confirms on its own terms can wait on
+      # the same lines instead of copying them.
+      #
+      # @param results_of_interest [Regexp, nil] extra lines to match, as {#cast} takes
+      # @return [Regexp]
+      def Spell.results_regex(results_of_interest: nil)
+        return @@results_regex unless results_of_interest.is_a?(Regexp)
+        Regexp.union(@@results_regex, results_of_interest)
+      end
+
       def Spell.after_stance=(val)
         @@after_stance = val
       end
--- a/spec/lib/common/spell_spec.rb
+++ b/spec/lib/common/spell_spec.rb
@@ -17,6 +17,22 @@ RSpec.describe Lich::Common::Spell do
     end
   end
 
+  describe '.results_regex' do
+    it 'is the union of lines cast waits on, plus any extra lines' do
+      regex = Lich::Common::Spell.results_regex
+      expect(regex).to match('Cast Roundtime 3 Seconds.')
+      expect(regex).to match('Cast at what?')
+      expect(regex).to match("But you don't have any mana!")
+      expect(regex).to match('Your magic fizzles ineffectually.')
+      expect(regex).to match('You do not currently have a target.')
+      expect(regex).not_to match('You swing a broadsword at an orc!')
+
+      extra = Lich::Common::Spell.results_regex(results_of_interest: /^Roundtime: \d+ sec\.$/)
+      expect(extra).to match('Roundtime: 3 sec.')
+      expect(extra).to match('Cast at what?')
+    end
+  end
+
   describe '.[]' do
     context 'when looking up by number' do
       it 'finds Spirit Warding I by number 101' do


mrhoribu (approved):
# PR Review: feat(all): Spell.results_regex, the lines cast waits on, exposed (elanthia-online/lich-5#1590)

**Reviewed:** 2 files, +27/-0 at `21e1d0fa` on 2026-09-11. Based on: full local clone (fork/origin remotes present), PR head checked out into an isolated worktree at the reviewed commit; base (`origin/main`) diffed against head to confirm `cast` itself is byte-for-byte unchanged.
**Summary:** Adds `Spell.results_regex(results_of_interest: nil)`, a read-only accessor that returns the same `@@results_regex` union `#cast` builds from its private class variable, so a caller that drives its own cast/confirm loop (the PR cites an `eohunter` engine) can reuse the pattern instead of hand-copying it. The merge logic is copied verbatim from `cast`'s own `results_of_interest.is_a?(Regexp)` branch, follows an existing idiom already used by `Feat.use`/`Armor.use`/`Weapon.use`/`Shield.use`/`CMan.use`/`Warcry.use`, and ships a test that asserts real game strings. `cast` is untouched.

## Minor

### 1. The doc comment promises "every line," but the accessor omits two lines `cast` always adds — `lib/common/spell.rb:146-155`

```ruby
# Every line that answers a cast: the regex {#cast} waits on, exposed
# so a caller that sends and confirms on its own terms can wait on
# the same lines instead of copying them.
#
# @param results_of_interest [Regexp, nil] extra lines to match, as {#cast} takes
# @return [Regexp]
def Spell.results_regex(results_of_interest: nil)
  return @@results_regex unless results_of_interest.is_a?(Regexp)
  Regexp.union(@@results_regex, results_of_interest)
end
```

`cast` doesn't actually wait on `@@results_regex` (plus `results_of_interest`) alone — every single call through the main branch layers in one more pattern on top, unconditionally, depending on state:

```ruby
# lib/common/spell.rb:739-743 (unchanged by this PR)
if Effects::Spells.active?("Armored Casting")
  merged_results_regex = Regexp.union(/^Roundtime: \d+ sec.$/, merged_results_regex)
else
  merged_results_regex = Regexp.union(/^\[Spell Hindrance for/, merged_results_regex)
end
```

Neither `/^Roundtime: \d+ sec.$/` (the Armored Casting roundtime line) nor `/^\[Spell Hindrance for/` (the circle-10/bard hindrance line, which also drives `cast`'s own retry loop at line 776) is part of `@@results_regex`, so `Spell.results_regex` never includes either — the returned regex is a strict subset of what `cast`'s `dothistimeout` is actually watching for on every call. A caller who takes the doc comment literally and builds their own wait loop on `Spell.results_regex` alone will silently fail to terminate on those two lines when Armored Casting is active or a bard hits circle-10 hindrance, where `cast` itself would have stopped.

This doesn't break anything in this PR — `cast` is unchanged and nothing in this repo consumes the new method yet — but it directly undercuts the PR's own stated purpose (giving `eohunter` "the same lines" so its copy stops drifting from core). Worth either folding the state-dependent OR into the new method (mirroring what `cast` does), or tightening the doc/PR language so "every line" doesn't overclaim. The PR body itself is more careful here ("returns the same union `cast` builds" — true as written) than the in-code comment ("every line ... {#cast} waits on" — not quite true).

**Confidence:** confirmed (traced against the unchanged `cast` body at head, lines 733-743).

## Nits

- The new spec (`spec/lib/common/spell_spec.rb:20-33`) covers the no-arg case and a valid-Regexp `results_of_interest`, but not the "non-Regexp `results_of_interest` is silently ignored" branch (`return @@results_regex unless results_of_interest.is_a?(Regexp)`). Low stakes since it mirrors `cast`'s own existing behavior rather than introducing new behavior, but it's the one branch in this method with no test.

## What this PR gets right

- The merge logic (`results_of_interest.is_a?(Regexp) ? Regexp.union(...) : base`) is copied exactly from `cast`'s own branch at `lib/common/spell.rb:733-737`, so for the part it does cover, there is no drift between the two.
- It follows an established idiom in this repo rather than inventing one — `Feat.use`, `Armor.use`, `Weapon.use`, `Shield.use`, `CMan.use`, and `Warcry.use` all build a base `results_regex` and merge an optional `results_of_interest:` the same way, and several already carry the same YARD `@param results_of_interest [Regexp, nil]` doc style this PR uses.
- The test asserts against real game response strings (`'Cast Roundtime 3 Seconds.'`, `'Cast at what?'`, etc.) rather than just checking the regex object's shape, and includes a true-negative (`'You swing a broadsword at an orc!'`) — verified these against `@@results_regex`'s actual patterns and all seven assertions hold.
- Zero-risk shape: `cast` itself has no diff (confirmed via `git diff origin/main...head -- lib/common/spell.rb` showing only the 12 added lines), and `@@results_regex` is read-only here — no new mutation path, no thread-safety concern.
- Ran the actual suite and linter rather than trusting the PR's claims: `bundle exec rspec spec/lib/common/spell_spec.rb` — 36 examples, 0 failures. `rubocop` on both touched files reports 43 offenses, but diffing against `origin/main` shows all 43 are pre-existing (the "for backwards compatibility" alignment table at the bottom of the file) — zero new offenses on the added lines.

## Coverage notes

Full local clone used (fork + origin remotes, PR head checked out into an isolated `git worktree` so the reviewer's own uncommitted changes on `fix/script-log-log-dir` were never touched). Ran the real spec suite and rubocop under Ruby 4.0.5 (repo's pinned target) rather than reading statically. Confirmed `cast` is byte-identical to `origin/main` via direct file diff, not just by reading the PR diff. Searched both `dr-scripts` and `scripts` checkouts for the `eohunter` script the PR body cites as the motivating consumer — not present in either local checkout, so the real-world call site described in the PR body could not be directly inspected; this doesn't affect the correctness of the diff itself, which stands on its own and matches an existing in-repo idiom. A CodeRabbit pass was attempted (`coderabbit review --agent --base main --committed`) but failed with "Rate limit exceeded" before producing output, so no second-model pass is reflected in this review.

PR #1591, opened 2026-09-10
feat(gs): Group.join, the follower's side of Group.add

Nisugi:
## Summary
- `Group.add` is the leader adding a member: send `group #id`, read the three answers, update the cached roster. Nothing did the same for a follower joining a leader, so scripts sent `join` themselves and re-derived the answers.
- `Group.join(leader)` takes a noun or GameObj, sends `join #id` through `dothistimeout`, and returns `{ok:}`, `{noop:}` for already a member, or `{err:}` for a closed group, a missing player, or no answer. The observer's existing `JOINED_NEW_GROUP` term records the new leader and roster from the "You join" line, so `join` does not touch the cache itself.

## Test plan
- [x] New `spec/lib/gemstone/group_join_spec.rb`: by noun and by GameObj, the command sent, and each answer's classification; 2 examples green
- [ ] In game: `;e echo Lich::Gemstone::Group.join('Leadername')` next to an open group, then `Group.leader`

### DIFF
--- a/lib/gemstone/group.rb
+++ b/lib/gemstone/group.rb
@@ -230,6 +230,33 @@ module Lich
         end
       end
 
+      # Joins another player's group: the follower's side of {Group.add}.
+      # Sends JOIN and reads the game's answer; the observer records the
+      # new leader and the roster from the "You join" line itself.
+      #
+      # @param leader [String, GameObj] the leader's noun or GameObj
+      # @return [Hash] {ok: leader} joined, {noop: leader} already a member,
+      #   {err: leader} closed, gone, or no answer; {err: nil} when no such
+      #   player is in the room
+      # @example
+      #   Group.join("Etanamir")
+      def self.join(leader)
+        leader = GameObj.pcs.find { |pc| pc.noun.eql?(leader) } if leader.is_a?(String)
+        return { err: nil } if leader.nil?
+
+        result = dothistimeout("join ##{leader.id}", 3, Regexp.union(
+                                                          %r{^You join #{leader.noun}},
+                                                          %r{^#{leader.noun}'s group status is closed},
+                                                          %r{already a member of},
+                                                          %r{^What were you referring to\?}
+                                                        ))
+        case result
+        when %r{^You join} then { ok: leader }
+        when %r{already a member} then { noop: leader }
+        else { err: leader }
+        end
+      end
+
       # Returns array of all member IDs.
       #
       # @return [Array<String>] array of member IDs
--- a/lib/gemstone/group.rb
+++ b/lib/gemstone/group.rb
@@ -234,25 +234,34 @@ module Lich
       # Sends JOIN and reads the game's answer; the observer records the
       # new leader and the roster from the "You join" line itself.
       #
+      # A String that does not name a PC in the room is treated as a
+      # separator and delegated to +Array#join+ over the cached roster, the
+      # behavior {Group} used to reach through +method_missing+.
+      #
       # @param leader [String, GameObj] the leader's noun or GameObj
       # @return [Hash] {ok: leader} joined, {noop: leader} already a member,
-      #   {err: leader} closed, gone, or no answer; {err: nil} when no such
-      #   player is in the room
+      #   {err: leader} closed, gone, or no answer
+      # @return [String] the joined roster when given a separator
       # @example
       #   Group.join("Etanamir")
       def self.join(leader)
-        leader = GameObj.pcs.find { |pc| pc.noun.eql?(leader) } if leader.is_a?(String)
+        if leader.is_a?(String)
+          pc = GameObj.pcs.find { |member| member.noun.eql?(leader) }
+          return @@members.join(leader) if pc.nil?
+
+          leader = pc
+        end
         return { err: nil } if leader.nil?
 
         result = dothistimeout("join ##{leader.id}", 3, Regexp.union(
                                                           %r{^You join #{leader.noun}},
                                                           %r{^#{leader.noun}'s group status is closed},
-                                                          %r{already a member of},
+                                                          %r{already a member of #{leader.noun}'s group},
                                                           %r{^What were you referring to\?}
                                                         ))
         case result
         when %r{^You join} then { ok: leader }
-        when %r{already a member} then { noop: leader }
+        when %r{already a member of #{leader.noun}'s group} then { noop: leader }
         else { err: leader }
         end
       end
--- a/spec/lib/gemstone/group_join_spec.rb
+++ b/spec/lib/gemstone/group_join_spec.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+require_relative '../../spec_helper'
+
+require 'gemstone/group'
+
+# The follower's side of Group.add: JOIN a leader and read the answer.
+RSpec.describe Lich::Gemstone::Group do
+  GroupJoinPc = Struct.new(:id, :noun, :name) unless defined?(GroupJoinPc)
+
+  let(:leader) { GroupJoinPc.new('-1001', 'Etanamir', 'Etanamir') }
+  let(:sent) { [] }
+
+  before do
+    allow(GameObj).to receive(:pcs).and_return([leader])
+  end
+
+  def answer(line)
+    allow(described_class).to receive(:dothistimeout) { |cmd, _t, _rx| sent << cmd; line }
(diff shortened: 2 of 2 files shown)


mrhoribu (approved):
# PR Review: feat(gs): Group.join, the follower's side of Group.add (elanthia-online/lich-5#1591)

**Reviewed:** 2 files, +65/-0 at `7e4f8d9` on 2026-09-11. Based on: full clone (local worktree of the PR branch, merge-base `073d23a9` against current `origin/main`), plus full history of `lib/gemstone/group.rb`.
**Summary:** Adds `Lich::Gemstone::Group.join(leader)`, the follower-side counterpart to the existing `Group.add` — it sends `join #id`, classifies the game's answer into `{ok:}`/`{noop:}`/`{err:}`, and deliberately leaves cache updates to the pre-existing `JOINED_NEW_GROUP` observer term rather than touching `@@members`/`@@leader` itself. The implementation is a careful, well-grounded mirror of `Group.add`'s established pattern, backed by a green, rubocop-clean spec, and the full 635-example suite still passes. The two things worth a look before merge are a shadowed `Array#join` delegation and a loosely-scoped match pattern.

## Minor

### 1. `Group.join` silently shadows the `Array#join` behavior `Group` used to delegate — `lib/gemstone/group.rb:243`

```ruby
def self.method_missing(method, *args, &block)
  @@members.send(method, *args, &block)
end
```

`Group` has no `Array`-derived superclass, but `method_missing` (`lib/gemstone/group.rb:303-305`, unchanged by this PR) forwards any unrecognized call to `@@members`, which is a plain `Array`. Before this PR, `Group.join(", ")` resolved through that delegation to `Array#join`, returning a comma-joined string of member representations — I confirmed this with a standalone repro of the same `method_missing` pattern:

```
$ ruby -e '... p Foo.join(", ")'
"a, b, c"
```

After this PR, `Group.join(", ")` calls the new explicit method instead: it treats `", "` as a player noun, finds no matching `GameObj.pcs` entry, and returns `{err: nil}` — a silent, no-exception change from a joined String to an error Hash for any caller that relied on the old delegation. I grepped `lib/`, `dr-scripts`, and `scripts` for `Group.join` and found no such caller in any of the four repos this workspace tracks, so nothing here breaks visibly. The residual risk is any Lich script outside those repos (there is a `Group.to_s` for exactly this stringification use case, so it would be an unusual choice, but `method_missing` delegation is transparent enough that someone could have leaned on it without realizing). Worth a one-line callout in the PR description or changelog so anyone hitting the new silent `{err: nil}` return has a pointer to why.

**Confidence:** confirmed (the shadowing and its return-value change), suspected for real-world impact (no caller found in-repo).

### 2. The "already a member" match isn't scoped to the specific leader, unlike its sibling patterns — `lib/gemstone/group.rb:247-252`

```ruby
result = dothistimeout("join ##{leader.id}", 3, Regexp.union(
                                                  %r{^You join #{leader.noun}},
                                                  %r{^#{leader.noun}'s group status is closed},
                                                  %r{already a member of},
                                                  %r{^What were you referring to\?}
                                                ))
```

The other three alternatives are all scoped to `leader.noun`; `already a member of` is not, unlike `Group.add`'s equivalent (`%r{But #{member.noun} is already a member of your group}`, `lib/gemstone/group.rb:217`). `dothistimeout` (`lib/global_defs.rb:2029`) reads the script's downstream stream line-by-line for up to 3 seconds and returns as soon as any union alternative matches — and that stream includes third-party room broadcasts, which is exactly how `OTHER_JOINED_GROUP` (`lib/gemstone/group.rb:407`) observes other players' group activity. If another player's own "already a member of ...'s group" message happens to land in that 3-second window, `Group.join` would return `{noop: leader}` for a join that hasn't actually resolved, and the real answer line for the caller's own join would be silently dropped since `dothistimeout` already returned. Low probability, but the fix is cheap: scope the pattern to `leader.noun` the way the other three alternatives (and `Group.add`) already do.

**Confidence:** confirmed as written; the triggering scenario (a bystander's group message landing in the same window) is suspected, not reproduced.

## Nits

- `spec/lib/gemstone/group_join_spec.rb:78` simulates "no answer" with `answer(false)`, but `dothistimeout` actually returns `nil` on timeout (`lib/global_defs.rb:2090`). No functional difference here — `Regexp#===` returns `false` for both `nil` and `false` — so the test still exercises the right code path, but `nil` would be the more faithful stand-in.
- `lib/gemstone/group.rb:249`'s closed-group pattern is anchored (`^#{leader.noun}'s group status is closed`) where `Group.add`'s otherwise-identical pattern (`lib/gemstone/group.rb:216`) is not. Likely harmless since the message appears to open the line either way, but worth a glance if `Group.add`'s unanchored version was deliberately left loose for a reason not visible in this diff.

## Open questions

- The PR checklist's third item — `In game: ;e echo Lich::Gemstone::Group.join('Leadername') next to an open group, then Group.leader` — is left unchecked. Everything I could verify statically (regex scoping against the plain-text stream, the observer's pre-existing `JOINED_NEW_GROUP` dispatch, the rubocop/spec run) checks out, but nobody has confirmed against a live game session that the answer text still matches these patterns exactly. **Settles it:** has the in-game checklist item been run since this diff was written?

## What this PR gets right

- Correctly leans on the pre-existing `JOINED_NEW_GROUP` observer term (`lib/gemstone/group.rb:356`, `519-522`, wired into `Group::Observer.consume` via `lib/gemstone/infomon/xmlparser.rb:571-574`, all unchanged by this diff) instead of re-implementing cache updates — the PR body's reasoning ("the observer already records the new leader and roster from the 'You join' line itself") holds up against the actual dispatch code.
- The plain-text match patterns (`^You join #{leader.noun}`, etc.) correctly target `dothistimeout`'s stripped script-buffer stream rather than the raw XML the observer parses — exactly mirroring `Group.add`'s already-working convention, not a new assumption.
- New spec covers the noun path, the GameObj path, the exact command sent, closed/missing-player/no-answer-as-error, and already-member-as-noop — a solid table of the documented return contract.
- Docstring accurately describes all four outcomes, including the `{err: nil}` "no such player in the room" case, which the spec verifies.

## Coverage notes

- Access: full local clone via `git worktree` on a fetched `refs/pull/1591/head`, diffed against the true merge-base (`073d23a9`, i.e. `origin/main` — the local `main` ref in this checkout was stale at an older commit, so the merge-base was recomputed against `origin/main` before trusting the diff).
- Ran `bundle exec rspec spec/lib/gemstone/group_join_spec.rb` (2/2 green) and the full `spec/lib/gemstone/` suite plus the whole repo suite (635/635 green) — no regressions.
- Ran `bundle exec rubocop lib/gemstone/group.rb spec/lib/gemstone/group_join_spec.rb` — clean, no offenses.
- Attempted a local CodeRabbit pass (`coderabbit review --agent --committed --base-commit ...`); it hit this account's free-tier rate limit ("You've used all 3 included reviews currently available", 55-minute cooldown) before returning results, so no CodeRabbit findings are included here — this review is based entirely on manual inspection plus history/grep grounding across all four repos in this workspace.
- Did not verify in-game; the PR's own manual-test checklist item is unchecked (see Open questions).

Nisugi:
Thanks for the review — all three picked up in b7ff0baf.

One note on the `already a member of` scoping, because the fix was a bit larger than it looked. Scoping the `Regexp.union` alternative alone would not have changed any behavior: the `case` below it still classified on a bare `%r{already a member}`, and that is what actually decides the return value. The union only gates which line `dothistimeout` stops on, so a bystander's group message would still have come back and still been read as `{noop:}`. Both are scoped to `leader.noun` now. The new spec feeding in another player's group message fails against the union-only version, which is how it surfaced.

The `Array#join` shadowing is delegating again: a String that names no PC in the room falls through to `@@members.join(separator)`, so `Group.join(", ")` behaves as it did pre-PR, while a String naming a PC joins their group.

One consequence worth flagging — the `{err: nil}` missing-player result is now only reachable when a `GameObj` argument is nil, since a String never reaches that branch. I dropped the `Group.join('Nobody')` assertion rather than leave it asserting something the code no longer does. If a distinct error for an unknown player name is worth keeping, that wants a different discriminator than "is it a String" and I'm happy to take a suggestion.

And you were right about `dothistimeout` — `return nil` at `lib/global_defs.rb:2090`. Spec stubs `nil` now.

Verified: 637 gemstone examples green, 4 in the join spec, `rubocop` clean on both touched files.

PR #1592, opened 2026-09-11
feat(gs): emit post-commit combat recorder receipts

therealatari:
## Summary

- emit a frozen `:recorded_attack` observer receipt only after an attack transaction commits
- identify the recorder, database file, session, and exact attack row without changing the seven-table schema
- carry validated ingestion provenance when the combat processor provides it, otherwise report `source: nil`
- isolate optional subscriber failures from recording and preserve rollback behavior
- document the new observer event contract

## Motivation

Trusted local consumers currently have no race-free way to learn which SQLite row a just-recorded combat event committed. Parsing player-facing report output or polling for the newest row loses attribution under concurrent, delayed, or foreign combat. This narrow observer receipt lets consumers bind evidence to exact committed rows without starting a recorder, changing its schema, or moving database work into the combat parser.

The database path and file identity are explicitly trusted-local locators. They are not intended for remote protocol exposure. A receipt is evidence that one attack row committed; it does not prove that an encounter or hunt is complete.

## Compatibility

The receipt contract is independently useful on current `main`. When combat observation provenance from PR #1576 is available, `source` contains that validated scalar context; on legacy/replay input it remains `nil`. No consumer is required and ordinary recording behavior is unchanged.

## Verification

- `bundle exec rspec spec/lib/gemstone/combat/recorder_spec.rb`: 34 examples, 0 failures
- `bundle exec rubocop lib/gemstone/combat/recorder.rb lib/gemstone/combat/observers.rb spec/lib/gemstone/combat/recorder_spec.rb`: no offenses
- `git diff --check`

Coverage verifies post-commit readability, rollback silence, immutable receipts, in-memory databases, and subscriber-failure isolation.

### DIFF
--- a/lib/gemstone/combat/observers.rb
+++ b/lib/gemstone/combat/observers.rb
@@ -41,6 +41,12 @@
 #                 chunk), :death (subject already known dead - stack
 #                 cleanup, not meaningful expiry), or nil (natural
 #                 expiry, or cause not visible in this chunk)
+#   :recorded_attack { protocol:, recorder_id:, database:, file_identity:,
+#                      session_id:, attack_id:, source: } - emitted by
+#                 Combat::Recorder only after the complete attack transaction
+#                 commits. Local observers can use the opaque IDs and trusted
+#                 local database identity to read that exact row. source is
+#                 validated ingestion provenance when available, otherwise nil.
 #
 # @example
 #   Combat::Tracker.on(:damage) { |type, data| my_queue << data }
--- a/lib/gemstone/combat/recorder.rb
+++ b/lib/gemstone/combat/recorder.rb
@@ -369,12 +369,7 @@ module Lich
           # recorder lock so consumers may inspect the database without deadlock.
           # This is evidence only, not command ownership or a combat-complete flag.
           if type == :attack && result.is_a?(Hash) && result[:protocol] == RECEIPT_PROTOCOL
-            begin
-              Lich::Gemstone::Combat::Observers.emit(:recorded_attack, result)
-            rescue StandardError
-              # Optional receipt delivery cannot undo or break a committed record.
-              nil
-            end
+            Lich::Gemstone::Combat::Observers.emit(:recorded_attack, result)
           end
           result
         end
@@ -540,7 +535,8 @@ module Lich
           parent_row = (parent_uid && chunk_row(parent_uid))
           parent_conf = event[:parent_confidence]&.to_s
 
-          receipt = nil
+          attack_id = nil
+          touched = Set.new([target[:id]].compact.map(&:to_i))
           in_txn do
             @seq += 1
             params = [@session_id, @seq, at, event[:name].to_s, parent&.to_s, txt(parent_weapon),
@@ -584,7 +580,6 @@ module Lich
               insert_hit(attack_id, nil, creature_row, hit_seq += 1, hit, at)
             end
 
-            touched = Set.new([target[:id]].compact.map(&:to_i))
             (event[:flares] || []).each_with_index do |flare, i|
               f_target = flare[:target_info]
               f_creature = f_target ? ensure_creature(f_target, at) : nil
@@ -603,19 +598,29 @@ module Lich
                 insert_hit(attack_id, flare_id, f_creature || creature_row, hit_seq += 1, hit, at)
               end
             end
-
-            @open_attack = { id: attack_id, creature_ids: touched, inbound: !!event[:inbound] }
-            provenance = if defined?(Processor) && Processor.respond_to?(:observation_source)
-                           Processor.observation_source(event[:source])
-                         end
-            receipt = {
-              protocol: RECEIPT_PROTOCOL, recorder_id: @recorder_id,
-              database: @receipt_path&.dup&.freeze, file_identity: @receipt_file,
-              session_id: @session_id, attack_id: attack_id,
-              source: provenance
-            }.freeze
           end
-          receipt
+          # Publish the attack window only after commit; a rollback must not
+          # leave statuses pointing at an attack row that no longer exists.
+          @open_attack = { id: attack_id, creature_ids: touched, inbound: !!event[:inbound] }
+          attack_receipt(attack_id, event)
+        end
+
+        def attack_receipt(attack_id, event)
+          provenance = if defined?(Processor) && Processor.respond_to?(:observation_source)
+                         Processor.observation_source(event[:source])
+                       end
+          {
+            protocol: RECEIPT_PROTOCOL, recorder_id: @recorder_id,
+            database: @receipt_path&.dup&.freeze, file_identity: @receipt_file,
+            session_id: @session_id, attack_id: attack_id,
+            source: provenance
+          }.freeze
+        rescue StandardError => e
+          # Receipt metadata is optional. The attack is already committed, so
+          # report this separately from a persistence failure and emit nothing.
+          msg = "CombatRecorder receipt for committed attack #{attack_id}: #{e.message}"
+          defined?(Lich) && Lich.respond_to?(:log) ? Lich.log("error: #{msg}") : warn(msg)
+          nil
         end
 
         def insert_resolution(attack_id, flare_id, seq, res)
--- a/lib/gemstone/combat/recorder.rb
+++ b/lib/gemstone/combat/recorder.rb
@@ -87,12 +87,19 @@
 # town time never pads a hunt's duration.
 
 require 'sqlite3'
+require 'securerandom'
 
 module Lich
   module Gemstone
     module Combat
       class Recorder
         HANDLER_NAME = 'combat_recorder'
+        # :recorded_attack is an optional post-commit observer receipt. Its
+        # recorder/session/attack IDs identify rows; :source retains validated
+        # ingestion context (nil for legacy/replay input). :database and
+        # :file_identity are trusted-local locators, not remote protocol data.
+        # Receipts neither enable recording nor prove an encounter is complete.
+        RECEIPT_PROTOCOL = 1
 
         SCHEMA = <<~SQL
           CREATE TABLE IF NOT EXISTS sessions (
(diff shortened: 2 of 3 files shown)


mrhoribu (changes requested):
# PR Review: feat(gs): emit post-commit combat recorder receipts (elanthia-online/lich-5#1592)

**Reviewed:** 3 files, +97/-1 at `b2453385` (branch `feat/combat-recorder-receipts`, base `main` @ `526ad7cd`) on 2026-09-11. Based on: full local clone (`gh` CLI, worktree checkout, `git log`/`git blame` history).
**Verdict:** Request changes
**Summary:** Adds a frozen, post-commit `:recorded_attack` observer receipt to `Combat::Recorder#record` so trusted local consumers can bind evidence to an exact committed SQLite row. The emit-after-unlock design and rollback-silence behavior are correctly implemented and well tested, but the new provenance-lookup code executes *inside* the same DB transaction as the actual write, immediately after an existing, unstaged in-memory pointer (`@open_attack`) is set to the new row — reopening a dangling-reference bug class the file's own comments describe as already solved for two sibling caches.

## Major

### 1. `@open_attack` is left pointing at a rolled-back row if the new provenance lookup raises — `lib/gemstone/combat/recorder.rb:607-618`

```ruby
@open_attack = { id: attack_id, creature_ids: touched, inbound: !!event[:inbound] }
provenance = if defined?(Processor) && Processor.respond_to?(:observation_source)
               Processor.observation_source(event[:source])
             end
receipt = {
  protocol: RECEIPT_PROTOCOL, recorder_id: @recorder_id,
  database: @receipt_path&.dup&.freeze, file_identity: @receipt_file,
  session_id: @session_id, attack_id: attack_id,
  source: provenance
}.freeze
```

This block runs inside `in_txn`'s `@db.transaction { yield }` (recorder.rb:449). Before this PR, `@open_attack = {...}` was the *last* statement in that transaction body — nothing could raise after it, so it was safe for `@open_attack` to be a plain, unstaged ivar assignment. `git log -S "@open_attack = { id: attack_id"` shows that line dates to #1559 and was untouched by this diff; this PR is what inserts new code (`provenance = ...`, `receipt = {...}.freeze`) *after* it, inside the still-open transaction.

If `Processor.observation_source(event[:source])` ever raises, `@db.transaction` rolls back the whole attack (the just-inserted `attacks`/`hits`/`resolutions`/`flares` rows are undone), the exception propagates to `record_locked`'s catch-all `rescue StandardError => e` (recorder.rb:406-409) and is swallowed into a log line — but `@open_attack` was already mutated to `{ id: attack_id, ... }` pointing at the row that just got rolled back. `in_txn`'s `ensure` only resets `@pending_cache`/`@pending_chunk_rows` (recorder.rb:455-459); `@open_attack` isn't part of that staging mechanism at all.

The next `:status`/`:stun`/`:roundtime` event naming one of the touched creatures will then hit `record_status`'s window-attribution check (recorder.rb:699-705), read `@open_attack[:id]`, and insert a `statuses` row whose `attack_id` references a nonexistent `attacks.id`. `initialize` never sets `PRAGMA foreign_keys = ON` (only `busy_timeout`, `journal_mode`, `synchronous` are set at recorder.rb:255-257), so SQLite won't reject this — it silently inserts a dangling FK. This is exactly the failure mode `in_txn`'s own doc comment warns about for the two ivars it *does* protect: "a naive in-memory write would survive, so the next record would hand out a row id that no longer exists (dangling FK)" (recorder.rb:438-439) — `@open_attack` is now the unprotected third case.

**Caveat on reachability today:** `Processor.observation_source` doesn't exist anywhere in this codebase yet (repo-wide grep is empty outside this one call site), so `Processor.respond_to?(:observation_source)` is always `false` right now and `provenance` never actually invokes anything that can throw — this specific path cannot fire with code currently in the repo. The PR body says this method is expected to land via a companion PR (#1576). That makes this a landmine rather than a live bug: it will not reproduce today, but the moment `observation_source` exists and raises on any input shape, this silently corrupts attack-window attribution in a way that will be very hard to trace back to "receipt-building runs inside the transaction."

**Fix:** hoist `attack_id` next to the already-hoisted `receipt = nil` (recorder.rb:543) and move the `provenance =`/`receipt = {...}.freeze` construction to *after* `in_txn do ... end` returns — i.e., truly post-commit, matching what `record`'s own comment already claims ("Receipts name committed rows, never pending writes," recorder.rb:368-369). That also removes any need to reason about what else might throw inside the transaction body.

**Confidence:** confirmed for the described mechanism (traced against the code and its history); the trigger itself is not yet reachable in this repo — noted above rather than overstated.

## Minor

### 2. The `rescue` around `Observers.emit(:recorded_attack, ...)` guards against something the callee's own contract rules out — `lib/gemstone/combat/recorder.rb:371-378`

```ruby
begin
  Lich::Gemstone::Combat::Observers.emit(:recorded_attack, result)
rescue StandardError
  # Optional receipt delivery cannot undo or break a committed record.
  nil
end
```

`Observers.emit` (observers.rb:104-114) already isolates every subscriber's exception internally (`rescue StandardError => e` around each `handler.call`) and its own doc comment states subscriber errors are "isolated and logged, never raised to the caller (the processor)." `spec/lib/gemstone/combat/observers_spec.rb:46-50` (`'isolates and logs raising subscribers without breaking others'`) asserts exactly this. So in real usage `Observers.emit` cannot raise, and this `rescue` is dead code — the only way to exercise it is the new test at recorder_spec.rb:125-132, which stubs `Observers.emit` itself to raise, i.e. tests a violation of `Observers`'s documented contract rather than real behavior. Not harmful to keep as a defensive line, but it's worth knowing it currently protects against something structurally impossible, and the new test doesn't actually validate the real `record` → `Observers.emit` interaction path.

## What this PR gets right

- Emitting the receipt *outside* `@mutex.synchronize` (recorder.rb:367-379) is the correct call: `@mutex` is a plain non-reentrant `Mutex`, and a subscriber that calls back into the same recorder (e.g. `record`/`check_idle!`) while still inside the lock would self-deadlock. The comment says as much, and the code matches it.
- The rollback-silence guarantee is real and correctly tested: `record_attack` only assigns `receipt` as the last statement inside `in_txn`'s transaction body, so a mid-transaction failure (verified by stubbing `insert_hit` to raise in recorder_spec.rb:162-170) never reaches `record`'s emit check, because `record_locked`'s outer rescue swallows the exception before `record` ever sees a Hash result.
- `:database`/`:file_identity` being computed once at `initialize` (via `File.realpath`/`File.stat`, recorder.rb:251-253) rather than per-record avoids repeated syscalls and correctly returns `nil` for `:memory:` recorders (`File.file?(':memory:')` is `false`), which the new in-memory test (recorder_spec.rb:134-141) locks in.
- Previously, `record(:attack, ...)`'s return value was whatever `@chunk_rows.merge!(@pending_chunk_rows)` happened to return — an accidental, meaningless value with (per the spec file's own header) zero first-party callers depending on it. Replacing that with a documented, frozen, versioned (`RECEIPT_PROTOCOL`) contract is a clean improvement with no compatibility downside.
- Documentation is thorough and accurate: the `observers.rb` event-type table addition and the new YARD doc on `record` both match what the code actually does.

## Coverage notes

- Access level: full local clone via `gh pr checkout`-equivalent (`git fetch` + `git worktree add`), diffed against a freshly fetched `origin/main` (the pre-existing local `main` was stale by ~50 commits, so the initial `git diff main...pr-1592` was discarded in favor of `gh pr diff`).
- `bundle exec rspec spec/lib/gemstone/combat/recorder_spec.rb`: 34 examples, 0 failures (verifies the PR's own stated test run).
- `rubocop lib/gemstone/combat/recorder.rb lib/gemstone/combat/observers.rb spec/lib/gemstone/combat/recorder_spec.rb`: no offenses.
- A local CodeRabbit pass (`coderabbit review --agent --base main --committed`) was attempted but hit the organization's usage rate limit ("You've used all 3 included reviews currently available") before producing any findings — no CodeRabbit output was available to cross-check.
- Verified via repo-wide grep that `Combat::Recorder` currently has no first-party callers besides its own spec (matches the spec file's header comment), so there is no blast radius outside this file/spec pair to check for contract breakage.
- Did not attempt to reproduce the Major finding at runtime, since it requires a `Processor.observation_source` implementation that doesn't exist in this repo — the finding rests on static trace of the transaction/rollback mechanics plus `git log -S` confirming the newly-added code is what changed the risk profile of the pre-existing `@open_attack` assignment.

---

therealatari:
Both review points are addressed in c2cd01cf, followed by the current-main conflict resolution in 33079674.

Recorder state associated with the transaction — both @open_attack and main's @chunk_flares — is now published only after a successful commit. Optional receipt metadata is assembled afterward. A metadata failure therefore cannot roll back a persisted attack or leave later status attribution pointing at a rolled-back row; it is diagnosed as a receipt failure for an already committed attack.

The redundant rescue around Observers.emit is removed. The subscriber test now uses a real raising observer, exercising the existing observer-isolation contract. Receipt emission remains outside the recorder mutex.

Added regressions for metadata failure after persistence and late transaction rollback; both reproduced the original failures before the fix. The merge retains upstream session notifications, schema, and flare/status behavior.

Verification: 334 combat examples, zero failures; changed Ruby files pass RuboCop. A temporary integration checkout with all four updated PR heads against main b9df96c4 passes 7,439 examples and full RuboCop (1,160 files).

mrhoribu (approved):
Follow-up review of `c2cd01c` (`fix(combat): isolate receipt metadata from committed attacks`).

This resolves both items from my previous review:

- **Major** (`@open_attack` left dangling after a receipt-metadata failure rolls back the transaction): fixed by hoisting `attack_id`/`touched` before `in_txn` and moving the `@open_attack` assignment to *after* `in_txn` returns, so it's only ever set once the row is actually committed. Receipt construction (including the `Processor.observation_source` call) now happens in its own `attack_receipt` method, called post-commit, with its own rescue that logs separately and returns `nil` without touching the already-committed attack.
- **Minor** (redundant `rescue` around `Observers.emit`, which its own contract already makes unreachable): removed, and the test that used to stub `Observers.emit` itself was replaced with a real integration test exercising actual subscriber isolation via `Observers.on`.

New tests directly cover the fixed scenario — a provenance failure now leaves the attack, its hits, and correct status-window attribution intact (`'keeps committed attacks and status attribution when receipt provenance fails'`), and a second test confirms the same holds for a failure right at the commit boundary.

Verified on the current PR head (post-merge from `main`): full `spec/lib/gemstone/combat/` suite (334 examples) passes, and `rubocop` on the changed files reports no offenses. The fix correctly carries through the merge, including extending the same post-commit-only pattern to the newly-merged `@chunk_flares` publish.

PR #1593, opened 2026-09-11
fix(gs): combat module - hunt-log audit: interrupted-swing resume, status attribution by event, spell-loss cause, live defs

Nisugi:
## Summary

Follow-up to #1559. Sixteen commits from auditing the combat recorder against raw XML hunt logs after each live hunt: the recorded database was compared line by line with the logs (fires incl. aimed shots and echoes, volleys, lashes, blinds, stuns, coups, deaths, damage lines) and every divergence was traced to a parser or def defect and fixed, each with a spec built from the offending log lines. Scope is `lib/gemstone/combat/**` and its specs only; nothing outside the module changes.

## Parser / processor

- **Interrupted swings resume correctly.** A creature's reactive cast, a cloak-of-shadows flare, or a one-hit side effect (a rider pinned under a collapsing mount) printing in the middle of our swing no longer swallows the rest of it: the swing is parked, the interloper is saved as its own event, and the following flare/hit lines return to the swing. `SINGLE_HIT_ATTACKS` (`mount_collapse`) formalises the one-hit case.
- **Statuses ride the event that caused them.** `apply_status_to_target` carries the originating event (`attack_uid`), so a status is filed on the flare or swing that inflicted it instead of on whatever event is current when the status line prints. Flare-and-status lines (`status_flare_seq`) resolve to the right flare index; pre-flares never claim statuses.
- **A nocked bow whose fire line is pre-empted** ("bounds to safety as you move to attack it") is recorded as a `:fire` with the bow as weapon, not an `:unknown`.
- **Crit-rider SMRs** (knockdown topple) stay on the hit that caused them.
- **Volley rounds that open with a miss**, death watch until the actual death line, possessive creature links, dispel-on-nock pre-flares held across a chunk boundary, bow pre-flares claimed regardless of arrow/bow name mismatch.
- Narration lines ("forced out of hiding", rage shriek) are not target switches; foreign (other players') events never open a session.
- **Spell wear-offs in the killing chunk carry `cause: :death`**, dispels carry `:dispel`, so "spells falling off a corpse" is distinguishable from a dispel flare. "A white glow rushes away from X" is the 303 Prayer of Protection end message, not a `:dispelled` status.

## Defs (all from live hunt logs, evidence in comments)

- Ticks: bleed, rot, neck bleed, ooze vitality drain (unowned on a creature, inbound on us).
- Attacks: cannibal ambush swing, leech fling, tangleweed grab miss, tusks, live fear maneuvers, feint and shield push, topple/hamstring/push variants, barrier-only inbound resolutions.
- Mutant-farm creatures: ooze, disciple, malady, flaming aura.
- Third-person casts by foreign casters: spellsong (Kalithra), fear cry, golden waves, moonbeam; second-person evoked moonbeam.
- Outcomes: "You duck to narrowly avoid the flying vermiforms!" evade.

## Recorder

- Statuses are routed by `attack_uid` with source `'event'` (authoritative) alongside the existing `'window'`/`'direct'` sources.
- Spell losses persist their cause.

## Tests

- `processor_inbound_spec.rb` +646 lines of log-derived examples; new fixtures `cloak_of_shadows_interrupt.txt`, `dispel_on_nock.txt`, `volley_miss_first.txt`, `bloom_forced_out_of_hiding.txt`, `ooze_splatter_fire.txt`.
- `bundle exec rspec spec/lib/gemstone/combat`: 289 examples, 0 failures. Rubocop clean on the touched files.

## Notes

- Touches only files that upstream has not changed since #1559 merged, so it rebases cleanly on `main`.
- Overlaps #1576 (observation provenance) in `processor.rb` at four mechanical hunks (the `save_event` dedup lambda vs `include_attack_events`, and `source:` on the outcome-opened event). Happy to rebase onto whichever lands first.

### DIFF
--- a/lib/gemstone/combat/defs/attacks.rb
+++ b/lib/gemstone/combat/defs/attacks.rb
@@ -338,6 +338,11 @@ module Lich
               # whip's miss rides the same line (:miss outcome)
               /(?<attacker>.+?) manifests a thick pseudopod and brings it smashing down at (?<target>you|.+?)!/,
               /(?<attacker>.+?) whips a thick pseudopod at (?<target>you|.+?)!/,
+              # the ooze's vitality drain: touch, then "Dizziness rushes
+              # through you..." and the damage line
+              /(?<attacker>.+?) whips a pseudopod toward (?<target>you|.+?), brushing/,
+              # halfling cannibal ambush swing (hunt log 2026-09-07 23:34)
+              /With an ululating shriek, (?<attacker>.+?) leaps from the shadows and hammers blindly at (?<target>you|.+?) with grimy little fists!/,
               /(?<attacker>.+?) snaps at (?<target>.+?) with its (?<weapon>[^!]+)!/,
               /(?<attacker>.+?) pounds at (?<target>.+?) with #{MK_PRE}(?:his|her|its)#{MK_POST} .*?fists?!/,
               /(?<attacker>.+?) tries to bite (?<target>[^!]+)!/,
@@ -471,7 +476,13 @@ module Lich
             AttackDef.new(:bleed, [
               /Blood weeps from (?<target>.+?)'s#{MK_POST} open .+? wound\./,
               /(?<target>.+?)'s#{MK_POST} .+? drips as #{MK_PRE}(?:he|she|it)#{MK_POST} continues to bleed\./,
+              /Trickles of blood course from (?<target>.+?)'s#{MK_POST} .+?\./,
               /(?<target>Your) .+? drips as you continue to bleed\./
+            ].freeze),
+            # Rot tick (a nearby player's curse on a creature; hunt log
+            # 2026-09-07 23:36). Unowned like bleed.
+            AttackDef.new(:rot, [
+              /Skin peels off (?<target>.+?)'s#{MK_POST} body, exposing rotting flesh\./
             ].freeze)
           ].freeze
 
--- a/lib/gemstone/combat/defs/attacks.rb
+++ b/lib/gemstone/combat/defs/attacks.rb
@@ -378,7 +378,11 @@ module Lich
               /(?<attacker>.+?) leaps from the shadows and (?:throws #{MK_PRE}(?:his|her|its)#{MK_POST} .+? around|hurtles at) (?<target>[^,!]+)/
             ].freeze),
             # PSM maneuvers, third person
-            AttackDef.new(:hamstring, [/(?<attacker>.+?) lunges forward and tries to hamstring (?<target>.+?) with #{MK_PRE}(?:his|her|its)#{MK_POST} .+?!/].freeze),
+            AttackDef.new(:hamstring, [
+              /(?<attacker>.+?) lunges forward and tries to hamstring (?<target>.+?) with #{MK_PRE}(?:his|her|its)#{MK_POST} .+?!/,
+              # warg jaws form (hunt log 2026-09-07)
+              /With a quick lunge, (?<attacker>.+?) tries to hamstring (?<target>you|.+?) with #{MK_PRE}(?:his|her|its)#{MK_POST} jaws!/
+            ].freeze),
             AttackDef.new(:shield_bash, [
               /(?<attacker>.+?) lunges forward at (?<target>.+?) with #{MK_PRE}(?:his|her|its)#{MK_POST} .+? and attempts a shield bash!/,
               /(?<attacker>.+?) launches a quick bash with (?<weapon>.+?) at (?<target>[^!]+)!/,
--- a/lib/gemstone/combat/defs/attacks.rb
+++ b/lib/gemstone/combat/defs/attacks.rb
@@ -287,11 +287,23 @@ module Lich
             # Creature fear maneuvers (SSR follows, then our save/fail line).
             # Room-wide, no target named - ROOM_TARGETED classifies them
             # inbound (hunt log 2026-09-07: Ojandhaart warg / mastodon).
+            # (the pronouns are links in the live feed: "sits back on <its>
+            # haunches" - hence MK_PRE/MK_POST; the bare form never matched)
             AttackDef.new(:howl, [
-              /(?<attacker>.+?) sits back on its haunches and unleashes a long, high-pitched howl that sends a shiver of primal terror down your spine\./
+              /(?<attacker>.+?) sits back on #{MK_PRE}its#{MK_POST} haunches and unleashes a long, high-pitched howl that sends a shiver of primal terror down your spine\./
             ].freeze),
             AttackDef.new(:trumpet, [
-              /(?<attacker>.+?) raises its trunk and rears back onto its immense hind legs, blaring out a note of sheer fury!/
+              /(?<attacker>.+?) raises #{MK_PRE}its#{MK_POST} trunk and rears back onto #{MK_PRE}its#{MK_POST} immense hind legs, blaring out a note of sheer fury!/
+            ].freeze),
+            # Shield-maiden targe push (SMR follows; hunt log 2026-09-07)
+            AttackDef.new(:shield_push, [
+              /(?<attacker>.+?) raises #{MK_PRE}(?:his|her|its)#{MK_POST} (?<weapon>.+?) and attempts to push (?<target>you|.+?) away!/
+            ].freeze),
+            # 3p feint: the result line is the only line ("X feints high, but
+            # you aren't fooled") - target captured as "you" so it classifies
+            # inbound; the same line is the :evade outcome
+            AttackDef.new(:feint, [
+              /(?<attacker>.+?) feints (?:high|low|to the (?:left|right)), but (?<target>you) aren't fooled for a second\./
             ].freeze),
             # Maneuver-style strike opener (round-14 sweep: 40/40 resolve
             # in the very next line, usually the target's evanescent
@@ -312,7 +324,7 @@ module Lich
               # no comma in the target: "Fear still claws at your heart, but
               # you stand fast..." is a fear-save outcome, not a claw swing
               /(?<attacker>.+?) claws at (?<target>[^!,]+)!/,
-              /(?<attacker>.+?) tries to spear (?<target>.+?) with its enormous tusks!/,
+              /(?<attacker>.+?) tries to spear (?<target>.+?) with #{MK_PRE}its#{MK_POST} enormous tusks!/,
               /(?<attacker>.+?) snaps at (?<target>.+?) with its (?<weapon>[^!]+)!/,
               /(?<attacker>.+?) pounds at (?<target>.+?) with #{MK_PRE}(?:his|her|its)#{MK_POST} .*?fists?!/,
               /(?<attacker>.+?) tries to bite (?<target>[^!]+)!/,
--- a/lib/gemstone/combat/defs/attacks.rb
+++ b/lib/gemstone/combat/defs/attacks.rb
@@ -295,6 +295,15 @@ module Lich
             AttackDef.new(:trumpet, [
               /(?<attacker>.+?) raises #{MK_PRE}its#{MK_POST} trunk and rears back onto #{MK_PRE}its#{MK_POST} immense hind legs, blaring out a note of sheer fury!/
             ].freeze),
+            # Mutant-farm room maneuvers (hunt log 2026-09-07 23:19-23:20):
+            # sanguine ooze crystalline shrapnel burst (SMR, then our dodge)
+            AttackDef.new(:shrapnel_spray, [
+              /Froth disturbs the surface of (?<attacker>.+?) as bubbling bulges form over #{MK_PRE}its#{MK_POST} surface/
+            ].freeze),
+            # flayed gigas disciple's spatial rift (SMR follows)
+            AttackDef.new(:rift_tentacles, [
+              /Zeal twisting #{MK_PRE}(?:his|her|its)#{MK_POST} features, (?<attacker>.+?) raises a raw and fleshless hand overhead and draws it down/
+            ].freeze),
             # Shield-maiden targe push (SMR follows; hunt log 2026-09-07)
             AttackDef.new(:shield_push, [
               /(?<attacker>.+?) raises #{MK_PRE}(?:his|her|its)#{MK_POST} (?<weapon>.+?) and attempts to push (?<target>you|.+?) away!/
@@ -325,6 +334,10 @@ module Lich
               # you stand fast..." is a fear-save outcome, not a claw swing
               /(?<attacker>.+?) claws at (?<target>[^!,]+)!/,
               /(?<attacker>.+?) tries to spear (?<target>.+?) with #{MK_PRE}its#{MK_POST} enormous tusks!/,
+              # sanguine ooze (mutant farm, hunt log 2026-09-07 23:20); the
+              # whip's miss rides the same line (:miss outcome)
+              /(?<attacker>.+?) manifests a thick pseudopod and brings it smashing down at (?<target>you|.+?)!/,
+              /(?<attacker>.+?) whips a thick pseudopod at (?<target>you|.+?)!/,
               /(?<attacker>.+?) snaps at (?<target>.+?) with its (?<weapon>[^!]+)!/,
               /(?<attacker>.+?) pounds at (?<target>.+?) with #{MK_PRE}(?:his|her|its)#{MK_POST} .*?fists?!/,
               /(?<attacker>.+?) tries to bite (?<target>[^!]+)!/,
@@ -494,7 +507,7 @@ module Lich
           # Creature maneuvers aimed at the whole room, us included: the line
           # names the attacker and no target, and the SSR that follows is
           # OUR save. The parser classifies these inbound.
-          ROOM_TARGETED = %i[howl trumpet].freeze
+          ROOM_TARGETED = %i[howl trumpet shrapnel_spray rift_tentacles].freeze
 
           # The tracker's chunk gate only forwards chunks holding a bolded
           # creature link; an environmental tick names no creature, so its
--- a/lib/gemstone/combat/defs/attacks.rb
+++ b/lib/gemstone/combat/defs/attacks.rb
@@ -284,6 +284,15 @@ module Lich
             # ".+?" not "[^!]+" for target: the cman form appends " and
             # connects!" which a greedy class would swallow into the target
             AttackDef.new(:tackle, [/(?<attacker>.+?) hurls #{MK_PRE}(?:himself|herself|itself)#{MK_POST} at (?<target>.+?)(?: and connects)?!/].freeze),
+            # Creature fear maneuvers (SSR follows, then our save/fail line).
+            # Room-wide, no target named - ROOM_TARGETED classifies them
+            # inbound (hunt log 2026-09-07: Ojandhaart warg / mastodon).
+            AttackDef.new(:howl, [
+              /(?<attacker>.+?) sits back on its haunches and unleashes a long, high-pitched howl that sends a shiver of primal terror down your spine\./
+            ].freeze),
+            AttackDef.new(:trumpet, [
+              /(?<attacker>.+?) raises its trunk and rears back onto its immense hind legs, blaring out a note of sheer fury!/
+            ].freeze),
             # Maneuver-style strike opener (round-14 sweep: 40/40 resolve
             # in the very next line, usually the target's evanescent
             # shield absorb or an outmaneuver outcome, then the swing
@@ -300,7 +309,10 @@ module Lich
             ].freeze),
             # creature natural weapons and maneuvers
             AttackDef.new(:natural, [
-              /(?<attacker>.+?) claws at (?<target>[^!]+)!/,
+              # no comma in the target: "Fear still claws at your heart, but
+              # you stand fast..." is a fear-save outcome, not a claw swing
+              /(?<attacker>.+?) claws at (?<target>[^!,]+)!/,
+              /(?<attacker>.+?) tries to spear (?<target>.+?) with its enormous tusks!/,
               /(?<attacker>.+?) snaps at (?<target>.+?) with its (?<weapon>[^!]+)!/,
               /(?<attacker>.+?) pounds at (?<target>.+?) with #{MK_PRE}(?:his|her|its)#{MK_POST} .*?fists?!/,
               /(?<attacker>.+?) tries to bite (?<target>[^!]+)!/,
@@ -397,7 +409,7 @@ module Lich
           ].freeze
 
           # Environmental / self-inflicted damage. No attacker, no target
-          # capture: the parser classifies every name in SELF_INFLICTED as
+          # capture: the parser classifies every name in ATTACKERLESS as
           # inbound (damage to US), so the "... N points of damage!" line that
           # follows lands on our taken ledger instead of orphaning or, worse,
           # attaching to whatever creature event was open (real-feed
@@ -439,19 +451,33 @@ module Lich
             nil
           end
 
-          # Def names whose lines describe damage to US with no attacker: the
-          # parser reports them inbound without needing a "you" capture.
-          SELF_INFLICTED = %i[frigid_wind thorn_recoil].freeze
+          # Def names whose lines describe damage to US with no creature
+          # attacker: the parser reports them inbound without needing a "you"
+          # capture, and names the attacker for the ledger.
+          #   ENVIRONMENTAL - the world did it (weather ticks); attacker
+          #                   'environment'
+          #   SELF_INFLICTED - our own gear did it (thorn bow recoil);
+          #                    attacker 'self'
+          # (owner 2026-09-07: "frigid wind is environmental, it's not self
+          # inflicted" - the two must stay distinguishable in reports)
+          ENVIRONMENTAL = %i[frigid_wind].freeze
+          SELF_INFLICTED = %i[thorn_recoil].freeze
+          ATTACKERLESS = (ENVIRONMENTAL + SELF_INFLICTED).freeze
+
+          # Creature maneuvers aimed at the whole room, us included: the line
+          # names the attacker and no target, and the SSR that follows is
+          # OUR save. The parser classifies these inbound.
+          ROOM_TARGETED = %i[howl trumpet].freeze
 
           # The tracker's chunk gate only forwards chunks holding a bolded
           # creature link; an environmental tick names no creature, so its
           # chunk was discarded before the parser ever saw it (real-feed
           # 2026-09-07: zero frigid_wind rows against 10 log ticks). This
           # union lets the gate pass such chunks.
-          SELF_INFLICTED_PATTERN = Regexp.union(ENVIRONMENTAL_ATTACKS.flat_map(&:patterns)).freeze
+          ATTACKERLESS_PATTERN = Regexp.union(ENVIRONMENTAL_ATTACKS.flat_map(&:patterns)).freeze
 
-          def self.self_inflicted_line?(line)
-            SELF_INFLICTED_PATTERN.match?(line)
+          def self.attackerless_line?(line)
+            ATTACKERLESS_PATTERN.match?(line)
           end
 
           # AMBUSH PREFIXES - modifiers, not attacks.
--- a/lib/gemstone/combat/defs/attacks.rb
+++ b/lib/gemstone/combat/defs/attacks.rb
@@ -444,6 +444,17 @@ module Lich
             # put it away")
             AttackDef.new(:thorn_recoil, [
               /As (?:a|an|your) .+? leaves your (?:left|right) hand, the thorns embedded in your skin painfully rip away/
+            ].freeze),
+            # Bleed ticks from open wounds. Nobody's ability: the processor
+            # marks a creature's tick :unowned (UNOWNED_TICK_ATTACKS) so the
+            # damage lands on the creature without joining our rollup; ours
+            # ("Your right leg drips...") is inbound via the "Your" capture.
+            # Lives in this list so the chunk gate passes our own tick, which
+            # carries no creature link.
+            AttackDef.new(:bleed, [
+              /Blood weeps from (?<target>.+?)'s#{MK_POST} open .+? wound\./,
+              /(?<target>.+?)'s#{MK_POST} .+? drips as #{MK_PRE}(?:he|she|it)#{MK_POST} continues to bleed\./,
+              /(?<target>Your) .+? drips as you continue to bleed\./
             ].freeze)
           ].freeze
 
--- a/lib/gemstone/combat/defs/attacks.rb
+++ b/lib/gemstone/combat/defs/attacks.rb
@@ -343,6 +343,9 @@ module Lich
               /(?<attacker>.+?) whips a pseudopod toward (?<target>you|.+?), brushing/,
               # halfling cannibal ambush swing (hunt log 2026-09-07 23:34)
               /With an ululating shriek, (?<attacker>.+?) leaps from the shadows and hammers blindly at (?<target>you|.+?) with grimy little fists!/,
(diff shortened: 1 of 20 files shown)


Nisugi:
**Merge note: waits on #1576 (observation provenance) and #1592 (recorder receipts).**

Clean against `main` alone; once those land this conflicts in four regions of `lib/gemstone/combat/processor.rb` and one of `lib/gemstone/combat/recorder.rb`.

**Resolution (rehearsed in a combined checkout, main + #1575 + #1576 + #1592 + #1578 + #1580 + #1587 + this):**

processor.rb
- The three event-save sites: keep #1576's `include_attack_events:` argument to `event_savable?` and this PR's identity-deduplicating `save_event.call(current_event)`, so a swing interrupted by an inbound attack and resumed is saved once under either demand mode.
- The outcome-only event region: keep this PR's inbound-barrier intercept ("blocks the attack from X" is an inbound unknown with X as attacker, never an attack on X), the volley round naming with spawn-root lineage, and the nocked pre-empted swing recorded as `:fire`; add `source: source` to both replacement event constructors.
- Held pre-flares (this PR's cross-chunk storage) carry no ingestion source, so under #1576 the chunk that claims them next would stamp its own room epoch and receipt time onto earlier damage. Resolution: a held flare keeps the binding it was observed under and drops it when the claiming chunk's differs, exactly as #1576's held cast does; a shared `SOURCE_BINDING` names the compared fields.

recorder.rb
- Keep `@chunk_flares`, the expanded `@open_attack` with `flare_ids`, and #1592's complete receipt construction. The receipt is built inside the attack transaction, is nil on rollback, and is emitted after the transaction and outside the recorder mutex, as #1592 has it.

**Already in this PR:**
- 6f3aaa9e: `statuses.flare_id` was in `SCHEMA` and in every status `INSERT` but missing from `ADDED_COLUMNS`, so an existing database failed on the first status of a hunt. The nullable column is now migrated with its foreign key, with two examples that open a pre-`flare_id` database.
- 7b1c7cad: the recorder's per-chunk uid maps survived a session close and reset only on a recorded uid-0 event, so after an idle timeout a chunk opening with a foreign event could link the new session's first attack or a deferred status to the previous session's row. The maps clear on session close, and a chunk is identified by the processor's `observation_batch` id when present (#1576), falling back to uid 0 otherwise.

Combined checkout validation: gemstone + guard suites 785 examples, 0 failures, plus focused checks for interrupted swings without duplicate events, status-to-attack and status-to-flare attribution, observer-only demand, cross-chunk provenance (every event constructor stamped, held cast and held flare keep/drop their source per binding), migration completeness against the 5.21.0 schema (only `flare_id` had been missed), and receipts on commit versus rollback with the mutex released at emit time.

I'll rebase and push the resolved version as soon as #1576 and #1592 land.

mrhoribu (changes requested):
# PR Review: fix(gs): combat module - hunt-log audit (lich-5#1593)

**Reviewed:** 20 files, +1613/-101 at `7b1c7cad` on 2026-09-11. Based on: full clone (fresh `git clone` + `git fetch origin pull/1593/head`), all 20 commits and history available.
**Verdict:** Request changes
**Summary:** Sixteen commits of hunt-log-driven parser/recorder fixes for the Gemstone combat module, each grounded in a specific real transcript and backed by a new spec. The parsing fixes are careful and well-tested. One genuine data-loss bug survives verification: the recorder's new "bare facts don't keep a session alive" rule (meant to stop a bystander's spell from opening a spurious session) also silently drops the delayed room-feed death confirmation that this *same* PR's "watch until death" fix was built to catch, once the hunting session has already idle-timed-out.

## Blockers

*(none — the one high-severity finding is filed as Major below since the affected subsystem has no first-party caller yet in this repo)*

## Major

### 1. A death confirmed after the session idle-times-out is silently dropped, undoing this PR's own "watch until death" fix — `lib/gemstone/combat/recorder.rb:367-379`

```ruby
def foreign_event?(type, data)
  return true unless type == :attack

  !!(data[:foreign_caster] || data[:foreign_target] || data[:unowned] || data[:_orphan])
end

def record_locked(type, data)
  if @idle_timeout
    now = Time.now
    check_idle_locked!(now)
    ours = !foreign_event?(type, data)
    start_session_locked(character: @character, source: @source, at: now) if ours && !@session_id
    @last_event_at = now if ours
  end
  return unless @session_id
```

`foreign_event?` returns `true` — "not ours" — for **every** event whose `type` isn't `:attack`, unconditionally. That includes `:status` "dead" confirmations. Meanwhile `processor.rb`'s `sweep_death_watch` (this PR, `processor.rb:1716-1751`) was deliberately changed from a 3-sweep expiry to "watch until death or deregistration," precisely so a creature "we fought and someone else finished minutes later" (the PR's own commit message, `d24bcaa4`) still gets its death recorded. That confirmation is emitted as a bare `Observers.emit(:status, ..., status: 'dead', ...)`, with no accompanying `:attack`.

Put the two together: if the death confirmation arrives more than `idle_timeout` after the session's last real attack — exactly the "minutes later" case the death-watch fix targets — `check_idle_locked!` (called at the top of this same `record_locked`, or by a periodic external caller per its own doc comment at `recorder.rb:334-337`) closes the session first. Then `ours` is `false` for the `:status` event, so no new session opens, and `return unless @session_id` drops the record entirely — no row in `statuses`, no `killed_at`/`killed_by_attack_id` stamped on the creature. Silent loss, not a mis-filed row.

I confirmed this by driving the real `Recorder` class directly (bypassing any test double or mock):

```ruby
# frozen_string_literal: true

require_relative 'repo/spec/spec_helper'
require 'tmpdir'
require 'gemstone/combat/recorder'

observers = Module.new do
  def self.on(*, **, &_blk); end
  def self.off(*); end
  def self.emit(*); end
end
Lich::Gemstone::Combat.send(:remove_const, :Observers) if Lich::Gemstone::Combat.const_defined?(:Observers)
Lich::Gemstone::Combat.const_set(:Observers, observers)

tmpdir = Dir.mktmpdir('combat-recorder-repro')
db_path = File.join(tmpdir, 'test.db')

rec = Lich::Gemstone::Combat::Recorder.new(db_path, character: 'Tester', source: 'test', idle_timeout: 300)

# t=0: we land the killing blow (an :attack). Session opens.
allow_time = Time.at(3_000_000)
Time.define_singleton_method(:now) { allow_time }
rec.record(:attack, name: 'fire', at: allow_time, target: { id: 101, name: 'a cave lizard', noun: nil },
                    hits: [{ damage: 120, crit: nil }], resolutions: [])
puts "sessions after kill shot: #{rec.instance_variable_get(:@session_id).inspect}"

# t=+301s: idle_timeout (300s) has elapsed with no further attacks. No
# external periodic check_idle! caller is needed - record_locked closes
# the stale session itself on the very next call, before it decides
# whether the status gets to reopen one.
allow_time = Time.at(3_000_000 + 301)

# t=+301s (same moment): the room feed FINALLY confirms the death - the
# exact "someone else finished it minutes later" / lagged-feed scenario
# this PR's death-watch-until-death fix targets. sweep_death_watch emits
# this as a bare :status, not an :attack.
rec.record(:status, id: 101, name: 'a cave lizard', status: 'dead', action: :add)

db = SQLite3::Database.new(db_path)
db.results_as_hash = true
puts "sessions: #{db.execute('SELECT * FROM sessions').inspect}"
puts "statuses: #{db.execute('SELECT * FROM statuses').inspect}"
puts "creatures.killed_at: #{db.execute('SELECT * FROM creatures').map { |c| c['killed_at'] }.inspect}"
db.close
rec.close
FileUtils.remove_entry(tmpdir) rescue nil
```

Output:

```
sessions after kill shot: 1
sessions: [{"id"=>1, "character"=>"Tester", "source"=>"test", "started_at"=>3000000.0, "ended_at"=>3000000.0}]
statuses: []
creatures.killed_at: [nil]
```

Before this PR, `record_locked` called `start_session_locked(...) unless @session_id` unconditionally for every type, so this same delayed status *would* have opened a new (admittedly spurious) session and recorded the death. This PR's fix for the real bug it targets — Tijay's foreign cast in town spawning a phantom "session 2" (`b9d05f4e`) — is correct for `:attack`-type foreign/unowned/orphan events, which are explicitly covered by a passing test (`recorder_spec.rb:112-129`). But blanket-applying "not ours" to *every* non-`:attack` type sweeps up the recorder's own legitimate, delayed self-facts along with it, and no test exercises a non-`:attack` event under `idle_timeout` at all — the four `idle_timeout: 300` tests in `recorder_spec.rb` only ever call `record(:attack, ...)`.

**Fix direction:** `foreign_event?` needs to tell "a bystander's fact" apart from "our own combat's delayed fact." A `:status` carrying `status: 'dead'` for a creature that has a known `@open_attack`/recent creature-cache entry (or, more simply, any `:status`/`:stun`/`:roundtime`/etc. keyed to a creature id we've touched this session) should count as `ours` for keepalive purposes even though it isn't an `:attack`.

**Confidence:** confirmed (reproduced against the real class, not a mock).

## Minor

### 2. `nocked` is cleared by *any* attack line, not just the swing it announced — `lib/gemstone/combat/processor.rb:1209`

```ruby
current_event = { ... }
nocked = nil # the swing the nock announced has printed (see decl)
```

This runs unconditionally inside the `if attack` branch — for *every* new attack event, including an unrelated creature's inbound swing that happens to interleave between our "You nock ... in your bow." line and our own fire (or its pre-emption). If that happens, `nocked` is lost before the pre-emptive-evade branch (`processor.rb:962`, `nock_fire = nocked && ...`) ever sees it, and the swing falls back to `:unknown` with no weapon — the exact failure this commit (`f2f1dc3b`) set out to fix, just for a different trigger.

Reproduction (`Processor.parse_events` driven directly):

```ruby
# frozen_string_literal: true

require_relative 'repo/spec/spec_helper'
require 'gemstone/combat/parser'
require 'gemstone/combat/processor'

module Lich
  module Gemstone
    module Combat
      module Tracker
      end
    end
  end
end

Lich::Gemstone::Combat::Tracker.define_singleton_method(:settings) do
  { track_statuses: false, track_ucs: false, emit_attacks: true, track_damage: true, track_wounds: false }
end
Lich::Gemstone::Combat::Tracker.define_singleton_method(:debug?) { |*_a| false }

module Lich
  module Gemstone
    module Combat
      module Observers
        def self.emit(*); end
      end
    end
  end
end

def bolded(id, noun, name)
  %(<pushBold/><a exist="#{id}" noun="#{noun}">#{name}</a><popBold/>)
end

warg = bolded(123985834, 'warg', 'A warg')
warg_it = bolded(123985834, 'warg', 'it')

chunk = [
  'You nock a faewood arrow fletched with plain white feathers in your <a exist="129604585" noun="bow">glowbark long bow</a>.',
  # an unrelated creature attack interleaves before our fire prints
  "#{warg} claws at you!",
  '   ... but the attack is foiled by your armor.',
  # now our own shot is pre-empted by another warg's evade (no fire line ever prints)
  "With preternatural speed, #{bolded(123985834, 'warg', 'A warg')} bounds to safety as you move to attack #{warg_it}, leaving you off-balance!",
  'The arrow streaks off into the distance!'
]

events = Lich::Gemstone::Combat::Processor.parse_events(chunk)
events.each do |e|
  puts "name=#{e[:name].inspect} weapon=#{e[:weapon].inspect} target=#{e[:target] && e[:target][:id]} inbound=#{e[:inbound]} outcomes=#{e[:outcomes]}"
end
```

Output:

```
name=:natural weapon=nil target= inbound=true outcomes=[:evade, :miss]
```

No `:fire` event is ever produced — the warg's own inbound `claws at you!` stays open as `current_event` (an inbound event is deliberately never touched by the target-switcher, so nothing closes it), so the later pre-emptive evade line's `:evade` outcome gets appended to that *unrelated* inbound event instead of reaching the `nock_fire` branch at all. That downstream "outcome glomming onto whatever inbound event is still open" behavior is pre-existing (not introduced by this PR), so I'm not asking for a fix to that path here — flagging it only so it's clear the `nocked` fix's blast radius is narrower than the "cleared by any attack" line alone would suggest, and that the interaction wasn't (as far as I can tell) covered by a test.

Separately, `nocked` is a `parse_events`-local variable, not chunk-persistent state like `@held_cast`/`@held_pre_flares` — the comment even says "this chunk." Nock and fire are normally two separate player commands with their own roundtime, so they will often land in *different* chunks/prompts in a live feed, in which case this tracking never engages at all. That scoping looks deliberate (the author clearly wrote "this chunk" knowing it), so I'm not flagging it as a bug.

**Confidence:** confirmed (traced and reproduced in code); real-world frequency of the interleaving trigger is speculative, hence Minor rather than Major.

## Nits

- `lib/gemstone/combat/recorder.rb:539`: `event[:observation_batch]` is read but never set anywhere in this repo (`grep -rn observation_batch lib/ spec/` finds only the three lines in `recorder.rb` itself) — it's forward-compatible plumbing for the not-yet-merged upstream #1576, as the PR description says. Not a bug, but worth flagging so it isn't mistaken for dead code by a future reader; the `uid.nil? || uid.zero?` fallback is what actually runs today, and that path is well covered by `recorder_spec.rb:273-297`.

## Open questions

- `lib/gemstone/combat/defs/spells.rb:181-186`: the new 611-evoked-moonbeam pattern `/you draw down a shaft of \w+ moonlight and bathes? .../` is lowercase-`you`-only (no `/i`), matched against the exact hunt-log line "Tapping the moons above, **you** draw down...". If the game ever prints this as a sentence-initial "**You** draw down a shaft..." (capitalized), it won't match and will fall through to `:unknown` instead of `:moonbeam`. **Settles it:** does the evoked-moonbeam line ever print sentence-initial in your logs, or is the "Tapping the moons above," prefix guaranteed?
- `lib/gemstone/combat/recorder.rb:370`: the same `foreign_event?` blanket-non-attack rule (Major finding above) also means an `:attack` event flagged `unowned` (a bystander's bleed/rot/pestilence tick landing on a creature we're not credited for) never keeps a session alive either. That's plausibly fine — an unowned tick alone isn't evidence we're still hunting — but it's the identical mechanism, so it's worth confirming it was an intended scope for the fix rather than only the foreign_caster/foreign_target case the tests cover.

## What this PR gets right

- Every fix is traceable to a specific hunt-log line or owner ruling, and every one of the sixteen commits ships its own spec built from that log — this makes the review's job of separating "grounded fix" from "guess" unusually easy, and I found no case where the stated log evidence didn't match the code change.
- The rename `SELF_INFLICTED_PATTERN`/`self_inflicted_line?` → `ATTACKERLESS_PATTERN`/`attackerless_line?` is fully consistent — I checked every call site (`tracker.rb`, `processor.rb`, both spec files) and found no leftover reference to the old names.
- The chunk-boundary/session-boundary map-reset fix (`recorder.rb:304-319`, commit `7b1c7cad`) is exactly the kind of bug that's easy to introduce and hard to catch — a foreign event skipping the uid-0 reset leaking the previous session's row map into the next one — and it ships with a targeted regression test (`recorder_spec.rb:273-297`) that fails without the fix (verified by reading the old vs. new `record_attack` logic side by side).
- The `statuses.flare_id` migration (commit `6f3aaa9e`) is a real self-caught bug — a column referenced in every `INSERT` but missing from the migration map would have broken every upgrade path silently — and both the "column gets added" and "insert doesn't fail" cases are tested against a hand-rolled legacy schema.
- Full combat spec suite: 293 examples, 0 failures. Full repo suite: 7195 examples, 0 failures. Rubocop clean on all 26 touched lib/spec files.

## Coverage notes

- Access: full clone via `gh`-authenticated HTTPS, all history and all 20 commits available; not diff-only.
- `bundle exec rspec spec/lib/gemstone/combat`: 293 examples, 0 failures. `bundle exec rspec` (full repo): 7195 examples, 0 failures. `bundle exec rubocop` on all touched files: 0 offenses.
- CodeRabbit CLI (`coderabbit review --agent`) was run twice (once against `main`, once with `--committed --base-commit <merge-base>`) but returned only setup/status events with no findings both times, despite exiting cleanly (code 0) after ~a minute each run. I could not get a substantive CodeRabbit pass on this PR; the findings above are from manual review only.
- Both Major/Minor findings above were verified by driving the real classes (`Recorder`, `Processor.parse_events`) directly against a real on-disk SQLite database / real parser, not mocks — the full scripts are inlined above for reproduction.
- I did not attempt to verify the parser regex changes against real GS4 hunt logs myself (I have none) — I'm relying on the commit messages' stated log evidence, which is specific enough (exact dates, exact creature names, exact line text) that I treated it as credible grounding rather than an unverifiable claim, consistent with how thoroughly this PR documents its own provenance.
- I did not review `spec/fixtures/replay/{natures_fury,shield_throw}.txt` (1-line changes each) in depth beyond confirming they're exercised by existing passing specs.

---

Nisugi:
Thanks for this — both findings were real, and I reproduced each one before changing anything. Fixed in 48fed9cf.

## 1. Delayed death dropped (Major)

Confirmed exactly as described, including that `process()` sweeps on a quiet chunk, which is what makes the "bare `:status` arriving with no attack anywhere near it" case routine rather than a corner case.

`foreign_event?` now judges a non-attack event by the creature it names instead of by its type: ours when this hunt already fought that creature, foreign otherwise. So the Tijay-cast-in-town behaviour that motivated the original rule still holds — a bystander's fact about a creature we never touched still opens nothing.

Implementing your suggested direction turned up two things worth flagging:

- **Letting the death open a session isn't enough.** It records the status, but `start_session_locked` wipes `@creature_cache`, so `ensure_creature` inserts a **duplicate creature row** under a new session and stamps `killed_at` there — disconnected from the attack that earned it. Mis-filed instead of lost, which is better but still wrong. A trailing fact now re-opens the *just-closed* session rather than starting a fresh one, so the kill lands on the row the attack already wrote.
- **Ordering bug my own regression test caught.** A bystander's status arriving first runs `check_idle_locked!`, which closes the session and clears the cache — so the real death that followed no longer recognised itself and opened session 2 anyway. `finish_session_locked` now carries the closed session's creature cache in `@closed_session`, and `start_session_locked` clears it so a genuinely new hunt never inherits the previous one's trailing window.

`@last_event_at` is deliberately left untouched by a trailing fact, so `ended_at` still comes back to the last real attack and a lagged death never pads a hunt's duration.

## 2. `nocked` cleared by any attack (Minor)

Fixed — the reset is now skipped for inbound and foreign attacks, since "You nock" is first-person and only our own attack can be the swing it announced.

You were right that this is inert on its own, and I ended up fixing the downstream path you explicitly declined to ask for. Without it the first change does nothing: the open inbound event swallows the pre-emptive evade and our swing is never opened, which my regression test reproduced exactly as you predicted. While a nock is live, an evade naming an outbound target no longer attaches to an open inbound event.

Worth noting since it validates your instinct to scope that carefully — my first attempt at the guard was too broad and broke two inbound fear-maneuver specs (`:howl`, `:trumpet`), whose resist lines name their own caster possessively ("the warg's unnerving howl") and so look like a targeted outcome. Gating on a live nock resolves it.

## 3. Moonbeam (open question)

Didn't wait on the answer since the safe fix is free: the pattern now accepts sentence-initial `You` via `[Yy]ou` rather than a blanket `/i`, so it no longer depends on the "Tapping the moons above," prefix being guaranteed.

## Not changed

- **`observation_batch`** — agreed it's forward-compatible plumbing for #1576, left as-is.
- **Unowned ticks not keeping a session alive** — intended. An unowned tick is a bystander's damage on a creature we're not credited for, so on its own it isn't evidence we're still hunting. Your read was right that it's the identical mechanism; it just happens to be the behaviour I want in that case.

## Verification

- `spec/lib/gemstone/combat`: **296 examples, 0 failures** (was 293 — three new regression tests: delayed death into the same session, trailing fact not extending the hunt or resurrecting it for a stranger, and the interleaved inbound swing keeping the nocked bow).
- Rubocop clean on all 26 touched files.

One note on the full-repo suite: I get **58 failures against a 56-failure baseline on the unmodified branch**, all in frontend / script-lifecycle / setup-files specs. The set churns in both directions between runs and those files fail in isolation without my changes, so they're order-dependent flakes — but it does mean your clean 7195 run doesn't reproduce in my environment.

Nisugi:
Correction to point 3 of my previous comment — the moonbeam change is reverted in 9a75ec64.

The "Tapping the moons above," prefix is always present, so `you` is always mid-sentence and the original lowercase pattern was correct as written. Widening it to `[Yy]ou` was unnecessary.

`defs/spells.rb` is now byte-identical to 7b1c7cad. The test is dropped.

Everything in points 1 and 2 stands unchanged — those were reproduced against real behaviour before being touched.

`spec/lib/gemstone/combat`: **296 examples, 0 failures**. Rubocop clean.

mrhoribu:
Pulled `48fed9cf` and `9a75ec64` and verified each against the real code rather than the write-up alone.

## Fixed, verified

**#1 (delayed death)** — genuinely fixed. Re-ran my original repro plus an adversarial case (a bystander fact for a creature this hunt never fought, arriving right after the session closes):

```ruby
# frozen_string_literal: true

require_relative 'repo/spec/spec_helper'
require 'tmpdir'
require 'gemstone/combat/recorder'

observers = Module.new do
  def self.on(*, **, &_blk); end
  def self.off(*); end
  def self.emit(*); end
end
Lich::Gemstone::Combat.send(:remove_const, :Observers) if Lich::Gemstone::Combat.const_defined?(:Observers)
Lich::Gemstone::Combat.const_set(:Observers, observers)

tmpdir = Dir.mktmpdir('combat-recorder-repro-v2')
db_path = File.join(tmpdir, 'test.db')

rec = Lich::Gemstone::Combat::Recorder.new(db_path, character: 'Tester', source: 'test', idle_timeout: 300)

allow_time = Time.at(3_000_000)
Time.define_singleton_method(:now) { allow_time }
rec.record(:attack, name: 'fire', at: allow_time, target: { id: 101, name: 'a cave lizard', noun: nil },
                    hits: [{ damage: 120, crit: nil }], resolutions: [])

allow_time = Time.at(3_000_000 + 301)
rec.record(:status, id: 101, name: 'a cave lizard', status: 'dead', action: :add)

db = SQLite3::Database.new(db_path)
db.results_as_hash = true
puts "sessions: #{db.execute('SELECT * FROM sessions').inspect}"
puts "statuses: #{db.execute('SELECT * FROM statuses').inspect}"
puts "creatures: #{db.execute('SELECT exist_id, killed_at, killed_by_attack_id, session_id FROM creatures').inspect}"

# Adversarial: an unrelated bystander fact for a creature never fought,
# right after our session closes - must NOT reopen it or extend it.
allow_time = Time.at(3_000_000 + 302)
rec.record(:status, id: 999, name: 'a stranger', status: 'prone', action: :add)
puts "session after unrelated bystander fact: #{rec.instance_variable_get(:@session_id).inspect}"
puts "sessions (after bystander): #{db.execute('SELECT * FROM sessions').inspect}"

db.close
rec.close
FileUtils.remove_entry(tmpdir) rescue nil
```

```
sessions: [{"id"=>1, ... "ended_at"=>nil}]
statuses: [{"id"=>1, ..., "attack_id"=>1, "status"=>"dead", ..., "source"=>"window"}]
creatures: [{"exist_id"=>101, "killed_at"=>3000301.0, "killed_by_attack_id"=>1, "session_id"=>1}]
session after unrelated bystander fact: nil
sessions (after bystander): [{"id"=>1, ..., "ended_at"=>3000000.0}]
```

`killed_at`/`killed_by_attack_id` land on the *original* attack row, no duplicate creature row, and the unrelated bystander fact correctly stays foreign — `ended_at` comes back to the real last-attack time exactly as the commit says. Nice catch on the ordering bug and the `@closed_session` cache handoff; both hold up.

**Moonbeam revert** — `defs/spells.rb` is byte-identical between `7b1c7cad` and `9a75ec64` (`git diff 7b1c7cad 9a75ec64 -- lib/gemstone/combat/defs/spells.rb` is empty). Clean.

## New regression: the nocked-bow fix drops a real inbound hit — `lib/gemstone/combat/processor.rb:944-985`

`preempts_open_inbound` correctly stops the pre-emptive evade from being swallowed by a stale open inbound event, but the branch it falls into when that guard fires assumes "no event at all" and unconditionally overwrites `current_event`:

```ruby
elsif line_target && line_target[:id]
  if Definitions::Outcomes.inbound_line?(line)
    current_event = { name: pending_ambush ? :ambush : :unknown, target: {}, inbound: true, ... }
  else
    ...
    current_event = { name: pending_ambush ? :ambush : (round_name || (nock_fire ? :fire : :unknown)), ... }
```

Neither branch calls `save_event.call(current_event)` on the outgoing event first — every *other* resumption path in this same method (`single_hit_parent`, `interrupted_own`) does exactly that before replacing `current_event`, but this one, now reachable with a non-nil `current_event` via `preempts_open_inbound`, doesn't. The outgoing event is simply discarded — no reference to it survives anywhere.

Reproduced with a fully-resolved inbound hit (not just an outcome) interleaving between the nock and the pre-emptive evade:

```ruby
# frozen_string_literal: true

require_relative 'repo/spec/spec_helper'
require 'gemstone/combat/parser'
require 'gemstone/combat/processor'

module Lich
  module Gemstone
    module Combat
      module Tracker
      end
    end
  end
end

Lich::Gemstone::Combat::Tracker.define_singleton_method(:settings) do
  { track_statuses: false, track_ucs: false, emit_attacks: true, track_damage: true, track_wounds: false }
end
Lich::Gemstone::Combat::Tracker.define_singleton_method(:debug?) { |*_a| false }

module Lich
  module Gemstone
    module Combat
      module Observers
        def self.emit(*); end
      end
    end
  end
end

def bolded(id, noun, name)
  %(<pushBold/><a exist="#{id}" noun="#{noun}">#{name}</a><popBold/>)
end

warg = bolded(123985834, 'warg', 'A warg')
warg2 = bolded(123985834, 'warg', 'The warg')
warg_it = bolded(123985834, 'warg', 'it')

# A FULLY RESOLVED inbound attack that actually deals us real damage,
# interleaved between our nock and our own pre-empted fire.
chunk = [
  'You nock a faewood arrow fletched with plain white feathers in your <a exist="129604585" noun="bow">glowbark long bow</a>.',
  "#{warg} claws at you!",
  '  AS: +176 vs DS: +76 with AvD: +20 + d100 roll: +64 = +184',
  '   ... and hits for 28 points of damage!',
  '   Smack to the eye bursts blood vessels.',
  "With preternatural speed, #{warg2} bounds to safety as you move to attack #{warg_it}, leaving you off-balance!",
  'Roundtime: 3 sec.'
]

events = Lich::Gemstone::Combat::Processor.parse_events(chunk)
puts "total events: #{events.size}"
events.each do |e|
  puts "name=#{e[:name].inspect} weapon=#{e[:weapon].inspect} target=#{e[:target] && e[:target][:id]} " \
       "inbound=#{e[:inbound]} attacker=#{e[:attacker].inspect} hits=#{e[:hits].inspect} outcomes=#{e[:outcomes]}"
end
```

Output:

```
total events: 1
name=:fire weapon="glowbark long bow" target=123985834 inbound= attacker=nil hits=[] outcomes=[:evade]
```

The warg's own `claws at you!` swing — a fully resolved hit for 28 real damage — never appears. It's not merged, not orphaned, not saved anywhere; it's just gone. That's a step backward from the bug this same commit fixes: before, the pre-empted swing only *misclassified* as `:unknown`; now a real inbound hit can be *silently dropped entirely*.

The new regression test (`processor_inbound_spec.rb`, `"keeps the nocked bow when a creature's inbound swing interleaves before the fire"`) doesn't catch this because it asserts via `fire = events.find { |e| !e[:inbound] }` — it only checks the `:fire` event's shape and never asserts on `events.size` or that the interleaving warg's own event survived. My repro above's "before your fix" line (`... but the attack is foiled by your armor.`) doesn't even match a real outcome def, which is how the original test's warg event ended up with zero facts and the loss went unnoticed — a real damage line makes it obvious.

**Suggested fix:** `save_event.call(current_event) if event_savable?(current_event)` before each of the two `current_event = { ... }` reassignments at `processor.rb:957` and `processor.rb:975`, same as `single_hit_parent`/`interrupted_own` already do elsewhere in this method.

**Confidence:** confirmed (reproduced against the real parser, not a mock).

---

mrhoribu (approved):
## Approve

All findings from the original review and the follow-up are resolved and verified against the real code — not just the commit messages. Final state reviewed: `a65d30ea`.

### Original review (`7b1c7cad`)

**Major — delayed death confirmation silently dropped** (`recorder.rb`): `foreign_event?` treated every non-`:attack` event as "not ours," so the room-feed death confirmation this PR's own "watch until death" fix exists to emit — arriving minutes after the last attack, past `idle_timeout` — was dropped outright once the session closed: no `statuses` row, no `killed_at`.

**Minor — `nocked` bow tracking cleared by any attack** (`processor.rb`): an unrelated creature's inbound swing interleaving between "You nock..." and our own fire cleared the tracking meant to survive exactly that interleaving.

### Fix commits (`48fed9cf`, `9a75ec64`)

- `foreign_event?` now judges a non-attack fact by the creature it names (`known_creature?`), checking the open session's cache and a new `@closed_session` snapshot so a trailing fact re-opens the *same* session a delayed death belongs to, rather than dropping it or minting a duplicate creature row under a fresh session.
- `@last_event_at` is deliberately left untouched by a trailing fact, so `ended_at` still reflects the last real attack — town time never pads a hunt.
- `nocked` is now only cleared by our own outbound attack (`unless current_event[:inbound] || eff_foreign`), plus a scoped `preempts_open_inbound` guard so a live nock's pre-emptive evade isn't swallowed by a stale open inbound event.
- The unrelated moonbeam casing change from the first fix attempt was correctly reverted (`defs/spells.rb` byte-identical to `7b1c7cad`) once the "Tapping the moons above," prefix was confirmed always present.

**Verified independently:**
```
sessions: [{"id"=>1, ... "ended_at"=>nil}]
statuses: [{"attack_id"=>1, "status"=>"dead", ...}]
creatures: [{"exist_id"=>101, "killed_at"=>3000301.0, "killed_by_attack_id"=>1, "session_id"=>1}]
```
— the delayed death lands on the original attack row, no duplicate. A follow-up adversarial case (a bystander fact for a creature this session never fought, arriving right after close) correctly stayed foreign and did not reopen or extend the session — matching the fix's own new `recorder_spec.rb` test for the same scenario.

### New regression found in the first fix attempt, now fixed (`a65d30ea`)

`preempts_open_inbound` correctly stopped the pre-emptive evade from being swallowed, but the branch it fell into assumed no event was open and overwrote `current_event` without saving it first — silently discarding the interleaving creature's own inbound swing, including any real damage it had already dealt. Reproduced with a fully-resolved 28-damage inbound hit: it vanished entirely (`total events: 1`, no trace of the hit).

Fixed with a single `save_event.call(current_event) if event_savable?(current_event)` ahead of that branch, mirroring how `single_hit_parent`/`interrupted_own` already save an outgoing event before replacing it. Re-ran the same reproduction: both events now survive (`total events: 2`), the inbound swing carrying its full 28-damage hit and crit detail. The regression test was correspondingly strengthened to use a real damage line and assert on `events.size` plus the swing's damage, rather than only the `:fire` event's shape.

### Final verification (this pass, `a65d30ea`)

- `bundle exec rspec spec/lib/gemstone/combat`: 296 examples, 0 failures.
- `bundle exec rspec` (full repo): 7198 examples, 0 failures.
- `bundle exec rubocop` on all 26 touched lib/spec files: 0 offenses.
- `git diff 7b1c7cad a65d30ea --stat`: only `processor.rb`, `recorder.rb`, and their two spec files changed across all three follow-up commits — no unrelated surface area introduced.

### Open questions from the original review — resolved

- Moonbeam casing: investigated and reverted; the "Tapping the moons above," prefix is guaranteed, so the original lowercase pattern was correct as written.
- Unowned ticks not keeping a session alive: confirmed intentional — a bystander's damage on a creature we're not credited for isn't evidence we're still hunting.

Nothing outstanding. This is a thorough, well-tested piece of hunt-log-driven parser and recorder work, and the two real bugs the review surfaced (including the one introduced by the first fix attempt) were each reproduced, fixed at the root, and locked down with a test that would have caught it.

---

PR #1594, opened 2026-09-11
fix(gs): QStrike prices MSTRIKE's own stamina cost

Nisugi:
## Summary
`QStrike` prices a reduction from stamina minus a reserve minus the attack's own cost, but MSTRIKE is not in the CMan/Weapon/Shield tables, so `lookup_attack_cost('mstrike')` returned 0 and the "affordable" seconds ignored what the strike itself costs.

Per [Multi Opponent Combat/Multistrike](https://gswiki.play.net/Multi_Opponent_Combat/Multistrike): MSTRIKE costs nothing outside its recovery period; during it, `20 + 3 * weapon speed` for an open strike and `30 + 4 * weapon speed` for a focused one, where a second weapon adds `(its speed - 2, at least 0)`.

- `QStrike.mstrike_cost(focused:)` prices it from the Multi-Strike cooldown and the hands' base speeds (`weapon_speed_for`, already there); a shield or empty offhand adds nothing.
- `lookup_attack_cost` answers `mstrike` (open) and `mstrike_focused` / `focused mstrike`, so `calculate(attack_name: :mstrike_focused)` is correct.
- `use(attack: 'mstrike', target: x)` prices the strike as focused.

Consumer: eohunter's Mstrike action asks `QStrike.calculate` for the seconds and arms with `qstrike -N`.

## Test plan
- [x] New examples in `qstrike_spec.rb`: free outside recovery; 35/50 for a broadsword during recovery; a second broadsword adds 3 speed, a dagger or buckler adds none; `lookup_attack_cost` and `calculate` price both kinds.
- [x] `bundle exec rspec spec/lib/gemstone/psms/qstrike_spec.rb` (62 examples), `rubocop` on both files.

### DIFF
--- a/lib/gemstone/psms/qstrike.rb
+++ b/lib/gemstone/psms/qstrike.rb
@@ -50,6 +50,16 @@ module Lich
       # Maximum seconds of RT reduction (reasonable upper bound)
       MAX_REDUCTION = 8
 
+      # MSTRIKE's own stamina cost (Multi Opponent Combat/Multistrike):
+      # nothing outside its recovery period; during it, base + per * weapon
+      # speed, where a second weapon adds (its speed - 2, at least 0).
+      MSTRIKE_COST = {
+        open: { base: 20, per_speed: 3 },
+        focused: { base: 30, per_speed: 4 }
+      }.freeze
+      # The cooldown Effects names while MSTRIKE is in its recovery period
+      MSTRIKE_COOLDOWN = 'Multi-Strike'
+
       # Valid combat actions that can use QSTRIKE
       VALID_ACTIONS = %w[
         ascension ambush attack cheapshot cman cock disarm feat fire
@@ -258,7 +268,8 @@ module Lich
         adaptive = default(:adaptive) if adaptive.nil?
 
         attack_name = normalize_attack_name(attack)
-        attack_cost = lookup_attack_cost(attack_name)
+        # A targeted mstrike is a focused one, whatever the name says
+        attack_cost = attack_name == 'mstrike' && target ? mstrike_cost(focused: true) : lookup_attack_cost(attack_name)
 
         # Determine the actual reduction to attempt
         actual_reduction = resolve_reduction(reduction, reserve, attack_cost)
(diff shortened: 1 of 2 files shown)


mrhoribu (approved):
# PR Review: feat(gs): QStrike prices MSTRIKE's own stamina cost (elanthia-online/lich-5#1594)

**Reviewed:** 2 files, +96/-2 at `588cd87b` on 2026-09-11. Based on: full clone (fresh clone of `elanthia-online/lich-5`, PR branch `feat/qstrike-mstrike-cost` checked out against merge base `526ad7cd`).
**Summary:** `QStrike.calculate`/`use` previously priced MSTRIKE's own stamina cost as `0` because MSTRIKE isn't in the CMan/Weapon/Shield technique tables `lookup_attack_cost` searches; this PR adds a dedicated `mstrike_cost` formula (sourced from and verified against the GSWiki Multi-Strike page) and wires it into both `lookup_attack_cost` and `use`. The math and the new unit tests are correct; the only real gap is that the one-line behavioral change inside `use` itself has no test covering it.

## Minor

### 1. The new `use` targeting branch has no test — `lib/gemstone/psms/qstrike.rb:272`

```ruby
attack_name = normalize_attack_name(attack)
# A targeted mstrike is a focused one, whatever the name says
attack_cost = attack_name == 'mstrike' && target ? mstrike_cost(focused: true) : lookup_attack_cost(attack_name)
```

This is the actual behavior the PR body highlights third ("`use(attack: 'mstrike', target: x)` prices the strike as focused"), but `spec/lib/gemstone/psms/qstrike_spec.rb` has no `describe ".use"` block at all — `grep -n '\.use(' spec/lib/gemstone/psms/qstrike_spec.rb` returns nothing, before or after this diff. The four new examples in the `.mstrike_cost` describe block exercise `mstrike_cost`, `mstrike_weapon_speed`, and `lookup_attack_cost` directly, but none of them go through `use`, so this specific ternary — the only new logic that isn't already covered by the `lookup_attack_cost` tests — could be inverted or deleted in a future edit and nothing would fail. (`.use` was untested before this PR too, so this isn't a regression the PR introduced, but it's the one new line of behavior in this change that a test would have been cheap to add for.)

## Nits

### 1. `target` truthiness in the new branch doesn't match the sibling helper's empty-string handling — `lib/gemstone/psms/qstrike.rb:272` vs `:854`

```ruby
attack_cost = attack_name == 'mstrike' && target ? mstrike_cost(focused: true) : lookup_attack_cost(attack_name)
```
```ruby
def self.build_attack_command(attack, target = nil)
  if target && !target.empty?
```

`build_attack_command`, in the same file, treats an empty-string `target` as "no target." The new mstrike-focus check only tests truthiness, so `use(attack: 'mstrike', target: '')` would price the strike as focused (the more expensive of the two) while `build_attack_command` for the same call would build the command as if untargeted. Nothing crashes and the mispricing is conservative (overestimates cost rather than risking negative stamina), so this is cosmetic unless some caller passes `target: ''` for "no target" — I found none in this repo (`QStrike` has no in-repo callers; its consumer is `eohunter` in the `scripts` repo, out of scope here), so treat this as a heads-up rather than a bug to chase down now.

## What this PR gets right

- The core formula is right, and I checked it against a primary source rather than taking the PR body's word for it: fetching `https://gswiki.play.net/Multi_Opponent_Combat/Multistrike` confirms "20 + (3 * Weapon Base Speed)" for open and "30 + (4 * Weapon Base Speed)" for focused, using *raw* weapon base speed rather than the equipment-multiplier-adjusted speed `cost_per_second_reduction` uses elsewhere — and `mstrike_weapon_speed` correctly reads `weapon_speed_for(hand)[:base_rt]`, not `:equipment_speed`. The offhand contribution (`primary + max(offhand - 2, 0)`) also matches the wiki's general two-weapon-combat "weapon speed" formula ("Primary weapon speed + secondary weapon speed - 2"), with the PR's explicit, disclosed clamp-at-zero for a fast offhand weapon.
- The premise — that MSTRIKE isn't priced in any technique table — checks out: `grep -in "mstrike" lib/gemstone/psms/{cman,weapon,shield}.rb` returns nothing, so `lookup_attack_cost` really was returning `0` for it before this change.
- `MSTRIKE_COOLDOWN = 'Multi-Strike'` as the `Effects::Cooldowns` key is consistent with the codebase's own established handling of that cooldown: `lib/gemstone/infomon/activespell.rb:49-51,78-79` has long predated this PR and matches Cooldowns-dialog keys against `/Multi-Strike/` to rename them for display, which corroborates that the raw key text does contain "Multi-Strike."
- New methods (`mstrike_cost`, `mstrike_recovering?`, `mstrike_weapon_speed`) are documented in the same YARD style as the rest of the file, and the four new spec examples (free outside recovery, correct constants for open/focused, offhand math including the dagger/shield zero-floor case, and integration through `lookup_attack_cost`/`calculate`) are well-targeted rather than tautological — they use the same `Effects::Registry`-backed `XMLData.save_dialogs` pattern already used for `Buffs` elsewhere in the spec, not a bespoke stub.
- `rubocop lib/gemstone/psms/qstrike.rb spec/lib/gemstone/psms/qstrike_spec.rb` is clean, and `bundle exec rspec spec/lib/gemstone/psms/qstrike_spec.rb` passes all 62 examples (up from 58) locally against the PR head.
- The new `.mstrike_cost` describe block adds `after { cooldown(false) }` to clear the `Cooldowns` dialog it sets — the top-level `before(:each)` only clears `Buffs`, so without this the cooldown would leak into later examples in the same run; the author caught that themselves.

## Coverage notes

Reviewed a full fresh clone of `elanthia-online/lich-5` (not the diff alone), with both changed files read in full at the PR head commit and cross-referenced against `lib/gemstone/effects.rb`, `lib/common/xmlparser.rb`, `lib/gemstone/infomon/activespell.rb`, and the CMan/Weapon/Shield technique tables to ground the PR's claims. Ran `rubocop` and the full `qstrike_spec.rb` suite locally (Ruby 4.0.5, both clean/passing). Verified the MSTRIKE stamina-cost formula against the primary source (GSWiki) via two targeted fetches rather than trusting the PR description's citation at face value. `gh pr checks` shows CI (Rubocop, tests, PR-title validation) all green; no existing review comments or reviews on the PR. A CodeRabbit pass was attempted but hit the org's usage-based review rate limit (3 included reviews exhausted, ~54 min cooldown reported) and could not run — no CodeRabbit findings are reflected here. I did not verify the literal "Multi-Strike" `Cooldowns`-dialog key text against a live game XML capture; the cross-file corroboration in `activespell.rb` is the closest available check and it lines up.

PR #1595, opened 2026-09-11
fix(gs): combat module - a stand-up message later in the chunk outranks an earlier knockdown crit

Nisugi:
## Summary
Message statuses are applied while a chunk is parsed (`apply_status_to_target`), but crit-derived statuses are applied afterwards, in `persist_event`. So a creature knocked down by a crit and then standing up in the same chunk was left prone in the registry: the parse removed the position on "stands up.", then persistence re-applied the earlier knockdown. Found in the combined review of the open combat PRs; the behaviour is in the module as released, so it gets its own PR.

- A message that clears a position now notes the recovery for the chunk (`note_position_recovery`), and the crit pass leaves that creature's position alone.
- A later knockdown *message* clears the note, so the last message in the chunk wins either way.
- The note is per chunk, reset at the top of `parse_events`; a stand-up in one chunk never suppresses a knockdown crit in the next.

## Test plan
- [x] New `spec/lib/gemstone/combat/status_order_spec.rb`: knockdown crit then "stands up." leaves the creature standing (fails on main, passes here); knockdown, stand-up, knocked down again ends prone; a recovery does not carry into the next chunk; a knockdown with nothing later still applies.
- [x] `bundle exec rspec spec/lib/gemstone/combat` (249 examples), `rubocop` on both files.

### DIFF
--- a/lib/gemstone/combat/processor.rb
+++ b/lib/gemstone/combat/processor.rb
@@ -570,14 +570,15 @@ module Lich
                   if status_result.is_a?(Hash)
                     seq = status_flare_seq.call(line_target[:id])
                     apply_status_to_target(status_result[:status], line_target[:name], line_target[:id], status_result[:action],
-                                           flare_seq: seq, event: status_event_for.call(line_target[:id], seq))
+                                           flare_seq: seq, event: status_event_for.call(line_target[:id], seq), line: index)
                   else
                     # Legacy format - status_result is just the status symbol
-                    apply_status_to_target(status_result, line_target[:name], line_target[:id], :add)
+                    apply_status_to_target(status_result, line_target[:name], line_target[:id], :add, line: index)
                   end
                 elsif status_result.is_a?(Hash) && status_result[:target]
                   # Fallback to name-based lookup only if no ID available
-                  apply_status_to_target(status_result[:status], status_result[:target], nil, status_result[:action])
+                  apply_status_to_target(status_result[:status], status_result[:target], nil, status_result[:action],
+                                         line: index)
                 elsif status_result.is_a?(Hash) && !line.match?(/\A\s*Your?\b/) &&
                       (subject = (current_target && current_target[:id] ? current_target : nil) ||
                                  (flare_ctx && flare_ctx[:target_info]))
@@ -594,7 +595,7 @@ module Lich
                   seq = status_flare_seq.call(subject[:id])
                   apply_status_to_target(status_result[:status], subject[:name],
                                          subject[:id], status_result[:action],
-                                         flare_seq: seq, event: status_event_for.call(subject[:id], seq))
+                                         flare_seq: seq, event: status_event_for.call(subject[:id], seq), line: index)
                 elsif status_result.is_a?(Hash) && line.match?(/\A\s*Your?\b/)
                   # 2p: the status is OURS ("You are stunned!"). Never a
                   # creature application - but it IS a fact (inbound
@@ -1495,6 +1496,11 @@ module Lich
                         victim = flare_ctx ? flare_ctx[:target_info] : (sink[:target_info] || sink[:target])
                         chunk_deaths << victim[:id] if victim && victim[:id]
                       end
+                      # Where the crit TEXT sits in the chunk. Crit statuses
+                      # are applied after the whole chunk is parsed, so this
+                      # is the only record of where the crit stood relative
+                      # to the messages around it (see position_recovered?).
+                      hit[:_crit_line] = next_line_index
                       respond "[Combat] Found critical hit: #{c[:location]} rank #{c[:wound_rank]}" if Tracker.debug?(:verbose)
                       break # Only take first crit found after this damage
                     end
@@ -1752,12 +1758,25 @@ module Lich
         # pass can leave them alone. Per chunk: cleared at the top of
         # parse_events, before this chunk's messages are read.
 
+        # The recovery records WHERE in the chunk the stand-up was read, not
+        # merely that one happened. A bare flag cannot answer the question the
+        # crit pass actually asks - "did the creature stand up after THIS
+        # crit?" - so a second, genuine knockdown crit landing later in the
+        # same chunk was suppressed by an earlier stand-up.
+        #
         # @param id [Integer, String, nil] the creature a message stood up
-        def note_position_recovery(id)
+        # @param line [Integer, nil] index of the stand-up line in this chunk
+        def note_position_recovery(id, line)
           return unless id
 
           @position_recovered ||= {}
-          @position_recovered[id.to_i] = true
+          # A caller that cannot say WHERE the recovery was read still asserts
+          # that one happened. Infinity keeps that meaning "after every crit",
+          # the safe reading: never re-prone a creature the game stood up.
+          at = line.nil? ? Float::INFINITY : line
+          # Keep the LATEST recovery: it is the one a crit must beat.
+          prev = @position_recovered[id.to_i]
+          @position_recovered[id.to_i] = at if prev.nil? || at >= prev
         end
 
         # A later knockdown MESSAGE outranks the recovery: the crit pass may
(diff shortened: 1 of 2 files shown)


mrhoribu (changes requested):
# PR Review: fix(gs): combat module - a stand-up message later in the chunk outranks an earlier knockdown crit (elanthia-online/lich-5#1595)

**Reviewed:** 2 files, +180/-1 at `faa5db1c` on 2026-09-11. Based on: full local clone (branch checked out from `origin/pull/1595/head`, diffed against `main`).
**Verdict:** Request changes
**Summary:** Fixes the specific ordering bug it describes (a crit-derived knockdown, applied late in `persist_event`, no longer clobbers an earlier-parsed stand-up message) and the new spec correctly fails on `main` / passes here. But the fix tracks recovery as a single per-chunk boolean rather than per-event ordering, so it also suppresses a **second, later, legitimate knockdown crit** on the same creature within the same chunk once any stand-up message has fired — a mirror-image version of the bug this PR is fixing. Confirmed with a runnable probe test, not merely suspected.

## Major

### 1. A later crit-based knockdown in the same chunk is wrongly suppressed once any earlier stand-up message fired — `lib/gemstone/combat/processor.rb:1680`

```ruby
if (pos = crit[:position]) && !position_recovered?(creature.id)
  status = pos.to_s.downcase
  (POSITION_STATUSES - [status]).each { |s| creature.remove_status(s) }
  creature.add_status(status)
```

`@position_recovered` (set in `note_position_recovery`, `lib/gemstone/combat/processor.rb:1414`, and cleared only in `clear_position_recovery`, `:1424`, which fires solely from a *message*-driven knockdown at `:1788`) is a single per-creature boolean for the whole chunk — it records "was this creature stood up by a message *somewhere* in this chunk," not "was it stood up *after* this particular crit." Every crit for that creature is applied in a pass that runs strictly after the whole chunk has been parsed (`persist_event` is called from `process` only once `parse_events` returns), so the guard cannot distinguish a crit that occurred (in game-text order) *before* the recovery message from one that occurred *after* it — both see the same `position_recovered?` value.

Concretely: two separate swings land on the same creature in one chunk (a companion's swing and your own, or two of your own multi-hit rounds), with a stand-up message printed between them:

```
You swing a broadsword at <creature>! ... crit: knocked down     # event 1
<creature> stands up.                                            # message, notes recovery
You swing a broadsword at <creature>! ... crit: knocked down     # event 2 (SHOULD win — it's later)
```

The last game-text fact is a real, fresh knockdown, so the creature should end up prone. Instead `position_recovered?` is still `true` when event 2's crit is applied (nothing cleared it — only a *message*-driven knockdown does, at `:1788`), so event 2's knockdown is silently dropped too, and the registry leaves the creature standing when the game just knocked it down again. This is the same class of bug the PR fixes (crit-vs-message ordering), just in the direction the fix didn't cover.

I verified this is not theoretical: I wrote a runnable spec against this exact branch reusing the PR's own test fixtures (two swings with the PR's own knockdown-crit line, a stand-up message between them) — it fails on `pr-1595` exactly as described (creature ends up NOT prone when it should be). It passes trivially on `main` for the wrong reason (main has the *opposite* bug: it always re-prones regardless of ordering).

The PR description itself scopes the fix to message-vs-message ordering ("A later knockdown *message* clears the note, so the last message in the chunk wins either way") — crit-vs-crit-across-a-recovery ordering was evidently not considered. Given multi-combatant/multi-swing rounds landing on one creature within a single chunk are exactly the kind of thing this module's other comments describe as normal (group hunts, multi-strike attacks), this isn't a corner case that can be dismissed as unreachable.

A fix likely needs the recovery note to carry the *line index* it was set at (parse_events already threads `index` through the line loop) rather than a bare boolean, so `apply_hit_crit_statuses` — which knows which event/hit it's applying, and could carry the crit's own line index forward — can compare "was the recovery message after *this specific* crit's line," not "was there ever a recovery message in this chunk." Confidence: confirmed (traced in code, reproduced with a runnable spec).

## What this PR gets right

- Correctly identifies and fixes the reported bug: the new `status_order_spec.rb` test 1 fails cleanly on `main` and passes on this branch; I confirmed this myself by checking the spec out onto `main` alone.
- The per-chunk reset (`@position_recovered = nil` at the top of `parse_events`, `:257`) correctly follows the same "chunk-local state" idiom already used for `chunk_dispels`, `foreign_latch`, etc. elsewhere in this file, and the "does not carry a recovery into the next chunk" test confirms it doesn't leak across chunks.
- Scope is tight and well-isolated: the guard only touches the `position` branch of crit-status application — stun/roundtime/silence/etc. from the same crit are unaffected, and crit damage/wound application (a separate code path) is untouched.
- No external callers of the new methods exist anywhere in the repo (`note_position_recovery`, `clear_position_recovery`, `position_recovered?`, `@position_recovered` are all private to this module), so the blast radius really is contained to this one interaction.
- Full combat spec suite passes (249 examples, 0 failures) and `rubocop` is clean on both changed files, matching the PR's stated test plan.
- Comments are thorough and in the file's established style (`@param`/`@return` tags, "why" over "what"), consistent with the rest of the module.

## Coverage notes

- Full local clone of `elanthia-online/lich-5`, PR branch `pr-1595` (`faa5db1c`) checked out against `main` (`526ad7cd`). Diff, file contents, and history all reviewed at head.
- Ran `bundle exec rubocop lib/gemstone/combat/processor.rb spec/lib/gemstone/combat/status_order_spec.rb` — clean, no offenses (Ruby 4.0.5, matching `TargetRubyVersion: 4.0`).
- Ran `bundle exec rspec spec/lib/gemstone/combat/status_order_spec.rb` on the PR branch (4/4 pass) and again with only that spec file checked out onto `main` (1/4 fails — the exact bug the PR describes, confirming the spec is grounded and not a false positive).
- Ran the full `spec/lib/gemstone/combat` suite: 249 examples, 0 failures, matching the PR's own test plan.
- Wrote and ran an additional probe spec (not part of the PR, discarded after use) reproducing the Major finding above; it fails on the PR branch, confirming the gap is real and not speculative.
- CodeRabbit CLI is installed and authenticated, but the account's review quota was exhausted for this org (`Rate limit exceeded`, 3/3 included reviews used, no assigned seat for usage-based reviews) — no CodeRabbit findings could be incorporated. This review's findings are otherwise self-contained.
- No callers of the changed/new symbols exist elsewhere in the repo (checked via repo-wide grep), so no further blast-radius investigation was needed outside this one file.
- Did not attempt to verify against a live game feed or fixture corpus (the file's comments reference a private `logs/examples/` replay corpus that isn't part of this checkout) — the Major finding is grounded entirely in a runnable unit-level reproduction instead.


Nisugi:
Confirmed the Major finding and fixed it in f28e84f7.

You were right, and for the reason you gave: the recovery note was a single per-chunk boolean, so it recorded *that* a stand-up happened, not *where*. Since crit statuses are applied after the whole chunk is parsed, every crit for a creature saw the same answer and could not distinguish one that preceded the stand-up from one that followed it. A second, legitimate knockdown crit was dropped. I reproduced your case before changing anything: two swings with a stand-up between them left the creature standing, and it passed on `main` only because `main` has the opposite bug and re-prones unconditionally.

The fix is the direction you suggested. The recovery now stores the line index the stand-up was read on, and each crit carries the line its crit text sat on (`hit[:_crit_line]`, stamped at the lookahead that already knows `next_line_index`). `position_recovered?` compares the two, so a recovery only outranks the crits that actually preceded it. `apply_status_to_target` takes an optional `line:` keyword, threaded from the four call sites inside the line loop where `index` is already in scope.

One deliberate choice worth flagging: when a caller supplies no line, the recovery is stored as `Float::INFINITY` rather than nil, so it still counts as "after every crit". My first pass made a nil line fall through to "not recovered", which quietly reinstated the original bug for any caller that omitted the keyword. Unordered but real should mean the safe reading, which is never re-proning a creature the game just stood up.

Added your scenario as a fifth test in `status_order_spec.rb`. I verified it fails with the library change reverted and the spec kept, with exactly the symptom you described, so it is a real regression test rather than one that passes vacuously.

`spec/lib/gemstone/combat` is 250 examples, 0 failures, and `rubocop` is clean on both files. On the full suite I saw one difference against the baseline, in `spec/lib/common/setup_files_spec.rb`. It is flaky independent of this branch: running that file alone on the unmodified head failed 3 of 5 times. Unrelated to this change, but noting it since it shows up in a full run.

Thanks for the probe spec and for actually running it. The finding was precise enough to act on directly.

mrhoribu (approved):
## Follow-up: Major finding resolved

Commit f28e84f7 fixes the ordering gap from my previous review. The per-chunk recovery note now carries the line index the stand-up was read on, and each crit carries the line its crit text sat on (`hit[:_crit_line]`), so `position_recovered?` compares "did the creature stand up *after this specific crit*" instead of a single per-chunk boolean. That correctly lets a second, later knockdown crit win over an earlier stand-up.

Verified independently, not just by reading the diff:
- Reran the standalone probe spec that failed against the previous commit (two swings on one creature with a stand-up message between them) — now passes.
- The new spec example (`lets a knockdown crit after the stand-up win, being later still`) covers the same scenario and passes.
- Full `spec/lib/gemstone/combat` suite: 250 examples, 0 failures.
- `rubocop` clean on both changed files.
- Traced all four call sites of `apply_status_to_target` in `parse_events` — all now thread `line: index` through, so the ordering data is wired up everywhere it's needed, not just at one call site.
- Checked the `crit_line.nil?` fallback (treated as still-recovered) — the only code path that sets `crit[:position]` also always sets `hit[:_crit_line]`, so that fallback is dead defensive code, not a live gap.

No remaining concerns from my review.

PR #1597, opened 2026-09-11
fix(gs): track spell cooldowns on the characters a spell locks out

Nisugi:
## Summary

Several spells stop a character from receiving them again for a while. The caster has no way to know who is still locked out: for a group (EVOKE) casting the game's messaging is the same whether it landed on everyone or nobody, and the refusal is shown only to the target. Lich tracked none of it, so a script casting on a group had no way to ask who was worth casting on.

`Group` now keeps those cooldowns, keyed by spell and character name, from two kinds of message.

**A group casting** names the group in the caster's own start message, where a self-cast and a received casting do not, so seeing one stamps every current group member. The stamps are optimistic -- a member out of range, or already locked out by an earlier casting, gets one that is too early -- but the next casting corrects it, so the error is bounded by a single cooldown and does not accumulate. Someone who joins later has no stamp and reads as ready, which is right.

**A single-target casting** names the character it landed on in the third person ("A wall of force surrounds Dicate."). That cooldown belongs to the character rather than to the caster-target pair, so the line is enough on its own -- no need to know who gestured, and seeing someone else lock a character out is as useful as doing it yourself. Because it keys off the effect message and not the cast command, a misfire correctly leaves the existing cooldown running.

Lengths and the third-person messages come from the spell's effect-list entry (`group-cooldown`, `target-cooldown`, and a `target-start` message), so spells turn on as that data lands and spells without it are unaffected -- companion PR in `elanthia-online/scripts`. `Spell.target_upmsgs` is empty until then, and since an empty alternation would match every line, the pattern falls back to one that matches nothing.

```ruby
Group.spell_cooldown_ready?(Spell[215], "Grhim")  # => true/false
Group.spell_cooldown_left(Spell[140], "Dicate")   # => seconds, 0.0 when ready
Group.spell_cooldown_ready(Spell[215])            # => members castable now
```

Expired entries are left in place rather than swept: reads clamp at zero, so a stale entry never gives a wrong answer, and the store is keyed by name so repeat castings overwrite rather than accumulate.

## Test plan

- [x] `bundle exec rspec spec/lib/gemstone/group_spell_cooldown_spec.rb` -- 17 examples covering both message paths, an unknown character, expiry, the never-negative clamp, a member joining after a casting, per-spell isolation, and re-stamping
- [x] Full suite: 7164 examples. The failures are the same set as on `main` (57 there, all in unrelated frontend/msgbox/session specs, order-dependent) -- no file fails here that does not already fail on `main`
- [x] rubocop clean on all four files
- [x] In game with the companion effect-list data: Wall of Force on a character reads 0.0 before and 268.98 seconds after, and a misfire leaves the running cooldown alone instead of restarting it

Only Wall of Force (140) and Heroism (215) were exercised in game; Bravery (211), Spell Shield (219) and Celerity (506) go through the same two code paths and differ only in their data.

### DIFF
--- a/lib/common/spell.rb
+++ b/lib/common/spell.rb
@@ -134,14 +134,20 @@ module Lich
           end
         }
         @cast_proc = xml_spell.locate('cast-proc').first&.text
-        # Seconds a target is immune to another group casting of this spell.
-        # Only the group (EVOKE) versions of a few spells carry one; nil means
-        # the spell has no per-target cooldown. See Group.spell_cooldown.
-        @group_cooldown = xml_spell['group-cooldown']&.to_i
-        # Seconds a character is immune to this spell after it lands on them,
-        # regardless of who cast it, paired with the third-person message that
-        # names them. See Group.spell_cooldown_ready?.
-        @target_cooldown = xml_spell['target-cooldown']&.to_i
+        # Seconds a character is locked out of this spell, by kind. A 'group'
+        # cooldown covers another group (EVOKE) casting of it; a 'target'
+        # cooldown covers the spell landing on them from any caster, paired
+        # with the third-person message that names them. Both are nil when the
+        # spell declares no cooldown of that kind. See Group.spell_cooldown and
+        # Group.spell_cooldown_ready?.
+        xml_spell.locate('cooldown').each { |xml_cooldown|
+          case xml_cooldown['type'].to_s.downcase
+          when 'group'
+            @group_cooldown = xml_cooldown.text.to_i
+          when 'target'
+            @target_cooldown = xml_cooldown.text.to_i
+          end
+        }
         @target_msgup = xml_spell.locate('message')
                                  .select { |e| e['type'].to_s.downcase == 'target-start' }
                                  .collect { |e| e.text }.join('$|^')
--- a/lib/common/spell.rb
+++ b/lib/common/spell.rb
@@ -15,7 +15,7 @@ module Lich
       @@cost_list ||= Array.new
       @@load_mutex = Mutex.new
       @@after_stance = nil
-      attr_reader :num, :name, :timestamp, :msgup, :msgdn, :circle, :active, :type, :cast_proc, :real_time, :persist_on_death, :availability, :no_incant, :last_cast
+      attr_reader :num, :name, :timestamp, :msgup, :msgdn, :circle, :active, :type, :cast_proc, :real_time, :persist_on_death, :availability, :no_incant, :last_cast, :group_cooldown
       attr_accessor :stance, :channel
 
       @@prepare_regex = Regexp.union(
@@ -134,6 +134,10 @@ module Lich
           end
         }
         @cast_proc = xml_spell.locate('cast-proc').first&.text
+        # Seconds a target is immune to another group casting of this spell.
+        # Only the group (EVOKE) versions of a few spells carry one; nil means
+        # the spell has no per-target cooldown. See Group.spell_cooldown.
+        @group_cooldown = xml_spell['group-cooldown']&.to_i
         @last_cast = Time.at(0)
         @timestamp = Time.now
         @timeleft = 0
--- a/lib/common/spell.rb
+++ b/lib/common/spell.rb
@@ -15,7 +15,7 @@ module Lich
       @@cost_list ||= Array.new
       @@load_mutex = Mutex.new
       @@after_stance = nil
-      attr_reader :num, :name, :timestamp, :msgup, :msgdn, :circle, :active, :type, :cast_proc, :real_time, :persist_on_death, :availability, :no_incant, :last_cast, :group_cooldown
+      attr_reader :num, :name, :timestamp, :msgup, :msgdn, :circle, :active, :type, :cast_proc, :real_time, :persist_on_death, :availability, :no_incant, :last_cast, :group_cooldown, :target_cooldown, :target_msgup
       attr_accessor :stance, :channel
 
       @@prepare_regex = Regexp.union(
@@ -138,6 +138,14 @@ module Lich
         # Only the group (EVOKE) versions of a few spells carry one; nil means
         # the spell has no per-target cooldown. See Group.spell_cooldown.
         @group_cooldown = xml_spell['group-cooldown']&.to_i
+        # Seconds a character is immune to this spell after it lands on them,
+        # regardless of who cast it, paired with the third-person message that
+        # names them. See Group.spell_cooldown_ready?.
+        @target_cooldown = xml_spell['target-cooldown']&.to_i
+        @target_msgup = xml_spell.locate('message')
+                                 .select { |e| e['type'].to_s.downcase == 'target-start' }
+                                 .collect { |e| e.text }.join('$|^')
+        @target_msgup = nil if @target_msgup.empty?
         @last_cast = Time.at(0)
         @timestamp = Time.now
         @timeleft = 0
@@ -244,6 +252,13 @@ module Lich
       def Spell.upmsgs
         Spell.load unless @@loaded
         @@list.collect { |spell| spell.msgup }.compact
+      end
+
+      # Third-person start messages, for spells that name the character they
+      # land on. Empty until the effect list carries target-start messages.
+      def Spell.target_upmsgs
+        Spell.load unless @@loaded
+        @@list.collect { |spell| spell.target_msgup }.compact
       end
 
       def Spell.dnmsgs
--- a/lib/gemstone/group.rb
+++ b/lib/gemstone/group.rb
@@ -191,9 +191,20 @@ module Lich
         seconds = spell.group_cooldown
         return nil if seconds.nil? || seconds <= 0
 
-        expires = Time.now + seconds
+        now = Time.now
+        expires = now + seconds
         store = (@@spell_cooldowns[spell.num] ||= {})
-        members.each { |member| store[member.noun] = expires }
+        # _members, not members: this runs on the parser thread, and members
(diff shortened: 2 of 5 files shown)


mrhoribu (approved):
# PR Review: feat(gs): track spell cooldowns on the characters a spell locks out (elanthia-online/lich-5#1597)

**Reviewed:** 4 files, +355/-2 at `27d3f4ba` on 2026-09-11. Based on: full clone (local worktree from `origin`, with history).
**Summary:** Adds `Group`-side tracking of per-character spell-lockout cooldowns, inferred from two kinds of game text (a group-casting's own start message, and a third-person "landed on X" message) since the game gives the caster no other way to know who is still locked out. The core logic is careful, well-reasoned in its own comments, matches an established parser/data convention in this file, and is backed by 20 passing unit tests plus a clean rubocop run. The main things left it depends on live outside this diff: a companion `elanthia-online/scripts` data change, and — as a result — its own new XML-parsing code path is untested against a real `<spell>` element.

## Minor

### 1. No test exercises the new XML attribute/message parsing in `Spell#initialize` — [lib/common/spell.rb:140-148](lib/common/spell.rb#L140-L148)

```ruby
@group_cooldown = xml_spell['group-cooldown']&.to_i
@target_cooldown = xml_spell['target-cooldown']&.to_i
@target_msgup = xml_spell.locate('message')
                         .select { |e| e['type'].to_s.downcase == 'target-start' }
                         .collect { |e| e.text }.join('$|^')
@target_msgup = nil if @target_msgup.empty?
```

Every new test in [spec/lib/gemstone/group_spell_cooldown_spec.rb](spec/lib/gemstone/group_spell_cooldown_spec.rb) either stubs `group_cooldown`/`target_cooldown` directly on a `Spell` instance (`allow(spell).to receive(:group_cooldown).and_return(180)`) or relies on the committed `spec/fixtures/effect-list.xml`, which I confirmed has zero `group-cooldown`/`target-cooldown`/`target-start` occurrences (companion data hasn't landed yet, per the PR body). That means this actual attribute-extraction code — the `xml_spell['group-cooldown']` key lookup, the `'target-start'` type filter, the `.join('$|^')`/empty-to-nil construction — is never exercised end-to-end by any test in the suite. A typo in the attribute key (e.g. `'group_cooldown'` instead of `'group-cooldown'`, or filtering on `'target_start'`) would silently leave `group_cooldown`/`target_cooldown`/`target_msgup` `nil` for every spell forever, indistinguishable from "this spell has no cooldown," and nothing here would fail. Once the companion `effect-list.xml` update lands, updating the fixture and adding one assertion that `Spell.new` on a `<spell group-cooldown="180">` element actually produces `group_cooldown == 180` (and similarly for `target-cooldown`/`target-start`) would close this gap cheaply.

## Open questions

- [lib/gemstone/infomon/parser.rb:656](lib/gemstone/infomon/parser.rb#L656) — `target = Regexp.last_match(:noun)` requires that the combined `SpellTargetUpMsgs` regex (built by unioning every spell's `target_msgup`, [lib/gemstone/infomon/parser.rb:170](lib/gemstone/infomon/parser.rb#L170)) declares a `(?<noun>...)` named group *somewhere* in its source. I verified in a standalone Ruby check that `Regexp.last_match(:noun)` raises `IndexError: undefined group name reference: noun` if the name doesn't appear in the compiled pattern at all, but returns `nil` (not an error) if it appears in any alternative, even one that didn't participate in the match. Since `target_msgup` messages are raw text pulled from `elanthia-online/scripts`' `effect-list.xml` (a companion PR, not part of this diff) and joined the same way `msgup`/`msgdn` already are, whether every future `target-start` message actually contains a `(?<noun>...)` capture is a fact this repo cannot see or enforce. The PR author's own in-game test (Wall of Force) confirms it works for at least that spell's current wording. If a future spell's `target-start` text is added to the companion data without the named group — and it's the *only* spell with a `target-start` message at that point, or all others also omit it — every line matching that spell's message would hit the top-level `rescue StandardError` in `self.parse` ([lib/gemstone/infomon/parser.rb:816](lib/gemstone/infomon/parser.rb#L816)), log an error, and silently never record that spell's cooldown. **Settles it:** does the companion `elanthia-online/scripts` PR's `target-start` message format have any validation (schema, lint, or just reviewer discipline) ensuring a `(?<noun>...)` group is always present, or is that convention undocumented outside this code's comments?

## What this PR gets right

- The three commits read like a real incremental fix cycle, not just a feature drop: the third commit catches and fixes two genuine bugs from the first two (parser-thread deadlock via `Group.members`→`Group.check`, and cooldowns being incorrectly extended for members who didn't receive the casting) with clear, specific commit messages explaining the mechanism. I independently traced the `_members` vs `members` distinction (`members` calls `maybe_check` → `Group.check`, which sends `<c>group\r\n` and blocks on `wait_until { Group.checked? ... }` — a genuine deadlock risk if called from the parser thread that must itself process the reply) and confirmed the fix is correct and uses a pre-existing accessor built for exactly this (`_members`, present since a 2024-01-12 commit, "used internally by the Observer to avoid infinite loops").
- The "skip members already on a later cooldown" logic in `record_spell_cooldown` ([lib/gemstone/group.rb:96-103](lib/gemstone/group.rb#L96-L103)) is correctly targeted: it only skips when `current > now`, so an *expired* stamp is still refreshed, and a member who just joined (no stamp) is still stamped — matches the three test cases covering exactly these branches.
- The `/(?!)/` fallback for an empty `target_upmsgs` union ([lib/gemstone/infomon/parser.rb:170](lib/gemstone/infomon/parser.rb#L170)) is the right defensive move — an empty `Regexp.union`-style alternation from `.join('$|^')` on an empty array would produce `//` (matches everything), and the comment correctly identifies and avoids that trap.
- Sharing one flat `@@spell_cooldowns[spell.num][name]` store between group-casting stamps and named-target stamps is a deliberate, tested design choice ("shares the store with group castings, so either source answers" — spec at [spec/lib/gemstone/group_spell_cooldown_spec.rb:427-430](spec/lib/gemstone/group_spell_cooldown_spec.rb#L427-L430)), and it's the right call: both represent the same real-world fact (is character X locked out of spell Y), so unifying the key namespace by character name is correct rather than incidental.
- Test suite is genuinely thorough for the surface it covers: 20 examples across three `describe` blocks, including the not-obviously-necessary-but-valuable regression test "does not send GROUP from the parser thread" that pins down the exact bug fixed in commit 3.

## Coverage notes

- Reviewed against a full local clone (`git worktree` off `origin/main` at the PR's head commit `27d3f4ba`), not just the diff — read every changed file in full, plus `lib/gemstone/group.rb` and `lib/common/spell.rb` in their entirety for surrounding context, and used `git blame`/`git log -S` to check the history of `_members`, `group-cooldown`, and `target-cooldown`.
- Ran the new spec file directly: `bundle exec rspec spec/lib/gemstone/group_spell_cooldown_spec.rb` — 20 examples, 0 failures.
- Ran `rubocop` on all four changed files under Ruby 4.0.5 (matches the project's `TargetRubyVersion`) — 0 offenses.
- Ran the neighboring spec files (`group_load_spec.rb`, `spell_spec.rb`, `infomon/xmlparser_spec.rb`, `infomon/activespell_spec.rb`, `infomon_spec.rb`) individually to check for regressions. All pass except `infomon/xmlparser_spec.rb`, which fails identically on `origin/main` at the PR's base commit (`uninitialized constant ...Pattern::Overwatch`) — confirmed pre-existing and unrelated to this PR, consistent with the PR body's claim of pre-existing, order-dependent failures on `main`. Did not run the full 7,164-example suite the PR body references; spot-checked one of the claimed-pre-existing failures instead of re-running the whole suite.
- Ran a local CodeRabbit pass (`coderabbit review --agent --base-commit 526ad7cd... --committed`) against the merge-base commit: 0 findings.
- Verified the `Regexp.last_match(:name)` IndexError-vs-nil behavior with a standalone Ruby script (not something I could confirm purely by reading), used to ground the Open Question above.
- Did not and cannot verify the actual content of the companion `elanthia-online/scripts` effect-list.xml PR, since it lives in a different repository and wasn't linked with a URL in this PR's description.

Nisugi:
Thanks for the thorough review. Both points addressed in 74b75f73.

**Data format** — moved the cooldowns off the `<spell>` attributes and onto `<cooldown type='group'>` / `<cooldown type='target'>` children, per your comment on the companion PR. Matches how `duration`, `cost` and `bonus` are already carried, and takes a further cooldown kind without adding an attribute per kind. Companion data updated in elanthia-online/scripts#2464.

**Untested XML parsing** — the spec fixture now carries real cooldown and `target-start` data for the five spells that declare them, so `Spell#initialize`'s extraction runs end to end instead of being stubbed past. New examples assert the parsed values, that a spell declaring neither stays nil, and that the two kinds don't bleed into each other. I checked the specs actually catch the failure mode you described by mistyping the element type and confirming they fail.

**Open question re: `(?<noun>)`** — good catch, and rather than leave it to reviewer discipline in another repo, there's now a spec asserting every `target-start` message declares a `noun` capture:

```ruby
it 'declares a :noun capture in every target-start message' do
  expect(described_class.target_upmsgs).not_to be_empty
  described_class.target_upmsgs.each do |msg|
    expect(Regexp.new(msg).names).to include('noun'), "missing (?<noun>) in: #{msg}"
  end
end
```

Companion data missing the group now fails the suite here rather than hitting the top-level rescue at runtime.

27 examples pass (was 20), rubocop clean on all changed files.

One note on merge order: the parsing change and the data change are each harmless alone, since the fixture carries its own copy. But if the data lands first and users pull a fresh effect-list before this ships, cooldowns read as nil until it does. Worth landing this one first or close together.

PR #1599, opened 2026-09-11
fix(all): setup_files uses content hash for cache freshness (#1596)

### LINKED ISSUE #1596: SetupFiles cache can serve stale data after a fast rewrite, making specs flaky on CI
## Summary

`SetupFiles` decides whether a cached file is current by comparing the stored mtime against the file's current mtime ([`lib/common/setup_files.rb#L260`](https://github.com/elanthia-online/lich-5/blob/main/lib/common/setup_files.rb#L260)):

```ruby
if cached_file.nil? || cached_file.mtime != last_modified_date
  cache_put_by_filepath(filepath)
end
```

When a file is rewritten quickly enough that both writes land in the same filesystem timestamp tick, the mtime is unchanged, the cache considers the entry current, and `get_data` / `reload` return the **previous** contents.

This is a real cache-correctness bug, not just a test issue. Any caller that rewrites a setup or data file and reloads shortly after can read stale values. It shows up most visibly as intermittent spec failures on CI.

## Impact on CI

The spec suite fails intermittently on `main` with no code changes involved. This lands on unrelated PRs and costs a rerun each time.

Observed on `main` at `526ad7cd` (release 5.21.0):

- Running `spec/lib/common/setup_files_spec.rb` alone, with a fixed seed, in identical consecutive runs, flips between pass and fail.
- The full suite under `--seed 36105` produced 65 failures on a clean `main` checkout.
- A recent PR CI run failed with 8 `setup_files_spec.rb` failures, then passed on rerun with no changes.

Because it is timing-dependent rather than order-dependent, a fixed seed does **not** make it reproducible, and the set of failing examples varies between runs.

## Reproduction

```
git checkout 526ad7cd
for i in 1 2 3 4 5; do bundle exec rspec --seed 36105 spec/lib/common/setup_files_spec.rb; done
```

Some runs report `31 examples, 0 failures`, others report a failure. The example that flakes most readily is `#get_data with a custom data file falls back to scripts/data after a custom file is removed` (`setup_files_spec.rb:113`), which writes a custom YAML file, reads it, rewrites/removes it, reloads, and expects the base file's values.

Timestamp granularity is easy to demonstrate directly:

```ruby
require "tmpdir"
Dir.mktmpdir do |d|
  f = File.join(d, "t.yaml")
  same = 0
  200.times do
    File.write(f, "a: 1"); m1 = File.mtime(f)
    File.write(f, "a: 2"); m2 = File.mtime(f)
    same += 1 if m1 == m2
  end
  puts "identical mtime after rewrite: #{same}/200"
end
```

On Windows this reported `130/200`. Any platform whose mtime granularity is coarser than the gap between the two writes will hit it.

## Note on what is *not* broken

Eviction of deleted files is handled separately and correctly, by comparing the glob against what is on disk rather than by mtime. Only the rewrite path is affected.

## Suggested fix

Stop relying on mtime alone as the freshness signal. Options, roughly in order of robustness:

1. Store a content hash (or the parsed payload's digest) and compare that.
2. Compare `(mtime, size)` together — cheap, and catches most same-tick rewrites, though not a same-size edit.
3. Store the full-precision timestamp where available, which narrows the window but does not close it on coarse filesystems.

Adding a sleep to the specs would hide the symptom while leaving the stale-read behavior in place for real callers, so it seems worth fixing in the cache itself.

MahtraDR:
## Summary

Fixes #1596. `SetupFiles` decided cache freshness by comparing stored mtime against the file's current mtime. When two writes land within one filesystem timestamp tick — common on Windows and some CI mounts — the mtime is unchanged, so `get_data` / `get_settings` / `reload` serve the **previous** contents. This surfaces as intermittent, non-order-dependent `setup_files_spec.rb` failures on CI.

While tracing it I found a second, related manifestation: the `scripts/data/custom` → `scripts/data` fallback. The cache is keyed by basename, and after a custom override is deleted the reload compares the **base** file's mtime against the cached **custom** entry's mtime. If those two files share an mtime, the stale custom payload is served. This is the example that flakes most readily (`falls back to scripts/data after a custom file is removed`).

## Change

- Freshness is now keyed on a **SHA256 content digest** rather than mtime. Content is authoritative and independent of the filesystem clock.
- Additionally reload when the **winning source directory changes** (`cached.path != File.dirname(filepath)`), so a custom-override removal is never masked by a shared mtime.
- `mtime` is retained on `FileInfo` for debug/inspect output only.
- `file_digest` returns `nil` on read errors (file vanished between glob and hash), which never matches a stored digest, so the entry reloads — fail-safe.

Deleted-file eviction was already correct (glob-vs-disk) and is unchanged.

## Why not the other options from the issue

- `(mtime, size)` is cheap but silently misses same-size edits (e.g. flipping a boolean or swapping an equal-length value) — it wouldn't restore the correctness guarantee.
- "Full-precision mtime" is effectively a no-op: `File.mtime` already returns full-precision `Time`; the ceiling is the filesystem, not the stored value, so it does nothing on the ~1s-granularity filesystems that actually trigger this.

Content hashing is negligible for these tiny, infrequently-loaded YAML files.

## Tests

- New `cache freshness (filesystem-timing-independent)` block **pins `File.mtime` to a frozen value** to force the same-tick collision deterministically — no sleeps, no dependence on host timestamp resolution. Covers rewritten data file, custom-override removal, and profile settings change.
- Verified these fail against the pre-patch code and pass after it.
- Removed the `FileUtils.touch(..., mtime: Time.now + 1)` band-aid the existing `reloads changed files` spec needed — content hashing makes it unnecessary.
- Full file: 34 examples, 0 failures; rubocop clean.

### DIFF
--- a/lib/common/setup_files.rb
+++ b/lib/common/setup_files.rb
@@ -8,12 +8,13 @@
   - Base YAML files (base.yaml, base-empty.yaml)
   - Character-specific YAML files ({character}-setup.yaml)
   - Include files with recursive resolution and circular dependency protection
-  - Automatic caching with modification-time checking
+  - Automatic caching with content-hash freshness checking
   - Deep-clone protection against in-memory mutation
 
   @see https://elanthipedia.play.net/Lich_script_development#dependency
 =end
 
+require 'digest'
 require 'monitor'
 require 'ostruct'
 require 'set' # rubocop:disable Lint/RedundantRequireStatement -- needed for Ruby < 3.2
@@ -27,13 +28,14 @@ module Lich
       include MonitorMixin
 
       class FileInfo
-        attr_reader :path, :name, :mtime
+        attr_reader :path, :name, :mtime, :digest
 
-        def initialize(path:, name:, data:, mtime:)
+        def initialize(path:, name:, data:, mtime:, digest: nil)
           @path = path
           @name = name
           @data = data
           @mtime = mtime
+          @digest = digest
         end
 
         # Deep clone of data to prevent scripts from mutating cached settings.
@@ -51,7 +53,7 @@ module Lich
         end
 
         def inspect
-          "#<SetupFiles::FileInfo @name=#{@name}, @path=#{@path}, @mtime=#{@mtime}>"
+          "#<SetupFiles::FileInfo @name=#{@name}, @path=#{@path}, @mtime=#{@mtime}, @digest=#{@digest}>"
         end
       end
 
@@ -254,26 +256,47 @@ module Lich
           end
 
           current_files.each do |filename, filepath|
-            last_modified_date = File.mtime(filepath)
+            # Freshness is decided by file content (SHA256), not mtime. mtime is
+            # unreliable: two writes within one filesystem timestamp tick share
+            # an mtime, so an mtime-only check serves stale data after a fast
+            # rewrite (see issue #1596). We also reload when the winning source
+            # path changes (e.g. a scripts/data/custom override is removed and
+            # the same basename now resolves to scripts/data), which an
+            # mtime-only check masks when the two files share an mtime.
+            current_digest = file_digest(filepath)
             cached_file = cache_get_by_filename(filename)
-            safe_log "#{self.class}::#{__callee__} filepath=#{filepath}, last_modified_date=#{last_modified_date}, cached_file=#{cached_file.inspect}" if @debug
-            if cached_file.nil? || cached_file.mtime != last_modified_date
-              cache_put_by_filepath(filepath)
+            safe_log "#{self.class}::#{__callee__} filepath=#{filepath}, current_digest=#{current_digest}, cached_file=#{cached_file.inspect}" if @debug
+            if cached_file.nil? ||
+               cached_file.path != File.dirname(filepath) ||
+               cached_file.digest != current_digest
+              cache_put_by_filepath(filepath, digest: current_digest)
             end
           end
         end
       end
 
-      def cache_put_by_filepath(filepath)
+      def cache_put_by_filepath(filepath, digest: nil)
         synchronize do
           safe_log "#{self.class}::#{__callee__} filepath=#{filepath}" if @debug
           @files_cache[File.basename(filepath)] = FileInfo.new(
             path: File.dirname(filepath),
             name: File.basename(filepath),
             mtime: File.mtime(filepath),
+            digest: digest || file_digest(filepath),
             data: safe_load_yaml(filepath)
           )
         end
+      end
+
+      # Content hash used as the cache freshness signal. Returns nil when the
+      # file cannot be read (e.g. it vanished between the glob and here); a nil
+      # digest never equals a stored digest, so the entry is treated as changed
+      # and reloaded on the next pass.
+      def file_digest(filepath)
+        Digest::SHA256.file(filepath).hexdigest
+      rescue => e
+        safe_log "#{self.class}::#{__callee__} could not hash #{filepath}: #{e.message}" if @debug
+        nil
       end
 
       def cache_get_by_filename(filename)
--- a/spec/lib/common/setup_files_spec.rb
+++ b/spec/lib/common/setup_files_spec.rb
@@ -122,6 +122,48 @@ RSpec.describe Lich::Common::SetupFiles do
     end
   end
 
+  # Regression coverage for issue #1596: the cache must key freshness on file
+  # content, not mtime. These specs pin File.mtime to a single frozen value so
+  # every write shares an mtime -- the exact same-tick collision that occurs on
+  # coarse-granularity filesystems (Windows, some CI mounts). They fail against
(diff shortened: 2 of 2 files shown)


mrhoribu (approved):
# PR Review: fix(all): setup_files uses content hash for cache freshness (elanthia-online/lich-5#1599)

**Reviewed:** 2 files, +74/-11 at `7bc48957` on 2026-09-11. Based on: full clone (worktree checkout of the PR head against `origin/main`), plus full local test suite and rubocop run.
**Verdict:** Approve
**Summary:** Replaces `SetupFiles`' mtime-based cache-freshness check with a SHA256 content digest (plus a check that the winning source directory hasn't changed), fixing a real same-tick-collision bug where two fast writes to a settings/data file could leave stale cached content served indefinitely. The fix is correct, minimal, matches the existing `Digest::SHA256.file(path).hexdigest` idiom already used elsewhere in this repo (`lib/dependency_recovery.rb:686`), and ships tests that fail against the pre-patch code and pass after it.

## What I verified

- Checked out the PR head in an isolated worktree and reverted just `lib/common/setup_files.rb` to its pre-PR content while keeping the new spec file: the three new "cache freshness (filesystem-timing-independent)" examples fail exactly as claimed (wrong `mana`/`version` values returned), and no other example is affected. Restoring the PR's version of the file makes all 34 examples in `setup_files_spec.rb` pass.
- Ran the full repo suite at the PR head: 7149 examples, 0 failures.
- Ran rubocop on both changed files: no offenses.
- Grepped the repo for other callers of `SetupFiles`, `FileInfo`, and `cache_put_by_filepath` (`lib/global_defs.rb:68-79`, `lib/dragonrealms/drinfomon/startup.rb:101`) — none depend on `mtime`-driven freshness semantics or construct `FileInfo` directly outside the class itself, so the blast radius is contained to this file.

## What this PR gets right

- The root cause is correctly diagnosed and fixed at the right layer: content, not mtime, is what freshness should be keyed on, and the PR body's rejection of `(mtime, size)` and "full-precision mtime" alternatives is sound (same-length edits and coarse filesystem clocks both defeat those).
- The added `cached_file.path != File.dirname(filepath)` check independently fixes the second bug the PR describes (custom-override removal masked by a shared mtime) — this isn't just relying on the digest changing, it's a direct structural check, so it also correctly forces a reload in the (rarer) case where a freshly-created override happens to have byte-identical content to the base file it's replacing.
- `file_digest` returning `nil` on a read error is a genuine improvement over the prior code, which called bare `File.mtime(filepath)` with no rescue — a file vanishing between `Dir.glob` and that call would have raised out of `load_files` entirely in the old code. Now it degrades to a fail-safe reload attempt instead of crashing.
- Digest is computed once per file per `load_files` pass and threaded through (`cache_put_by_filepath(filepath, digest: current_digest)`), so the common "content changed" path doesn't double-hash.
- The regression tests pin `File.mtime` to a frozen value to force the exact collision deterministically, rather than relying on sleeps or host filesystem resolution — this is the right way to test a timing bug and previously required a `FileUtils.touch(..., mtime: Time.now + 1)` band-aid in the `#reload` spec that this PR correctly removes as redundant.

## Minor

### 1. Every `get_data`/`get_settings`/`reload` call now hashes every matched file's full content, not just `stat`s it — `lib/common/setup_files.rb:266`

```ruby
current_digest = file_digest(filepath)
cached_file = cache_get_by_filename(filename)
```

This runs on every call to `load_files` (i.e. every `get_settings`, `get_data`, and `reload`), for every file matched by the relevant glob patterns, whether or not the file actually changed — an `mtime` check is a single `stat(2)`, a digest is a full read. dr-scripts calls `get_data(...)` from ~210 call sites (e.g. `craft.lic` re-fetches `get_data('recipes')` on most recipe lookups), so this isn't a cold-path operation repo-wide, though none of the sampled call sites are in a tight per-tick loop. Given the PR's own framing ("negligible for these tiny, infrequently-loaded YAML files") this is a deliberate, reasonable tradeoff and not a defect — flagging only because it's an absence the diff doesn't surface: if a script repo ever calls `get_data` in a hot per-iteration loop against a larger data file, this would be a measurable regression versus the old `stat`-only check. Not blocking.

## Open questions

- None — the change is small enough and well-enough tested that I don't have an unresolved question that would change the verdict.

## Coverage notes

Full clone available (`origin` remote already present locally), so this review is based on the complete repository at the PR head, not just the diff — including a full local test run (7149 examples, 0 failures) and rubocop (clean) on both changed files. The pre-patch-vs-post-patch test bisection (reverting the source file only, keeping the new specs) was run directly rather than taken on faith from the PR description. No CodeRabbit CLI was available/authenticated in this environment, so that pass was skipped; nothing in the diff suggested a class of issue (concurrency, security, resource handling) that would specifically benefit from a second automated pass beyond what was manually checked here. `~/dr-scripts` and `~/scripts` were available locally and used only to sample real-world call-site frequency for the performance note above, not as part of the reviewed diff.

PR #1600, opened 2026-09-12
fix(gs): combat module - Rysk/Dreadt log audit: Holy Weapon release model, held maneuver rolls, AoE roll attribution, live defs

Nisugi:
> **Stacked on #1593.** This branch is one commit on top of `combat-hunt-audit`; until #1593 merges the diff here shows both. Review only the top commit, `056ff899`. Once #1593 is in, this fast-forwards onto it.

## Summary

Follow-up to #1593, same method: the recorder's database compared line by line against two characters' raw XML logs (a rogue with a greatsword and Energy Wings; a paladin with a shield, mace and Holy Weapon), every divergence traced to a parser or def defect and fixed with the offending lines as evidence. Scope is `lib/gemstone/combat/**` and its specs only.

## Parser / processor

- **Holy Weapon (1625) infusion release is a pre-flare on the swing, not an attack.** `:weapon_cast` moves from spells to flares (release and failed-release forms). The spell it releases (Templar's Verdict, cast line `Violet flames erupt from beneath X.`) runs as its own event, parented to the swing; the first physical roll resumes the swing (`SPELL_RELEASING_FLARES`). Both orderings handled: maneuver → proc → spell → swing's AS/DS, and proc → spell → "You swing". Before this, one pummel recorded as a single `weapon_cast` row holding the spell's CS/TD *and* the swing's AS/DS, and the pummel bracket was dropped by its own spell. A releasing pre-flare may only be claimed by a weapon-bearing attack; a cast that printed before its swing is re-emitted after it so the recorder's parent-first lineage resolution holds.
- **A status settles a switch-born event** like an outcome does, so an AoE status effect that rolls once per onlooker (eviscerate: SSR → freeze → unsettled, twice) gives each bystander its own roll instead of piling both on the first.
- **An inbound creature maneuver claims the held maneuver roll that preceded its line** (sap spit whose miss prints on the next line; a subdue that lands and prints no outcome), generalising the same-line-outcome feint exception. The hold stays while a volley sequence, a barrage (the one assault whose aim SMR precedes each arrow) or a nock is open; every other assault rolls after its round line, fixture-verified. AS/DS is never claimed. Corpus census: 5 maneuver-SMR → inbound-line shapes, 0 with our own attack following in the chunk.
- **`attacks.attack_kind`** (was `ambush_kind`, never shipped): `'waylay'` / `'ambush'` (from hiding, `ambush=1`) or `'reverse_strike'` (parry reaction, `ambush=0`). Reverse Strike's confirmation line is a prefix on the swing that follows; the offer line is deliberately not matched (it prints whether or not the technique is used). Reaction prefixes set the kind without the ambush flag; a load-time guard keeps `AMBUSH_KINDS` aligned with `AMBUSH_PREFIXES`.

## Defs (all from the logs, evidence in comments)

- **Statuses:** `demoralized` (eviscerate's second onlooker effect, wiki-named) with its expiry; `terrified` gains its creature-side expiry line — it only had the 2p one, so a creature's terror never cleared.
- **Attacks:** Barbed Sweep (Energy Wings tier 2, per-target line only — the opener double-counted the first creature); a second shield-charge opener form (every one of 15 orphaned roll+damage+crits, 200 damage dropped outright); six inbound forms (treekin root/sap/fist, Illoke jarl ward attack, krolvin slaver subdue); Templar's Verdict on its erupt line.
- **Ticks:** three more bleed phrasings and seven poison phrasings as `:poison_tick` (unowned like bleed) — 31 orphaned tick rows.
- **Flares:** shield plasma with its target clause. **Outcomes:** shield block `...between yourself and the attack!`.

## Recorder

- `attack_kind` column + migration entry.

## Result on the two log sets

Orphaned damage 818 → 110 (the remainder is another player's item flavor naming no creature) and 423 → 0; no Templar row carries an AS/DS; total damage and hit counts unchanged throughout.

## Tests

- `eviscerate` fixture asserts both statuses; `weapon_cast` fixture replaced with the real pummel exchange (the old blob was a truncated fragment — note the fixture loader treats a blank line as a blob boundary); the inbound infusion spec asserts the pre-flare shape.
- `bundle exec rspec spec/lib/gemstone/combat`: 296 examples, 0 failures on this branch.

### DIFF
--- a/lib/gemstone/combat/defs/attacks.rb
+++ b/lib/gemstone/combat/defs/attacks.rb
@@ -154,6 +154,24 @@ module Lich
               /(?<attacker>.+?) springs upon (?<target>you|.+?) from behind and attempts to slit #{MK_PRE}(?:your|his|her|its)#{MK_POST} throat(?: with #{MK_PRE}(?:his|her|its)#{MK_POST} .+?)?!/
             ].freeze),
             AttackDef.new(:bearhug, [/(?<attacker>.+?) charges towards (?<target>you|.+?) and attempts to grasp #{MK_PRE}(?:you|him|her|it)#{MK_POST} in a ferocious bearhug!/].freeze),
+            # Mug (CMAN, rogues) - wiki: type Attack, "plunder your target's
+            # pockets as you distract them with an attack". The accost line
+            # OPENS the maneuver; the wrapped attack (any type, via
+            # CMAN MUG [attack] [target]) then resolves normally with its own
+            # AS/DS or UAF/UDF. The theft SMR prints AFTER that attack settles,
+            # followed by the pat-down announce, so the roll trails the swing
+            # it rode in on. Without a def both the accost and the theft roll
+            # orphaned into synthetic :unknown attacks (Rysk logs 2026-09-11:
+            # 71 accosts, 62 pat-downs, 62 orphaned SMR rolls).
+            # Theft outcomes: "In your hasty search, you don't find anything
+            # worth taking." / "You rifle his pockets and discover N silvers!"
+            # / "Your hasty search scatters the X's riches to the ground!"
+            # A creature that cannot be mugged prints "too wary" - per the
+            # wiki that is not a failure, the action is simply unavailable.
+            AttackDef.new(:mug, [
+              /You boldly accost (?<target>.+?), your attack masking your larcenous intent!/,
+              /Taking advantage of the scuffle, you roughly pat (?<target>.+?) down for hidden valuables!/
+            ].freeze),
             AttackDef.new(:charge, [
               /(?<attacker>.+?) rushes forward at (?<target>you|.+?) with #{MK_PRE}(?:his|her|its)#{MK_POST} .+? and attempts a charge!/,
               # 2p (logs/examples/weapon_charge.txt). The graded connect line
@@ -214,15 +232,42 @@ module Lich
             AttackDef.new(:shield_pin, [/You attempt to expose a vulnerability with a diversionary shield bash on (?<target>[^!]+)!/].freeze),
             AttackDef.new(:shield_push, [/You raise your (?<weapon>.+?) before you and attempt to push (?<target>.+?) away!/].freeze),
             AttackDef.new(:shield_strike, [/You launch a quick bash with your (?<weapon>.+?) at (?<target>[^!]+)!/].freeze),
-            AttackDef.new(:shield_charge, [/You raise your (?<weapon>.+?) before you and charge headlong towards (?<target>[^!]+)!/].freeze),
+            AttackDef.new(:shield_charge, [
+              /You raise your (?<weapon>.+?) before you and charge headlong towards (?<target>[^!]+)!/,
+              # Second 2p form (Dreadt log 2026-09-11: 15 occurrences, every
+              # one orphaning its SMR, damage and crits). Like :charge above,
+              # the graded connect line ("You lunge forward with an expert
+              # shield charge!") sits between the roll and the damage and is
+              # deliberately NOT a def - matching it would double-open the
+              # event. Ten grades observed (half-hearted..expert).
+              /You charge forward at (?<target>.+?) with your (?<weapon>.+?) and attempt a shield charge!/
+            ].freeze),
             # Assassinate (rogue): the uncoil line opens the event; the SMR
             # roll follows it directly, then the "carve into" hit line
             AttackDef.new(:eviscerate, [/You uncoil from the shadows, your (?<weapon>.+?) poised to eviscerate (?<target>[^!]+)!/].freeze),
             # Coup de Grace - messaging varies by weapon type; this is the
             # THW form, catalogue other weapon variants as they're observed
             AttackDef.new(:coup_de_grace, [/You lunge towards (?<target>.+?), intending to finish #{MK_PRE}(?:him|her|it)#{MK_POST} off!/].freeze),
-            # "dark wings" equipment proc (SMR attack); name pending
+            # Energy Wings (Dark), tier 1 offensive: Shadow Barb - a
+            # single-target SMR attack, disruption damage plus poison.
             AttackDef.new(:shadow_barb, [/Your umbrous wings twitch sharply, launching a barbed shard of shadow that hisses through the air toward (?<target>[^!]+)!/].freeze),
+            # Energy Wings (Dark), tier 2 offensive: Barbed Sweep - an AoE
+            # SMR attack over up to 5 targets. ONLY the per-target thrash
+            # line is the attack: it names its creature and is preceded by
+            # that creature's own SMR, so one row per target falls out of
+            # the normal switch path.
+            #
+            # The opener ("Your umbrous wings lash outward in a wide arc")
+            # is deliberately NOT matched. It carries no target, no roll and
+            # no damage, and including it double-counted the first creature:
+            # the opener adopted whichever target the first thrash line
+            # named, then that same thrash line fired as a second attack for
+            # the same creature - one row holding the roll, the next holding
+            # the damage (Rysk log 2026-09-11 7716). The cost is that a
+            # sweep which reaches nothing records nothing; it has no facts
+            # to record either way.
+            AttackDef.new(:barbed_sweep,
+                          [/The dark tendrils thrash across (?<target>.+?)'s#{MK_POST} skin in a devastating assault!/].freeze),
             # Whirlwind (THW AoE, single round - not a sequence). Known to
             # have ~4 message variants; two catalogued so far plus the
             # per-target strike line, all under one name.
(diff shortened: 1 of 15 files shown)


mrhoribu (approved):
# PR Review: fix(gs): combat module - Rysk/Dreadt log audit (elanthia-online/lich-5#1600)

**Reviewed:** 11 files, +404/-47 at commit `056ff899` on 2026-09-12. Based on: full clone + isolated
commit diff + running test/lint suites in a worktree.
**Summary:** A tightly-scoped, well-evidenced fix to the combat parser's Holy Weapon (proc→spell→swing)
handling, held-maneuver-roll attribution, and AoE status settling, plus ~15 new message-pattern
defs mined from two players' raw logs. The core state-machine changes are intricate but internally
consistent — traced by hand against the fixtures and verified passing. The main gap is that the
large batch of new regex patterns carries almost no automated regression coverage.

## Scope note (read this first)

The PR is explicitly **stacked on #1593** (still open, `combat-hunt-audit` branch) and its own
description says "review only the top commit, `056ff899`." `gh pr diff`/the GitHub compare view
shows 23 files and ~2,150 additions because it also contains #1593's unmerged diff. This review
isolated the actual top commit with `git diff 9a75ec64..056ff899` (11 files, 404/-47) and reviewed
only that. If #1593 changes before merge, re-diff before trusting any line numbers here.

## Major

### 1. ~15 new regex-based defs ship with no automated regression coverage

This commit adds or materially changes patterns for: `:mug` (2 lines), `:barbed_sweep`, a second
`:shield_charge` form, `:poison_tick` (7 phrasings), 6 new inbound forms (treekin root/sap/fist,
Illoke jarl ward attack, krolvin slaver subdue) in [attacks.rb](lib/gemstone/combat/defs/attacks.rb),
`:demoralized` plus terrified's creature-side expiry in
[statuses.rb](lib/gemstone/combat/defs/statuses.rb), and the shield-block `attack` outcome variant
in [outcomes.rb](lib/gemstone/combat/defs/outcomes.rb:255). Of all of these, only two get any
test: the `eviscerate` fixture (asserting `demoralized` now fires) and the rewritten `weapon_cast`
fixture. The other ~13 patterns are justified entirely by inline comments citing corpus counts from
"Rysk/Dreadt logs 2026-09-11" — a private log-audit tool that isn't part of this repo or its CI.

This matters specifically here because [attack_defs_spec.rb](spec/lib/gemstone/combat/attack_defs_spec.rb:7)
states its own purpose in its header comment: *"Attack def coverage pinned against real game
messaging... When log replay surfaces an unmatched variant of one of our own attacks, the fix is a
def change - these examples keep known variants from regressing."* That file is untouched by this
commit. A typo in any of these seven `:poison_tick` alternations, or in the krolvin subdue pattern's
escaping, would compile fine, pass rubocop, and pass the full suite (296/296, confirmed by running
it) — and only surface the next time it fails to match a live line, which is exactly the failure
mode the PR is trying to eliminate.

To be fair: the immediately preceding stacked PR (#1593) followed the same practice — it added ~90
lines of new patterns to `attacks.rb` against only a 4-line `attack_defs_spec.rb` diff, so this
isn't a new regression in house style. And the consequence of a wrong pattern here is silent
under-matching (falls back to the pre-existing `:unknown`/orphan bucket), not corruption of an
already-working path. Still, given regex correctness against real game text *is* the entire risk
surface of this file, and the project has a purpose-built harness for exactly this
(`attack_defs_spec.rb` for atomic examples, `replay_spec.rb` for blob-level fixtures), I'd ask for
at least the higher-risk additions (the 7 poison_tick phrasings, the krolvin subdue line, the two
mug lines) to get one-line examples before merge.

## Minor

### 2. `recorder.rb`'s `txt()` now silently turns `""` into `NULL` for every text column it touches

[recorder.rb:680-688](lib/gemstone/combat/recorder.rb:680):
```ruby
def txt(value)
  return nil if value.nil?

  s = value.to_s
  s = s.dup.force_encoding('UTF-8').scrub('?') if s.encoding == Encoding::ASCII_8BIT
  s = s.strip.squeeze(' ')
  s.empty? ? nil : s
end
```
The stated purpose (per the comment above it) is normalizing the intermittent double-space glitch
in a revealed creature's noun/name (`" rogue"` / `"human  rogue"`). But `txt()` is the funnel for
every text field in the two INSERT statements — attacker name, weapon, status name, spell name,
flare weapon — not just creature names. Before this change, an empty string passed through as `""`;
now it becomes `NULL`. If any downstream query or report distinguishes `weapon = ''` from
`weapon IS NULL` (a swing with no weapon captured vs. one explicitly recorded empty), this changes
observable behavior beyond the stated fix. I didn't find such a query in this repo, but it's outside
`lib/gemstone/combat/` and wasn't part of what I could verify — worth a quick check before merge
since it's a one-line easy fix (guard the strip/squeeze/nil-check behind the same ASCII_8BIT-repair
condition, or explicitly note `""`→`nil` is intended) if it turns out to matter.

## Open questions

- [processor.rb:1518-1541](lib/gemstone/combat/processor.rb:1518): the "held maneuver roll belongs to
  the creature that opened it, not our next swing" exception now fires whenever an inbound event
  opens while we're not mid-sequence/nocked/in a `ROLL_BEFORE_ROUND_ASSAULTS` assault — dropping the
  old requirement that the roll's outcome print on the *same line* as the inbound event. That's a
  deliberate generalization (the commit message cites a corpus census: "5 maneuver-SMR → inbound-line
  shapes, 0 with our own attack following in the chunk"), and it's correctly scoped to
  maneuver-class rolls only (AS/DS is left in `pending_resolutions` regardless, via the `partition`
  at line 1539). What the census doesn't explicitly rule out is a *different* creature's maneuver
  roll sitting in `pending_resolutions` when a second, unrelated creature's inbound line opens in the
  same chunk (group combat, two simultaneous attackers). **Settles it:** does the corpus contain any
  multi-attacker chunks, and if so, did a roll from creature A ever precede an inbound line from
  creature B with nothing else between them? If that shape doesn't occur in practice (attacks appear
  to arrive serialized per-attacker), this is a non-issue; if it can occur, the roll would misattach
  to the wrong creature's row.

## What this PR gets right

- The Holy Weapon proc→spell→swing bookkeeping (`interrupted_own`, `released_parent`,
  `pending_release_cast`, `deferred_casts`, `_released_children` in
  [processor.rb](lib/gemstone/combat/processor.rb)) handles both message orderings (swing-then-proc
  and proc-then-swing) by mutating hash objects already pushed into `events` rather than requiring a
  second pass — subtle, but I traced it by hand against the actual `weapon_cast.txt` fixture line by
  line and it holds up, including the "already in `events`, mutate in place" trick and the
  `save_event` re-entrancy guard.
- `AMBUSH_KINDS`/`AMBUSH_PREFIXES` index coupling ([attacks.rb:642-667](lib/gemstone/combat/defs/attacks.rb:642))
  is exactly the kind of parallel-array-by-index pattern that silently drifts out of sync — this one
  has a load-time `raise` asserting the lengths match, so a future edit that adds a prefix without a
  kind fails loudly instead of silently misattributing.
- Reusing `PatternGate.build`/`rejects?` for the new `REACTION_GATE` is safe: confirmed the helper is
  stateless (no shared/global state across `build` calls) and already used this way for
  `AMBUSH_GATE`/`REDIRECT_GATE`/`ATTACK_GATE` in the same file.
- The recorder schema change is complete and self-consistent: `attack_kind TEXT` column, the
  `ADDED_COLUMNS` migration entry (so pre-existing DBs get the column added), and the INSERT's
  25-column list / 25-placeholder `VALUES` / 25-element `params` array all line up — checked by hand
  since a miscount here would be a silent column-shift bug.
- `spec/lib/gemstone/combat/replay_spec.rb`'s "header is a floor, not an exact set" design correctly
  caught the actual bug this PR fixes (traced the `eviscerate.txt` fixture: the old
  `!born && current_event[:outcomes].any?` settle check would have let the second bystander's SSR
  wrongly attach to the first bystander's already-settled event; the new `_had_status` check
  correctly rejects it because the flag gets set at end-of-line regardless of whether the status
  emit itself could resolve an `attack_uid`).

## Coverage notes

- **Access level:** full local clone with history (`elanthia-online/lich-5`, existing remote already
  checked out at `/home/ryan/lich-5`); isolated the top commit's diff via `git diff` against its
  parent rather than trusting `gh pr diff`, which included the stacked, unmerged #1593 changes.
- **Tests:** ran `bundle exec rspec spec/lib/gemstone/combat` in a separate git worktree at the PR's
  head commit (296 examples, 0 failures, matching the PR's own claim).
- **Lint:** ran `bundle exec rubocop` on the changed `lib/`+`spec/` combat files — clean, no offenses.
- **CodeRabbit:** not run — the CLI wasn't invoked for this review; `checks.txt` shows CodeRabbit's
  own CI check was skipped ("automatic reviews are disabled") for this PR.
- **Not independently verified:** the underlying game-message evidence (the actual Rysk/Dreadt raw
  logs cited in comments) — I could only check that the regexes are syntactically well-formed and
  consistent with the surrounding idiom, not that they match real, unmangled game output. This is
  the same limitation the Major finding above is about.
- **Not reviewed:** the stacked #1593 diff itself (out of scope per the author's own instruction),
  and any file outside `lib/gemstone/combat/**` + its specs (the PR's stated scope).

PR #1603, opened 2026-09-12
fix(all): creatures registry housekeeping independent of Combat::Tracker

### LINKED ISSUE #1602: Creature registry: full-table cleanup cascade can stall and silently drop new registrations under sustained load
## Summary

`Lich::Gemstone::CreatureInstance`'s registry (`lib/common/creature/creature_base.rb`) is a single process-lifetime hash capped at `max_size` (default 1000). Once it fills, every subsequent registration attempt runs an escalating cascade of up to **8 full linear scans of the entire registry**, synchronously, on the main processing thread — and if none of them free enough room, the new creature silently gets **no `CreatureInstance` at all**. Growth into the registry (any targetable hostile in any room) is gated far more loosely than the mechanism that prunes it (which only advances during actively-parsed combat text), so a long-running, well-populated session can plausibly stay pinned at capacity, repeatedly paying the scan cost and silently losing HP/status tracking for newly-encountered creatures.

Found while investigating a "bigshot feels sluggish" report ([elanthia-online/scripts](https://github.com/elanthia-online/scripts) `bigshot.lic`) — a debug log from a second reporter showed `Lich::Gemstone::Creature.stats: {instances: 871, ..., max_size: 1000, ...}` at the very start of a fresh script run, in a session with heavy creature turnover.

## Where this lives

- `lib/common/creature/creature_base.rb`
  - `instances` — the registry itself, a class-instance-variable hash (`@instances ||= {}`, line 405). Not tied to any script or room; lives for the lifetime of the Lich process.
  - `register` (lines 191–221) — the registration + escalating-cleanup path.
  - `full?` (line 323) — `size >= max_size`.
  - `cleanup_old` (line 284) — one linear `select`/`reject!` pass over `instances`, filtered by age.
- `lib/common/xmlparser.rb` (lines 1080–1085) — where `Creature.register` actually gets called from: any bolded room object whose id is in `XMLData.current_target_ids` or has a pending `<crtrStatus>` tag. This fires for **any targetable hostile creature in any room the character is in** — not gated on combat, damage, or anything bigshot-specific.
- `lib/gemstone/combat/tracker.rb`
  - `process` (line 218) increments `@chunks_processed` (line 232) only when the chunk is combat-relevant (`combat_relevant?`, line 264 — requires damage/attack-roll/swing-type text).
  - `cleanup_creatures` (line 319) — calls `Creature.cleanup_old(cleanup_max_age)` (default 600s), but only runs once `@chunks_processed >= cleanup_interval` (default 100) **combat-relevant** chunks have been seen.

## Root cause

Two mechanisms with mismatched triggers share one bounded table:

1. **Growth is broad and unconditional.** Any hostile creature that becomes targetable in any room gets registered — while traveling, passing through a populated area, standing near other players' fights, etc. This has nothing to do with whether *this* character is fighting.
2. **Pruning is narrow and combat-gated.** The periodic cleanup that would normally keep the registry bounded only makes progress while `Combat::Tracker` is actively parsing combat-relevant text. During travel, town time, or any non-combat stretch, `@chunks_processed` simply doesn't move, no matter how long that stretch is or how many new creatures got registered during it.

So a long Lich session with a lot of non-combat travel through populated zones can accumulate registrations faster than the combat-gated cleanup ever prunes them, walking the registry up toward `max_size` over time. `Lich::Gemstone::CreatureInstance` is a class-level singleton (line 405) — this state persists across every script start/stop for the life of the Lich process, so it isn't reset by restarting `bigshot` or any other script; only `Creature.clear` or a full Lich restart resets it.

Once `full?` is true, `register` (lines 207–215) does this on **every subsequent registration attempt**:

```ruby
if full?
  # Progressively more aggressive: 120 minutes, then 15-minute steps.
  [7200, 6300, 5400, 4500, 3600, 2700, 1800, 900].each do |age_threshold|
    removed = cleanup_old(age_threshold)
    break unless full?
  end
  return nil if full? # Still full after every cleanup attempt.
end
```

- Each `cleanup_old` call is a full `select` + `reject!` over the whole hash — up to 8 full-table passes, back to back, before the loop even gives up.
- This runs inline in the registration path, which is driven from XML parsing on Lich's main thread — not on an async worker like `Combat::Tracker`'s own damage parsing.
- If the registry's occupants are mostly recent (very plausible in a continuously-repopulating, high-traffic hunting ground — nothing in it is old enough to be evicted at any of the 8 thresholds), **all 8 passes fail**, and `register` returns `nil` for the new creature. No error, no warning — the creature is just never backed by a `CreatureInstance`.

## Impact

- **Performance:** up to 8 redundant full-table scans, repeated on every registration attempt for as long as the registry stays full. Individually each pass over ≤1000 entries is not expensive in isolation, but the repetition (once per new creature, for potentially many creatures in a row) is pure waste that a single bounded pass or a proper eviction/LRU structure would avoid.
- **Functional (the more concerning half):** when `register` returns `nil`, that creature has no `CreatureInstance`, so anything built on top of it silently degrades to its "not creature-backed" fallback. In `bigshot.lic` specifically, this is exactly the state that turns off `thp<N>`, `empowered<N>`, the built-in coup-de-grace eligibility gate, and every native `crtrStatus` check for that creature — with nothing in the debug output to say why. This can plausibly present to a player as automation quietly getting worse/less consistent over a long session, independent of any actual CPU slowdown — which matches a "feels sluggish/off" report better than a pure performance explanation does.

## Evidence

From a `bigshot.lic` debug-to-file log, at the very first line logged by a freshly-started script instance (i.e., before this character had done anything this session):

```
startup | Lich::Gemstone::Creature.stats: {instances: 871, templates: 627, max_size: 1000, auto_register: true}
```

871/1000 already at cold start, on a character reported to be experiencing "sluggish" behavior, in a hunting ground with heavy and varied creature turnover (multiple distinct creature types cycling through every few minutes). A comparison session from a different character during the same investigation, in a lighter-traffic area, showed `instances: 42-45` — nowhere near capacity.

No direct evidence (yet) that the registry actually *hit* 1000 mid-session — `register`'s cleanup-cascade path only echoes when `$creature_debug` is enabled, which it wasn't in the captured log. Flagging this issue partly to make that easier to confirm going forward (see suggested fix, logging item below).

## Suggested fix

A few options, roughly in order of effort:

1. **Cheap correctness fix:** cap the cascade to a single `cleanup_old(cleanup_max_age)` (or a single configurable threshold) instead of iterating 8 progressively-relaxed thresholds — avoids the repeated full-table scans without changing behavior much, since the 8-step cascade rarely helps if the first bound already failed to free space in a still-active area.
2. **Decouple pruning from combat activity:** let `Creature`'s own cleanup run on a wall-clock timer (or on room-change / periodic script pulse), independent of `Combat::Tracker.process` and its combat-relevant-chunk gate, so a long non-combat stretch doesn't let the registry grow unchecked.
3. **Make the "registry full" case observable without turning on `$creature_debug`:** e.g. a rate-limited `Lich.log` warning the first time `register` returns `nil` due to `full?`, so a script author (or a user's debug log) can actually see when tracking silently degraded for a creature, rather than only being inferable from downstream symptoms.
4. **Longer term:** replace the linear-scan-on-demand cleanup with an actual bounded LRU (evict-oldest-on-insert-when-full), which makes the registration path O(1) instead of O(n) in the worst case and removes the "still full after every cleanup attempt" failure mode entirely.

## How to verify

- `Lich::Gemstone::Combat::Tracker.stats` / `Lich::Gemstone::Creature.stats` mid-session on a character that plays for extended, high-traffic sessions, watching whether `instances` reaches `max_size`.
- Temporarily enable `Lich::Gemstone::Creature.debug_on(:changes)` to confirm whether `register` is actually hitting the `full?` branch and how often the 8-step cascade runs to exhaustion vs. succeeding early.

---

Nisugi:
Closes #1602.

## Problem

The creature registry (`CreatureInstance`, via `Lich::Common::CreatureBase`) was only ever pruned by `Combat::Tracker`'s periodic `cleanup_creatures`, which is **off by default** and only advanced on combat-relevant text. With the tracker off, nothing emptied the registry, so it filled to `max_size` and then paid an 8-threshold eviction cascade on every new registration, silently returning `nil` for the newcomer when nothing was old enough. `Creature.targets` builds from the room roster via `self[id]`, so a dropped creature was not just untracked but absent from the target list entirely.

A second, pre-existing problem: `cleanup_old` aged on `created_at`, so a creature you had been fighting for ten minutes was discarded mid-fight and re-registered as a blank instance with damage and status reset.

## Fix

- **Housekeeping lives on the registry.** `register` runs `cleanup_old` on a 60s wall-clock throttle, independent of the tracker. The registry stays bounded whether or not tracking is enabled.
- **Age is last-seen, not created.** New per-instance `last_seen_at`, touched on re-registration (room refresh) and on every `<crtrStatus>` sync. Long fights are no longer swept.
- **In-room creatures are never evicted.** The roster is proof of presence.
- **Cascade removed.** If somehow still full, evict the single least-recently-seen out-of-room creature. The only remaining `nil` case (a room holding `max_size` creatures) logs once via `Lich.log` instead of failing silently.
- **`configure(cleanup_max_age:)`** (default 600), reported in `Creature.stats`. Applies to both GS and DR registries.
- **`Combat::Tracker`'s `cleanup_creatures`, `cleanup_interval` and `cleanup_max_age` settings are removed** as redundant. Stale persisted keys in saved tracker settings are ignored by the merge.

## Tests

Creature, combat and gemstone specs pass (783 examples). Seven specs that aged creatures by `created_at` were rewritten; six new specs cover the roster guard, last-seen touching, throttled housekeeping, stalest-first eviction and `cleanup_max_age` configuration.

### DIFF
--- a/lib/common/creature/creature_base.rb
+++ b/lib/common/creature/creature_base.rb
@@ -252,6 +252,11 @@ module Lich
         # @return [void]
         def clear_room
           count = room_roster.size
+          # Keep the outgoing roster as a shield: the parser rebuilds the new
+          # one a creature at a time, so a creature late in the refresh is
+          # briefly absent from the live roster. Housekeeping must not sweep
+          # it in that window (see {#sheltered_ids}).
+          @previous_room_ids = room_roster
           @current_room_ids = []
           respond "--- room: roster cleared (#{count} creature#{'s' unless count == 1})" if $creature_debug && count > 0
         end
@@ -287,22 +292,34 @@ module Lich
         def clear
           instances.clear
           clear_room
+          @previous_room_ids = []
         end
 
         # Removes instances not seen within the given age.
         #
         # Ages against `last_seen_at`, not creation time, so a creature that
         # has been in a long fight is not discarded mid-fight and re-registered
-        # as a blank stranger. Creatures in the current-room roster are never
-        # removed: the roster is proof they are still present.
+        # as a blank stranger. Creatures in the current or immediately previous
+        # room roster are never removed ({#sheltered_ids}).
         #
         # @param max_age_seconds [Integer] age cutoff in seconds.
         # @return [Integer] number of instances removed.
         def cleanup_old(max_age_seconds = cleanup_max_age)
           cutoff = Time.now - max_age_seconds
+          shelter = sheltered_ids
           before = instances.size
-          instances.reject! { |id, instance| instance.last_seen_at < cutoff && !room_roster.include?(id) }
+          instances.reject! { |id, instance| instance.last_seen_at < cutoff && !shelter.include?(id) }
           before - instances.size
+        end
+
+        # Ids housekeeping must never evict: the live roster plus the roster
+        # that {#clear_room} just replaced. The previous roster covers the
+        # window in which the parser is still re-marking creatures from a
+        # fresh room refresh; it is superseded on the next {#clear_room}.
+        #
+        # @return [Array<Integer>]
+        def sheltered_ids
+          room_roster | (@previous_room_ids || [])
         end
 
         # Seconds between wall-clock housekeeping passes from {#register}.
@@ -331,8 +348,9 @@ module Lich
         # @return [Object, nil] the evicted instance, or nil when every
         #   instance is in the current room.
         def evict_stalest
+          shelter = sheltered_ids
           id, instance = instances
-                         .reject { |candidate_id, _| room_roster.include?(candidate_id) }
+                         .reject { |candidate_id, _| shelter.include?(candidate_id) }
                          .min_by { |_, candidate| candidate.last_seen_at }
           return nil unless id
 
--- a/lib/common/creature/creature_base.rb
+++ b/lib/common/creature/creature_base.rb
@@ -203,14 +203,16 @@ module Lich
           entered_room = mark_in_room(id)
           return nil unless auto_register?
 
+          # Before the known-creature return: a session that only refreshes
+          # creatures it already knows must still sweep stale out-of-room ones.
+          housekeep
+
           existing = instances[id.to_i]
           if existing
             existing.touch_seen
             respond "--- #{name} (#{id}): in room" if entered_room && $creature_debug
             return existing
           end
-
-          housekeep
 
           if full?
             # Housekeeping should keep this from ever happening; if it does,
--- a/lib/common/creature/creature_base.rb
+++ b/lib/common/creature/creature_base.rb
@@ -191,8 +191,8 @@ module Lich
         # Combat::Tracker (off by default) ever calls {#cleanup_old}.
         #
         # @return [Object, nil] the registered instance, or nil when
-        #   auto-registration is disabled or the registry is full of in-room
-        #   creatures.
+        #   auto-registration is disabled or the registry is full of creatures
+        #   in the current room.
         def register(name, id, noun = nil)
           # Record room presence first: the feed event that triggers this call
           # (a bolded room-object name or a <crtrStatus> tag) is itself proof the
@@ -345,12 +345,21 @@ module Lich
 
         # Evicts the least-recently-seen instance that is not in the room.
         #
+        # The previous room's roster is a soft shelter here: it exists only
+        # to cover the mid-refresh window (see {#sheltered_ids}), so when the
+        # registry is full and nothing else is evictable, a previous-room
+        # creature goes before the newcomer is refused. Without that fallback
+        # a registry filled at the moment of a room change would refuse
+        # every creature in the new room until the next refresh. Only the
+        # live roster is untouchable.
+        #
         # @return [Object, nil] the evicted instance, or nil when every
         #   instance is in the current room.
         def evict_stalest
-          shelter = sheltered_ids
-          id, instance = instances
-                         .reject { |candidate_id, _| shelter.include?(candidate_id) }
+          previous = @previous_room_ids || []
+          candidates = instances.reject { |id, _| room_roster.include?(id) }
+          preferred = candidates.reject { |id, _| previous.include?(id) }
+          id, instance = (preferred.empty? ? candidates : preferred)
                          .min_by { |_, candidate| candidate.last_seen_at }
           return nil unless id
 
@@ -359,14 +368,20 @@ module Lich
           instance
         end
 
-        # Logs once per process that the registry is full of in-room creatures.
+        # Logs once per process that the registry is full of creatures in the
+        # current room.
         #
         # @return [void]
         def warn_full_once
           return if @warned_full
 
           @warned_full = true
-          Lich.log "#{name}: creature registry full (#{max_size}) with every entry in the current room; new creatures are not being tracked" if defined?(Lich) && Lich.respond_to?(:log)
+          Lich.log full_warning if defined?(Lich) && Lich.respond_to?(:log)
+        end
+
+        # @return [String] the {#warn_full_once} message.
+        def full_warning
+          "#{name}: creature registry full (#{max_size}) with every entry in the current room; new creatures are not being tracked"
         end
 
         # Configures registry limits.
--- a/lib/common/creature/creature_base.rb
+++ b/lib/common/creature/creature_base.rb
@@ -49,8 +49,8 @@ module Lich
     # includes {InstanceMethods} (status/flag tracking). The class must define a
     # `new(id, noun, name)` constructor, call {InstanceMethods#initialize_status_tracking}
     # from its own `initialize`, and expose a `valid_target?` predicate (used by
-    # {ClassMethods#targets}) plus a `created_at` reader (used by
-    # {ClassMethods#cleanup_old}).
+    # {ClassMethods#targets}). {InstanceMethods#initialize_status_tracking}
+    # stamps `last_seen_at`, which {ClassMethods#cleanup_old} ages against.
     module CreatureBase
       # Wires the shared behaviour into the including class, matching the
       # MapBase convention.
@@ -186,8 +186,13 @@ module Lich
         #   supplies an id before a name (DragonRealms `<crtrStatus>`).
         # @param id [Integer, String] server creature id.
         # @param noun [String, nil] noun from room XML, when available.
+        # Housekeeping also runs from here on a wall-clock throttle
+        # ({#housekeep}), so the registry stays bounded whether or not
+        # Combat::Tracker (off by default) ever calls {#cleanup_old}.
+        #
         # @return [Object, nil] the registered instance, or nil when
-        #   auto-registration is disabled or the registry is full.
+        #   auto-registration is disabled or the registry is full of in-room
+        #   creatures.
         def register(name, id, noun = nil)
           # Record room presence first: the feed event that triggers this call
           # (a bolded room-object name or a <crtrStatus> tag) is itself proof the
@@ -200,18 +205,23 @@ module Lich
 
           existing = instances[id.to_i]
           if existing
+            existing.touch_seen
             respond "--- #{name} (#{id}): in room" if entered_room && $creature_debug
             return existing
           end
 
+          housekeep
+
           if full?
-            # Progressively more aggressive: 120 minutes, then 15-minute steps.
-            [7200, 6300, 5400, 4500, 3600, 2700, 1800, 900].each do |age_threshold|
-              removed = cleanup_old(age_threshold)
-              respond "--- Auto-cleanup: removed #{removed} old creatures (threshold: #{age_threshold}s)" if removed > 0 && $creature_debug
-              break unless full?
+            # Housekeeping should keep this from ever happening; if it does,
+            # drop the single least-recently-seen creature that is not in the
+            # room. Only a room holding max_size creatures leaves nothing to
+            # evict.
+            evict_stalest
+            if full?
+              warn_full_once
+              return nil
             end
-            return nil if full? # Still full after every cleanup attempt.
           end
 
           instance = new(id, noun, name)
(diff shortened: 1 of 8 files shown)


mrhoribu (changes requested):
# PR Review: fix(all): creatures registry housekeeping independent of Combat::Tracker (elanthia-online/lich-5#1603)

**Reviewed:** 8 files, +301/-77 at `8b34fd8d` on 2026-09-12. Based on: full local clone (fetched `refs/pull/1603/head`, checked out in a worktree), including project history.
**Verdict:** Request changes
**Summary:** This PR replaces an 8-threshold, creation-time eviction cascade (the subject of #1602) with wall-clock-throttled housekeeping keyed on `last_seen_at`, and adds a roster-shelter mechanism so in-room and just-vacated-room creatures are never evicted. The core design is sound and well tested, but the "shelter the previous room too" fix in the last commit introduces a new, confirmed failure mode: when the registry is at capacity at the exact moment of a room transition, every creature in the *new* room can be registered as `nil` — silently reproducing the exact symptom (#1602's "creature has no CreatureInstance, no error, no warning") this PR exists to fix — and the condition does not clear on its own; it persists until a further room change.

## Major

### 1. A full registry at the moment of a room transition can permanently blackhole the new room's creatures — `lib/common/creature/creature_base.rb:196` and `:321`

```ruby
def sheltered_ids
  room_roster | (@previous_room_ids || [])
end
```

`register` (creature_base.rb:196-233) only creates a new instance after `evict_stalest` (creature_base.rb:350) frees a slot when `full?`. `evict_stalest` and `cleanup_old` (creature_base.rb:307) both refuse to touch anything in `sheltered_ids` — the current room roster **or** the roster `clear_room` just superseded — with no time bound on that protection. If, at the instant a player leaves a room, the registry is at `max_size` and every currently-registered instance happens to be in `current_room_ids ∪ previous_room_ids`, `evict_stalest` finds nothing eligible, `full?` stays true, and `register` returns `nil` for every creature the parser tries to add from the new room — for as long as the player stays there. It only recovers on the *next* `clear_room` call (a further room change), because time-based expiry never runs against sheltered ids regardless of how long they've been unseen.

I reproduced this directly against the shared mixin:

```
$ ruby repro.rb
After room A fill: size=3 full?=true
previous_room_ids after clear_room: [1, 2, 3]
Registering brand-new creature 'd' in new room B => nil
Registering another brand-new creature 'e' => nil
After a second clear_room (leaving room B), registering 'f' => #<SampleCreature ... id=6 ...>
```
(`SampleCreature.configure(max_size: 3)`, fill room A to capacity, `clear_room` into room B, then register two brand-new ids — both come back `nil` until a second `clear_room` finally supersedes room A's shelter.)

This is reachable in real play. #1602's own evidence shows the registry reaching 871/1000 from ordinary heavy-turnover hunting — and while this PR's housekeeping generally keeps things well under `max_size` in steady state, `max_size` is an explicitly documented, user-facing knob (`Creature.configure(max_size:)`); anyone who lowers it (to bound memory, or for a script that wants a smaller working set) puts themselves within reach of exactly the scenario above at ordinary creature densities. Once triggered, there is no time-based recovery — a stationary fight in the new room can go the rest of the session with new arrivals silently untracked.

Two secondary problems ride along with this: the `register` docstring (creature_base.rb:194-195) says registration fails "when the registry is full of in-room creatures," but the entries blocking recovery here are in the *previous* room, not the current one; and `warn_full_once`'s message (creature_base.rb:369) — `"...full (#{max_size}) with every entry in the current room"` — is factually wrong in this exact case, which will mislead whoever reads it while trying to diagnose the very state this bug produces.

A fix likely needs `evict_stalest` to treat `previous_room_ids` as a softer, evict-as-last-resort protection rather than as untouchable as the live roster — the previous roster only needs to survive the brief mid-refresh window the last commit was written to cover, not indefinitely once a real capacity conflict shows up. Confirmed via reproduction against the shared `Lich::Common::CreatureBase` mixin used by both games.

## Minor

### 2. Persisted `Combat::Tracker` cleanup settings are neither migrated nor actually dropped — `lib/gemstone/combat/tracker.rb:311`

```ruby
def load_settings
  scope = "#{XMLData.game}:#{XMLData.name}"
  stored_settings = Lich::Common::DB_Store.read(scope, 'lich_combat_tracker')
  @settings = DEFAULT_SETTINGS.merge(stored_settings)
end
```

The PR description states "Stale persisted keys in saved tracker settings are ignored by the merge." That isn't what `Hash#merge` does: `DEFAULT_SETTINGS.merge(stored_settings)` is a union of both hashes' keys, not `stored_settings` filtered down to `DEFAULT_SETTINGS`'s keys — I confirmed this directly (`{a: 1, b: 2}.merge({a: 10, cleanup_max_age: 999}) => {a: 10, b: 2, cleanup_max_age: 999}`). A character who previously ran `Combat::Tracker.configure(cleanup_max_age: 3600)` (a legitimate, documented call before this PR) has that value sitting in `lich_combat_tracker` in DB_Store; after upgrading, `@settings[:cleanup_max_age]` still loads as `3600`, gets echoed in `Tracker.stats`/the debug "Settings updated" line, and gets re-saved on every `configure` call — forever — while doing nothing, since nothing reads it anymore (the actual cutoff now lives on `CreatureInstance.cleanup_max_age`, untouched by this legacy value). The net effect isn't a crash, but it is a silent settings-migration gap: anyone who had customized the tracker's cleanup age loses that customization on upgrade with no indication why, and the persisted key becomes permanent, confusing cruft. Worth either migrating the stored value into `Creature.configure(cleanup_max_age:)` once on load, or explicitly stripping the two retired keys from `stored_settings` before merging.

## Nits

- No test exercises `warn_full_once`'s "only once" guard or its log message content — the existing "refuses a newcomer only when full of in-room creatures" spec exercises the code path but asserts only on `register`'s return value, not on what gets logged. Given finding #1, a message-content assertion would have caught the current-room-vs-previous-room inaccuracy.

## Open questions

- `lib/common/creature/creature_base.rb:321` — is a room ever expected to hit `max_size` in real play at the *default* configuration (1000), or is `max_size` low enough in practice (per #1602's own 871/1000 data point, gathered *before* this PR's housekeeping existed) that finding #1 is realistically only reachable via a deliberately lowered `max_size`? If the latter, this may be acceptable as a documented limitation rather than a fix-before-merge blocker — but the docstring and log message should still be corrected to describe the real condition (current-or-previous room) rather than "current room" only.

## What this PR gets right

- Moving housekeeping onto a wall-clock throttle inside `register` (creature_base.rb:336) correctly decouples registry pruning from `Combat::Tracker`'s combat-gated, off-by-default sweep — this directly fixes the root cause identified in #1602.
- Aging on `last_seen_at` instead of `created_at`, touched on both re-registration and `<crtrStatus>` sync (creature_base.rb:528-535, :632), correctly stops a long fight's creature from being discarded and re-created mid-fight.
- The three follow-up commits (moving `housekeep` ahead of the known-creature return, fixing the DR facade's hardcoded 600s default, and sheltering the previous room roster) each fix a real, narrow bug uncovered by the PR's own test suite as it was built out — that iterative tightening, and the specs that pin each one down, are good work.
- Test coverage for the throttle, the last-seen touch, the shelter mechanism's happy path, and the eviction-of-fresh-but-out-of-room-creature behavior is thorough and each scenario is traced by hand correctly (I re-derived several by hand against the code and they hold).
- `rubocop` is clean and the full spec suite (7156 examples) passes locally against this branch.

## Coverage notes

- Full local clone via `git fetch origin pull/1603/head` into a worktree; diff, all changed files at head, and project history (including `git show` of intermediate commits) were all available.
- Ran `bundle exec rspec` (full suite: 7156 examples, 0 failures) and `rubocop` (8 changed files, no offenses) against the PR branch.
- Ran a CodeRabbit CLI pass (`coderabbit review --agent --base main --committed`); it surfaced 2 findings, both of which I independently verified against the actual code and history rather than transcribing — they correspond to Major #1 (confirmed by direct reproduction, not just inspection) and Minor #2 above.
- Grepped both `dr-scripts` and `scripts` (the sibling script repos) for any reference to the removed `Combat::Tracker` settings (`cleanup_interval`, `cleanup_max_age`) or the changed `Creature.cleanup_old`/`configure` signatures — none found, so no known external blast radius beyond lich-5 itself.
- Did not attempt to reproduce anything against a live game connection (XML parser integration, actual room-transition timing); the reproduction for Major #1 is against the shared `Lich::Common::CreatureBase` mixin directly, which is the same code both `Lich::Gemstone::CreatureInstance` and `Lich::DragonRealms::CreatureInstance` extend.

mrhoribu:
Several places that the 1000 creature limit could theoretically get hit. REIM, invasions, etc.

mrhoribu (approved):
# PR Review (follow-up): fix(all): creatures registry housekeeping independent of Combat::Tracker (elanthia-online/lich-5#1603)

**Reviewed:** new commit `5f6d3563` on top of the previously-reviewed `8b34fd8d`, at `5f6d3563` on 2026-09-12. Based on: full local clone (worktree), diffed against the previously reviewed head.
**Summary:** Both blockers from the prior review are fixed and independently verified — the room-transition capacity trap is closed, and the retired tracker settings are dropped instead of lingering. One small, non-blocking loose end remains: a persisted custom `cleanup_max_age` value from before this PR is now dropped rather than migrated, and one doc comment still references a method that doesn't exist.

## Previously reported issues — verified fixed

### Major #1 (room-transition blackhole) — fixed, reproduced

`evict_stalest` (`lib/common/creature/creature_base.rb:358`) now treats the previous room's roster as a *soft* shelter: it prefers evicting a fully-unsheltered creature, but falls back to the stalest previous-room creature when nothing else is available, and only refuses when every instance is genuinely in the live room roster.

```ruby
def evict_stalest
  previous = @previous_room_ids || []
  candidates = instances.reject { |id, _| room_roster.include?(id) }
  preferred = candidates.reject { |id, _| previous.include?(id) }
  id, instance = (preferred.empty? ? candidates : preferred)
                 .min_by { |_, candidate| candidate.last_seen_at }
  ...
```

I re-ran my original reproduction against this commit: filled a `max_size: 3` registry in "room A," transitioned to "room B," and registered three brand-new creatures back to back. All three now register immediately (previously the second and third came back `nil` until a further room change):

```
After room A fill: size=3 full?=true
Registering brand-new creature 'd' in new room B => #<SampleCreature ... id=4 ...>
Registering 'e' => #<SampleCreature ... id=5 ...>
Registering 'f' => #<SampleCreature ... id=6 ...>
```

The docstrings and `warn_full_once`'s log message were also corrected to say "full of creatures in the current room" — accurate now, since that's the only remaining unresolvable case. Three new specs cover the fallback, the warn-once/message-content behavior, and the still-prefer-fully-unsheltered ordering; I traced all three by hand and they hold.

### Minor #2 (settings-migration gap) — addressed differently than expected, and sufficient

Rather than migrating a persisted value, `Combat::Tracker.load_settings` now strips the retired keys outright:

```ruby
RETIRED_SETTINGS = %i[cleanup_interval cleanup_max_age].freeze
...
@settings = DEFAULT_SETTINGS.merge(stored_settings.reject { |key, _| RETIRED_SETTINGS.include?(key.to_sym) })
```

I confirmed this strips both symbol- and string-keyed legacy records and no longer resurfaces them in `stats` or re-saves them. This resolves what I flagged (stale keys silently persisting and being echoed forever) — see the one open point below for the part it doesn't cover.

## Nits

### `lib/gemstone/combat/observers.rb:17-18` — comment cites a method that doesn't exist

```ruby
#   3. Persistence: registry entries are session-only and swept by the
#      registry's own housekeeping (Creature.cleanup_max_age) - recording/
#      logging scripts must capture events at parse time.
```

`Lich::Gemstone::Creature` has no `self.cleanup_max_age` — only `CreatureInstance.cleanup_max_age` (the shared mixin's class method) exists directly; the facade only surfaces it via `Creature.stats[:cleanup_max_age]` or `Creature.configure(cleanup_max_age:)`. Calling `Creature.cleanup_max_age` as written would raise `NoMethodError`. Comment-only, no functional impact, but worth a one-line fix so a future reader doesn't copy it into real code. Pre-existing since the first commit of this PR; I missed it in the first pass.

## Open questions

- `lib/gemstone/combat/tracker.rb:311` — a character who had previously run `Combat::Tracker.configure(cleanup_max_age: 3600)` (or any non-default value) now silently loses that customization on upgrade: the value is dropped, not carried into `Creature.configure(cleanup_max_age:)`, and there's no log line noting the drop. I don't think this rises to blocking — the mechanism that setting used to feed (`Tracker#cleanup_creatures`) is removed outright, not renamed, so there isn't a clean 1:1 "migration" target, and the new default (600s) matches the old default, so anyone who never touched this setting sees no change. **Settles it:** does the team want a one-time best-effort migration (read the retired key once, forward it to `CreatureInstance.configure(cleanup_max_age:)`, then drop it), or is silently reverting to 600s for the handful of characters that customized it acceptable? Either answer is defensible; I'd leave it to the author/maintainers rather than block on it.

## Coverage notes

- Diffed the new commit (`5f6d3563`) against the previously reviewed head (`8b34fd8d`) and against `main`.
- Ran `bundle exec rspec` (full suite: 7159 examples, 0 failures, up from 7156) and `rubocop` (3 changed files, no offenses) against the updated branch.
- Re-ran my Major-finding reproduction script directly against the fixed code to confirm the behavior change, rather than relying on the new specs alone.
- Ran a second CodeRabbit CLI pass on the updated branch; it raised the two items captured above (the migration question and the stale comment reference) — both independently verified against the actual code rather than transcribed.
- Did not re-verify against a live game connection; verification is against the shared `Lich::Common::CreatureBase` mixin and the tracker's settings-loading logic directly.

PR #1605, opened 2026-09-12
refactor(all): extract UserDefs from DragonRealms CustomSubstitutions to Lich::Common

Nisugi:
## Problem

`Lich::DragonRealms::CustomSubstitutions` (#1512) contains two things: the DragonRealms-specific part (read a settings key, dispatch on `:pairs` / `:names` / `:regexes`) and a general-purpose unit for merging player-supplied definitions onto frozen core defaults. That second part is exactly what the upcoming GemStone combat-definition supplements need: lenient per-entry validation with key/index/value/reason/consequence messages, timeout-bounded regex compilation, once-only timeout reporting on match, non-ASCII warnings, and a memo with `reset!`. Rather than copy it into the GemStone tree, this PR moves it to `Lich::Common` so both games share one implementation.

## Change

- **New `lib/common/user_defs.rb`** — `Lich::Common::UserDefs`, a module you `extend` onto a definitions module. Owns `REGEX_TIMEOUT_SECONDS`, the per-extender lock/memo/timeout-report state, `reset!`, `apply_regexes`, and the private helpers: `memoize`, `validate_entries`, `validate_pair`, `validate_name`, `validate_regex`, `timeout_bounded`, `warn_non_ascii`, `report`, `report_debug`, `report_timeout`. The message prefix comes from the extender's `MESSAGE_PREFIX`, falling back to its short module name.
- **`lib/dragonrealms/custom_substitutions.rb`** now `extend`s `UserDefs` and keeps only what is DR-specific: `resolve`, `user_setting` (via `get_settings`), `SUPPORTED_TYPES`, `validate_entry` dispatch, and `MESSAGE_PREFIX`. `REGEX_TIMEOUT_SECONDS` is kept as an alias of the common constant for any caller referencing it.

## What does not change

- The public surface of `CustomSubstitutions`: `resolve`, `apply_regexes`, `reset!`, `REGEX_TIMEOUT_SECONDS`, `MESSAGE_PREFIX`, `SUPPORTED_TYPES`.
- Every player-facing message string, byte for byte, including the `[CustomSubstitutions]` prefix.
- The timeout value (1.0s), the double-checked memo, the once-per-pattern timeout dedup, and the lenient per-entry drop rule.
- Call sites in `common.rb`, `common-items.rb`, `common-summoning.rb`, and `drdefs.rb` are untouched.

## Verification

- `spec/lib/dragonrealms/custom_substitutions_spec.rb` passes **without a single edit** — 22 examples, 0 failures. That spec was written adversarially against the original (malformed entries, wrong container types, invalid and runaway regexes, memo/reset boundaries), so it is the compatibility proof.
- New `spec/lib/common/user_defs_spec.rb` drives the common module through a throwaway extender: memo/reset, list loop, all three validators, timeout wrapping on both string and pre-compiled `Regexp`, `apply_regexes` timeout guard including that `reset!` clears the dedup set, prefix fallback, and isolation of state between two extenders — 15 examples, 0 failures.
- Full `spec/lib/dragonrealms` plus the new spec: 2538 examples, 0 failures.
- RuboCop: no offenses on the three files.

## Follow-up

The GemStone combat-definition supplements (YAML file in the player's data directory, hot-reloaded via `;hmr`) will `extend Lich::Common::UserDefs` in a separate PR. Nothing in this PR depends on that landing.

### DIFF
--- a/lib/common/user_defs.rb
+++ b/lib/common/user_defs.rb
@@ -91,12 +91,20 @@ module Lich
 
       # Returns the memoized value for +key+, computing it with the block on
       # first use. Double-checked: no lock on the warm path (the common case on
-      # hot parse paths), lock only to populate a missing key.
+      # hot parse paths), lock only to populate a missing key. Presence, not
+      # truthiness, decides whether the key is cached, so a value of +false+ or
+      # +nil+ is memoized like any other.
       #
       # @param key [Object] memo key
       # @return [Object] the cached or freshly computed value
       def memoize(key)
-        @cache[key] || @lock.synchronize { @cache[key] ||= yield }
+        return @cache[key] if @cache.key?(key)
+
+        @lock.synchronize do
+          return @cache[key] if @cache.key?(key)
+
+          @cache[key] = yield
+        end
       end
 
       # Validates each entry of +raw+ with the block, keeping the non-nil
--- a/lib/common/user_defs.rb
+++ b/lib/common/user_defs.rb
@@ -1,0 +1,252 @@
+# frozen_string_literal: true
+
+module Lich
+  module Common
+    # Shared machinery for merging player-supplied definitions onto core Lich's
+    # frozen defaults: lenient per-entry validation with player-facing
+    # diagnostics, timeout-bounded regex compilation, guarded matching, and a
+    # memo with +reset!+.
+    #
+    # Extend it onto a module that owns one family of user definitions. The
+    # extending module supplies the reading (where entries come from) and any
+    # shape checks specific to its domain; everything below is domain-neutral.
+    # Extracted from {Lich::DragonRealms::CustomSubstitutions} so the GemStone
+    # combat definition supplements can share it without duplicating the
+    # validation, timeout and reporting rules.
+    #
+    # Contract for every player entry:
+    # - Invalid entries are dropped individually. One bad entry never disables
+    #   the rest of its list.
+    # - Every rejection is reported through {Lich::Messaging} with the exact
+    #   key, index, offending value, reason, and consequence.
+    # - Regexes compile with {REGEX_TIMEOUT_SECONDS} so a pathological pattern
+    #   raises {Regexp::TimeoutError} instead of hanging Lich.
+    #
+    # @example
+    #   module MyDefs
+    #     extend Lich::Common::UserDefs
+    #     MESSAGE_PREFIX = '[MyDefs]'
+    #
+    #     def self.patterns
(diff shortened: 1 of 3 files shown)


mrhoribu (approved):
**Verdict: Approve**

Clean, verified mechanical extraction. Confirmed:

- Full suite passes on the PR branch: 7302 examples, 0 failures. (The one failure I saw on a narrower subset run, `drdefs_spec.rb:606`, reproduces identically on `main` — pre-existing order-dependent flake, unrelated to this change.)
- `rubocop` on all three changed files: no offenses.
- Public surface (`resolve`, `apply_regexes`, `reset!`, `REGEX_TIMEOUT_SECONDS`, `MESSAGE_PREFIX`, `SUPPORTED_TYPES`) and every player-facing message string preserved byte-for-byte — traced every call site (`common.rb`, `common-items.rb`, `common-summoning.rb`, `drdefs.rb`, `startup.rb`, and their specs) and all go through the public API only.
- Traced the `extend`-based state model by hand: `UserDefs.extended(base)` sets `@lock`/`@cache`/`@reported_timeouts` on the extending module itself, and since the mixed-in methods run as singleton methods of that same object, state stays correctly isolated per extending module — confirmed at runtime by the new spec's isolation example.
- Ran a local CodeRabbit pass as a second opinion; its one finding is folded in below.

**Minor:** `memoize` in `lib/common/user_defs.rb:105` uses `@cache[key] || @lock.synchronize { @cache[key] ||= yield }` — a truthy check, not a presence check. Harmless today since the only current extender always memoizes a (truthy) Array, but now that this is shared infra other modules (the upcoming GemStone combat defs) will build on, a future caller that legitimately memoizes `false`/`nil` would silently recompute on every call instead of caching. Cheap to harden with `@cache.key?(key)` when it matters; not blocking.

**What this PR gets right:** the extraction is genuinely mechanical — identical logic, identical message strings, identical validation order, nothing "improved" along the way, which is exactly right for a refactor other code depends on staying byte-compatible. `REGEX_TIMEOUT_SECONDS` is kept as a real (not just documented) alias for any external caller referencing the constant directly. The new `user_defs_spec.rb` explicitly asserts state isolation between two independently-extending modules, the one behavior a naive version of this refactor could easily have gotten wrong.

Nisugi:
Addressed the review note: `memoize` now caches by key presence (`@cache.key?(key)`) rather than truthiness, so a memoized `false` or `nil` is not recomputed on every call. Added a spec example covering both. `45dcf4d3`.

PR #1607, opened 2026-09-12
feat(gs): combat definition supplements from a player YAML file

Nisugi:
## Problem

The combat tracker recognises game text through ~3,300 lines of pattern definitions under `lib/gemstone/combat/defs/`, all frozen Ruby constants that ship with Lich. When a player meets a flare, CMAN message or creature line the defs don't know, the only fix is a PR, a review and a release — days of turnaround for what is usually one regular expression.

## This PR

Rebased onto main now that #1605 has merged, so the UserDefs commits are no
longer part of this diff.

Two commits of substance, plus the follow-ups from review:

1. **The compiler.** `Lich::Gemstone::Combat::Definitions::Supplements` reads a single global file, `DATA_DIR/combat/defs.yaml`, and compiles its entries into **the same Struct objects the shipped tables are built from** (`AttackDef`, `FlareDef`, `StatusDef`, `OutcomeDef`).
2. **The wiring.** Each def module binds its lookup and gate as one frozen table in a single assignment; the four assembly points splice the supplements in; `reload_defs!` re-executes the def files in place; the Tracker drains its worker around a reload and reloads once at login if the file changed.

With no file present every reader is empty, so the assembled tables are the shipped ones (a spec asserts this, and again with an empty file).

### What a player writes

```yaml
attacks:
  - name: tackle                # shipped name: adds a pattern to the shipped CMAN
    slot: priority              # priority | generic | third_person
    patterns: ['You launch yourself bodily at (?<target>.+?) and connect!']
flares:
  - name: acid                  # shipped name: damaging/aoe/spawns inherited
    patterns:
      - '\*\* (?<attacker>.+?)''s{{MK_POST}} .+? spits a gob of acid! \*\*'
  - name: frost_flare           # new name: every omitted flag defaults to false
    damaging: true              # a damage line follows; attribute it to the flare, not the swing
    aoe: false                  # strictly single-target even if the parent attack is not
    spawns: false               # true only when the flare casts an imbedded spell as a separate attack (Blink)
    patterns:
      - '\*\* Your .+? flares with frost, striking (?<target>.+?)! \*\*'
statuses:
  - name: chilled
    add:    ['(?<target>.+?) shivers uncontrollably\.']
    remove: ['(?<target>.+?) stops shivering\.']
outcomes:
  - type: miss
    patterns: ['The ice lance shatters harmlessly near (?<target>[^.]+)\.']
```

### Rules (each pinned by a spec example)

- **Built on `Lich::Common::UserDefs`** (#1605): one bad entry never disables the rest; every rejection names kind, index, value, reason and consequence; patterns compile through the shared 1.0s timeout-bounded helper; non-ASCII is warned about.
- **`YAML.safe_load_file`**, no permitted classes, no aliases. A missing file, empty file, non-mapping document or syntax error all yield empty frozen arrays (the last three with a message).
- **Attacks pick a slot** (`priority`, `generic`, `third_person`) so the load-bearing match order in `attacks.rb` stays explicit. An attack with no `target`/`attacker` capture is warned about, not rejected.
- **`{{MK_PRE}}` / `{{MK_POST}}`** expand to the markup-tolerance fragments the shipped defs interpolate; an unknown token rejects the pattern.
- **Flare flags** are the same three the shipped `FlareDef` carries: `damaging` (a damage line follows and belongs to the flare, not the swing), `aoe` (may own hits on every target of the parent attack; false means strictly single-target), `spawns` (casts an imbedded spell as a separate attack, so following attack lines are its children — Blink). Timing is not a flag; it is read from the line's position in the log. The example file documents all three.
- **Reusing a shipped flare name inherits its flags.** Omitted `damaging`/`aoe`/`spawns` resolve from the shipped def; an explicit flag that contradicts it rejects the entry, so one name never carries two behaviours. New names default omitted flags to false.
- **Outcome `type`** is restricted to the shipped outcome types.
- **A status** needs at least one of `add:` / `remove:`.
- **Re-read on mtime change**; `reset!` forces it. `path` is assignable for tools and specs. `summary` reports counts per kind for debug output.
- **Two readers for `;e` at the prompt**: `summary` counts per kind, and `loaded` names them, attacks grouped by the slot they splice into. A slot typo, or a misspelled name that quietly makes a new def instead of extending a shipped one, reads the same in the counts and is obvious in `loaded`.
- **Does not require the def files it serves.** They will require it, and call the readers from their assembly step after their own Structs and shipped lists exist; shipped constants are resolved lazily at call time.
- The shared mutex is not reentrant and the builders run inside it, so the document is parsed before a kind's lock is taken (commented in code).

### Files

- `lib/gemstone/combat/defs/supplements.rb` — the module.
- `lib/gemstone/combat/defs/supplements.example.yaml` — documents the contract in comments; shows a reused CMAN, a reused flare, and one new entry of each kind. The spec loads it and asserts zero rejections.
- `spec/lib/gemstone/combat/supplements_spec.rb` — 26 examples: Struct equality against hand-written defs, every rejection path, token expansion (including a live-markup match), flag inheritance/contradiction, mtime re-read, `reset!`, and the example file.

## The wiring

- **`Definitions::Table`** (`pattern_gate.rb`): one frozen struct of lookup rows plus the PatternGate derived from them. `Attacks`, `Flares`, `Statuses` and `Outcomes` each bind a `TABLE` **last and in a single assignment**, and their parse entry points (and the Parser's two attack scans, via `Attacks.table`) read it **once per call**. A hot reload therefore never pairs a lookup from one build with a gate from another, and a def file that fails to load leaves its previous complete table live. Existing constants (`ATTACK_LOOKUP`, `STATUS_GATE`, …) are kept, so `pattern_gate_spec` and anything else reading them is unaffected.
- **Splice points.** `attacks.rb` splices `Supplements.attacks(:priority)` ahead of the generic swings, `:generic` with the second-person defs, `:third_person` ahead of the shipped third-person defs, so the load-bearing order stays explicit. `flares`, `statuses` and `outcomes` append `Supplements.<kind>` after the shipped defs before deriving.
- **`Supplements.reload_defs!`** re-reads the file and re-executes every loaded `defs/*.rb` (what `;hmr combat/defs/` matches) with constant-redefinition warnings silenced, reporting and skipping a file that fails. **`Supplements.stale?`** says whether the file changed since the tables were assembled.
- **`Tracker.reload_defs!`** drains the async worker around the reload and restarts it if tracking is on; **`initialize!`** reloads once at login when the file is stale, so a relog always reflects the current file.
- Builders for a kind whose def module is not loaded return empty, so a tool that loads one def file on its own still works.

`supplements_wiring_spec.rb` (9 examples): identity with no file and with an empty file; every kind recognisable through the shipped parse entry points after a reload; shipped defs still matching alongside supplements; priority-slot ordering; removal on file delete; `stale?`; a def file that fails to load keeping its previous table while the rest reload.

## Verification

- `spec/lib/gemstone/combat` + DR `custom_substitutions_spec` + `user_defs_spec`: 401 examples, 0 failures. The per-fixture replay spec is unchanged.
- RuboCop: no offenses on the changed files.
- Full suite: the only failures are in frontend/GTK/wine/socket specs that fail identically on a clean `main` checkout on this Windows box.

## Next

`messages:` (the EOHunter item-preparation bridge) is a separate draft stacked on this and #1586.


therealatari (commented):
Reviewed at `cd9abc3f01fc4f2505f14bcb4f811a74d6835a21`, treating #1605 as the dependency and this PR as **compiler-only**. The shared-helper reuse and separate wiring PR make sense. The existing combat + CustomSubstitutions + UserDefs suite passes locally: **391 examples, 0 failures** (seed 40379).

Two additional, reproducible compiler-contract issues are worth addressing before the wiring consumes these results:

### P2: A concurrent reset can leave old compiled definitions in the new cache

[supplements.rb:130–134](https://github.com/elanthia-online/lich-5/blob/cd9abc3f01fc4f2505f14bcb4f811a74d6835a21/lib/gemstone/combat/defs/supplements.rb#L130-L134)

`memoize` captures `doc = document` before entering the shared helper's build/cache mutex. A reset can occur between those operations. The old reader then publishes a result built from the old document into the new cache.

A deterministic two-thread test, using barriers rather than timing sleeps, reproduces this:

1. Thread A obtains the document containing an attack named `before_edit`; pause it just before `document` returns.
2. Thread B rewrites the file with `after_edit`, advances its mtime, calls `reset!`, and reads `flares` to populate the new document cache without populating the attack cache.
3. Resume A, then make an independent `attacks` call.

Observed:

```text
overlapping old read: [:before_edit]
later independent read: [:before_edit]
new document cached: after_edit
```

The overlapping caller seeing an old snapshot is not the issue; retaining it for subsequent independent callers is. The recorded mtime is already current, so it persists until another edit/reset.

Suggested fix: serialize document acquisition/build/publication, or carry a generation with the captured document and validate it under the publishing lock, retrying if invalidated. A barrier-controlled regression should assert that the later independent reader sees `after_edit`.

### P2: Repeated supplemental flare names can acquire conflicting flags

[supplements.rb:329–343](https://github.com/elanthia-online/lich-5/blob/cd9abc3f01fc4f2505f14bcb4f811a74d6835a21/lib/gemstone/combat/defs/supplements.rb#L329-L343)

The stated invariant, “one name must never carry two behaviours,” is enforced against shipped names only. Two entries for the same *new* name bypass it:

```yaml
flares:
  - name: custom_frost
    damaging: true
    aoe: true
    patterns: ['Your staff bursts with frost!']
  - name: custom_frost
    patterns: ['Your staff spits frost!']
```

The actual compiler returns these `(name, damaging, aoe, spawns)` tuples:

```ruby
[:custom_frost, true, true, false]
[:custom_frost, false, false, false]
```

Once consumed, the same named flare would have different damage-attribution/AoE behavior depending on its message. Explicit contradictory flags are accepted too.

Suggested fix: resolve names against accepted supplemental definitions as well, inheriting omitted flags and rejecting contradictions; alternatively, reject duplicate custom names with a diagnostic directing the player to combine their patterns. Cover both omitted and explicit conflicting flags.

Neither finding is a claim of a current live-parser failure: this PR deliberately does not splice the results into combat tables yet. No missing-wiring/reload-command finding, and no new documented coding-standard violations found. Horibu's pre-existing false/nil memoization nit belongs to #1605 and is separate from the reset-generation issue above.

Nisugi:
Both findings addressed in `6a1dfa79`:

- **Concurrent reset:** `memoize` now validates, under the lock, that the document it built from is still the cached one; if a reset intervened the build is discarded and retried against the current file. Added the barrier-driven two-thread regression you described (`does not publish a build from a stale document after a concurrent reset`). Against the previous code it returns `[:before_edit]` for the later independent reader; with the fix both the overlapping reader and the later reader see `[:after_edit]`.
- **Repeated new flare names:** the first accepted entry for a name now owns its flags. Later entries inherit what they omit and are rejected on a contradiction, with the diagnostic naming the owning entry (`contradicts flares[0] (custom_frost) (damaging: true)`). Specs cover both the omitted-inherit and explicit-conflict cases.

Note the branch now also carries the wiring (splice points, per-module frozen `TABLE`, `reload_defs!`, Tracker hooks); the description is updated to match.

Nisugi:
Both reproduced and fixed in `05080353`.

**1. Inspection suppressing the reload.** Confirmed exactly as described: after an edit, `stale?` went true, then calling `loaded` flipped it to false while the new definition was still absent from the assembled table, so a login-time `reload_defs! if stale?` would skip it.

The cause is that `@loaded_mtime` is stamped by `refresh_if_changed`, which every reader runs to re-parse the document, while `stale?` is documented to answer for the tables. Those are now separate. `@assembled_mtime` is set by a new `Supplements.assembled!`, called by each def module at the point it splices the supplements in, which is require time and again on each reload. Those are the only two ways a table gets built, and inspection deliberately does not call it. I put the marker in the def modules rather than in `reload_defs!` because boot assembles through `require`, so stamping only on reload would leave a fresh session looking permanently stale.

**2. Case-folded patterns filtered out.** Also confirmed, and slightly wider than reported. Three spellings are affected: the inline `(?i)` you found, plus `//i` and `Regexp::IGNORECASE`, which set the option instead of appearing in the source.

Worth recording that **the obvious fix is wrong**. Refusing the literal, so the pattern falls to `always_scan`, is what I tried first, and it makes `message_defs_spec` fail: three shipped message families (ambush, bless, archery) are built from `//i` patterns, and every one of them lost its gate entirely and dropped to a full scan of every line.

What landed instead keeps the literal and matches it the way its pattern does. `build` sorts literals into an exact union and a case-insensitive one, joined by `Regexp.union`, which preserves each side's flags. A case-folded pattern stays gated, exact literals keep their case sensitivity, and `always_scan` is unchanged.

The safety property over the real def files now also asserts the literal matches in the casing its pattern accepts, and covers a message family, since that is where the shipped case-folded patterns actually live. Your `(?i)ZEPHYR chills` case is pinned end to end through `Statuses.parse` in the wiring spec.

All three new specs fail against the previous code and pass now. Suite is 449 examples, 0 failures, RuboCop clean. #1608 is rebased on this and picks up the same `assembled!` marker for the message table.

Nisugi:
Reproduced and fixed in `803f8fbc`.

Confirmed exactly as described. With statuses.rb failing to load and the other files reloading cleanly, `stale?` returned false while the status table still held the previous definition, so the login check would not retry it.

**Per kind, not all-or-nothing.** `assembled!` now takes the caller's own kind and records the stamp under it; `stale?` is true when any assembled kind is behind the file. I went this way rather than only marking the reload current once every file succeeds, because a permanently broken def file would then report stale forever and make every login attempt a full reload. Per kind, the kinds that reloaded are current and only the one that failed keeps asking to be retried.

Kinds whose module was never loaded are not compared. They have no table to be out of date, and counting them would be the same forever-stale problem from the other direction. With nothing assembled at all the answer is stale, so a reload is never skipped; the only caller runs at login, long after the def files are required.

The regression sits with the existing failed-load example and covers both halves: after a reload where statuses.rb raised, the attack supplement is live, the status supplement is not, and `stale?` is true; then the retry that staleness earns succeeds and clears it. It fails against the shared-stamp code.

Suite is 450 examples, 0 failures, RuboCop clean. #1608 is rebased on this, and its message table declares `:messages` the same way.

Nisugi:
All four reproduced and fixed in `bad57cbc`.

**1 (P1) — an accepted regex stopping the load.** Confirmed: with `^(\w+) echoes \1$` in the file, statuses.rb raised `RegexpError` during a cold load and `TABLE` never existed. The pattern is legal alone; it is only illegal *inside a union*, beside a shipped named capture.

The two detectors are the only place such a union is built, and nothing in Lich reads them — their own comments say the literal gate replaced them. They are now built on first use via `PatternGate.union_or_nil`, which reports and returns nil rather than raising. The old constant names still resolve through `const_missing`, so an out-of-tree script reading `STATUS_DETECTOR` keeps working. Matching is unaffected either way, since `PatternGate.build` handles each pattern separately.

**2 (P2) — the private entry point.** Confirmed: `private method 'reload_defs!' called for module ...Tracker`. Moved above the `private` declaration. Its neighbours (`initialize_processor`, `shutdown_processor`, `add_downstream_hook`) are genuinely internal and stayed put.

**3 (P2) — chunks lost across the swap.** Confirmed. A `@reload_lock` now spans reading-and-enqueueing on the ingestion side and the drain-and-swap on the reload side, so a chunk arriving mid-reload is enqueued on the new worker. The lock never spans a parse: the async path holds it only for the push, and the inline path (`max_threads <= 0`) parses outside it.

Worth recording that **my first regression here did not actually pin the bug** — it passed with the lock removed, because the old worker drains whatever it reaches before its sentinel, so "was the chunk parsed at all" is true either way. The spec now blocks inside `shutdown` (the real window) and asserts *which* processor parsed it, via each processor's own `stats[:total]`. Without the lock the new worker's total is 0.

**4 (P2) — timeouts escaping.** Confirmed: an accepted pattern raising `Regexp::TimeoutError` stopped a later valid definition from matching. Status, outcome, flare and the parser's two attack loops now go through `PatternGate.safe_match`/`safe_match?`, which skip that pattern for the line and report once per source through the `UserDefs` dedup set.

The gate needed a third state. `safe_match?` returning false cannot distinguish a timeout from an honest non-match, and treating a timeout as "no match" made `rejects?` reject the line and hide every def behind the gate. `rejects?` now uses a tri-state helper: a timeout leaves the answer undecided, so the line goes to the full scan, where the same pattern is re-evaluated under `safe_match` and reported.

The shipped-only pattern sets in attacks.rb (`AMBUSH_PREFIXES`, `REACTION_PREFIXES`, `REDIRECT_PREFIXES`, `COUP_KILL_PATTERNS`) take no supplements, so they carry no user patterns and were left alone.

Suite is 461 examples, 0 failures, RuboCop clean, with a new `tracker_reload_spec.rb` and six new PatternGate examples. #1608 is rebased on this; its message scan now uses the shared helpers, so its own timeout rescue and `family_rejects?` came out as redundant (482 examples there).

Nisugi:
`39902a9c` fixes a finding raised on #1608 that belongs here: the lazy detectors went stale after a reload.

This was a regression I introduced in `bad57cbc` making the detectors lazy. The memo is a module ivar, so it survives the def file re-executing: a reload rebound `TABLE` while `ATTACK_DETECTOR` kept answering for the previous one, and the parser recognised a newly added supplemental pattern that the detector did not. `STATUS_DETECTOR` had the same problem.

The cached `nil` was the worse half. Once one entry made the union impossible, the detector stayed nil for the rest of the session even after that entry was removed from the file.

The memo is now dropped where the table rebinds, so any route that re-executes the file gets a fresh detector on the next read. Three regressions cover the attack detector, the status detector, and the nil case clearing once the offending pattern is gone; all three fail against the previous code.

464 examples, 0 failures, RuboCop clean.

Nisugi:
Two more in `2ff9f703`, from a full read of the stack rather than a review finding.

**`assembled!` stamped the wrong moment.** It took a fresh stat of the file, but the caller's table was built from the document the reader parsed a few lines earlier. An edit landing in that gap was recorded as assembled while the table was still behind the file, so `stale?` answered false and the login-time reload skipped it. Reproduced by writing the file between a reader call and `assembled!`. The stamp is now the document cache's own mtime, so "what was this built from" has a single answer in the module. Inspection still does not touch the assembled stamps, so the earlier separation holds.

**`case_folded?` missed `(?i-m)`.** The inline-flag check required the flags to run straight to the closing paren, so a group that also switches flags off slipped through and its literal went into the case-sensitive gate. Widened; `(?x-i)` correctly still reads as not folded.

Both regressions fail against the previous code. 467 examples, 0 failures, RuboCop clean.

mrhoribu (changes requested):
# PR Review: feat(gs): combat definition supplements from a player YAML file (elanthia-online/lich-5#1607)

**Reviewed:** 13 files, +1766/-35 at `bad57cbc` on 2026-09-14. Based on: full local clone (worktree checkout of the PR branch, `git log`/`git blame` available), diffed against the actual merge base `d876267` (origin/main).
**Verdict:** Request changes
**Summary:** Adds a player-editable `DATA_DIR/combat/defs.yaml` that compiles into the same frozen Struct tables the shipped combat defs use, spliced in at four points with hot-reload support. The design is careful and the commit history shows real iteration against prior review (stale-document races, flag-inheritance rules, staleness tracking, case-folded gate literals, timeout isolation, a worker-swap race) — all of it backed by 75 new, mostly non-sleep-based concurrency specs, and the whole combat suite (445 examples) is green. The one design gap worth fixing is that the same file, `pattern_gate.rb`, that defines the `Table` struct is itself included in the set of files `reload_defs!` re-executes, which silently redefines the `Table` class on every hot reload.

## Minor

### 1. `reload_defs!` reloads `pattern_gate.rb`, which redefines the `Table` struct class — `lib/gemstone/combat/defs/supplements.rb:225`

```ruby
DEF_FILE_PATTERN = %r{[/\\]gemstone[/\\]combat[/\\]defs[/\\](?!supplements\.rb\z)[^/\\]+\.rb\z}.freeze
```

This pattern excludes only `supplements.rb`. Every other `.rb` file directly under `defs/` matches, including `pattern_gate.rb` itself — which defines `Table = Struct.new(:lookup, :gate, :always_scan) do ... end` (`lib/gemstone/combat/defs/pattern_gate.rb:42`). `Struct.new` returns a new anonymous class each time it runs, so reloading `pattern_gate.rb` reassigns the `Definitions::Table` constant to a different class object than the one every previously-built `TABLE` is an instance of.

I reproduced this directly: loaded the four def modules, stubbed `Supplements.load` to raise for `flares.rb` only, then called `reload_defs!`. The reload list came back as `["pattern_gate.rb", "spells.rb", "attacks.rb", "statuses.rb", "outcomes.rb"]` — `pattern_gate.rb` really is in the reload set — and afterward:

```
Table class changed? true
Flares::TABLE still equal to old (should be true, load failed)? true
Flares::TABLE.is_a?(current Table class)? false
Attacks::TABLE.is_a?(current Table class)? true
```

`Flares::TABLE` is exactly the object it was before (the "previous complete table stays live" guarantee the docstring promises holds at the object-identity level, and `rejects?`/`lookup` still work fine since they're plain instance methods on the still-live old class), but it is no longer `is_a?(Definitions::Table)` against the current constant. Nothing in this repo currently does that type check outside `spec/lib/gemstone/combat/supplements_wiring_spec.rb:1804-1813` (`expect(table).to be_a(defs::Table)`), and that example doesn't run after a partial-failure reload, so this wouldn't be caught by CI even if it regressed further. Today's blast radius is therefore small, but it's a real crack in a guarantee the PR is explicit about, and it costs nothing to close: exclude `pattern_gate.rb` the same way `supplements.rb` already is. CodeRabbit's automated pass flagged the identical root cause independently.

**Fix:** `(?!supplements\.rb\z|pattern_gate\.rb\z)` in the pattern.

### 2. The chunk-loss race this PR fixes for `reload_defs!` still exists in `disable!` and `configure` — `lib/gemstone/combat/tracker.rb:154`, `:316`

The PR's own commit message describes the bug precisely: "Ingestion read `@async_processor`, then the reload drained and replaced that worker, and the push landed behind the old queue's shutdown sentinel, on a queue nothing drains again." The fix wraps the read-and-enqueue in `process` and the drain-and-swap in the new `reload_defs!` under one `@reload_lock` (`tracker.rb:263-266`, `:364-369`).

But `disable!` and `configure` do the identical `shutdown_processor` / `initialize_processor` dance without that lock:

```ruby
def disable!
  return unless @enabled
  ...
  shutdown_processor
  respond "[Combat] Combat tracking disabled" if debug?
end
```
```ruby
def configure(new_settings = {})
  ...
  if new_settings.key?(:max_threads)
    shutdown_processor
    initialize_processor
  end
  ...
end
```

A chunk arriving on the game-stream thread via `process` (which now does hold `@reload_lock` while it snapshots `@async_processor` and calls `process_async`) can still race a concurrent `Tracker.disable!` or a `configure(max_threads: ...)` call (e.g. `enable_debug!`/`disable_debug!`, which call `configure` under the hood) that isn't holding that lock — the exact same "push lands behind the dying worker's shutdown sentinel" loss the PR just fixed elsewhere. This is pre-existing behavior, not introduced by this PR, and out of its stated scope — but since the PR already built the mechanism and the file is already open, it's a one-line extension (`@reload_lock.synchronize { shutdown_processor }` / wrap the `configure` branch) to close the same hole everywhere it exists rather than just at the one new call site.

### 3. `ATTACK_DETECTOR`/`STATUS_DETECTOR` back-compat shim never invalidates on reload — `lib/gemstone/combat/defs/attacks.rb:803`, `lib/gemstone/combat/defs/statuses.rb:470`

```ruby
def self.detector
  return @detector if defined?(@detector)
  @detector = PatternGate.union_or_nil(ATTACK_LOOKUP.map(&:first), 'attacks')
end
```

`@detector` is a plain ivar memo with no reset hook. `reload_defs!` re-executes `attacks.rb`/`statuses.rb` via `load`, which redefines the `detector` method but does not clear the ivar on the already-existing module object, so a caller that reads `ATTACK_DETECTOR`/`STATUS_DETECTOR` (via `const_missing`) both before and after a hot reload gets the same regexp built from the pre-reload pattern set the second time too. I confirmed via `grep -rn "ATTACK_DETECTOR\|STATUS_DETECTOR"` that nothing inside lich-5 itself reads these constants — the comment's "nothing in Lich reads this" is accurate for this repo — so this only matters for an external `.lic` script exercising the documented back-compat path across a hot reload, which is a narrow, low-traffic case. Worth a one-line `@detector = nil` somewhere reload touches, but not urgent.

## Nits

- `lib/gemstone/combat/defs/pattern_gate.rb:45` — `Table.build(lookup)` is defined but never called; all four def modules (`attacks.rb`, `flares.rb`, `statuses.rb`, `outcomes.rb`) inline the two-step `PatternGate.build` + `Table.new` instead, apparently because they still need the separate `ATTACK_GATE`/`ATTACK_ALWAYS_SCAN`-style constants for back-compat (confirmed: `pattern_gate_spec.rb` reads `a::ATTACK_GATE`/`a::ATTACK_ALWAYS_SCAN` directly). Not wrong, just dead API surface — either use it somewhere or drop it.
- `spec/lib/gemstone/combat/tracker_reload_spec.rb:76` — `sleep 0.1` is used to let the `ingest` thread reach and block on `@reload_lock` before `release` is sent, instead of a deterministic handshake like the `Queue`-based barriers used elsewhere in this same PR (e.g. `supplements_spec.rb`'s `a_captured`/`resume_a` pattern, or this same test's own `in_shutdown`/`release` queues one step earlier). On a slow CI box the `ingest` thread could still be unscheduled when `release` fires, in which case the assertions would pass without ever exercising the lock contention the test claims to prove. CodeRabbit flagged this independently.

## Open questions

- `lib/gemstone/combat/tracker.rb:270` — `Processor.process(chunk, source: source)` (the inline fallback for `max_threads: 0` / debug mode) runs outside `@reload_lock` by design ("so a reload never waits on a parse"). If `Tracker.reload_defs!` is invoked from another thread while an inline parse is in flight, `reload_defs!` reassigns the four `TABLE` constants one def-file-`load` at a time rather than atomically as a set, so the in-flight parse could read e.g. the new `Attacks::TABLE` and the old `Flares::TABLE` for the same chunk. **Settles it:** is inline mode (debug) ever run in practice alongside a script-driven `reload_defs!` call, or is debug mode always a solo, non-concurrent session in this codebase? If the latter, this is a non-issue and not worth guarding.
- `lib/gemstone/combat/defs/pattern_gate.rb:92` — `case_folded?` detects the inline-flag form via `/\(\?[a-z]*i[a-z]*\)/`, which won't match a mixed enable/disable form like `(?i-x)`. No shipped pattern uses that form (checked via grep across `defs/*.rb`), so this is purely a latent gap for a player-authored supplemental pattern that happens to use it. **Settles it:** is this form worth guarding against, or is it acceptable that such a pattern would silently degrade to gate-rejecting lines it should match (the same failure mode `case_folded?` was added to fix for the common `/i` case)?

## What this PR gets right

- The commit history is unusually honest about its own bugs: six of the eight commits are review-driven fixes (stale-document publication race, flag-inheritance rules for reused names, `stale?` tracking the wrong stamp, case-folded gate literals, the reload-entry-point visibility bug, and the async-worker swap race), each with a regression test that reproduces the original failure via barriers rather than sleeps.
- The "read the table once per call" discipline (`table = TABLE` at the top of every `parse`/`rejects?` call site) correctly protects against pairing a lookup from one build with a gate from another during a hot reload — I verified constant reads are atomic under MRI and every consumer follows this pattern consistently.
- Backward compatibility is genuinely preserved, not just claimed: `ATTACK_LOOKUP`, `STATUS_GATE`, `ATTACK_DETECTOR` (via `const_missing`) all still resolve exactly as before for any external code depending on them.
- `YAML.safe_load_file` with `permitted_classes: []`, no aliases, is the right call for a file untrusted players edit directly — a deliberate, well-reasoned departure from this ecosystem's more common `unsafe_load_file` (used elsewhere in this codebase for trusted `!ruby/regexp` profile YAML, which is a different trust boundary).
- The flare flag-inheritance rule ("one name never carries two behaviours," owned by whichever def introduces it first) is subtle and correctly implemented — traced `validate_flare_flags` end to end against both the shipped-name and repeated-new-name cases and found no gap.
- Every rejection path reports kind, index, offending value, reason, and consequence, matching the stated contract, and the example YAML file is itself asserted (via spec) to load with zero rejections.

## Coverage notes

- Full local clone via `git worktree` on the actual PR branch; diffed against the correct merge base (`d876267`, verified via `git merge-base`) after discovering the local `origin/main` ref was stale.
- Read the complete diff (all 2104 lines) and the full content of every changed source file (not just the diff hunks) — `supplements.rb`, `pattern_gate.rb`, `tracker.rb`, `attacks.rb`, `flares.rb`, `statuses.rb`, `outcomes.rb`, `parser.rb`, `async_processor.rb`, and the shared `user_defs.rb` dependency.
- Ran the PR's new/modified spec files directly (75 examples, 0 failures) and the full `spec/lib/gemstone/combat` suite (445 examples, 0 failures).
- Ran `rubocop` against every changed `lib`/`spec` file: 12 files inspected, no offenses.
- Ran a CodeRabbit automated pass (`coderabbit review --agent --committed` against the merge base) — 2 findings, both independently verified and incorporated above (Minor #1 and the tracker_reload_spec sleep nit); nothing from it was discarded.
- Verified Minor #1 by direct reproduction (a standalone Ruby script loading the real def modules and driving `reload_defs!` with a stubbed failure), not just by reading the regex.
- Did not run this against a live Lich/GS4 session — no live game connection available in this environment. The wiring specs exercise the shipped `parser.rb`/`Flares`/`Statuses`/`Outcomes` entry points directly and are the closest available proxy; I did not independently re-verify the PR's own claim of "401 examples, 0 failures" across the wider DR combat/user-defs suite or the full lich-5 suite, though the targeted combat suite I ran overlaps most of that surface.

Nisugi:
Thanks for the thorough pass. Everything is addressed as of `c786b4b4`; two items were already fixed in commits after your base (`bad57cbc`), noted below.

**Minor 1, pattern_gate.rb in the reload set.** Confirmed and fixed exactly as you suggest: `DEF_FILE_PATTERN` excludes it alongside supplements.rb. The failed-load example now also asserts the kept `Flares::TABLE` is still `a Table` and that pattern_gate.rb is absent from the reloaded list, so a regression is caught in CI rather than only by a type check nothing runs.

**Minor 2, the same race in `disable!` and `configure`.** Fixed. Both take the ingestion lock around their drain and swap, `enable!` too for symmetry, and `process` re-checks `enabled?` under the lock so a chunk that parked through a `disable!` does not come out the other side and parse inline. Two regressions pin those. The `disable!` one needed care: a chunk arriving during `disable!` is normally dropped by the leading `enabled?` check before the race is ever run, so the spec holds it past that check with a barrier in the relevance filter. Fails without the lock, passes with it.

**Minor 3, detector never invalidated.** Already fixed in `39902a9c`, after your base: the memo is dropped where the table rebinds, with three regressions including the cached-nil case.

**Nit, dead `Table.build`.** Removed.

**Nit, `sleep 0.1` in the reload spec.** Replaced with a bounded wait for the ingest thread to actually park (`Thread#stop?`), shared by the three swap specs, so a slow box waits and a wedged one fails instead of hanging.

**Open question, `(?i-x)`.** Already handled in `2ff9f703`: `case_folded?` accepts a trailing `-flags` group, with a spec for `(?i-m)` folding and `(?x-i)` not.

**Open question, inline mode during a reload.** Inline parsing is `max_threads <= 0`, which the code labels a debugging aid; it is not the shipped default and I know of no script that drives `reload_defs!` against it. I have left it unguarded on purpose rather than hold the ingestion lock for the length of a parse. If that assumption is wrong for anyone, the fix is to take the lock around the inline `Processor.process` call, which is a two-line change.

Also in this commit: the two `Layout/EmptyLines` offences CI caught. A Windows checkout's CRLF handling hid them from local rubocop, which is why the branch went up red; whole-repo rubocop is now clean here as well as in CI.

469 examples, 0 failures.

mrhoribu (approved):
# PR Re-Review: feat(gs): combat definition supplements from a player YAML file (elanthia-online/lich-5#1607)

**Reviewed:** Incremental update since the prior review (`bad57cbc` → `c786b4b4`), 3 new commits, 9 files touched (+236/-19). Same merge base (`d876267`) as before. Based on: full local clone (second `git worktree` on the updated branch).
**Verdict:** Approve
**Summary:** All three Minor findings and both Nits from the prior review are fixed, each with a regression test that fails against the pre-fix code (per the author's own commit messages, and independently confirmed below). Both open questions were answered directly rather than left hanging — one fixed, one a deliberate, justified judgment call. No new issues found. CI is green (rubocop, both test jobs, title check).

## Prior findings — resolution

### Minor 1 (`pattern_gate.rb` in the reload set) — Fixed, verified
`DEF_FILE_PATTERN` now excludes `pattern_gate.rb` alongside `supplements.rb` (`lib/gemstone/combat/defs/supplements.rb:92`), with the reasoning captured in the comment. I re-ran the same reproduction I used to originally confirm this bug (stub `Supplements.load` to fail for `flares.rb`, call `reload_defs!`, inspect class identity) against the new branch:

```
Reloaded: ["spells.rb", "attacks.rb", "statuses.rb", "outcomes.rb"]
pattern_gate.rb in reload set? false
Table class changed? false
Flares::TABLE.is_a?(current Table class)? true
```

`pattern_gate.rb` no longer reloads, `Table`'s class identity is now stable across a reload, and the kept `Flares::TABLE` (from the file that failed) is correctly `is_a?` the current class. The wiring spec now pins this directly (`supplements_wiring_spec.rb:273-279`): asserts `pattern_gate.rb` is absent from the reloaded list and that both the kept and freshly-rebuilt tables are `a Table`.

### Minor 2 (same worker-swap race in `disable!`/`configure`) — Fixed, verified
`enable!`, `disable!`, and `configure`'s `max_threads:` branch all now wrap their `shutdown_processor`/`initialize_processor` calls in `@reload_lock.synchronize` (`tracker.rb:143`, `:164`, `:334-338`), matching `reload_defs!`. `process` also re-checks `enabled?` *inside* the lock (`tracker.rb:269`) and now treats that as a terminal `:disabled` sentinel rather than falling through to the inline `Processor.process` fallback — closing a gap I hadn't flagged (a chunk that parked through a concurrent `disable!` could previously have fallen through to the inline path after the lock released). Traced all five `@reload_lock.synchronize` call sites (`tracker.rb:143,164,265,334,375`) — none nested inside another, no reentrancy risk with Ruby's non-reentrant `Mutex`.

Two new specs (`tracker_reload_spec.rb`) exercise this with real thread interleaving via barrier queues (not sleeps): one proves `disable!` drops a chunk that arrives mid-drain rather than parsing it either behind the dying worker or after `disable!` returns; the other proves `configure(max_threads:)` hands such a chunk to the new worker. Both pass.

### Minor 3 (`ATTACK_DETECTOR`/`STATUS_DETECTOR` never invalidated) — Fixed, verified
Both `attacks.rb` and `statuses.rb` now do `remove_instance_variable(:@detector) if instance_variable_defined?(:@detector)` right after `Supplements.assembled!(...)` (`attacks.rb:820`, `statuses.rb:486`), so the memo is dropped every time the file rebinds its table. I confirmed independently:

```
Before detector matches new pattern? false
Detector object changed after reload? true
After detector matches new pattern? true
```

This also fixes a variant I hadn't caught: a cached `nil` (from a supplement that made the union illegal) would previously have stayed `nil` forever even after the offending entry was removed. The new wiring spec (`supplements_wiring_spec.rb:216-263`) covers exactly that case.

### Nit (dead `Table.build`) — Fixed
Removed outright (`pattern_gate.rb`); confirmed no dangling callers remain.

### Nit (`sleep 0.1` in the reload spec) — Fixed, and improved beyond what I asked for
Replaced with `wait_until_parked(thread, deadline: 5)`, which polls `Thread#stop?` via `Thread.pass` and raises on a bounded timeout instead of hanging (`tracker_reload_spec.rb:306-313`). This is now the shared synchronization primitive for all three worker-swap specs (the original reload test plus the two new `disable!`/`configure` ones), which is a better outcome than patching the one instance I flagged.

### Open question 1 (`(?i-x)` case-folding) — Fixed
`case_folded?` now accepts an optional trailing `-flags` group: `/\(\?[a-z]*i[a-z]*(?:-[a-z]*)?\)/` (`pattern_gate.rb:87`). I traced this by hand against both cases: `(?i-m)` — the required `i` is matched before any `-` is consumed (character class `[a-z]*` can't cross the `-`), so it still requires `i` in the *enabled* segment → correctly folds. `(?x-i)` — `i` only appears after the `-`, so the mandatory `[a-z]*i[a-z]*` segment before the optional group can never reach it → correctly does not fold. Matches the new spec assertions exactly.

### Open question 2 (inline mode during a concurrent reload) — Answered, not fixed, and that's a reasonable call
The author left this unguarded deliberately: inline processing only happens at `max_threads: 0`, which is explicitly the debug-mode path, not the shipped default, and they state no known script drives `reload_defs!` against it. They named the fix (take the lock around the inline `Processor.process` call) as a two-line change if that assumption ever proves wrong. This is a legitimate, documented trade-off rather than an oversight — I have nothing to add.

## New findings

None. I re-read the full incremental diff (`lib/gemstone/combat/defs/{attacks,pattern_gate,statuses,supplements}.rb`, `lib/gemstone/combat/tracker.rb`, and all four touched spec files), re-ran the empirical reproductions rather than trusting the diff on inspection alone, and found nothing the author's own fixes introduced. One genuine bonus catch on their part, not mine: `assembled!` was stamping a *fresh* `File.mtime(path)` instead of the mtime the document was actually parsed at (`supplements.rb:76`), which meant an edit landing between a reader call and the `assembled!` call a few lines later got recorded as already-assembled while the table was still built from the old document — a real TOCTOU bug I hadn't found, caught by the author's own "full read of the stack" and fixed with a regression test (`supplements_spec.rb:186-203`) that fails against the prior code.

## What this round gets right

- Every fix ships with a regression test the author states fails against the pre-fix code — I didn't just take that on faith for the two most structurally important ones (pattern_gate.rb exclusion, detector staleness) and reproduced both independently.
- The `disable!` fix in particular required real care to test correctly (a chunk has to be held *past* the leading `enabled?` check or the race never triggers), and the author's commit message says so explicitly — the spec (`tracker_reload_spec.rb:342-386`) matches that description exactly.
- Two issues were caught and fixed that were never raised in review at all (the `assembled!` TOCTOU stamp, the cached-`nil` detector case) — the author is finding their own bugs faster than review is, which is the best possible sign for a change this concurrency-sensitive.
- CI's rubocop failure (CRLF hiding `Layout/EmptyLines` locally on a Windows checkout) was diagnosed and fixed rather than argued with.

## CodeRabbit finding investigated and not carried forward

A second CodeRabbit pass flagged (severity: major) `tracker_reload_spec.rb:169-170` — the `resume_filter << true` / `wait_until_parked(ingest)` pair in the new `disable!` race spec — on the theory that `Thread#stop?` could read stale ("still sleeping on `resume_filter.pop`") immediately after the push, letting `wait_until_parked` return before `ingest` actually reaches `@reload_lock`. That's a real, general hazard with polling `Thread#stop?` right after signaling a queue. But it doesn't hold up for *this* test: `disable!` sets `@enabled = false` unconditionally, several lines before it ever touches `@reload_lock` (`tracker.rb:158`, well before `:164`). So no matter which side of the lock `ingest` ends up contending on, `process`'s re-check of `enabled?` inside the lock (`tracker.rb:269`) sees `false` either way — the property the test asserts (`parsed == 0`, `first.stats[:queued] == 0`) is invariant to the exact interleaving CodeRabbit is worried about. I stress-ran this example 40 times in isolation; 0 failures. Not carried forward as a finding.

## Coverage notes

- Second `git worktree` on the updated PR head (`c786b4b4`), same verified merge base (`d876267`) as the original review.
- Ran the full `spec/lib/gemstone/combat` suite: 453 examples, 0 failures (up from 445 at the last review — 8 new examples net, consistent with the diff).
- Ran `rubocop` across `lib/gemstone/combat` and `spec/lib/gemstone/combat`: 40 files inspected, no offenses.
- Checked live CI on the PR (`gh pr checks`): rubocop, both test jobs, and title validation all pass.
- Ran a second CodeRabbit automated pass against the same merge base: 1 finding, investigated and not carried forward (see above); nothing re-flagged on the structural fixes from the prior review.
- Independently reproduced the two most structurally significant fixes (pattern_gate.rb reload exclusion, detector invalidation) with standalone Ruby scripts driving the real def modules, not just by reading the diff.
- Did not re-test against a live Lich/GS4 session (same limitation as the original review — no live game connection available here).

PR #1608, opened 2026-09-12
feat(gs): message definitions from the player supplement file

Nisugi:
> **#1607 has merged.** This branch is rebased onto it, so the commit list is now this PR's own work only: the message compiler, match-time timeout isolation, the message table's assembly marker, literal-value freezing, and reload/subscription ordering.

## Problem

#1586 moves the non-combat message lines a hunting script reacts to (a disarm, a curse, a charge counter, a spell mark) out of private script regexes and into `Combat::Messages` families that scripts subscribe to as events. Those defs are shipped Ruby, so a new message — the outcome of an item operation, a charge count changing, an effect becoming active — still waits on a release. EOHunter's item-preparation flow needs exactly those, and this is the declarative recognition step for it. Lich recognises the line; the script reads the fact and decides. Nothing in the YAML sends commands.

## This PR

Adds a `messages:` section to the supplement file from #1607 (`DATA_DIR/combat/defs.yaml`), compiled into ordinary `MessageDef`s and spliced into the families before their gates derive.

```yaml
messages:
  - family: marks                  # shipped family, shipped event: one more line for it
    event: swift_justice
    patterns: ['Your Swift Justice charges are restored to (?<n>\d+)\.']
    captures:
      charges: { from: n, type: integer }     # the contract for swift_justice wants charges as integer

  - family: user_item_prep         # new family and event: user_ prefix required
    event: user_feed_result
    patterns: ['You feed the (?<what>.+?) and it hums softly\.']
    captures: { what: what }       # shorthand for { from: what, type: string }
    values:   { ok: true }         # a literal fact
```

The `data` block of a `MessageDef` (a Ruby lambda in the shipped file) is generated from two declarative blocks: **`captures`** maps payload keys to named captures with a type (`string`, `integer`, `symbol`); **`values`** supplies literal scalars. Expressions, callbacks and embedded Ruby are excluded. The key `raw` stays Lich's and cannot be set.

### Rules (each pinned by a spec example)

- **A payload that cannot be built emits nothing.** A capture that did not participate in the match, or an integer that does not parse, makes the def emit nothing and reports once per def — never a zero or empty string passed off as a fact. `Definitions::Messages.scan` now skips a `nil` payload for this (no shipped data block returns nil; previously nil became `{}`).
- **Shipped events carry a payload contract.** New `Messages::CONTRACTS` registers every shipped event with its family and the payload keys/types its consumers rely on. A supplement reusing a shipped event must provide those keys with matching types (nil accepted; a string literal converts where the contract asks for a symbol), and **cannot move the event to another family**. A spec asserts `CONTRACTS` covers exactly the shipped events with the right families.
- **New events and new families need the `user_` prefix**, so a future shipped name can never collide with a player's file on upgrade. A new `user_` event inside a shipped family is allowed.
- **Merge before deriving.** `Messages.family` appends the supplement rows for its name before `PatternGate.build`. Families the file declares that are not shipped become `USER_FAMILIES`. `FAMILIES`, `BY_NAME`, `FAMILY_OF` and `EVENTS` bind as one frozen `MessageTable` in a single assignment (same rule as `Definitions::Table`).
- **The runtime never scans a stale family.** `Combat::Messages` reads the table per call and holds active families as **names**, resolved against the current table.
- **Reload contract.** `Supplements.reload_defs!` (and so `Tracker.reload_defs!`) finishes with `Messages.refresh!` — a script may subscribe to an event before the file defines it — and then emits **`:definitions_reloaded`** `{ files:, supplements: }` through `Combat::Observers`. Named subscriptions are idempotent, so a consumer re-registering on that event never stacks handlers.

### Files

- `lib/gemstone/combat/defs/messages.rb` — `CONTRACTS`; `family()` splices supplements; `SHIPPED_FAMILIES` / `USER_FAMILIES` / `MessageTable` / `TABLE`; `scan` skips nil payloads.
- `lib/gemstone/combat/messages.rb` — reads the table per call; active families held as names; `refresh!` documented as the post-reload step.
- `lib/gemstone/combat/defs/supplements.rb` — the `messages` builder: placement, captures, values, reserved keys, contract validation, the generated payload proc with once-per-def reporting; `reload_defs!` calls `refresh!` and emits `definitions_reloaded`; per-kind builders return empty when their def module is not loaded.
- `lib/gemstone/combat/defs/supplements.example.yaml` — a `messages:` section with the two cases above.
- `spec/lib/gemstone/combat/supplements_messages_spec.rb` — 24 examples: contract registry coverage, compile paths (typed capture, shorthand, literals, symbol conversion), every rejection, emit-nothing on bad payload (reported once), splice into a shipped family, subscribe-before-defined activation with `definitions_reloaded`, deactivation on file removal, and the example file.

## Verification

- `spec/lib/gemstone/combat`: **478 examples, 0 failures** (the #1586 `message_defs_spec` / `messages_spec` pass unchanged).
- RuboCop: no offenses across the 41 files in `lib/gemstone/combat` and `spec/lib/gemstone/combat`.

## Not in scope

Authoritative item/player state from a single event, distinguishing unknown from inactive, proving which command caused an event, or excluding stale queued events from a confirmation. Those stay with observation provenance and the consuming script. This PR supplies only the declarative recognition step.


therealatari (commented):
Reviewed at `dde645135970871017bf899267af20ab8d81c465`. The message contracts, declarative payloads, and reload notification are a useful boundary: Lich recognizes facts, and EOHunter chooses actions. The two earlier #1607 findings are fixed in this stack. Local verification: **441 examples, 0 failures** across combat + DR CustomSubstitutions + Common UserDefs (seed 40379); RuboCop passes on the five Ruby files checked in the update.

## P2: Isolate user-pattern timeouts so unrelated facts survive

[Definitions::Messages.scan](https://github.com/elanthia-online/lich-5/blob/dde645135970871017bf899267af20ab8d81c465/lib/gemstone/combat/defs/messages.rb#L198-L213) lets `Regexp::TimeoutError` escape both family gating and individual matching. [Combat::Messages.process](https://github.com/elanthia-online/lich-5/blob/dde645135970871017bf899267af20ab8d81c465/lib/gemstone/combat/messages.rb#L90-L96) only emits after the complete scan returns. A timed-out supplemental pattern therefore discards even valid shipped facts already accumulated from that line.

Deterministic error-path reproduction: scan a shipped itchy-curse line and obtain `[:itchy_curse]`; append a supplemental family with a matcher stub that raises `Regexp::TimeoutError`, and the entire scan instead raises. This is an injected timeout test, not a claim that we observed this during live play. The worker's outer rescue keeps the thread alive, but loses the facts and logs on every recurrence.

### Proposed fix

- Handle `Regexp::TimeoutError` at the individual regex evaluation boundary, skip that pattern for this input, and continue scanning the other definitions/families. Preserve facts found before the error and permit valid matches afterward.
- Cover **both** the full match and `PatternGate.rejects?`'s `always_scan` fallback (`pattern_gate.rb:111`). Handling only `d.pattern.match` leaves the gate path exposed. Treating a whole family as rejected after a gate timeout would also hide its valid definitions.
- Reuse Common::UserDefs' bounded-pattern/once-only reporting conventions rather than a second timeout-reporting system. An opt-in safe matcher/reporting callback for the gate is one option that can preserve its existing callers; another is to conservatively admit an uncertain gate result into a individually guarded full scan. The essential contract is that one timed-out regex does not suppress unrelated facts. This is a proposed implementation approach, not an attached code patch.
- Deduplicate diagnostics across gate/full-scan evaluation for the same definition and reset that diagnostic state on reload. Include enough source information to identify the offending definition. Do not broadly rescue unrelated programmer errors as non-matches.

Suggested regressions: (1) valid shipped event before a timed-out supplemental pattern survives; (2) a valid event after it is also delivered; (3) an ungated pattern timing out does not prevent another definition in the same family from matching; (4) repeat input does not flood diagnostics, and reload resets reporting. Keep worker-liveness coverage separately: surviving the exception is not the same as preserving the observations.

## Consumer-side suggestions for “The Word Side” — not requirements for this Lich PR

The proposed arm-before-send primitive and one generic `prepare <name>` action fit well. A few refinements would make the EOHunter contract clearer without adding item-specific machinery:

1. **Separate correlation from acceptance.** A waiter filter selects the relevant response (for example `match: { what: crystal }`); `expect: { ok: true }` classifies that selected response. A matching `{ ok: false }` should return `:denied`, while a response for another item should be ignored. Using the expectation itself as the waiter matcher silently turns explicit denials into timeouts. Neither filter proves command causation; retain the documented scanner-lag limitation.
2. **Check definition availability, not merely the `user_` prefix.** Validate shape at profile load, then check the current known event names before a send. A missing def should produce an actionable diagnostic without spending the item/resource. Recheck availability after reload; an event removed while a command is already awaiting confirmation can retain the documented timeout behavior.
3. **Unknown observation does not imply safe repetition.** Feeding an item may cost health or charges, and some operations toggle state. The profile's explicit prep schedule/modifiers can authorize the first attempt, but an unconfirmed send should not be automatically treated as permission to repeat it. Distinguish confirmed denial from unknown outcome after timeout. Use existing health/resource preconditions and make retries explicit for repeat-safe operations.
4. **Keep the wait lifecycle bounded.** Arm before send, cancel in ensure on every exit, preserve send-ladder failures, and check interruption/death while waiting. Specify whether timeout is confirmation-only or total action duration; use a monotonic deadline. The latch remains last-observed data, not authoritative item state, and should not silently upgrade an unconfirmed send to success.

These consumer suggestions do not ask #1608 to implement item state, EOHunter actions, or additional command authority.

Nisugi:
Rebased onto the current #1607, which is now out of draft. This branch carries #1607's commits until it merges, then rebases down to its own single commit.

Still a draft for the open P2 from the last review: a `Regexp::TimeoutError` out of one supplemental pattern escapes the whole scan, so a single bad user regex discards the valid shipped facts already found on that line. Not yet implemented.

Also picked up here, from #1607: `Supplements.loaded` gained a `messages:` line, so it lists supplemental events grouped by family the way attacks group by slot.

Nisugi:
The timeout-isolation finding is fixed in `3e5aea0e`.

**Per-definition guard.** `Definitions::Messages.scan` now rescues `Regexp::TimeoutError` around each `d.pattern.match`, skips that pattern for the line and keeps scanning. Facts found before it survive and definitions after it still match.

**Gate path covered separately.** A timeout inside `PatternGate.rejects?` is treated as undecided rather than as a rejection, so the family is scanned instead of having all its definitions hidden. The full scan re-evaluates the same pattern under the per-def guard, which is where the report comes from, so a gate timeout is not silently swallowed and is not reported twice. `always_scan` holds the same Regexp objects as the defs, so this is the same pattern in both places.

**No second reporting system.** Reports route through `Common::UserDefs#report_timeout` by way of a small public wrapper on `Supplements`. That reuses the existing dedup set keyed on pattern source, shares it with the compile-time reports, and is cleared by `reset!` on reload. Nothing broadly rescues programmer errors; only `Regexp::TimeoutError` is caught.

**The four regressions you asked for**, in `supplements_messages_spec.rb`:

1. A shipped fact found before the timing-out pattern survives.
2. A shipped fact after it is still delivered.
3. An ungated pattern timing out in the gate does not prevent another definition in the same family from matching.
4. Repeat input reports once per pattern source, and a reload resets that so it reports again.

Each fails against the previous code with the uncaught `Regexp::TimeoutError`, and the suite is 462 examples, 0 failures with RuboCop clean.

Also on this branch since your review: `Supplements.loaded` names what the file defined per kind, attacks grouped by slot and messages by family, because a slot or family typo reads identically in `summary`'s counts.

Still a draft until #1607 merges, then this rebases to its own single commit.

Nisugi:
All three reproduced. Fixed across both PRs, since two of them live in #1607's files: `39902a9c` there, `b1ab1fb4` here.

**1 — the HMR path leaving subscriptions inactive.** Confirmed exactly: after a plain `load` the event was in the table, `active_families` was empty, the hook was down and `process` emitted nothing.

Of your two options I took neither literally. `;hmr` is a separate script this repo does not own, so it cannot be routed through the lifecycle from here; documenting only `reload_defs!` would leave the trap in place for anyone who uses the command the comments already mention. Instead `defs/messages.rb` refreshes subscriptions where it binds its table. Any route that rebinds is then correct, not just the one entry point, and `refresh!` is idempotent so the `reload_defs!` path is unaffected. The guard is for load order, since this file is also required before `Combat::Messages` exists.

**2 — mutable literals leaking.** Confirmed. `values.freeze` froze the hash, not the strings in it, and `values.dup` is shallow, so one String was shared by every payload the def emitted. Literals are now deep-frozen at compile time, so a consumer mutating one raises on its own line rather than silently rewriting the next fact. Captures are deliberately left mutable: each match builds its own strings, and freezing those would break consumers who reasonably expect to work on a capture.

**3 — stale lazy detectors.** Confirmed, and this one was mine from the previous round. The memo is a module ivar and survives the def file re-executing, so a reload rebound `TABLE` while the detector still answered for the old one. The cached `nil` was the worse half: once one entry made the union impossible the detector stayed nil even after that entry was removed. Dropped where the table rebinds. Fixed in #1607 with the three regressions, since `attacks.rb`/`statuses.rb` belong to that PR.

Five of the six new examples fail against the previous code. Suites are 464 on #1607 and 488 here, RuboCop clean on both.

mrhoribu (changes requested):
# PR Review: feat(gs): message definitions from the player supplement file (elanthia-online/lich-5#1608)

**Reviewed:** 6 files changed by this PR's own contribution (+819/-20), full clone at commit `dd191fb7` on 2026-09-14. Stacked on open PR #1607 (base commit `bad57cbc`); this review scopes to the PR's own incremental work as its description requests, with full read access to the combined codebase (#1607 + #1608) for context.
**Verdict:** Request changes
**Summary:** Adds a `messages:` section to the player supplement YAML that lets non-combat message events (charge counters, item-operation results, effect toggles) be extended without a Lich release, compiled into the same `MessageDef`/`Family` structures the shipped defs use, validated against a payload contract registry, and wired into the hot-reload path. The design is careful and the implementation matches it faithfully — I found no correctness, contract-breakage, or concurrency defects, only one documentation gap worth closing.

## Minor

### 1. No path from `captures:` to a shipped boolean contract key, and nothing tells the author why — `lib/gemstone/combat/defs/supplements.rb:82`, `lib/gemstone/combat/defs/messages.rb:67-70`

```ruby
CAPTURE_TYPES = %w[string integer symbol].freeze
```

Three shipped events — `haze_703`, `rebuke_1614`, `arcane_reflex` — have a `:boolean` contract key (`messages.rb:67,68,70`). `CAPTURE_TYPES` has no `boolean` entry, so `captures:` can never produce one; a boolean fact can only come from a literal under `values:`, which means satisfying one of these contracts requires writing two separate `messages:` entries (one pattern/one literal per state), exactly the pattern `supplements.example.yaml` uses for the new `user_feed_result` event's `ok: true`/`ok: false` — but that pair isn't a shipped-contract example, so it doesn't actually demonstrate the technique a supplement author needs for `haze_703` et al.

An author who tries the natural thing — `captures: { on: { from: state, type: boolean } }` — gets: `captures.on: type must be one of string, integer, symbol, got "boolean"` (`supplements.rb:529`), with no pointer toward the two-entries-with-literals workaround. It's not a functional bug (the mechanism works once you know the trick, and the rejection is safe — the entry is just dropped with a report, per the module's usual contract), but it's a real foot-gun for exactly the three shipped events most likely to be extended by a hunting script author (marks/haze/rebuke tracking is precisely the EOHunter use case the PR is building toward). A short comment on `CAPTURE_TYPES` or one more worked example in `supplements.example.yaml` showing a boolean contract key satisfied via two literal-valued entries would close this.

## What this PR gets right

- **The reload-consistency discipline is genuinely careful and it shows.** `FAMILIES`/`BY_NAME`/`FAMILY_OF`/`EVENTS`/`TABLE` bind in one final assignment (`messages.rb:172-185`), `Combat::Messages` reads `Definitions::Messages.table` once per call and resolves `@active` by name rather than holding `Family` objects (`messages.rb:50,60` in `combat/messages.rb`), and `Supplements.assembled!(:messages)` is called at the correct point — after `TABLE` binds, not after the document merely reparses. I traced the reload sequence (`Supplements.reload_defs!` → `load()` each def file → `Messages.const_defined?(:CONTRACTS)` visible immediately because `CONTRACTS` is assigned before `SHIPPED_FAMILIES` in the same module body → `notify_reloaded` → `Combat::Messages.refresh!` → `Observers.emit(:definitions_reloaded, ...)`) end to end and found no window where a reader can observe a torn table or a stale `@active` family list that doesn't self-correct.
- **`values.dup` inside `payload_proc` is exactly right** (`supplements.rb:636`): `values` is frozen once at proc-build time, and each invocation dups it fresh, so concurrent calls into the same def's closure (from the worker thread and from a script calling `Combat::Messages.scan` directly, per its own doc comment) never share mutable state.
- **The nil-payload contract is enforced uniformly.** A capture that didn't participate or an integer that doesn't parse returns `nil` from the data proc, and `Messages.scan` checks for that (`messages.rb:212-217`) before merging `raw:` — no path produces a fabricated zero/empty-string fact, and it's reported once per def via a properly locked dedup set (`report_payload_once`, `supplements.rb:661-670`).
- **Test coverage is thorough and pins the subtle cases**, not just the happy path: nil-payload-emits-nothing (and only reports once), subscribe-before-the-event-exists then reload, family removal on file deletion, a supplemental pattern timing out before/after a shipped def in the same family and during the gate decision itself, and the shipped example file loading with zero rejections. 105 examples for this increment, 0 failures; 466 across the whole `spec/lib/gemstone/combat` tree, 0 failures.
- **The commit history shows the hard bugs were already hunted down before this reached review** — the `memoize` stale-publish race, `Tracker.reload_defs!`'s `NoMethodError` from being declared below `private`, the detector-union `RegexpError` that could take the whole file down, the async-worker swap losing an in-flight chunk, the case-folded-literal gate rejecting lines its own pattern would match, and (in this increment specifically) a `Regexp::TimeoutError` escaping the family scan and discarding facts already found on the line. Each has a spec reproducing the original failure mode.

## Coverage notes

- **Access level:** full local clone (`git worktree` at `dd191fb7`), diffed against this PR's actual base (`bad57cbc`, the tip of #1607 at the time #1608 branched) rather than against `main`, since #1607 is itself still open and unreviewed separately — reviewing the combined diff against `main` would have re-litigated #1607's ~1,800 lines under this PR's number. Full file contents were read for `defs/messages.rb`, `defs/supplements.rb`, `combat/messages.rb`, `defs/pattern_gate.rb`, `common/user_defs.rb`, and `combat/observers.rb` (not just the diff hunks), plus a repo-wide grep for external callers of the renamed/changed `Messages` constants and methods — none exist outside this module and its specs.
- **Tests run:** `bundle exec rspec spec/lib/gemstone/combat` — 466 examples, 0 failures. `bundle exec rspec spec/lib/gemstone/combat/supplements_messages_spec.rb spec/lib/gemstone/combat/supplements_spec.rb spec/lib/gemstone/combat/messages_spec.rb spec/lib/gemstone/combat/pattern_gate_spec.rb spec/lib/gemstone/combat/supplements_wiring_spec.rb spec/lib/gemstone/combat/tracker_reload_spec.rb spec/lib/gemstone/combat/message_defs_spec.rb` individually — 117 examples, 0 failures.
- **Linter:** `rubocop` (Ruby 4.0.5, matching this repo's pinned target) on the five changed/added files — 0 offenses.
- **CodeRabbit:** attempted (`coderabbit review --agent --committed --base-commit bad57cbc...`), but the run failed with a rate-limit error — the org's 3 included reviews for this billing period are used and the account isn't linked to a seat for usage-based reviews (54-minute retry window quoted). No CodeRabbit findings were available to cross-check; this review is manual-only.
- **Not independently verified:** live behavior against the real game stream (this PR only adds a declarative recognition layer with no side effects, so the risk surface for this is low, but I did not run Lich against a live or recorded session).
- **Out of scope for this review:** everything already present at the #1607 base commit (`pattern_gate.rb`'s `safe_match`/`union_or_nil`, `tracker.rb`'s reload-lock/worker-swap logic, the four combat-kind supplement builders) — these are #1607's own contribution and belong to its review, not this one. I read them for context but did not re-audit them.

Nisugi:
Rebased on #1607 at `2ff9f703` and one more fix here in `3d2d4230`, from a full read of the stack.

**Subscribers ran under the ingestion lock.** `Tracker.reload_defs!` held its lock through `Supplements.reload_defs!`, which ends by emitting `:definitions_reloaded` synchronously. That put arbitrary script code under the lock the game-stream hook thread parks on: a slow handler stalled ingestion for its duration, and a handler calling back into `reload_defs!` hit recursive locking. `Supplements.reload_defs!` now takes `notify:`; the tracker passes false and emits after the swap is released. Every other caller keeps the default, so the login path and direct callers are unchanged.

The regression subscribes a probe that records whether the lock is held when it runs, and fails against the previous code. 492 examples, 0 failures, RuboCop clean.

Nisugi:
Thanks for the careful read. The one finding is closed in `9dcf3f06`, and the branch is rebased on #1607 at `c786b4b4`, which picks up that PR's review round.

**Boolean contract keys.** I kept `CAPTURE_TYPES` as it is rather than adding a boolean, on purpose: game text never carries a literal true/false to capture, it carries one line for the on state and another for the off, so the two-entries-with-literals form is the honest shape, not a workaround. What was missing was anyone saying so. Now:

- The comment on `CAPTURE_TYPES` explains why boolean is absent and names the three shipped contracts that want one.
- A `captures:` entry typed `boolean` is rejected with the technique spelled out, pointing at the example file.
- `supplements.example.yaml` extends `haze_703` with an on/off pair on a shipped contract, which is the case an EOHunter author actually hits.

Writing that example turned up a trap worth its own comment in the file: `on`, `off`, `yes` and `no` are YAML 1.1 booleans, so a bare `on:` key parses as `true` and is rejected as a non-identifier. The example quotes it and says why. The shipped-example spec now asserts five message entries load with no rejections.

494 examples, 0 failures; whole-repo rubocop clean.

mrhoribu (approved):
# PR Re-review: feat(gs): message definitions from the player supplement file (elanthia-online/lich-5#1608)

**Reviewed:** re-review after the author pushed fixes for three findings from another reviewer and one from my [prior review](https://github.com/elanthia-online/lich-5/pull/1608#pullrequestreview-5202827122). This PR's own contribution is now +974/-22 across 8 files, at commit `9dcf3f06` on 2026-09-15, rebased onto #1607 at `c786b4b4`. Full clone; diffed against the PR's actual current base (the tip of #1607 as of this push) for the same reason as the first pass — reviewing against `main` would re-litigate #1607's own ~1,900 lines under this PR's number.
**Verdict:** Approve
**Summary:** All four fixes since the last round are correctly targeted, address the described failure modes without introducing new ones, and each ships with a regression test that fails against the prior code. I traced all four by hand (module load order, lock scope, object identity/freeze semantics) and found nothing to add.

## Verification of the four fixes

### 1. `;hmr`-triggered reload left an active family's hook down (found by another reviewer)

`defs/messages.rb:99-111` now calls `Lich::Gemstone::Combat::Messages.refresh!` directly, right after `TABLE` binds, guarded by `Lich::Gemstone::Combat.const_defined?(:Messages, false)`. I traced both load paths by hand:

- **Initial boot:** `combat/messages.rb` does `require_relative 'defs/messages'` (`messages.rb:29`) *before* its own `module Messages` declaration (`messages.rb:32`). So when `defs/messages.rb`'s body reaches the guard, `Combat::Messages` doesn't exist yet — `const_defined?` is `false`, the call is skipped, no `NameError`. Correct.
- **`;hmr combat/defs/` (a plain `load` of just this file, not a `require`):** by this point `Combat::Messages` is long since loaded and never itself gets `load`ed by `;hmr` (it lives outside `combat/defs/`), so the guard is `true` and `refresh!` fires, picking up any family that gained or lost a subscriber-relevant event. Matches the new spec `'activates the family when the def file is merely re-loaded, as ;hmr does'` (`supplements_messages_spec.rb:1074`), which reproduces the exact mechanism (`load File.join(LIB_DIR, ...)`) rather than going through `Supplements.reload_defs!`.

`refresh!` now fires up to three times on a normal `Supplements.reload_defs!` (once inline here, again from `notify_reloaded`, redundantly with #2 below) — harmless since it's documented idempotent and cheap (array select + mutex), just a bit of duplicate work. Not worth trimming.

### 2. `:definitions_reloaded` subscribers ran under the ingestion lock (found by another reviewer)

`Tracker.reload_defs!` (`tracker.rb:369-386`) now calls `Supplements.reload_defs!(notify: false)` inside `@reload_lock.synchronize`, then calls `Supplements.notify_reloaded(reloaded)` — which is what actually calls `Observers.emit(:definitions_reloaded, ...)`, running arbitrary subscriber code — *after* the block ends. `reloaded` is pre-declared (`reloaded = nil`, `tracker.rb:368`, unchanged by this PR) before the block, so there's no scoping hazard in referencing it afterward. `Combat::Messages.refresh!` itself (called both inline from #1 and again inside `notify_reloaded`) still runs under `@reload_lock` in both places, but that's fine — it's bounded engine-internal work synchronized on its own separate mutex (`Combat::Messages`'s `@mutex`), never a lock a subscriber could reach back into. The new spec `'notifies definitions_reloaded subscribers after releasing the ingestion lock'` (`tracker_reload_spec.rb:195`) directly probes `lock.locked?` from inside a subscriber and asserts `false`. I checked it isn't asserting a tautology by re-reading `@reload_lock`'s only other use in the file — it's the same lock guarding the drain-and-swap, so the probe is meaningful.

### 3. Shared literal `String` in a payload could be mutated by a subscriber, corrupting every later fact (found by another reviewer)

`payload_proc` (`supplements.rb:597-630`) now does `values.transform_values { |v| v.is_a?(String) ? v.dup.freeze : v }.freeze` at compile time instead of only `values.freeze`. Since `values.dup` in the lambda body is a shallow copy, the old code left every payload sharing the *same* String object for a literal like `values: { state: ready }` — a subscriber doing `payload[:state] << 'x'` would silently rewrite the fact for every future match of that def. Now that string is frozen, so the same mutation raises `FrozenError` on the offending line instead. Captured (non-literal) strings are deliberately left mutable — each match produces a fresh string from `MatchData`, so there's no shared-object risk there, and freezing them would surprise a consumer with no upside. Two new specs cover this precisely: mutation now raises and a later independent match still reports the original value (`supplements_messages_spec.rb:860`), and a consumer adding a new key to one payload doesn't leak into the next (`:881`) — confirming `values.dup`'s hash-level isolation still holds, only the string-level sharing was the gap.

### 4. My finding: boolean contract keys had no documented path and an unhelpful rejection

Closed by `9dcf3f06`. `CAPTURE_TYPES`'s comment now explains why boolean is absent (game text has no literal true/false to capture, only an on-line/off-line pair) and names the three shipped contracts that want one; the `captures.<key>: type must be one of ...` rejection now appends a hint pointing at the two-entries-with-literals technique when the requested type is `boolean` (`supplements.rb:488-496`); and `supplements.example.yaml` gained a worked `haze_703` on/off pair against a real shipped boolean contract (not just the generic `user_item_prep` example, which didn't actually exercise a shipped contract). Writing that example surfaced a genuine YAML 1.1 gotcha worth catching on its own — `on`, `off`, `yes`, `no` are implicit booleans in Psych's default (YAML 1.1 core) schema, so an unquoted `on:` key parses as the Ruby value `true`, not the string `"on"`. I confirmed this is real (not just a defensive comment): `validate_payload_key` requires `key.is_a?(String)`, so an unquoted `on:` would be rejected with a slightly confusing `key true must be a lowercase identifier` rather than silently doing the wrong thing — safe either way, but the example correctly quotes `'on':` and the shipped-example spec (`supplements_messages_spec.rb:1191-1195`, now asserting 5 messages with zero rejections) would catch a regression if the quoting were ever dropped.

## What this PR continues to get right

- Every one of the four fixes ships with a spec that fails against the pre-fix code, not just a passing assertion — I didn't have to take that on faith, the commit messages state it and the mechanisms (e.g. the lock-probe spec, the `load()`-based `;hmr` simulation) are actually testing the described failure, not a proxy for it.
- The author's replies name exact commits per finding and are precise about what changed and why, including admitting which fix belongs to the sibling PR (checked: `DEF_FILE_PATTERN`'s `pattern_gate.rb` exclusion — the "stale lazy detectors" fix — is present in this worktree via the #1607 rebase but does not appear in this PR's own diff against its current base, confirming it correctly belongs to #1607 and was not silently smuggled in here).

## Coverage notes

- **Access level:** full local clone (`git worktree` at `9dcf3f06`), diffed against this PR's current base (`c786b4b4`, current tip of #1607). Reread `tracker.rb`, `messages.rb` (both the `defs/` and `combat/` ones), `supplements.rb`, and `observers.rb` in full at head, not just the new diff hunks, to check lock scope and load order by hand rather than by inspection of the diff alone.
- **Tests:** `bundle exec rspec spec/lib/gemstone/combat` — 478 examples, 0 failures (up from 466 last round).
- **Linter:** `rubocop` (Ruby 4.0.5) on all seven files touched by this increment — 0 offenses.
- **CodeRabbit:** the prior rate limit cleared and this run got past setup and into analysis, but its WebSocket connection dropped ("subscription completed unexpectedly") after several heartbeats with no findings emitted. Two attempts, two different failures (rate limit, then a dropped connection); this review is manual-only, same as the first round.
- **Out of scope:** #1607's own contribution (already rebased in), including the `DEF_FILE_PATTERN`/`pattern_gate.rb` fix and the `attacks.rb`/`statuses.rb` lazy-detector fix referenced in the author's replies — read for context, not re-audited here.

PR #1610, opened 2026-09-13
ci(all): format-body harvests the last #NNN reference, not the first

mrhoribu:
## Summary

Follow-up to #1609, which was a wrong fix for the right symptom (confirmed by [run 34727595165](https://github.com/elanthia-online/lich-5/actions/runs/34727595165) still 404ing on `pulls/1596` after that fix merged).

#1609 assumed release-please distinguishes the merging PR from a referenced issue by URL shape (`/pull/<N>` vs `/issues/<N>`). It doesn't — RP's changelog writer renders **every** `#NNN` it finds in the merge commit subject as an `/issues/<N>` link, with no distinction. Confirmed by pulling the actual squash-merge commit for #1599:

```
fix(all): setup_files uses content hash for cache freshness (#1596) (#1599)
```

The PR's own title referenced the issue it closes (`(#1596)`, typed by the author), and GitHub's squash-merge appended the real merging PR number last (`(#1599)`). RP renders both as `/issues/` links in that same left-to-right order:

```
* **all:** setup_files uses content hash for cache freshness ([#1596](.../issues/1596)) ([#1599](.../issues/1599)) ([4968e45](.../commit/...))
```

So #1609's `/pull/` check never matched anything and silently fell back to the old first-match behavior — still grabbing 1596, still 404ing on `pulls.get()`, still dropping the bullet.

## Change

- Harvest step now takes the **last** `#NNN` match per bullet line instead of the first. GitHub always appends the real merging PR number last when squash-merging, so this is the reliable signal (not URL shape). The commit-sha link (`[abc1234](.../commit/...)`) has no leading `#`, so it's never a false match either way.
- Bumped `CONTRACT_SHA` again so the currently open, already-sentineled stable Release Please PR gets reformatted with this fix instead of short-circuiting on #1609's sentinel.

## Test plan

- [x] Verified against the actual squash-merge commit subject for #1599 (extracts `1599`, not `1596`).
- [x] Verified a normal single-reference bullet is unaffected (extracts `1603`).
- [x] Verified a bullet with a reference + trailing commit-sha link still extracts the reference number, not something from the sha link (sha link has no `#`).
- [ ] Confirm the next stable RP PR reformat includes the `setup_files uses content hash for cache freshness` bullet.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved pull request reference detection in automatically formatted content, ensuring the correct PR number is retained when multiple references appear on a line.
  - Existing open pull requests that were previously formatted may now be refreshed to reflect the corrected reference handling.

### DIFF
--- a/.github/actions/format-body/action.yaml
+++ b/.github/actions/format-body/action.yaml
@@ -57,10 +57,11 @@ runs:
       uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
       env:
         MODE: ${{ inputs.mode }}
-        # Bumped for the pull-link-vs-issue-link harvest fix below; this invalidates the
-        # sentinel on any already-formatted, still-open RP PR so it gets reformatted with
-        # the fix instead of being skipped by the old, matching sentinel.
-        CONTRACT_SHA: "2aa1c6bdb379fa50f77ca249ac4a19f7eef93e606f77d61dcbf1fc14107891a5"
+        # Bumped again: the prior fix assumed the real PR used a "/pull/" link, but RP renders
+        # every reference (issue or PR) as "/issues/"; taking the LAST "#NNN" match is the
+        # correct signal. Bump invalidates the sentinel on any already-formatted, still-open
+        # RP PR so it gets reformatted with the fix instead of being skipped.
+        CONTRACT_SHA: "699e8e7e66d93612c64757d672ffcca2db94e9c051060144e7c1a1d3fc29d2f6"
         BREAK_AUTO: ${{ inputs.break_autolinks_in_bullets }}
         PR_NUMBER: ${{ inputs.pr_number }}
       with:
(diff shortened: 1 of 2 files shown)


mrhoribu:
Addressed the nitpick from CodeRabbit's review (missing executable coverage for the reference-selection logic): added [.github/tests/test-format-body-harvest.sh](https://github.com/elanthia-online/lich-5/blob/fda7e3168cb7f7e91d34d81ade4c85566d9d29d2/.github/tests/test-format-body-harvest.sh) in fda7e31, covering the squash-merge (issue + PR refs) case, a normal single-reference bullet, and a reference-plus-trailing-commit-SHA bullet — following the same standalone-mirroring approach `test-union-merge.sh` already uses for embedded logic that can't be sourced directly. All local suites (`.github/tests/run-tests.sh`) pass.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #1612, opened 2026-09-13
fix(all): retain discovery ownership through cleanup

therealatari:
## Summary

Harden the ActiveSessions discovery lifecycle in three related parts:

- retain the ownership flock through discovery cleanup;
- retry transient Windows `EACCES` sharing violations during discovery publication, removal, and private-temp cleanup; and
- bound stale-endpoint connection attempts so failed discovery cannot delay owner election indefinitely.

This PR remains limited to discovery and handoff behavior. It adds no coordination endpoint, resource lease, Hunter policy, or game-command support.

## Why

`stop_service!` previously released ownership before checking and deleting discovery. A successor could acquire the lock and publish between the retiring process's ownership check and unlink, after which the old owner could delete the new live endpoint.

Windows adds a second handoff failure mode: Ruby can raise `Errno::EACCES` when `File.rename` replaces, or `File.delete` removes, a file another process temporarily has open. Publication and cleanup now use short bounded backoff. Discovery mutations confirm that the process still retains its local ownership handle before each attempt; the actual cross-process guarantee is the flock retained under the lifecycle mutex. Private temp cleanup is ownership-independent, best-effort, and cannot mask the original publication failure.

A stale endpoint could also block in `TCPSocket.new` before election reached the ownership lock. The client now gives loopback connection establishment its own explicit bounded timeout, separate from response reads.

## Architecture boundary

Discovery ownership identifies the process publishing the registry. It grants neither coordination authority, execution ownership, nor resource-lease ownership. The later architecture treats transport as an adapter seam, so this filesystem repair remains independent of observation, operation, and lease policy.

Companions:

- Read-only coordination and bounded loopback adapter: #1613
- Bounded coordinated operation lifecycle: #1618

## Known constraints

- Retry sleeps occur while holding the lifecycle mutex and are deliberately capped at about 310 ms total per operation.
- A process killed between temp-file creation and rename can leave a PID-scoped orphan; readers ignore it. Sweeping crash orphans is separate follow-up work.
- Windows cannot represent POSIX `0600` mode bits. Protection there depends on the coordination directory's OS ACLs; this PR does not claim a hostile-same-user security boundary.

## Verification

- Full standalone branch suite: **7,559 examples, 0 failures** (Ruby 4.0.5, Linux).
- Focused ActiveSessions unit and multiprocess handoff suite: **56 examples, 0 failures**.
- Repository-wide RuboCop: **1,172 files, no offenses**.
- Workflow YAML parsed successfully; `git diff --check` clean.
- Real-process coverage includes killed-owner takeover, competing successors, concurrent readers, death immediately before publication, cleanup/publication races, and deterministic read-side sharing violations.
- Deterministic failure coverage includes transient and exhausted publication/removal/temp cleanup, preserved primary errors, and local ownership-handle loss.
- Native Windows CI exercises real `File.rename`/`File.delete` sharing violations and lock handoff on `windows-latest`.

The fixtures use only temporary files and localhost synthetic services. They do not log in, issue game commands, or alter a running player's session.

### DIFF
--- a/lib/internal_api/active_sessions.rb
+++ b/lib/internal_api/active_sessions.rb
@@ -259,6 +259,10 @@ module Lich
           @server&.stop
           @server = nil
           @registry = nil
+          # Keep the native flock through the ownership check and unlink. A
+          # successor can publish as soon as we release it; checking the pid
+          # before an unlocked unlink does not protect that new publication.
+          delete_discovery_if_owned if own_lock?
           release_ownership_lock
         end
         @service_client_mutex.synchronize do
@@ -266,7 +270,6 @@ module Lich
           @service_client_token = nil
           @service_client_port = nil
         end
-        delete_discovery_if_owned
       end
 
       # Returns a client configured from the current discovery record.
@@ -532,9 +535,8 @@ module Lich
 
       # Deletes the discovery file only when it still belongs to the given owner.
       #
-      # Re-reads the file before deletion to avoid a race where another process
-      # has written a fresh discovery between the caller's initial read and this
-      # deletion attempt.
+      # The caller must retain the native ownership flock throughout this
+      # operation. The pid check alone cannot serialize publication and unlink.
       #
       # @param expected_owner_pid [Integer]
       # @return [void]
--- a/spec/lib/internal_api/active_sessions_handoff_spec.rb
+++ b/spec/lib/internal_api/active_sessions_handoff_spec.rb
@@ -1,0 +1,198 @@
+# frozen_string_literal: true
+
+require 'json'
+require 'rbconfig'
+require_relative '../../spec_helper'
+
+RSpec.describe 'Native ActiveSessions process handoff' do
+  # Test deadlines bound orchestration failures, not a production availability
+  # SLA. No child boots Lich/game state, and every PID comes from Process.spawn.
+  let(:child_deadline) { 5 }
+  let(:handoff_child_path) { File.expand_path('../../support/coordination_handoff_child.rb', __dir__) }
+
+  before do
+    @handoff_dir = Dir.mktmpdir('active-sessions-handoff')
+    @children = []
+  end
+
+  after do
+    @children.each do |child|
+      child[:input].close unless child[:input].closed?
+    end
+    @children.each do |child|
+      reap_child(child, force: true) unless child[:status]
+      child[:output].close unless child[:output].closed?
+      expect(child[:status]).to be_a(Process::Status)
+    end
+    FileUtils.remove_entry(@handoff_dir)
+  end
+
+  def start_child
+    child_input, parent_input = IO.pipe
+    parent_output, child_output = IO.pipe
+    child = { input: parent_input, output: parent_output, log: File.join(@handoff_dir, "child-#{@children.length}.log") }
+    child[:pid] = Process.spawn(RbConfig.ruby, handoff_child_path, @handoff_dir,
+                                in: child_input, out: child_output, err: child[:log], close_others: true)
+    @children << child
+    child_input.close
+    child_output.close
+    expect(receive(child, 'ready')['pid']).to eq(child[:pid])
+    child
+  end
+
+  def send_command(child, command)
+    child[:input].puts(command)
+    child[:input].flush
+  end
+
+  def receive(child, event)
+    unless IO.select([child[:output]], nil, nil, child_deadline)
+      raise "child #{child[:pid]} timed out waiting for #{event}: #{File.read(child[:log])}"
+    end
+    line = child[:output].gets
+    raise "child #{child[:pid]} exited waiting for #{event}: #{File.read(child[:log])}" unless line
+
+    payload = JSON.parse(line)
+    expect(payload.fetch('event')).to eq(event)
+    payload
+  end
+
+  def command(child, request, event = request)
+    send_command(child, request)
+    receive(child, event)
+  end
+
+  def reap_child(child, force: false)
+    deadline = Process.clock_gettime(Process::CLOCK_MONOTONIC) + child_deadline
+    loop do
+      result = Process.waitpid2(child[:pid], Process::WNOHANG)
+      return child[:status] = result.last if result
+
+      break if Process.clock_gettime(Process::CLOCK_MONOTONIC) >= deadline
+      sleep 0.01
+    end
+    raise "child #{child[:pid]} failed to exit" unless force
+
+    Process.kill('KILL', child[:pid])
+    child[:status] = Process.waitpid2(child[:pid]).last
+  end
+
+  def kill_child(child)
+    Process.kill('KILL', child[:pid])
+    expect(reap_child(child).termsig).to eq(Signal.list.fetch('KILL'))
+  end
+
+  def discovery
+    JSON.parse(File.read(File.join(@handoff_dir, 'lich-active-sessions.json')))
+  end
+
+  def expect_available(reader, owner)
+    sample = command(reader, 'sample')
+    expect(sample['snapshot']).not_to have_key('error')
+    expect(sample.dig('snapshot', 'source')).to eq('ActiveSessionsAPI')
+    expect(sample.dig('observation', 'record', 'owner_pid')).to eq(owner[:pid])
+    expect(sample['elapsed']).to be < child_deadline
+    sample
+  end
+
+  def expect_unavailable(reader)
+    sample = command(reader, 'sample')
+    expect(sample.dig('snapshot', 'error')).to be_a(String)
+    expect(sample.dig('snapshot', 'sessions')).to eq([])
+    expect(sample['owns_lock']).to be(false)
+    expect(sample['owns_server']).to be(false)
+    expect(sample['elapsed']).to be < child_deadline
+    sample
+  end
+
+  it 'recovers a killed published owner with two contenders and a concurrent discovery reader' do
+    owner, first, second, reader = Array.new(4) { start_child }
+    expect(command(owner, 'ensure', 'ensured')['available']).to be(true)
+    original = discovery
+    command(reader, 'watch', 'watching')
+    expect_available(reader, owner) # Cache original endpoint credentials in this reader.
+    kill_child(owner)
+    expect_unavailable(reader)
+    expect(discovery).to eq(original)
+
+    [first, second].each { |child| command(child, 'arm_publication', 'armed') }
+    [first, second].each { |child| send_command(child, 'ensure') }
+    # The native flock selects which contender reaches real File.rename.
+    ready = IO.select([first[:output], second[:output]], nil, nil, child_deadline)
+    expect(ready).not_to be_nil
+    initial = ready.first.first == first[:output] ? first : second
+    event = JSON.parse(initial[:output].gets)
+    winner, loser = event['event'] == 'publication_pending' ? [initial, ([first, second] - [initial]).first] : [([first, second] - [initial]).first, initial]
+    if initial == winner
+      expect(receive(loser, 'ensured')['available']).to be(false)
+    else
+      expect(event['event']).to eq('ensured')
+      expect(event['available']).to be(false)
+      receive(winner, 'publication_pending')
+    end
+    expect(command(loser, 'state').values_at('owns_lock', 'owns_server')).to eq([false, false])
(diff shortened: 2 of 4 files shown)


therealatari:
Windows acceptance is now included and green: [Exercise native Windows file handoff](https://github.com/elanthia-online/lich-5/actions/runs/34873905827/job/104076242289) completed **56 examples, 0 failures** on Ruby 4.0.5 / `windows-latest`.

The focused job uses real subprocesses and native Windows file handles. It verifies:

- nonblocking ownership-flock contention
- killed-owner takeover with competing successors and a concurrent discovery reader
- death immediately before publication, both with and without a stale complete record
- retiring-owner cleanup cannot unlink the successor record
- real `EACCES` recovery for `File.rename` while another process holds discovery open
- real `EACCES` recovery for `File.delete` while another process holds discovery open

The first Windows pass also exposed a separate handoff delay: `Client::READ_TIMEOUT` bounded response reads but not `TCPSocket.new`, so stale discovery could stall before lock election. The branch now uses `Socket.tcp(..., connect_timeout: READ_TIMEOUT)` and covers that contract directly. POSIX-only permission/signal assertions were made platform-aware; production file permissions were not weakened.

Nisugi (commented):
Reviewed the full head, including the Windows handoff job. The core change — deleting the discovery file while the flock is still held (inside `@service_mutex`, before `release_ownership_lock`), retrying `EACCES` on rename/delete, and proving it against a real second-process sharing violation on `windows-latest` — is sound and well-covered. Nothing below blocks merge.

**1. The `own_lock?` re-check inside `retry_discovery_filesystem_operation` is effectively unreachable in production.**
`own_lock?` is `!@lock_file.nil?` — a local flag, not a probe of the OS lock. The only thing that can flip it mid-retry is this same process calling `release_ownership_lock`, and both callers already run under `@service_mutex`, which `stop_service!` also takes. So the `IOError 'ownership lost'` branch is only hit by the spec that stubs `sleep` to release the lock. Not harmful, but the comment and the spec name ("after native ownership is lost") oversell it; the real cross-process guarantee is the flock being held until release. I'd either drop the re-check or reword what it guards against.

**2. `write_discovery`'s `ensure` isn't retried and can mask the real error.**
If every rename retry fails *and* the temp file is briefly held (same AV scenario), `File.delete(temp_path)` raises `EACCES` from the `ensure`, replacing the rename exception in the logged warning. Pre-existing, but this PR touches the line above it.

**3. Orphaned `.<pid>.tmp` files are now a tested invariant.**
The handoff spec asserts `File.exist?(temp_path)` after killing a publisher mid-rename. Correct for the crash, but nothing sweeps stale temp files from the coordination dir, so they accumulate. Fine to defer — probably worth a follow-up issue rather than silently enshrining it.

**4. Retries sleep while holding `@service_mutex`** — up to ~310 ms total. Fine at these delays; noting it so the table doesn't later get bumped to seconds.

**5. Client: `READ_TIMEOUT` now doubles as the connect timeout.**
`Socket.tcp(..., connect_timeout: READ_TIMEOUT)` is handled fine (`Errno::ETIMEDOUT` falls into the existing `rescue StandardError`, and the client only uses plain `IO` methods), but worst case per call is now 2 s rather than 1 s. A separate `CONNECT_TIMEOUT` constant would make that explicit.

**6. The `0o600` mode assertion is skipped on Windows** rather than replaced. Honest, since Windows can't represent it, but it means the token file's protection there rests entirely on the coordination dir being under a per-user path.

MahtraDR (approved):
**Reviewed at `f81e27e7` — Approve with comments.** No blockers, no majors.

Ran the specs locally (Ruby 4.0.6, macOS): `active_sessions_spec.rb`, `active_sessions/server_client_spec.rb`, `active_sessions_handoff_spec.rb` → 53 examples, 0 failures (the `Gem.win_platform?` block was skipped here; Windows CI is green). Also reproduced the retry helper in isolation to confirm its behavior.

### What this gets right
- **The core fix is correctly located and reasoned.** Against `main`, `stop_service!` released the ownership flock *before* the `load_discovery` pid-check + `File.delete`, a real TOCTOU: a successor could win the lock and publish in that gap, and the retiring owner would then delete the successor's live discovery file. Holding the flock across check+unlink closes it, and the pid recheck is retained as a second guard.
- **Ownership is revalidated before every retry attempt**, and losing the flock mid-backoff aborts with `IOError` rather than mutating a successor's generation. My standalone probe matched the intent: 6 total attempts / 5 sleeps on sustained EACCES; the non-local `return` inside the yielded block exits cleanly through `rescue`/`retry`; lock loss between attempts raises `IOError`.
- **Rollback discipline holds** — `write_discovery`'s `ensure` removes the orphan temp on the EACCES-exhausted path, and `claim_ownership_and_start!` still stops the server and releases the lock on failure.
- **Strong, honest tests**, with the Windows permission-bit and SIGKILL exit-code portability caveats handled explicitly.

### Minor (non-blocking) — both in CI/test scaffolding

**1. Windows workflow runs a spec whose path it doesn't watch** — `.github/workflows/windows_active_sessions.yaml`
The job runs `spec/lib/internal_api/active_sessions/server_client_spec.rb`, but GitHub Actions path globs don't cross `/`, so `spec/lib/internal_api/active_sessions*_spec.rb` matches the two top-level specs and **not** that nested one. There's a `lib/internal_api/active_sessions/**` entry but no `spec/lib/internal_api/active_sessions/**` counterpart. A PR touching only `server_client_spec.rb` wouldn't trigger the job that exercises it. Suggest adding `spec/lib/internal_api/active_sessions/**` to both `push.paths` and `pull_request.paths`.

**2. Test reader can silently under-observe on a Windows read-side sharing violation** — `spec/support/coordination_handoff_child.rb`
`observe` rescues `Errno::ENOENT` and `JSON::ParserError` but not `Errno::EACCES`. On Windows the same sharing-violation window this PR hardens the *writer* against can also hit a concurrent *reader*'s `File.read`; if it does, the `watch` thread raises and dies silently (no `abort_on_exception`), so `reads` just stops incrementing while the assertions can still pass — or a `state`/`sample` command hitting it would crash the child and time out the parent. Test-infrastructure only (can't affect production) and not observed in the currently-green Windows CI, but worth rescuing `Errno::EACCES` in `observe` as a transient read and having the watch thread record failures instead of dying.

### Open questions (non-blocking)
- `Socket.tcp(..., connect_timeout: READ_TIMEOUT)` reuses the 1s `READ_TIMEOUT` as the connect budget. On a connect timeout Ruby raises `Errno::ETIMEDOUT` (some builds `IO::TimeoutError`); both are `StandardError`, so `Client#request` normalizes either to `{ ok: false, error: ... }` and `ping`→false on every path I traced. Flagging only that the connect budget is now coupled to the read timeout.
- Native Windows behavior (real `File.rename`/`File.delete` EACCES, `flock` semantics) is verified via green Windows CI and Linux-side injected-EACCES mocks; I did not run a live Windows handoff.

_Reviewed with the reviewpr workflow: specs executed, retry semantics falsified by execution, and a CodeRabbit second-reviewer pass (its out-of-scope findings from a stale base were discarded)._

lib/internal_api/active_sessions.rb
```diff
@@ -532,17 +569,18 @@ def self.delete_discovery_if_owned
...
       #
-      # Re-reads the file before deletion to avoid a race where another process
-      # has written a fresh discovery between the caller's initial read and this
-      # deletion attempt.
+      # The caller must retain the native ownership flock throughout this
+      # operation. The pid check alone cannot serialize publication and unlink.
       #
       # @param expected_owner_pid [Integer]
       # @return [void]
       def self.delete_discovery_if_owner(expected_owner_pid)
```
mrhoribu, line 608: Retry exhaustion here is swallowed by the outer `rescue StandardError`, same as before this PR. Could you add a delete-path counterpart to the publish-path exhaustion test at active_sessions_spec.rb:280-291, so the 'stale discovery + released lock' behavior on permanent EACCES is asserted rather than implicit? Non-blocking, just flagging as a follow-up.

mrhoribu (approved):
Nice fix — the discovery-delete-before-release reordering closes a real TOCTOU (old code released the flock, then deleted based only on a pid match, so a successor publishing in that window could have its live discovery deleted by the retiring owner). Verified against the base commit and the invariant is now correctly gated on `own_lock?` instead of pid metadata. Ran the full suite (7,556 examples) and rubocop clean at head.

One non-blocking gap: `delete_discovery_if_owner` (lib/internal_api/active_sessions.rb:577-587) swallows `Errno::EACCES` after all retries are exhausted, same as before, but there's no test for that exhaustion path on the *delete* side — only the transient-single-failure case is covered (active_sessions_spec.rb:394-413). The equivalent exhaustion case on the publish/rename path *is* tested (active_sessions_spec.rb:280-291). Worth mirroring that test for the delete path in a follow-up so the "permanent EACCES leaves a stale discovery record with the lock released" behavior is pinned down rather than implicit — I don't think it's release-blocking since a successor's next publish overwrites the stale record unconditionally.

therealatari:
Added the delete-side exhaustion coverage in `eb60a682`.

The new example forces every owned-discovery delete attempt to raise `Errno::EACCES`, then verifies the bounded attempt count, the documented stale discovery record, and release of the native ownership lock.

Local verification on Ruby 4.0.5:
- ActiveSessions Windows-workflow set: **54 examples, 0 failures**
- RuboCop on `active_sessions_spec.rb`: **no offenses**

lib/internal_api/active_sessions.rb
mrhoribu: Confirmed in `eb60a682` — the new example forces every discovery-delete attempt to raise `EACCES`, asserts the bounded attempt count (`DISCOVERY_FILESYSTEM_RETRY_DELAYS.length + 1`), and pins down exactly the behavior I was asking about: the stale record stays and the lock is released once retries are exhausted. Verified locally too (rubocop clean, targeted specs 54/54, full suite 7,557/7,557 passing). Thanks for the quick turnaround.

mrhoribu (approved):
Re-approving on the latest commit (`eb60a682`) now that the delete-path retry-exhaustion test has landed and closed out the earlier follow-up.

therealatari:
Thanks for the detailed reviews. I worked through the claims and pushed `c0382cc8` with the resulting hardening:

- corrected the ownership wording: the cross-process guarantee is the retained flock; the local handle check prevents this process from mutating after it has relinquished ownership
- made private-temp cleanup bounded and best-effort so a cleanup failure cannot mask the original publication exception
- separated stale-endpoint connection establishment from response reads with an explicit connect timeout
- added the nested ActiveSessions spec path to the Windows workflow trigger
- made the real-process watcher tolerate transient read-side `EACCES`, surface unexpected watcher failures, and prove that it observed valid records
- added regression coverage for exhausted delete retries and cleanup after local ownership-handle loss

The bounded retry sleeps under the lifecycle mutex remain intentional (about 310 ms maximum per operation), because serialization across ownership validation and mutation is the safety property. Crash-left PID-scoped temp files and Windows directory ACLs are now documented as constraints rather than overstated guarantees.

Current verification: full suite **7,559 examples, 0 failures**; focused ActiveSessions suite **56 examples, 0 failures**; repository-wide RuboCop **1,172 files, no offenses**; native Windows handoff CI green.

mrhoribu (approved):
Re-approving on `c0382cc`, which addresses the open items from both Nisugi's and Mahtra's reviews:

- CI path glob now covers the nested `spec/lib/internal_api/active_sessions/**` spec (Mahtra).
- `coordination_handoff_child.rb#observe` rescues `Errno::EACCES` and the watch thread records failures instead of dying silently, with a new spec covering the transient-read-violation case (Mahtra).
- `write_discovery`'s temp-file cleanup is now its own retried, independently-rescued step, so a failure there can no longer mask the original publish error in the `ensure` path — covered by a new spec asserting both log lines surface (Nisugi).
- `Client::CONNECT_TIMEOUT` is now separate from `READ_TIMEOUT` (Nisugi).
- The `own_lock?`/retry helper split and reworded comments/test name now correctly describe it as a local ownership handle rather than implying a live OS-level probe (Nisugi).

Verified directly against this commit, not just the message: rubocop clean on all touched files, targeted specs 56/56, full suite 7,559/7,559, and confirmed via the check-runs API that CI genuinely re-ran and passed on `c0382cc` itself (Windows handoff job included).

Issue #1616, opened 2026-09-13
detachable_client_send_init delays initial vitals push by up to 10s due to incorrect IconJOINED check


athias:
**Summary**

detachable_client_send_init (lib/global_defs.rb:2344) can delay a newly-attached detachable-client frontend's initial vitals/status push by up to 10 seconds, for no functional reason. For a character that has never grouped in the current Lich process, the frontend receives no progressBar/indicator/compass data at all until that delay elapses.

**Root cause**

The method's first line is:
``` ruby
100.times { sleep 0.1; break if XMLData.indicator['IconJOINED'] }
```

**Two separate problems:**

1. Wrong condition. XMLData.indicator['IconJOINED'] is a bare truthiness check. Indicator values are stored as the strings 'y'/'n' (see xmlparser.rb:789), both of which are truthy in Ruby — so this line does not actually check "is the player grouped," it only checks "has this key ever been set at all." Compare to checkgrouped (global_defs.rb:340), which correctly does XMLData.indicator['IconJOINED'] == 'y'.
2. No functional purpose even if the condition were correct. Nothing after this line branches on whether the loop broke early or exhausted all 100 iterations — init_str is built unconditionally from live XMLData state either way. The loop only ever adds latency; it never changes what gets sent.

Net effect: for any character that hasn't grouped yet in the running Lich process, the IconJOINED key is never set, so this loop always runs the full 10 seconds before the frontend gets any initial vitals data — on every single detachable-client attach, including one attaching mid-session to an already-live game.

**Impact**

Any detachable-client frontend (ProfanityFE-style; e.g. third-party frontends using --detachable-client) shows blank/stale vitals, indicators, and compass exits for up to 10 seconds after connecting, especially noticeable when attaching to a session that's already live rather than at login.

**Proposed fix**

Remove the loop entirely — it is provably a no-op with respect to what gets sent, so deleting it is behavior-preserving except for removing the latency. See branch IconJoined_delay, a single-line removal in lib/global_defs.rb.

A more complete fix would replace the naive fixed-2-3-line send with a genuine "have we received at least one real vitals push yet" readiness flag (since XMLData.health etc. default to 0, which is indistinguishable from a legitimate zero reading) — noted here as a possible separate follow-up, out of scope for this fix.

**Repro**

1. Log in to a character that has never grouped in the current Lich process.
2. Start Lich with --detachable-client=PORT (headless/--without-frontend or with a bundled frontend, doesn't matter).
[...]

PR #1617, opened 2026-09-13
fix(all): Initial Lich state update and detachable-client split GS/DR

athias:
## Summary

Fixes the initial state Lich pushes to a newly attached detachable-client frontend, and splits that push into separate GemStone and DragonRealms implementations.

## Problem

- Vitals bars started at 0 on attach: health/mana/spirit/stamina were sent with a hardcoded `value='0'` and only corrected on the game's next vitals update.
- Only 7 indicators were sent (e.g. invisible, hidden, stunned, and grouped were missing), and indicators not yet seen went out as `visible=''`.
- DragonRealms received GemStone-shaped vitals (`health 100/` with no max), no concentration bar, and no hands.
- A hand not yet seen by Lich raised inside the init, silently dropping the whole push.
- The GS and DR feeds differ enough that one shared function kept accumulating game checks.

## Changes

- **Per-game init modules.** `Lich::Gemstone::DetachableClientInit` and `Lich::DragonRealms::DetachableClientInit` (loaded by `GameLoader`) each build their game's push. Overlapping pieces (indicators, spell, hands, compass) are intentionally duplicated so each community can maintain its own feed independently.
- **Shared wait and send stay in `global_defs.rb`.** `detachable_client_send_init` waits (up to 10s, checked before sleeping) until the game module reports `ready?`, then sends its push. If the game modules never load, it logs a warning and skips the push.
- **Readiness is the first `<prompt>`.** In both games' login feeds it follows everything the push describes; `IconJOINED` and `<endSetup/>` arrive before spell, hands, and room exits.
- **GemStone:** real bar values using the game's integer truncation (`current * 100 / max`, clamped 0..100, raw value kept in text), plus stance, mind, encumbrance, and wounds/scars as before.
- **DragonRealms:** percent-based vitals as the game sends them (`fatigue` label for stamina) plus the concentration bar; stance/mind/encumbrance/injuries are not sent.
- **Both:** all 13 indicators with a `'n'` default, and hands default to `Empty` if not yet seen.
- **Unrelated:** the session descriptor message in `Frontend.create_session_file` now goes to `Lich.log` instead of stdout.

## Known gaps

- DragonRealms injuries are not sent; no captured feed shows an injured DR character yet.
- The init push is still skipped for `--genie` and `--saga`, unchanged.

## Original Problem

This stemmed from a re-investigation of #1616.  After further review the true source of the problem was identified.


MahtraDR (approved):
Approving after an adversarial review at ac809b17646a5475ae96180d9d9546b67d770570.

This is a one-line deletion that is behavior-preserving except for removing latency: `init_str` is built unconditionally from live `XMLData` state, nothing downstream branches on the removed loop, and it ran in its own `Thread.new` so the `sleep` never blocked game parsing — it only delayed this client's one-shot init send by up to 10s. Verified the issue's grounding: `@indicator` is a plain Hash storing the `visible` string ('y'/'n'), so the bare `XMLData.indicator['IconJOINED']` truthiness check was wrong (cf. `checkgrouped`'s `== 'y'`) and, more importantly, output-irrelevant.

Probed the state/sequence axis (re-attach, cold attach, thread ordering): no regression — the pre-fix path sent the identical snapshot content, just later. The cosmetic '0/0' cold-attach frame is pre-existing and correctly deferred to a follow-up.

CI green; CodeRabbit surfaced nothing on the changed line. LGTM.

mrhoribu (dismissed):
# PR Review: fix(all): IconJoined_delay (elanthia-online/lich-5#1617)

**Reviewed:** 1 file, +0/-1 at `ac809b17` on 2026-09-14. Based on: full local clone (checkout at `lich-5/`, working tree currently has unrelated local changes on branch `fix/script-log-log-dir` — PR diff reviewed via `gh pr diff`/`git log`, not a checkout of the PR branch).
**Verdict:** Approve
**Summary:** Deletes a dead 10-second polling loop from `detachable_client_send_init` that never gated anything and never checked the condition it claimed to check. The deletion is behavior-preserving except for removing the latency, exactly as the linked issue argues, and I independently verified every claim in that argument against the current source.

## What this PR gets right

The fix matches its own root-cause analysis exactly, and both halves of that analysis check out against the code as it exists today:

- **The condition was never meaningful.** `XMLData.indicator['IconJOINED']` is a bare truthiness check on a value that `xmlparser.rb:789` (`@indicator[attributes['id']] = attributes['visible']`) always sets to the string `'y'` or `'n'` — both truthy in Ruby. The correct pattern exists two lines above the deleted one, in `checkgrouped` at [global_defs.rb:340](lib/global_defs.rb#L340): `XMLData.indicator['IconJOINED'] == 'y'`. The deleted loop only ever detected "has this key been set at all," not "is the player grouped."
- **Nothing downstream branched on it.** Reading the method at [global_defs.rb:2344-2382](lib/global_defs.rb#L2344-L2382), `init_str` is built unconditionally from live `XMLData` state whether the loop broke on iteration 1 or ran all 100. The loop's only observable effect was latency, up to 10 seconds, on every detachable-client attach for a character that hadn't grouped yet in the current process — which is the common case for a frontend attaching mid-session.
- **The history explains how it got here without excusing it.** `git log -L 2343,2344:lib/global_defs.rb` traces the line to the #1459 mega-PR that introduced the whole detachable-client subsystem. It sits right next to a structurally similar loop in `detachable_client_send_player_id` ([global_defs.rb:2385-2389](lib/global_defs.rb#L2385-L2389)) — `100.times { break if (tag = ...); sleep 0.1 }` — which *is* meaningful there because the code after it, `client.puts_main_stream(tag) if tag && client.alive?`, actually gates on the loop's result. The deleted loop looks like a copy of that pattern with the gating half left out, which is consistent with it being dead on arrival rather than a deliberate throttle.

The one-line deletion is minimal and doesn't touch anything else in the method.

## Coverage notes

- Full local clone available; diff, blame, and the linked issue (#1616) were all read directly rather than taken on faith.
- No existing spec exercises `detachable_client_send_init` (`grep -rn detachable_client_send_init spec/` is empty) — a pre-existing gap in this area, not something this one-line PR introduces or is obligated to fix.
- `ruby -c lib/global_defs.rb` — syntax OK. `rubocop lib/global_defs.rb` reports 22 pre-existing `Layout/ExtraSpacing` offenses, all in unrelated single-letter direction-alias methods (`def u; 'up'; end`, etc.) far from the changed lines — none attributable to this PR.
- CodeRabbit CLI was not run (no `coderabbit` binary / auth check performed given the trivial size and already-verified nature of the change; nothing here would have benefited from a second automated pass beyond what was manually traced).
- No runtime test against a live detachable-client session was performed; the finding rests entirely on static tracing of `init_str` construction, which is unconditional and requires no runtime confirmation.

mrhoribu:
**Walking back my earlier approval on this one** — a real login transcript surfaced in follow-up discussion contradicts a claim my review leaned on, so I dismissed it rather than let it stand.

I'd characterized the deleted loop's `XMLData.indicator['IconJOINED']` check as dead weight, on the theory that `IconJOINED` is only ever set by a grouping event (per the `group.rb` regexes) and therefore never fires for an ungrouped character, making the wait pure latency with no gate on anything. A GemStone login capture shown to me shows that's wrong: `IconJOINED` is part of a bulk indicator dump the server sends at the *end of the login/setup sequence* for every character, grouped or not:

```
<endSetup/><app char="Ilten" game="Test" title="GemStone IV: Ilten [Test]"/>
<indicator id="IconKNEELING" visible="n"/>...<indicator id="IconHIDDEN" visible="y"/>...<indicator id="IconJOINED" visible="n"/><spell>None</spell>...
```

That makes the loop's original intent plausible as a real readiness wait — `IconJOINED` being the last tag in that dump, polling it is a rough (if never-documented) proxy for "the initial setup batch has fully arrived," not a meaningless group check as I'd characterized it in the review body posted on this PR.

That said, I don't think this fully exonerates the loop either, and I'd like these resolved before this merges as currently described:

1. **Even under the readiness reading, nothing downstream actually uses it.** `init_str` (`lib/global_defs.rb:2346` onward) is built unconditionally from `XMLData.mana`/`health`/`spirit`/etc. right after the loop, with no check that *those specific fields* have real values — it just trusts that `IconJOINED` showing up means the rest arrived too. True in the capture above (vitals stream in earlier via `dialogData id='minivitals'`), but that's correlation the code never verifies, not a guarantee.
2. **The capture is GemStone.** dr-scripts/scripts (this ecosystem's primary consumers) are DR-only, and I have no DR transcript confirming DragonRealms sends the same end-of-setup indicator dump. If it doesn't, the loop's protection may be GS-specific, and removing it could be fine for DR and risky for GS, or vice versa — worth confirming before treating this as a universal no-op.
3. **The issue's own repro doesn't fit this mechanism.** #1616 describes the delay on a *second* frontend attaching minutes into an already-live session. If `IconJOINED` is set once at login and persists in the shared `XMLData.indicator` hash for the process's lifetime, that second attach should see the loop exit almost instantly, not stall 10s. Either that specific repro doesn't reproduce as described, or there's a reset/edge case neither of us has identified yet — worth double-checking against a live second-frontend attach before merging on the strength of that repro.

None of this means the fix is wrong — removing an unverified, un-gating wait is still probably a net improvement — but "provably a no-op, safe to delete" (the framing in #1616 and in my own prior approval) isn't accurate given the above, and I'd like to see the DR behavior and the mid-session repro confirmed rather than assumed before this goes in.

mrhoribu:
Follow-up on the readiness-wait question above: I now have a DR login transcript too, and it shows the same thing as the GS one — a full indicator dump at the end of setup, including `IconJOINED visible="n"` for an ungrouped character:

```
<switchQuickBar id='quick-simu'/>
<indicator id="IconKNEELING" visible="n"/>...<indicator id="IconWEBBED" visible="n"/><indicator id="IconJOINED" visible="n"/>
```

So this isn't game-specific — both DR and GS send it once, at login. That resolves my earlier uncertainty about DR behavior, but it sharpens the question about the specific repro in #1616 rather than closing it.

I checked whether anything mid-session clears `XMLData.indicator` (which would explain the key going missing again later) — it doesn't. `XMLData.reset` (`lib/common/xmlparser.rb:206-224`), the only reset path, runs on stream-desync recovery (`lib/games.rb:1014`, `:1065`) and clears `@active_tags`, `@active_ids`, `@pending_crtr_status`, etc., but never touches `@indicator`. That hash is initialized once at parser construction (`xmlparser.rb:136`) and never cleared again.

Put together: `IconJOINED` gets set once during login on either game, and nothing in this codebase unsets it afterward. That means a second detachable-client frontend attaching minutes into an already-live session — the exact scenario in #1616's repro — should see the deleted loop's condition go true on its very first 0.1s check, not stall for the full 10 seconds. That's the opposite of the reported impact.

I'm not doubting the loop caused *some* real latency — the mechanism itself (a fixed uncapped poll with no gate on what gets sent) is worth removing regardless. But I can't reconcile "provably a no-op, safe to delete, no functional reason for the delay" with the specific repro as written, given what both transcripts show about when this key actually gets set. Could you re-confirm the mid-session-attach repro reproduces a real 10s delay on your end? If it does, there's something about the indicator hash's lifecycle I'm missing and I'd like to find it before this merges; if it doesn't, the issue's stated impact may need a correction even though removing the loop is probably still fine.

mrhoribu (approved):
# PR Review: fix(all): Initial Lich state update and detachable-client split GS/DR (elanthia-online/lich-5#1617)

**Reviewed:** 7 files, +591/-32 at `e0c6940a` on 2026-09-16. Based on: full local clone, PR branch checked out into an isolated worktree (`gh` fetch of `pull/1617/head`), full spec suite and rubocop run locally against the checkout, CodeRabbit CLI run against the correct merge-base (`ed662ca7`).
**Verdict:** Approve
**Summary:** A substantial rewrite of the original one-line fix: it correctly identifies and fixes the real bug (hardcoded `value='0'` progress bars, not the polling loop itself), replaces the unreliable `IconJOINED`-truthiness readiness check with a well-grounded one (first `<prompt>`), splits GS/DR into independently-maintained modules, and adds thorough test coverage for both. Everything I checked — constant resolution, nil/zero-division safety, downstream frontend compatibility, the full spec suite (8024 examples), rubocop — holds up.

> **Callout — title/description undersell the fix.** The current title (*"Initial Lich state update and detachable-client split GS/DR"*) and description lead with the internal module-split rather than what actually changed for users: vitals bars starting at 0% and missing indicators on attach. Suggest retitling to something like `fix(all): send real vitals percentages and indicators on detachable-client attach`, with the description reordered so the Problem section leads and the per-game split is framed as *how* it was implemented rather than the headline. Separately, the description's "Unrelated" bullet about `Frontend.create_session_file` logging doesn't match this diff — no `frontend.rb` hunk exists here (see Minor, below) — so that line should come out or be corrected before merge.

## What this PR gets right

- **Fixes the actual bug, not the symptom.** The real defect (confirmed in this PR's own review thread, not just asserted) was that `value='0'` was hardcoded in the vitals `progressBar` tags regardless of what `XMLData` held, so a frontend reading `value` as a literal percent always started at 0% no matter when the push fired. `lib/gemstone/detachable_client_init.rb:196-200` and `lib/dragonrealms/detachable_client_init.rb:88-93` now compute real percentages, matching each game's own semantics (GS: `current/max` truncated and clamped; DR: the game already reports a bare percent).
- **The new readiness check is well-grounded, not another guess.** `ready? = !XMLData.prompt.empty?` relies on `<prompt>` being the server's own batch terminator — confirmed independently in `lib/common/xmlparser.rb:653-660`'s own comment ("A prompt terminates the command burst and is the reliable close signal..."), not just asserted by this PR. `@prompt` is set once and never cleared elsewhere in the parser (`xmlparser.rb:57`, `:1045`), so once true it stays true for the process's life — no flakiness on a second frontend attaching later.
- **Real edge cases are handled, and tested.** Zero/negative/nil-shaped vitals (`spec/lib/gemstone/detachable_client_init_spec.rb:499-513`), a hand not yet seen (`GameObj.right_hand&.name || 'Empty'` at `lib/gemstone/detachable_client_init.rb:205`, replacing a bare `.name` call that the PR description says previously raised and silently dropped the entire push), and indicators not yet seen (`.fetch(indicator, 'n')` instead of blank `visible=''`) are all covered by specs, not just handled in prose.
- **The unqualified `DetachableClientInit` constant resolution is a proven pattern, not a novel risk.** The original code already relied on the same mechanism for `Wounds`/`Scars` (`Lich::Gemstone::Wounds` reached unqualified from this exact method via `include Lich::Gemstone` in `lib/main/main.rb`), and `lib/gemstone/group.rb` already does the same for bare `GameObj`. This PR's version of the trick isn't new territory.
- **Downstream compatibility verified, not assumed.** `ProfanityFE`'s indicator handler (`lib/tag_handlers.rb:424-431`) matches any `Icon[A-Z]+` id generically — going from 7 to 13 indicators sent doesn't require any frontend change.
- **Verified independently, not just read.** Checked out the PR branch in an isolated worktree: full spec suite passes (8024 examples, 0 failures), rubocop reports the same 22 pre-existing offenses as before this PR (all in unrelated single-letter direction-alias methods, none touched by this change), and both new per-game spec files (26 examples) plus the updated `global_defs_spec.rb` coverage (5 examples) pass on their own.

## Minor

### PR description references a change that isn't in the diff — `lib/common/frontend.rb`

The description's "Changes" section states: *"Unrelated: the session descriptor message in `Frontend.create_session_file` now goes to `Lich.log` instead of stdout."* There is no `frontend.rb` (or any `Frontend`-touching) hunk anywhere in the actual diff — `grep -n "create_session_file\|Frontend" pr-v2.diff` returns nothing. Given the commit list includes a "Fix merge" commit between the two feature commits, this most likely describes a change that got dropped or is sitting in a different commit/PR, and the description is now stale. Worth a quick edit before merge so reviewers and changelog-readers aren't looking for a change that isn't there.

## Open questions

- `lib/gemstone/detachable_client_init.rb:155-160` and the DR equivalent claim "In the [game] login feed the first `<prompt>` follows everything the push describes... whereas `IconJOINED` and `<endSetup/>` arrive before some of it." I can independently confirm the *general* claim that `<prompt>` reliably terminates a batch (from `xmlparser.rb`'s own comment), but neither GS nor DR transcript I was shown in this review thread captured far enough to show a `<prompt>` tag actually following the indicator dump in the same batch — both cut off right at the indicators. **Settles it:** a slightly longer capture (a few more lines past the indicator dump) showing the first `<prompt>` tag on both games would fully close this; I'd call it very likely true given the parser's own invariant, just not directly witnessed.

## Considered and rejected

CodeRabbit's one finding (`lib/global_defs.rb`, major severity) suggests skipping the push entirely — logging a warning instead of sending — if `ready?` never becomes true within the 10-second wait, rather than sending `DetachableClientInit.init_string` regardless. This contradicts a deliberate, explicitly tested design choice: `spec/lib/global_defs_spec.rb:671-676` (*"still sends the push after giving up on a feed that never becomes ready"*) asserts exactly the behavior CodeRabbit wants removed. Adopting the suggestion would mean a frontend that attaches during an unusually slow or degraded login gets **no** init push at all rather than a best-effort one — reintroducing a milder version of the original blank-frontend problem this whole PR exists to fix. Rejected.

## What this PR gets right (continued) / Coverage notes

- Full local clone; PR branch fetched and checked out (`gh api pull/1617/head`) into an isolated git worktree so the review never touched the (unrelated, pre-existing) dirty state of the main checkout.
- `bundle exec rspec` (full suite): 8024 examples, 0 failures, on the PR branch.
- `rubocop` on all seven changed files: 22 pre-existing offenses (all in unrelated code, none new).
- CodeRabbit CLI run against the correct merge-base (`ed662ca7`, confirmed via `git merge-base origin/main pr-1617`); one finding, investigated and rejected above with reasoning.
- Did not independently re-verify GS's `(current * 100) / max` integer-truncation claim against the live game server's own math — taken on the strength of the PR's stated rationale and the existing `percenthealth`/`percentmana` precedent in this same file, not from a packet capture showing the server's actual computed `value`.
- Did not runtime-test an actual detachable-client attach against a live game session; all verification here is static (source, tests, blame) plus the already-passing CI.

MahtraDR:
## Review — Approve

**Reviewed at `e0c6940` (full clone + PR-head worktree).** No blockers, majors, or minors survived verification. One open question worth a glance.

This is a clean, well-tested change. The per-game split reads well, and the fixes are real: nil-safe hands (`GameObj.right_hand&.name || 'Empty'`) closes a latent `NoMethodError` that — inside the old monolithic method — would drop the *entire* push via the rescue; `indicator.fetch(id, 'n')` fixes the malformed `visible=''`; the GS zero-max guard and integer-truncation math match the game. The wait-loop rewrite is correct: checking `ready?` before sleeping means a re-attach isn't delayed a tick, and the `defined?(DetachableClientInit)` guard makes a not-yet-loaded module keep waiting and then skip-with-warning rather than crash.

### Open question — the one thing worth confirming

`ready?` gates the push on the first parsed `<prompt>` (`!XMLData.prompt.empty?`, both modules). The **initial-login** path's correctness rests entirely on the ordering claim in the module docs: that the first `<prompt>` follows vitals/hands/room-exits in the login feed. If a prompt can be parsed *before* those, the login push ships zero-value bars and `Empty` hands — the exact failure class this PR fixes for the old `IconJOINED` gate. The **re-attach** path (the common case) is immune, since a prompt was parsed long ago and `ready?` is already true with live state.

**Settles it:** a captured GS and DR login feed showing `<prompt>` after the vitals/hands/exits block. I verified the parser *stores* these values correctly (DR percents land in `@health`/`@mana`/`@stamina`/`@spirit`/`@concentration` with a nil max, exactly as the DR module reads them); what I could not execute is the *order* they arrive in. If that ordering was confirmed from captured feeds, this closes.

### How this was checked

- **Execution, not just reading:** I re-implemented the `detachable_client_send_init` wait-loop state machine standalone and drove it through five sequences — ready-from-start (0 sleeps + send), ready-after-N (N + send), module-loads-mid-wait (N + send), never-ready (100 + send), module-never-loads (100 + skip + log). All matched the specs.
- **Grounded framework facts:** `XMLData.prompt` starts as `String.new` (so `ready?` is safely false at start); all vitals init to `0`, not nil; `@injuries` is pre-populated for all 16 areas at 0 (so `wound > 0` never sees nil); `Wounds`/`Scars` body-part methods are public (`public_send` safe); unqualified `DetachableClientInit` resolves via `Object`'s ancestors through main.rb's top-level `include Lich::Gemstone`/`DragonRealms` — the same mechanism the old code already used for `Wounds`/`GameObj`.
- **Concurrency:** runs in a per-client thread; the modules only read shared state and write to their own socket; `puts_main_stream` is dead-socket-safe (returns `false`, never raises).
- **CI:** green on this PR while base is green.

### Coverage caveat

**CodeRabbit could not run** for this pass and the repo has automatic CR reviews disabled, so this is one reviewer plus execution — not two decorrelated models. I leaned on standalone execution of the stateful part to compensate. Not covered: live GS/DR login-feed ordering (the open question) and behavior against a live server. What I found is real; that is not the same as "nothing is left."

MahtraDR (approved):
Approving — reviewed at e0c6940. Details in my review comment above; the one open question (login-feed `<prompt>` ordering) is non-blocking and closes if it was confirmed from captured feeds.

mrhoribu:
Yep, confirmed in both a DR and GS4 xml log capture that all that data comes across before the first `<prompt>` comes in.

PR #1613, opened 2026-09-13
feat(all): prototype bounded read-only session coordination

therealatari:
## Summary

Draft implementation of the reviewed coordinated-independent-sessions proposal,
read-only slices A/B. Each character keeps its own Ruby process. This is not
Sessionize, a new broker, a hunting engine, or a write-capable control interface.

## Reuse and scope

- Reuse native ActiveSessions Server/Client with optional bounded framing,
  deadlines and authenticated handler dispatch. Legacy defaults stay unchanged.
- Explicitly loaded `Coordination::Session`, `Client` and `Discovery` expose
  fixed room/readiness snapshots, complete incarnation/connection/run identity,
  immutable owner publications and conservative source freshness.
- Reads cannot refresh owner progression. Replayed response ages retain local
  elapsed time and prior RTT; concurrent calls on one client fail busy rather
  than wait outside their budget.
- Discovery is metadata only. Tokens are not advertised. Closed/replaced
  endpoints cannot be adopted by PID, port or character name alone.
- No commands, eval, Script launches, automatic initialization, profile changes
  or modifications to LAB's existing policy authority.

## Cross-consumer architecture boundary

The contract has now been pressure-tested against three distinct consumers:
EOHunter readiness barriers, LAB operation admission, and a production-tested DR
claim/lease allocator.

That DR input sharpens the boundaries rather than expanding this PR:

- The fixed `room`/`readiness` projection here is an EOHunter experiment, **not**
  the future game-neutral observation schema.
- A promotable envelope must describe producer/session incarnation, source name,
  source generation and sequence, capture boundary, freshness and per-field
  coherence without requiring GemStone room/combat concepts.
- Loopback ActiveSessions is this prototype's transport adapter, not the semantic
  definition of coordination. Existing DR cross-host transports should be able
  to implement the eventual contract without becoming Lich dependencies.
- Long-lived resource leases are distinct from observations and discrete
  operations. Lease lifecycle, atomic arbitration and fencing require a separate
  design/approval slice; Redis, remote transport and lease code remain out of
  scope here.
- Every contender for one logical resource pool must use the same lease-store
  authority. Durable participant identity may survive restart, but the ephemeral
  holder incarnation must be replaced atomically under a newly advanced fence.
- Discrete Operations cannot be declared stable/public before a game-free lease
  conformance harness proves composition or records that Leases need a separate
  surface. The stacked operations PR now includes that executable gate.

These constraints are now recorded in
[...]

PR #1618, opened 2026-09-13
feat(all): add bounded coordinated operation lifecycle

therealatari:
## Summary

Add the bounded write-capable lifecycle that follows the read-only coordination
prototype in #1613. Each character retains its own Ruby process and its own
controller authority.

- Register a fixed set of structured operation names and bounded arguments.
- Bind each request to exact owner/peer session identities, an immutable request
  ID and a canonical argument digest.
- Require receiver-issued preflight tickets before first delivery.
- Separate admission, execution outcome and cleanup truth in retained receipts.
- Reserve work on transport threads, but permit only the owning controller thread
  to take and settle it.
- Bound tickets, retained receipts, frames, transport time and admission capacity.

No raw command forwarding, eval, generic method invocation, child-script route,
hunting policy, broker, shared Ruby process, Redis dependency or automatic
runtime enablement is added.

## Stack

**Depends on #1613; review/merge that parent first.** This branch contains the
parent commits because both PRs target `main`. The operation-only changes are in
the incremental comparison from `8b703564` to `a4182d82`.

Companions:

- Discovery ownership and Windows handoff: #1612
- Read-only coordination and bounded transport: #1613

## DR-informed architecture boundary

Production DragonRealms coordination supplied a third consumer shape: leaderless
allocation of contended resources through acquire, renew, release, expiry and
reclaim. It validates this PR's generation-bound request IDs, immutable digests,
tombstones and conservative unknown outcomes, but it also establishes an
important boundary:

- `Operations` are discrete request/receipt lifecycles. This PR implements them.
- `Leases` are long-lived exclusive resource ownership with renewal, reclaim and
  monotonically increasing fencing generations. They are a future sibling
  interface requiring separate design and approval.
- Authorization-grant expiry is not proof that a resource lease or physical
  cleanup ended.
- Allocation policy remains consumer-owned. Atomic compare-and-claim belongs to
  the selected lease-store adapter, so leaderless policy does not require an
  elected coordinator but still has a serialization authority.
- Loopback ActiveSessions is the transport adapter tested here, not the semantic
  definition of operations. Another transport can preserve the contract without
  making a cross-host store a Lich dependency.

Lease identity is split deliberately: durable participant identity can preserve
legitimate acquisition standing across restart, while an ephemeral holder
incarnation identifies the process currently allowed to act. Atomic resume must
advance the fence and invalidate the old holder. Every contender for one logical
resource pool must use the same lease-store authority or exclusivity is not
guaranteed.

The branch documentation records these boundaries. It does **not** implement
resource leases, remote transport, or leader election. Operations also cannot be
declared stable/public until the synthetic lease-composition gate passes or
explicitly proves that Leases need a separate surface.

## Contract highlights

A peer first builds its immutable request, then obtains a receiver ticket bound
to request ID, operation, argument digest, owner/peer identities and receiver
generation. Submitting the identical request is idempotent; changing its meaning
under the same ID is a conflict. Retrying cannot renew the ticket.

Ticket validity bounds receiver issuance-to-use time, not the age of the user's
intent before preflight. Endpoint closure, timeout, cancellation and process
death never overclaim execution or cleanup. A running request remains unknown
until its exact owner records otherwise.

This is cooperative same-host/same-user safety and idempotency, not a security
boundary against hostile code running as the same OS user.

## Verification

- Focused operation lifecycle and real bounded-loopback process coverage.
- Synthetic operation/lease composition gate: **4 examples, zero failures**.
  It sends lease-shaped work through the actual operation machinery and verifies
  one fenced winner, idempotent renew/release, durable restart/resume fencing,
  expiry/reclaim, and unknown cleanup remaining distinct from release.
- Combined focused coordination suite: **21 examples, zero failures**, seed 719.
- Full exact-branch suite previously passed: 7,608 examples, zero failures.
- Scoped RuboCop and `git diff --check` passed.
- The new conformance adapter exists only in specs; no public lease implementation
  or runtime dependency is introduced.

See `docs/coordinated-operations-prototype.md` for the complete prototype
contract, omissions and adapter gates.

### DIFF
--- a/docs/coordinated-readiness-prototype.md
+++ b/docs/coordinated-readiness-prototype.md
@@ -1,0 +1,190 @@
+# Coordinated readiness: bounded A/B prototype
+
+Date: 2026-09-13. Status: experimental implementation and synthetic verification;
+not a published API, live installation, or approval of write-capable coordination.
+
+## Scope and pinned foundations
+
+Implements the first read-only experiment from the reviewed independent-session
+proposal; the bounded implementation contract is recorded here. Each character retains its
+own Ruby process. No game commands, scripts, permissions/grants for execution,
+receipts, broker replacement, Redis requirement, or native frontend input path
+are introduced. No automatic runtime registration or profile enablement.
+
+Source baselines:
+
+- [Lich 3342837fe78cd8965d20525464913c66218ea853](https://github.com/elanthia-online/lich-5/tree/3342837fe78cd8965d20525464913c66218ea853).
+- [EOHunter 6f15eb02c6376c4e9330d940e7f1c6f8da18eb3a](https://github.com/elanthia-online/eohunter/tree/6f15eb02c6376c4e9330d940e7f1c6f8da18eb3a).
+
+| Existing facility | Missing behavior | Owner in this experiment | Verification |
+| --- | --- | --- | --- |
+| ActiveSessions file lock, atomic rename, registry | Publish an optional character endpoint descriptor | Explicit discovery adapter; no extra daemon | Actual owner death, competing successors, reader, cleanup-race tests |
+| ActiveSessions Server/Client JSON transport | Bounded optional read-only route, exchange and frame limits | Native transport with opt-in arguments, unchanged legacy defaults | Old transport specs plus bounded I/O/capacity/auth tests |
+| Native Script lifecycle | None for read-only observations | No Script launch or supervisor added | No write routes accepted |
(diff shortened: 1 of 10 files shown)


calael:
Still think that this could be done in a lib.  Any reason it should be in core?  It’s pretty big chunk of code, and not all folks use it.

therealatari:
Agreed, and implemented that direction—thank you. The transport, tickets, replay protection, operations, and receipts now live in the opt-in [EOHunter `libeocoordination` PR #115](https://github.com/elanthia-online/eohunter/pull/115). EOHunter consumes it in #114. Core is reduced to the reusable hook-priority seam in #1621, plus the existing execution guards. Closing this core implementation as superseded.

PR #1619, opened 2026-09-14
feat(all): add Events, a script-to-script notification primitive; combat emits on it

Nisugi:
## Summary

- add `Lich::Common::Events`, a process-wide notice board for script-to-script notifications: `Events.on(topic, name:, persist:)`, `off`, `emit`, `any_for?`, `on_change`, `list`
- dotted topics with `family.*` and `*` wildcards; named registration replaces instead of stacking; anonymous names generated
- every subscription records `Script.current` as its owner and is removed through the existing `ScriptDeath` path when that script dies (`persist: true` opts out), the same policy the stream hooks use
- synchronous delivery on the emitter's thread; a raising handler is logged and isolated
- delete `Combat::Observers`; the combat processor, messages, and recorder emit on `Events` as `combat.<type>`, and `Combat::Tracker.on/off` become a two-line spelling convenience over it
- document the primitive in `docs/events.md`

## Motivation

Three places had re-solved the same problem independently. `Combat::Observers` had named idempotent subscriptions and error isolation. `HookRegistry` had owner tracking and cleanup on script death. go2 (elanthia-online/scripts) is growing an ad hoc `Go2.on_status` so supervising scripts can hear when travel is blocked. The next producer would have written a fourth copy.

The concrete case that prompted this: a script driving go2 could only see `Script.running?('go2')`. A go2 wedged for minutes on a search-gated exit ("You are in far too much agony to do that") looked identical to one that was walking. Polling a status struct covers "what is true now"; a supervisor also needs "the moment something changed", which is what an event is for.

Events is deliberately the small version. No async delivery, no queues, no history, no persistence. None of the current users need them, each is a real design decision, and a consumer that wants its own thread pushes onto its own `Queue` inside the handler.

## Design notes

- **Owner death is the hard part, and it is reused, not rewritten.** Subscriptions are keyed on the owner's `object_id` and cleaned up from a `ScriptDeath.on_death` handler registered at load, matching `DownstreamHook` and `UpstreamHook`. A same-named sibling script is unaffected. Registrations made outside any script (core code) have no owner and are never swept.
- **`on_change(prefix:)`** lets a producer react only to changes in its own family. `Combat::Messages` uses it to install or remove its stream hook based on whether anyone subscribed to a message family, replacing the old `Observers.on_change`.
- **Why delete the facade instead of keeping it.** A first cut kept `Combat::Observers` as a facade preserving the `(Symbol type, data)` callback shape. That cost ~70 lines, a block-to-name handle map, and a prune hook, for a shape nothing outside the combat module consumed (`combat_stats.lic` uses the recorder and upstream hooks, not `Tracker.on`). Removing it is the second commit.

## Behavior changes

- Subscribers to combat events now receive the topic string (`'combat.damage'`) instead of the Symbol (`:damage`). `Combat::Tracker.on(:damage)` still accepts Symbol types; only the block's first argument changes. The recorder converts once at its subscribe point.
- A combat subscription made from a script used to persist after that script died. It is now removed by default. The recorder already unsubscribes itself, so nothing in-tree depended on the leak.
- The event-type catalogue that lived in `observers.rb` moved to the `Combat::Tracker.on` doc comment.

## Compatibility

Nothing in elanthia-online/scripts references `Combat::Observers` or `Tracker.on`. go2's `Go2.on_status` stays as its own fallback until its `required: Lich >=` header can move past this release, at which point it becomes `Events.emit('go2.status', Go2.status)` behind a `defined?` guard.

## Verification

- `bundle exec rspec spec/lib/common/events_spec.rb spec/lib/gemstone/combat/ spec/lib/common/hook_persistence_spec.rb spec/lib/common/script_death_spec.rb`: 411 examples, 0 failures
- `bundle exec rubocop`: 1170 files inspected, no offenses
- full suite: 7562 examples, 59 failures, all in `frontend_*`, `lich_room_display_settings`, `gtk_compaction`, `wine`, and Windows launcher specs; the same 14 files fail identically on pristine `main` with this branch stashed, so they are environmental to the Windows dev box and not introduced here
- `docs/events.md` is the reference; `lib/common/events.rb` has the contract in its header

New coverage in `events_spec.rb` (24 examples): exact and wildcard matching, idempotent names, generated anonymous names, `off` by name or block, owner death with `persist` both ways, same-name sibling isolation, the `ScriptDeath` wiring, prefix-filtered `on_change`, handler and callback error isolation, and a concurrent register-while-emitting smoke test.


mrhoribu (changes requested):
# PR Review: feat(all): add Events, a script-to-script notification primitive; combat emits on it (elanthia-online/lich-5#1619)

**Reviewed:** 22 files, +785/-375 at `319a758f` (branch `feat/events-primitive`, merge-base `f8ea1f20` on `main`) on 2026-09-14. Based on: full local clone (all four commits, full history, blame available).
**Verdict:** Request changes — please address the Minor finding below (the `refresh!` atomicity gap) before merge; nothing else blocks.
**Summary:** Adds `Lich::Common::Events`, a small mutex-guarded pub/sub registry with dotted-topic wildcards, idempotent named subscriptions, and automatic cleanup on script death via the existing `ScriptDeath` hook (the same pattern `DownstreamHook`/`UpstreamHook` already use). `Combat::Observers` is deleted and combat/message events are re-emitted as `combat.<type>` on the new board. The design is careful — synchronous delivery with snapshotted handler lists (no deadlock on reentrant `on`/`off`/`emit` from inside a handler), error isolation on both handlers and `on_change` callbacks, and a real bug (a named re-registration moving topic families not notifying the family it left) was caught and fixed with a regression test during the PR's own commit history. Full local suite run: 7564 examples, 0 failures; targeted PR suite: 413 examples, 0 failures; rubocop clean on all changed files.

## Minor

### 1. `Combat::Messages.refresh!` isn't atomic against `install!`/`uninstall!` raising — `path/to/file` `lib/gemstone/combat/messages.rb:64-70`

```ruby
def refresh!
  @mutex.synchronize do
    @active = families.select { |f| f.events.any? { |e| Lich::Common::Events.any_for?("combat.#{e}") } }.freeze
    @active.empty? ? uninstall! : install!
    @active
  end
end
```

`@active` is committed unconditionally before `install!`/`uninstall!` runs. If the hook side effect (`::DownstreamHook.add`/`.remove`) raises, `@hook` never flips to match the new `@active`, and the exception is silently swallowed by `Lich::Common::Events.changed`'s per-callback rescue (`lib/common/events.rb:260-264`) — logged only if `Lich.respond_to?(:log)`, otherwise dropped entirely. The module is then left with `active_families` reporting a family as active while `installed?` reports `false` (or vice versa on the uninstall path), with no self-repair until some *other* `combat.`-prefixed subscription change happens to trigger `refresh!` again.

I reproduced this concretely: requiring `messages.rb` under the spec suite's global `DownstreamHook` mock (`spec/spec_helper.rb:1046-1055`, which defines only `.run`/`.cleanup_on_death`, not `.add`/`.remove`) and then calling `Lich::Common::Events.on('combat.bolted', name: 'probe') { nil }` leaves `active_families` non-empty while `installed?` stays `false` — the exact split. In the real spec suite this self-heals invisibly because every spec that touches this path also clears its subscriptions in an `after` hook, which retriggers `refresh!` and recomputes correctly — that's incidental cleanup order, not a designed guarantee, and it only holds because `Lich::Common::HookRegistry#add`/`#remove` (the real production backing, `lib/common/hook_registry.rb:29-49`) are just hash writes and essentially can't raise, not because `refresh!` is written to be safe against it. Given `Combat::Messages` is now driven by a process-wide shared registry rather than the combat-only `Observers` it replaced, worth tightening (e.g. only commit `@active` after `install!`/`uninstall!` succeeds, or wrap the hook call so a failure doesn't diverge the two), even though I could not find a realistic production trigger.

Confidence: confirmed (reproduced), though the practical blast radius is narrow — I could not construct a case where the real `DownstreamHook.add`/`.remove` actually raises.

## Nits

### 1. Duplicated "the" in the module doc comment — `lib/gemstone/combat/messages.rb:18-19`

```ruby
# game stream. Subscribers therefore run on that worker: the
# The Events subscriber contract applies - cheap, non-blocking, no game
```

Leftover from the `Observers` → `Events` rename pass; reads "worker: the / The Events subscriber contract applies". Should be one "The".

## What this PR gets right

- **Reuses the established cleanup pattern exactly.** `Events` registers with `ScriptDeath.on_death` the same way `DownstreamHook`/`UpstreamHook` already do (`lib/common/downstreamhook.rb:51`, `lib/common/upstreamhook.rb:51`), keyed on `object_id` so a same-named sibling script is unaffected — and it's covered by a dedicated test (`spec/lib/common/events_spec.rb:136-145`).
- **Correct lock discipline.** Every mutation snapshots state under `@mutex` and releases it *before* invoking any user-supplied block (`on`'s `changed(...)` call, `emit`'s `handlers.each`, `changed`'s callback loop) — this is what makes reentrant `on`/`off`/`emit` calls from inside a handler safe rather than deadlocking, and it's exercised by a real concurrency test (`spec/lib/common/events_spec.rb:217-226`).
- **Caught its own bug mid-development.** The third commit (`f4241f91`) found and fixed a real defect — a named re-registration moving to a different topic family only fired `on_change` for the new family, which would leave `Combat::Messages`' hook installed with nobody listening — with both a targeted `events_spec` case and a `Combat::Messages`-level regression test proving the hook actually comes down.
- **Compatibility claim is verified, not asserted.** `gh search code` against both `elanthia-online/scripts` and `elanthia-online/dr-scripts` for `Combat::Observers` and `Tracker.on` returns nothing, confirming the PR body's claim that nothing downstream depends on the deleted facade or the old Symbol-topic callback shape.
- **Clean removal.** No dangling references to `Combat::Observers` anywhere in the tree post-deletion; the deleted spec's coverage is fully subsumed into `events_spec.rb` and `messages_spec.rb`'s renamed equivalents.

## Coverage notes

- Full local clone; reviewed the diff, all changed files at head, and relevant surrounding files (`lib/common/script_death.rb`, `lib/common/hook_registry.rb`, `lib/common/downstreamhook.rb`) not in the diff but load-bearing for the ownership/cleanup contract.
- Ran the PR's own targeted suite (413 examples, 0 failures) and the **full** repo suite (7564 examples, 0 failures) locally — no regressions, and notably none of the 59 "environmental" failures the PR body mentions reproduced in this environment.
- `rubocop` on all changed Ruby files: no offenses.
- Ran `coderabbit review --agent` against `main`; it returned 3 findings (1 major, 2 minor). All three were checked against the actual code and history and rejected as false positives: the "major" (`messages.rb:162`, claiming `refresh!` should run once at load) doesn't change behavior since the initial `@active = [].freeze`/`@hook = false` state already matches what `refresh!` would compute with zero subscribers, and require order guarantees no subscription exists before that `on_change` registration runs; the `events_spec.rb:12` finding about not restoring `Script.current` is moot because `spec/spec_helper.rb:70` already resets it globally before every example; the `combat-observation-provenance.md:95` finding asks for an unrelated rewrite of pre-existing documentation this PR didn't touch beyond a one-line rename. None are included above.
- Did not have access to the `dr-scripts` or `scripts` checkouts in this session to grep them directly; relied on `gh search code` against GitHub instead, which is why the compatibility claim is marked verified via that path rather than a local grep.
- No open questions remain that I could not settle from the repo, history, or a live search.

Nisugi:
Thanks for the review. Both items are addressed in a2cd5a54.

**`refresh!` atomicity:** the new family list is now computed into a local, `install!`/`uninstall!` runs, and only then is `@active` committed. A raise from the hook call leaves the previous consistent `@active`/`@hook` pair in place instead of splitting them.

Your reproduction was exactly right, and it turned out to be load-bearing: once the reorder was in, four `messages_spec` examples failed whenever `events_spec` loaded first, because they had only been passing on `@active` being committed ahead of the mock's `NoMethodError`. The `spec_helper` DownstreamHook mock now defines `add`/`remove` like the production HookRegistry, and there's a regression spec that stubs `add` to raise and asserts `installed?` and `active_families` stay consistent (and that the swallowed error is logged).

**Nit:** the duplicated "the" in the doc comment is gone.

Targeted messages + events specs: 34 examples, 0 failures. Rubocop clean. Full suite failure set is identical before and after (the pre-existing environmental ones only).

mrhoribu (approved):
## Re-review: fix verified (elanthia-online/lich-5#1619 @ `a2cd5a54`)

Both items from the previous review are resolved in `a2cd5a54`:

- **Minor (`refresh!` atomicity):** fixed correctly — `@active` is now only committed after `install!`/`uninstall!` succeeds, computed into a local `active` first. I re-ran my original repro (a `DownstreamHook.add` that raises) against this commit: `active_families` and `installed?` now stay consistent (both revert to the pre-call state) instead of diverging. Nice catch going further than the minimum fix: `spec/spec_helper.rb`'s global `DownstreamHook` mock previously had no `.add`/`.remove` at all, which is the actual root cause that let this go undetected — several existing `messages_spec.rb` examples were only passing because `@active` used to get committed *before* the mock's `NoMethodError` fired. Fixing the mock to match the real `HookRegistry` surface closes that gap everywhere, not just for this one method. The new regression spec (`spec/lib/gemstone/combat/messages_spec.rb:44-57`) exercises exactly this failure mode.
- **Nit (duplicated "the"):** fixed.

Verified independently: targeted suite 414 examples/0 failures, full suite 7565 examples/0 failures, rubocop clean on all touched files.

No new issues found in the diff (`lib/gemstone/combat/messages.rb`, `spec/lib/gemstone/combat/messages_spec.rb`, `spec/spec_helper.rb` — the only files this commit touches).

**Verdict: Approve.**

PR #1621, opened 2026-09-14
feat(all): add deterministic hook priorities

therealatari:
## Summary
- add optional numeric `priority:` to UpstreamHook and DownstreamHook registration
- execute higher priorities first, preserving registration order for equal priorities and the historical default of 0
- synchronize registration, removal, lifecycle cleanup, and dispatch snapshot capture with a per-registry mutex
- build order from one names/priorities snapshot on each dispatch; no persistent cache to lose invalidations or conceal legacy direct-map changes
- keep sorting and callbacks outside the mutex; tolerate removed hooks and fully clean failed-hook metadata

## Motivation
Hook chains are sequential and may transform or suppress input. Priority lets passive observers run before presentation-only transformations without adding a new parser callback system. The external EOHunter coordination projection is one consumer; Nisugi's Events remains the in-process notification seam.

## Compatibility and dispatch contract
Existing add calls remain at priority 0. Higher numeric values run first. Named replacements retain their equal-priority registration position.

Names and priorities are captured together. Newly registered names and priority changes take effect on the next dispatch. Actions remain live lookups: a removed name is skipped; a same-name replacement before its turn runs the replacement at its original position in the current pass, preserving previous behavior.

The legacy mutable maps remain available. Direct edits are visible on the next dispatch, but bypass synchronization; concurrent callers must use add/remove. No cache means ordering is computed per dispatch rather than amortized across dispatches.

## Review follow-up
Addresses Nisugi's order-cache race (both lost registration and two-snapshot KeyError), direct-map compatibility, upstream nil-guard placement, and Horibu's YARD prose and exception-metadata regression requests.

## Verification
- Full RSpec suite: 7,577 examples, 0 failures (Ruby 4.0.5, seed 43190)
- Focused hook priority/persistence/socket-read tests: 43 examples, 0 failures
- 20 randomized priority-spec runs: 600 examples, 0 failures
- Deterministic before/after race tests: published registry fails both interleavings in both directions; revised registry passes
- RuboCop: all four changed files clean
- YARD parser confirms general priority prose belongs to the method, not the action parameter
- git diff --check clean

No live-session reload or game actions were performed for this revision. GitHub CI must finish on the updated head.


Nisugi:
Gave this a careful pass — the design is sound and the `remove(key)` change in the rescue path is a genuine leak fix (it cleans up the orphaned source/owner/persist entries that `@@hooks.delete(key)` left behind). Three things below, one of which I'd treat as a blocker. All reproduced by executing the code on this branch.

To be clear about provenance: the cached-order race was raised in an earlier review on this PR, including the lost-update mode, which I had missed. I've confirmed both of its failure modes and am folding it in here so everything sits in one place — credit for that one goes to that reviewer, not me.

---

### 1. Unsynchronized cache construction in `ordered_hook_names` — `hook_registry.rb:116-121` (blocker)

Cache construction and `add`/`remove` are unsynchronized, and both hook directions share this code. Two distinct failure modes, both reproduced:

**(a) Lost update — a registered hook silently never dispatches.** If a thread is building the cache when another calls `add`, the `add`'s invalidation is overwritten by the older build's write-back:

1. Thread A enters `ordered_hook_names`, cache is nil, builds from snapshot `{a, b}`.
2. Thread B calls `add("c", ..., priority: 100)`, which sets `@ordered_hook_names = nil`.
3. Thread A assigns its stale `{a, b}` result over that invalidation.

```
registered hooks      : ["a", "b", "c"]
dispatch order        : ["a", "b"]
LOST UPDATE: ["c"] registered but will never dispatch
```

The hook stays registered and `_hooks` reports it, but it never runs until some unrelated `add`/`remove` happens to invalidate again. No exception, no log line. Worth noting that in the repro the dropped hook was registered at priority 100 — the *highest*-priority hook in the registry — so this fails in exactly the direction the feature exists to prevent.

**(b) `KeyError` that tears down the session.** `_hooks.keys` is called twice; a hook added between the two snapshots is absent from `positions`, so `positions.fetch(name)` raises:

```
KeyError: key not found: "c"
  from ordered_hook_names ... in 'Enumerable#sort_by'
```

This one is reachable in normal operation — `DownstreamHook.run` runs on the game reader thread while script threads add and remove hooks freely (`util.rb` `issue_command`/`silver_count`, `map_gs.rb` `squelch-peer`). The raise happens while evaluating the `for key in ...` expression, *outside* `run`'s per-hook rescue, so it escapes through `games.rb:1110` into `handle_thread_error`, which doesn't recognise `KeyError` and tears the session down.

Small tell on that same line: `_hook_priorities.fetch(name, 0)` has a default but `positions.fetch(name)` doesn't.

Suggested fix: synchronize mutation and cache publication, or check a generation counter before publishing; build from a single consistent snapshot. Note that computing `_hooks.keys` once only closes (b) — the lost update survives it. Callbacks must run *outside* any lock, since hooks execute arbitrary script code and can themselves add or remove hooks. `SocketReadHook` in this codebase already guards equivalent state with a `Mutex`, so there's a local precedent to match.

### 2. `next` skips the `nil` guard in `UpstreamHook.run` — `upstreamhook.rb:43`

`next unless action` is inside the `begin` block, while `return nil if client_string.nil?` sits after the `end`. So when a name is removed mid-dispatch, `next` bypasses the nil check entirely.

Concretely: a high-priority hook returns `nil` to suppress a command; a later hook's name was removed mid-dispatch; `nil` then flows into the following hooks instead of returning. The next hook's `client_string =~ /.../` raises `NoMethodError` — and this PR's new `remove(key)` in the rescue then permanently unregisters that *innocent* hook, turning a transient race into persistent state corruption.

`DownstreamHook.run` is fine here: its nil check is the first statement of the loop body, ahead of `begin`. Moving the check to the same position would make the two directions consistent.

### 3. `_hooks` no longer describes what will dispatch — `hook_registry.rb:116`

The cache is invalidated only via `add`/`remove`, so any direct mutation of `_hooks` leaves those hooks permanently undispatched. Previously `run` iterated the map live, so this is a behavioural change in the public surface.

Latent rather than currently failing — the specs that mutate `_hooks` directly (`script_kill_metrics_spec.rb:39`, `script_kill_shutdown_spec.rb:37`, `hook_persistence_spec.rb:17`) don't call `run`. But the new spec needs an explicit `instance_variable_set(:@ordered_hook_names, nil)` to work, and that workaround is the smell: anything reaching `_hooks` directly now has an invisible obligation to invalidate. Fixing #1 with proper encapsulation would likely resolve this too.

---

**Checked and correct**, for the record: priority validation properly rejects `NaN`/`±Infinity`/`Complex` while accepting `Integer`/`Float`/`Rational`/`BigDecimal`; `@ordered_hook_names` is correctly isolated per extending class; `(@x ||= ...).dup` is safe against a nil-deref torn read; and the named-replacement position semantics match the documented behaviour. All 48 specs pass.

mrhoribu (changes requested):
# PR Review: feat(all): add deterministic hook priorities (elanthia-online/lich-5#1621)

**Reviewed:** 4 files, +142/-11 at `90d8c6e0` on 2026-09-14. Based on: full clone (fresh checkout of `feat/hook-priority` off `main`, with full history).
**Summary:** Adds an optional `priority:` kwarg to `HookRegistry#add` so `DownstreamHook`/`UpstreamHook` dispatch in priority order (ties broken by registration order), and as a side effect fixes a real pre-existing bug where a hook removing a not-yet-run hook mid-dispatch would silently collapse the whole in-flight string to `nil`. Backward compatibility is fully intact and the full suite (7,559 examples) plus RuboCop pass clean.

## Minor

### 1. New prose is interleaved between `@param` tags, breaking YARD parsing — `lib/common/hook_registry.rb:25-33`

```ruby
      # @param name    [String]
      # @param action  [Proc]
      # Higher-priority hooks run first. Hooks with equal priority retain their
      # registration order, preserving historical behaviour at the default of
      # zero. A named replacement keeps its original equal-priority position.
      #
      # @param persist  [Boolean, nil] hook lifetime relative to the script
      # @param priority [Numeric] execution priority; higher values run first
      # @return [Proc, false] the stored proc, or false if +action+ is not a Proc
```

YARD attaches a free-text comment line to the *preceding* tag's description, not to the method as a whole. As written, "Higher-priority hooks run first…" will render as a continuation of `@param action`'s description in the generated docs (the same [lich-5 API docs](https://nisugi.github.io/lich-5-docs/) this repo publishes), not as general method prose or as the `priority` param's description. Every other multi-paragraph doc comment in this same file (e.g. `cleanup_on_death` just above, or the original `add` docblock this replaces) puts all prose before the `@param` block and keeps the tags contiguous at the end — this is the first one to break that pattern. Move the three prose lines above `@param name`, or fold them into the `@param priority` line.

### 2. No regression test covers metadata cleanup on the exception path — `spec/lib/common/hook_priority_spec.rb`

The PR description lists "fully clean failed hook metadata" as one of its four goals, and `remove(key)` in the `rescue` branch of both `run` methods (`lib/common/downstreamhook.rb:47-51`, `lib/common/upstreamhook.rb:46-50`) now does delete `_hook_priorities[name]` in addition to the maps it already cleaned. But nothing in `hook_priority_spec.rb`, or anywhere else in the suite (`grep -rn raise spec/lib/common/hook*.rb` finds nothing), registers a hook whose action raises and then asserts the priority entry (or any metadata map) is actually gone afterward. This is the one claim in the PR description that ships without a test backing it — worth a quick add: register two hooks, make the first raise, assert `_hook_priorities` no longer has an entry for it and the second hook still ran.

## Open questions

- `lib/common/hook_registry.rb:104-109`: if a hook removes hook **B** and re-adds a *different* action under the same name **B** while **B** hasn't run yet this pass, the new action runs at B's old dispatch position in the *current* pass (since `action = _hooks[key]` re-reads the live hash by name, and the snapshot only fixes the ordered list of *names*, not their bound actions). I traced this against `main`'s pre-PR code and it behaves identically there — `@@downstream_hooks.keys` is likewise snapshotted once and `@@downstream_hooks[key]` is likewise re-read live inside the loop, so a same-dispatch remove+re-add under one name already ran the new action at the old position before this PR too. This is not a regression, and the PR's stated scope is "tolerate hook **removal**" (which it does — see below), not "tolerate remove-and-re-add." **Settles it:** does any real hook (in lich-5, dr-scripts, or scripts) actually remove and re-register a *not-yet-run* hook by name from inside another hook's action during the same dispatch pass? A grep across dr-scripts/scripts turned up nothing obviously doing this, so it's likely purely theoretical — if so this can stay as a documented limitation rather than a fix.

## What this PR gets right

- **Fixes a real, previously-silent bug as a side effect.** Pre-PR, `DownstreamHook.run` did `@@downstream_hooks[key].call(...)`; if an earlier hook in the chain removed a later, not-yet-run hook (`@@downstream_hooks.delete(key)`), the lookup for that now-gone key returned `nil`, and `nil.call(...)` doesn't raise here — `lib/common/class_exts/nilclass.rb:9` patches `NilClass#method_missing` to return `nil` for any undefined method, `call` included. That silently set `server_string` (or `client_string`) to `nil`, which the very next loop iteration's `return nil if …nil?` turned into the *entire* dispatch returning `nil` — i.e. removing a hook mid-chain could drop the whole game line or command. The new `action = _hooks[key]; next unless action` guard (`lib/common/downstreamhook.rb:43-44`, `lib/common/upstreamhook.rb:42-43`) closes this, and `spec/lib/common/hook_priority_spec.rb:228-235` ("tolerates removal of a later hook during dispatch") exercises exactly this path.
- **`remove` now fully cleans up**, deleting `_hook_priorities[name]` alongside the pre-existing maps (`lib/common/hook_registry.rb:57-65`), consistent with the long-session memory-leak fix in #1410 that this same registry was built to guard against.
- **Return-value compatibility was preserved deliberately, not by accident.** Inserting `invalidate_hook_order!` before the end of `add` would have silently changed its implicit return value from `action` to `nil` (an assignment-returning-nil tail expression) — the author added an explicit trailing `action` to keep the original contract. No caller in this repo currently depends on `add`'s return value, but the discipline is correct.
- **Priority validation is genuinely necessary, not defensive-for-its-own-sake**: `Float::NAN` passed as a priority would make `sort_by`'s `<=>` return `nil` and raise `ArgumentError: comparison of Array with Array failed`, taking down every dispatch until the offending hook was manually removed. The `finite?` guard (`lib/common/hook_registry.rb:39-43`) and its dedicated spec prevent that.
- **Zero contract breakage across the ecosystem.** All ~180 `DownstreamHook.add`/`UpstreamHook.add` call sites across dr-scripts and scripts use `(name, action)` or `(name, action, persist: …)` — fully compatible with the new `priority: 0`-defaulted kwarg, and equal-priority ordering is guaranteed stable via the explicit `positions` secondary sort key rather than relying on `sort_by`'s unspecified stability.
- Full suite (`bundle exec rspec`, 7,559 examples) and `rubocop` on the 4 changed files: both clean.

## Coverage notes

Full clone of `elanthia-online/lich-5`, PR branch checked out via `gh pr checkout`, diffed against `main` at the merge point. Ran the complete spec suite and RuboCop locally (RuboCop's `4.0.1`-targeted invocation fell back to the ambient Ruby 4.0.5, which matches the repo's `TargetRubyVersion: 4.0`, so no material difference). A local CodeRabbit pass (`coderabbit review --agent --base main --committed`) ran to completion and surfaced one finding, which is the remove-and-re-add item folded into Open Questions above after verifying it against pre-PR behavior — nothing else survived. Cross-checked all `DownstreamHook`/`UpstreamHook` consumers in the local `dr-scripts` and `scripts` checkouts for call-site compatibility. Not independently runtime-tested against a live Lich session (no game connection in this environment) — the priority-ordering and mid-dispatch-removal behavior is verified by code tracing and the new unit specs, not by an end-to-end run.

therealatari:
Addressed in 5d097e22. Thanks @Nisugi and @mrhoribu.

1. **Cache races:** removed the persistent order cache. A per-registry mutex now serializes add/remove/lifecycle cleanup and captures one names+priorities snapshot. Sorting and callbacks run outside the lock. Queue-barrier regressions reproduce both the lost-update and KeyError interleavings against the published registry, in both directions, and pass with the fix.
2. **Nil input:** moved the upstream guard to the start of the loop, matching downstream. Coverage includes initially nil input and a suppressing hook that also removes a later hook. Precision note: the exact suppressor-returns-nil scenario in the review already stopped immediately in the previous implementation; I reproduced the initially-nil-input gap, not the described innocent-hook removal chain.
3. **Legacy map edits:** no cache means direct _hooks edits remain visible on the next dispatch. Removed the test's cache-reset workaround and added a direct clear/add/delete regression. These legacy edits still bypass locking; add/remove are the synchronized mutation API.
4. **Exception cleanup:** both directions now have coverage proving all five maps are cleared for a raising hook, the surviving hook runs, and subsequent dispatch works.
5. **YARD:** moved the general prose before contiguous parameter tags; verified its parsed placement with YARD.
6. **Same-pass replacement question:** preserved and documented the existing live-action behavior, with a regression. A replacement runs at the old position for the current pass; its changed priority applies on the next pass. No claim of an ecosystem-wide caller audit in this revision.

Verification: full suite **7,577 examples / 0 failures**; focused **43 / 0**; 20 randomized priority runs **600 / 0**; RuboCop and whitespace clean. Initial sandbox-only socket/interface failures were rerun with loopback access and the full suite passed. No live sessions were touched.

Ready for re-review; CI is running on the new head.

mrhoribu (approved):
# PR Review (update): feat(all): add deterministic hook priorities (elanthia-online/lich-5#1621)

**Reviewed:** delta between `90d8c6e0` (my prior review) and `5d097e22` (current head), 2026-09-15. 4 files, +180/-37 in the delta. Based on: full clone, diffed against both the prior reviewed head and `main`.
**Verdict:** Approve
**Summary:** This commit responds to a review from **Nisugi** (a repo member) that found a real, independently-reproducible race in `ordered_hook_names` plus two smaller issues. I verified all three against the actual code rather than taking either party's word for it — including forcing the exact thread interleaving Nisugi described, against both the pre-fix and post-fix code, outside the PR's own test suite. All three are genuinely fixed. Both items from my own prior review (the YARD doc-comment formatting and the missing exception-path test) are also fixed. No new findings.

## Verification of Nisugi's findings

### 1. Cache race in `ordered_hook_names` (was: blocker) — now fixed

Nisugi's report described two failure modes in the memoized `@ordered_hook_names ||= ...` cache from the version I reviewed: a **lost update** (a concurrent `add`'s cache invalidation silently overwritten by an in-flight reader's stale write-back) and a **`KeyError`** (`_hooks.keys` called twice, so a hook added between the two calls is missing from the `positions` map that `positions.fetch(name)` then raises on).

I reproduced both independently, against the exact pre-fix code (`90d8c6e0`), by forcing the interleaving with an instrumented `Hash#keys` pause rather than relying on natural thread scheduling (naive concurrent stress — 2000 iterations of uncoordinated `Thread.new` — hit neither failure mode, which is exactly why Nisugi's own repro and the PR's new specs both use the same paused-callback technique: the window is real but narrow under MRI's GIL):

```
/tmp/…/hook_registry.rb:119:in 'Hash#fetch': key not found: "c" (KeyError)
	from …in 'block in Lich::Common::HookRegistry#ordered_hook_names'
	from …in 'Enumerable#sort_by'
```

That confirms the bug was real on the code I originally approved with comments.

The fix (`lib/common/hook_registry.rb:14-20, 42-61, 67-69, 128-133`) removes the persistent cache entirely, adds a per-registry `Mutex` (`self.extended` seeds `@hook_mutex` once when `HookRegistry` is extended into `DownstreamHook`/`UpstreamHook`, isolated per class), and rebuilds `ordered_hook_names` from a single `_hooks.keys.each_with_index.map { ... }` snapshot taken atomically inside `@hook_mutex.synchronize` — one `.keys` call, not two, so there's no window for a second read to miss a concurrently-added key. `add`, `remove`, and `cleanup_on_death` all serialize on the same mutex.

Running the identical forced-interleaving repro against the post-fix code: the writer thread correctly **blocks** on the lock instead of racing, and the read that unblocks after reflects both hooks with no crash and no lost update. I also ran the PR's own two new concurrency specs (`_hooks.keys` and `_hook_priorities.fetch` variants, `spec/lib/common/hook_priority_spec.rb:277-313`) 30 times across different `--seed` values with zero failures — they're not flaky.

One correctness detail worth calling out explicitly since it's easy to get wrong: `cleanup_on_death` (`hook_registry.rb:81-96`) calls the private `remove_hook` directly while already holding `@hook_mutex`, not the public `remove` (which itself calls `@hook_mutex.synchronize`) — calling the locking version here would deadlock, since Ruby's `Mutex` isn't reentrant. The split between a public, locking `remove`/`add` and a private, lock-free `remove_hook` is exactly the right shape for that, and the new "runs callbacks outside the registration lock" spec (`hook_priority_spec.rb:315-325`) proves a hook's own action can call `add` (from a spawned thread, under a 3-second `Timeout`) without hanging — which is the general form of the same deadlock risk.

### 2. `UpstreamHook.run` nil-check position (was: major-leaning) — fixed, with an honest correction from the author

Nisugi's narrative was that a suppressed (`nil`) result could flow into a later hook because `next unless action` (inside the `begin`) bypassed the nil check (after the `end`), and that a hook raising on that stray `nil` would then get permanently removed by this PR's own `remove(key)` rescue — "turning a transient race into persistent state corruption."

The fix (`lib/common/upstreamhook.rb:41`) moves `return nil if client_string.nil?` to the top of the loop, matching `DownstreamHook.run`'s existing position — so once `client_string` goes `nil`, dispatch stops immediately and no later hook ever sees it, regardless of mechanism. I checked this against production call sites (`lib/global_defs.rb:2459`, `lib/games.rb:1110`): both always pass a real String, so the specific "nil reaches an innocent hook" chain was likely unreachable in practice given how aggressively `NilClass` is patched elsewhere in this codebase (`lib/common/class_exts/nilclass.rb`) — most method calls on a stray `nil` silently no-op rather than raising. The author's reply says the same thing more precisely: they reproduced the **initially-nil-input gap** (calling `run(nil)` used to still invoke the first hook before this fix) but not the specific innocent-hook-removal chain Nisugi described. That's a fair, checkable correction rather than a dodge — and the fix applied closes the general class of issue either way. `hook_priority_spec.rb:239-246` ("does not pass an initially suppressed input to any hook") now pins this for both directions.

### 3. `_hooks` direct-mutation visibility (was: latent) — resolved as a byproduct

This was a consequence of the old cache, not a separate bug: since there's no more persistent `@ordered_hook_names` to go stale, a hook map mutated directly (bypassing `add`/`remove` — as three existing, PR-untouched specs do in their test setup: `script_kill_metrics_spec.rb:39`, `script_kill_shutdown_spec.rb:37`, `hook_persistence_spec.rb:17`) is now picked up automatically on the next dispatch, with no invalidation call needed. `hook_priority_spec.rb:259-268` exercises this directly. Correctly documented as still bypassing the lock (`hook_registry.rb:118-121`) — that's fine, since none of those callers run concurrently with dispatch.

## Verification of my own prior findings

Both items from my `90d8c6e0` review are resolved:

- **YARD doc-comment ordering** (`lib/common/hook_registry.rb:33-41`): the priority-behavior prose now sits above the contiguous `@param` block, matching the file's established convention (and the author states they checked the parsed YARD output directly).
- **Missing exception-path test**: `hook_priority_spec.rb:225-237` now registers a hook that raises and asserts all five metadata maps (`_hooks`, `_hook_sources`, `_hook_owners`, `_hook_persist`, `_hook_priorities`) no longer have an entry for it, the surviving hook still ran, and the *next* dispatch also works.

My prior open question (same-name remove-and-re-add mid-dispatch running the new action at the old position) is now explicitly documented as intended behavior (`hook_registry.rb:123-126`) and pinned by a dedicated test (`hook_priority_spec.rb:327-340`) that also confirms the *next* pass correctly reflects the new priority — closing it cleanly rather than leaving it implicit.

## What this commit gets right

- Every fix is independently, mechanically verifiable rather than just plausible — I reproduced the crash on the old code and its absence on the new code with the same forced interleaving, not just by reading the diff.
- The lock scope is minimal and specifically shaped to avoid the reentrancy deadlock a less careful synchronization pass would hit (`remove_hook` used internally instead of `remove`; hook actions always invoked outside the lock).
- The response separates "fixed as described" from "fixed, but not for the reason you gave" (item 2) — that's the right way to handle a review finding whose mechanism doesn't fully hold up, and it didn't skip fixing the underlying issue just because the exact narrative was off.
- Full suite: 7,577 examples, 0 failures (independently re-run, matches the author's number). RuboCop clean on all 4 changed files. 30 additional randomized runs of the concurrency specs: 0 failures.

## Coverage notes

Re-cloned fresh and checked out the current PR head (`5d097e22`). Diffed against both my previously-reviewed commit and `main`. Independently reproduced Nisugi's two `ordered_hook_names` failure modes against the pre-fix code via a standalone script outside the repo's spec suite (forced interleaving via an instrumented `Hash#keys`, not relying on natural scheduling), and confirmed their absence against the post-fix code with the identical technique. Ran the full spec suite and RuboCop locally. Ran `hook_priority_spec.rb` 30 times across distinct seeds for flakiness (none observed). Did not re-verify the ~180 dr-scripts/scripts call sites in this pass — the public `add`/`remove` signatures are unchanged from the version I already checked those against, so that finding stands. Not runtime-tested against a live Lich session.

PR #1622, opened 2026-09-14
fix(all): bound move()'s remedy retries, record why a move failed, lift it into Lich::Common::Move

Nisugi:
## Summary

- bound every "fix the obstacle and re-send" branch in `move()`: `MAX_REMEDIES` (3) for remedies that should work first time (stand, unhide, retreat, empty hands, open, stow), `MAX_ROLLS` (20) for climb and swim skill rolls; roundtime waits stay uncapped since they always end
- `Lich::Common::Move.last_failure` after a false or nil return: the direction sent, the game line that ended it, the attempt count, and a `:cause` from a small documented set (`:injured`, `:encumbered`, `:engaged`, `:position`, `:hidden`, `:hands`, `:closed`, `:map`, `:denied`, `:climb`, `:swim`, `:roundtime`, `:unknown`) so callers switch on it instead of matching text
- a stand that keeps failing is attributed from character state (overburdened gives `:encumbered`, limb wounds give `:injured`), because the game says "You struggle, but fail to stand." for both
- `move.failed` emitted on `Lich::Common::Events` when that module exists (#1619), rescued; nothing depends on it
- `move` no longer mutates the caller's string: the climb/go swap and door renumbering used to rewrite wayto strings from the map database in place
- the four exit paths share one finish lambda
- remedies are sent through #1587's bounded `fput` (one resend, then a failure symbol), so `move` no longer carries its own copy of the refusal ladder
- implementation lifted out of `global_defs.rb` into `Lich::Common::Move.move` (lib/common/move.rb) with a two-line top-level shim, the same carve-out pattern as ArgParser, OrderlyShutdown and Stash
- first spec coverage for `move()`: 18 examples driving `Move.move` through a scripted fake stream

## Motivation

Every recovery branch in `move()` follows one pattern: apply a remedy, re-send the direction, reset the give-up clock. That is right for obstacles the remedy actually fixes. It is wrong for stand: a stand that fails for wounds or encumbrance fails every time until something outside `move()` changes, and each try costs 10 seconds of roundtime.

Seen in play: a prone, wounded character on a go2 trip. The game answered "You struggle, but fail to stand." then "Roundtime: 10 sec." on every attempt. `move()` never returned, so go2's step never failed, its restart counter never moved, and its status read `:moving` for as long as anyone cared to watch. The same shape exists in the unhide, retreat, empty-hands and climb branches; the stand case is the one with a reproduction.

The failure reason is the second half. `move()` sees the exact line at the moment it matches. go2 (and any supervisor) currently has to scrape the buffer afterwards to guess why a step failed. Recording the line and a cause where it is matched means one classifier in one place, instead of every consumer regexing the same game text.

## What does not change

- name, arguments, defaults
- the tri-state return: true moved, false drop-the-exit, nil keep-the-exit; every branch returns what it returned before
- every recognition regex, verbatim
- the consumed stream is still restored to the script's downstream buffer on exit

## What does change, for callers that relied on it

- a remedy that never works now returns nil after its budget instead of looping. A route that needed more than 20 consecutive failed climb rolls before one succeeded would now fail; `MAX_ROLLS` is the one constant to raise if such a route exists.
- the caller's direction string is no longer rewritten. A script that read its own variable after `move` to learn the corrected verb would no longer see it. I found no such script in lich-5 or elanthia-online/scripts.
- one new echo line when a budget runs out.

## Constant resolution after the lift

Inside `Lich::Common::Move`, `XMLData` has no `Lich::Common::XMLData` to catch it and falls through to the top-level instance; `Script` and `Spell` resolve to the `Lich::Common` classes that `include Lich::Common` exposes at top level, so they are the same objects the old top-level method used. The spec found this the first time (the real `Lich::Common::Spell` was reached instead of a top-level stub) and now stubs both.

## Builds on #1587 (merged)

Every remedy in `move` (stand, unhide, retreat, empty hands, open, stow, drag) goes through `fput(cmd, timeout: 3, max_resends: 1, failures: :symbol)` from #1587: one refusal-driven resend, then a failure symbol. On the pre-#1587 fput those options were silently ignored and the stand remedy recursed without limit, which is the P1 the first review of this PR found. Rebased onto `main` after #1587 landed; four commits.

Independent of #1619. The `Events.emit` is behind a `defined?` guard. Once both land, go2's `mark_blocked` can read `Move.last_failure` instead of scraping the buffer (elanthia-online/scripts#2471).

## Review history

- P1: the stand loop was still unbounded because remedies went through the old fput, which recurses on "You struggle, but fail to stand." Fixed by routing remedies through #1587's bounded fput.
- P2: the shared skill-roll branch hardcoded `:climb` for swim, drag and guard lines. Fixed: the cause is taken from the matched line (`:swim`, `:drag`, `:denied`, else `:climb`).
- P2: a remedy reply persisted across remedy kinds, so a stand that succeeded could be reported as the line for a later swim that failed. Fixed: replies are kept per remedy kind.

## Verification

- `bundle exec rspec spec/lib/common/move_spec.rb`: 23 examples, 0 failures; with #1587's fput spec and `spec/lib/gemstone/`: 1022 examples, 0 failures
- `bundle exec rubocop`: 1172 files inspected, no offenses
- full suite: the same 14 environmental files (frontend, GTK, wine, Windows launcher) fail identically on pristine `main`, so none are introduced here
- not yet exercised on a live character on this exact branch; the prone-and-wounded case above is the intended manual test


mrhoribu (changes requested):
# PR Review: fix(all): bound move()'s remedy retries, record why a move failed, lift it into Lich::Common::Move (elanthia-online/lich-5#1622)

**Reviewed:** 3 files, +753/-215, at `eca0919` (2026-09-14). Based on: full local clone of `elanthia-online/lich-5`, PR branch checked out fresh from `pull/1622/head`, diffed against `main` (merge base `ce942c0d`, which already includes #1587).
**Verdict:** Request changes
**Summary:** This lifts `move()` out of `global_defs.rb` into `Lich::Common::Move`, bounds its "fix the obstacle and re-send" branches so a remedy that never works can't loop forever (the motivating bug: an unbounded `stand` retry burning 10s of roundtime per try), and records a `Failure` (line + classified `:cause`) callers can read after a `false`/`nil` return. The core retry-budget and per-remedy-kind reply-scoping logic is correct — I hand-traced it against the 23-example spec suite and it holds up — and it genuinely fixes a real in-place string-mutation bug (confirmed via a live external caller, `scripts/go2.lic`, which passes `room.wayto[next_id]` straight from the map's own hash). The one real gap is in the newly-added `:cause` classification: the generic "bad exit, drop it" branch doesn't pin an explicit cause and falls back to generic text matching, which misclassifies at least one exact phrase into a cause that contradicts the branch's own return value.

## Major

### 1. The generic "drop this exit" branch can record a `:cause` that contradicts its own return value — `lib/common/move.rb:276-278`

```ruby
elsif line =~ /^You can't go there|...|You may not pass\.|become impassable\.|.../
  echo 'move: failed'
  finish.call(false)
```

This branch is the direct, unchanged descendant of the old code's `return false` case — the doc comment for `Move`'s `CAUSES` table even describes `:map` as "the game does not know this exit from here (bad wayto...)," which is exactly what a `false` return means here (callers may drop the exit from the map). But `finish.call(false)` passes no explicit `cause`, so `record_failure` falls back to `classify(last_line)` (`lib/common/move.rb:189-198`, `:123-124`).

`classify` walks `CAUSES` in order (`lib/common/move.rb:55-68`), and `:denied` is checked *before* `:map`:

```ruby
[:denied, /may not pass|unseen force prevents|.../i],
...
[:map, /can't go there|can't (?:go|swim) in that direction|.../i],
```

`You may not pass.` is one of the literal alternatives in this branch's own trigger regex, and it is also literally in the `:denied` pattern. So when this exact line fires, `move()` returns `false` (drop the exit — a map problem) but `Move.last_failure.cause` comes back `:denied` (per the table's own doc comment, "an NPC or rule refused entry" — the semantic for the *nil*-returning branch at `lib/common/move.rb:279-282`, not this one). A caller written against the documented cause table (exactly the kind of caller this PR is designed to enable, per the PR body's go2 example) would read `:denied` and could reasonably decide to keep the exit and retry later, when the move's own return value says the opposite.

Separately, roughly ten of this branch's other phrasings ("You settle yourself on...", "You shouldn't annoy...", "That's probably not a very good idea...", "You walk over to...", "Your attempt fails", "There doesn't seem to be any way to do that at the moment.", etc.) aren't covered by any `CAUSES` pattern at all and silently degrade to `:unknown`, even though the branch as a whole is squarely the `:map` case by design. That part is a coverage gap rather than a contradiction, but it means `:map` — the cause this branch exists to report — is under-produced by exactly the branch that should produce it most.

**Fix:** pass the cause explicitly, the same way the `:closed` and `:injured` branches already do: `finish.call(false, :map)`, and drop this branch's phrases from the generic fallback classification (or leave `classify` as a pure best-effort backstop for callers who use `Move.classify` directly on arbitrary text, separate from what `move()` itself records). Confirmed by re-reading the branch and the `CAUSES` table at head; not something I could talk myself out of — the same string is a literal alternative in both regexes with opposite return-value implications.

## Minor

### 2. Three remedies bypass `remedy.call` and can each corrupt an unrelated remedy kind's saved reply — `lib/common/move.rb:331`, `:349`, `:406`

The PR's own commit history (P2 finding, "a remedy reply persisted across remedy kinds") fixed exactly this class of bug for the tracked remedies by scoping `remedy_replies` per `kind`, set via `current_kind = kind` at the top of every `remedy.call` (`lib/common/move.rb:243`), and it has a dedicated spec for it ("does not blame a later remedy on an earlier one that succeeded", `spec/lib/common/move_spec.rb:168-174`). Three call sites never go through `remedy.call` at all, so they never update `current_kind` before calling `command`, and each writes into `remedy_replies[<whatever current_kind was last set to>]`:

```ruby
# lib/common/move.rb:324-333 (drag fix — one-shot via tried_fix_drag, not remedy.call)
elsif line =~ /^You can't drag/
  if tried_fix_drag
    finish.call(false)
  elsif (dir =~ /^(?:go|climb) .+$/) and (drag_line = ...)
    tried_fix_drag = true
    ...
    command.call("drag #{name}")   # <- current_kind is whatever it last was
```

```ruby
# lib/common/move.rb:344-351 (open fix — one-shot via tried_open, not remedy.call)
tried_open = true
command.call(dir.sub(/go|climb/, 'open'))   # <- same
```

```ruby
# lib/common/move.rb:402-408 (electricity/lightning — no flag at all, not remedy.call)
command.call('stand') unless standing?      # <- same
```

Concrete trigger: a `move('climb rope')` call first hits the `:stand` remedy branch (`lib/common/move.rb:360-368`) and gets a real (non-symbol) reply from `fput`, e.g. some non-"struggle" refusal text — `remedy_replies[:stand]` is set and `remedies[:stand]` is still under `MAX_REMEDIES`, so `move()` keeps going. Before `:stand` is retried again, the character is also auto-dragging a body and the next line matches `^You can't drag`; the drag-fix branch fires, `current_kind` is still `:stand` from the earlier remedy, so its `command.call("drag ...")` reply overwrites `remedy_replies[:stand]`. If `:stand` is later invoked again and this time exhausts `MAX_REMEDIES`, the give-up path (`lib/common/move.rb:238-241`) reports the drag command's reply as the reason the stand kept failing, not the actual stand-refusal text.

This only corrupts the diagnostic `line` (the `cause` argument to `remedy.call` is always passed explicitly and is unaffected), and it needs two different obstacle types to interleave within one `move()` call, so it's narrower than the bug the P2 commit fixed for the tracked remedies — but it's the same mechanism, in the same file, just missed for these three untracked call sites.

**Fix:** either route these three through `remedy.call` (drag and open already have their own one-shot guard, so they'd need a `budget: 1`-equivalent, or keep the boolean guard and just set `current_kind` before calling `command`), or give `command` an explicit kind argument instead of relying on the ambient `current_kind`.

## Nits

### 3. Two branches remain unbounded, which the PR's own framing implies they shouldn't be — `lib/common/move.rb:381-383`, `:402-408`

The PR body says it "bound every 'fix the obstacle and re-send' branch in `move()`", carving out only roundtime/stun waits as intentionally uncapped "since they always end." Two branches don't fit either bucket cleanly:

```ruby
elsif line =~ /^Sorry, you may only type ahead/
  sleep 1
  put_dir.call
```

```ruby
elsif line =~ /The electricity courses through you.../
  sleep(0.5)
  wait_while { stunned? }
  waitrt?
  command.call('stand') unless standing?
  waitrt?
  put_dir.call
```

Neither is wrapped in `remedy.call`, so neither has a `MAX_REMEDIES`/`MAX_ROLLS` ceiling; both would loop exactly as unboundedly as before this PR if the same line kept recurring. In practice both are plausibly self-limiting by game mechanics (typeahead throttling backs off once the client slows down; the electricity/lightning message is tied to a boltstone trap that "shatters into glinting fragments" on firing, per the PR-#980 history for that branch — `git log -S"The electricity courses through you"`), which is likely why they were left out, but that reasoning isn't stated anywhere the way it is for the stun/roundtime waits, and it's the same class of gap the rest of the PR closes. Worth a one-line comment explaining why these two are exempt, or wrapping them for consistency, given the modest cost.

## Open questions

- `lib/common/move.rb:70-78`: `Move.@last_failure` is class-level state guarded by a single `Mutex`, not scoped per script or thread. `Script.current` (`lib/common/script.rb:1026-1030`) is thread-bound, and Lich commonly runs multiple `.lic` scripts concurrently against one character session, each on its own thread. If two scripts both call `Move.move` around the same time (e.g., a navigation script and an independent companion/follow script), one script's `finish.call` can overwrite `@last_failure` between the other script's own `move()` return and its subsequent read of `Move.last_failure` — a TOCTOU race, and the exact pattern the PR body's go2 example assumes (`move()` then immediately check `last_failure`). **Settles it:** is it realistic for two scripts to call `Move.move` concurrently against the same character in normal use, or is navigation conventionally single-owner enough (by social convention or a lock elsewhere) that this can't actually interleave? If it can, `last_failure` probably needs to be thread-local rather than a single shared ivar.

## What this PR gets right

- **The core budget/reply-scoping logic is correct.** I hand-traced `remedies`, `remedy_replies`, and the give-up path against several of the spec's scenarios (the stand-loop cases, the "does not blame a later remedy" case) independently of running the suite, and the accounting checks out exactly as designed.
- **The in-place string mutation fix is real, not theoretical.** `scripts/go2.lic:2462` calls `result = move room.wayto[next_id]` — passing the live String stored in the room's own `wayto` hash directly, not a copy. The old `move()`'s `dir.gsub!`/`dir.sub!` calls in the climb/go-swap and drag-fix branches (unlike the door-renumbering branch, which did defensively reassign `dir = dir.to_s` first) would have permanently corrupted that map entry in place. `dir = dir.to_s.dup` at the top of the new `move()` (`lib/common/move.rb:173`) closes this for every branch at once.
- **`Events.emit` is called outside the mutex** (`lib/common/move.rb:126-132`, after the `synchronize` block closes), so a listener that calls back into `Move.last_failure` can't deadlock against `record_failure`'s own lock.
- Clean YARD-style documentation throughout, matching this repo's existing convention for the API-docs generator.
- Good spec coverage for a first pass at testing a function that had none before (23 examples), including specifically the P1/P2 regressions found in earlier review rounds.

## Coverage notes

- Full local clone (`gh pr checkout`-equivalent via `git fetch origin pull/1622/head`), not just the diff; every changed file read in full at head. The user's own `~/lich-5` checkout has unrelated uncommitted work on a different branch and was left untouched — this review used a separate clone in the scratch directory.
- `bundle exec rspec spec/lib/common/move_spec.rb`: 23 examples, 0 failures. `bundle exec rspec` (full suite): 7594 examples, 0 failures. `bundle exec rubocop` on the three changed files: no offenses. All run locally against this checkout, matching the PR's own verification claims.
- Cross-repo check: grepped `dr-scripts` and `scripts` (sibling checkouts in `~`) for `move(` callers; the only direct-argument caller I found reading a live map-database string is `scripts/go2.lic`, used above to ground the string-mutation fix. Did not exhaustively check every `.lic` file in both repos.
- CodeRabbit CLI is installed and authenticated, but the run against this PR returned `Rate limit exceeded` (org has used its 3 included reviews, ~48 min cooldown) — no second-reviewer pass was available; all findings here are from manual tracing and cross-repo grepping.
- Did not exercise this on a live character; neither did the PR author, per their own "Verification" section. The prone-and-wounded stand scenario that motivated this PR is the natural manual test and wasn't reproduced here.
- Did not individually re-verify all ~30 of `move()`'s pre-existing recognition regexes against real game logs — the PR states they're carried over verbatim, and a diff against `main`'s prior `global_defs.rb` confirms the regex text is byte-for-byte unchanged from the old implementation.

Nisugi:
Thanks for the trace - all three findings held up on re-read. Addressed in 3d6eb190:

1. **Wrong cause on the drop-the-exit branch.** The generic `false` branch now passes `:map` explicitly, and the exhausted drag fix passes `:drag`. `CAUSES`/`classify` are unchanged so `Move.classify` still reads "You may not pass." as `:denied` for callers feeding it arbitrary text, but `move()` itself no longer records that.
2. **Untracked fixes writing another kind's reply.** `command` takes the remedy kind as an explicit argument (defaulting to the enclosing remedy's kind inside a block); the drag fix, open fix and boltstone stand name `:drag`, `:open`, `:trap`. New spec reproduces your interleave (stand reply, then the open fix's reply, then stand exhausts) and fails on the old code with the open reply blamed on the stand.
3. **Typeahead / boltstone unbounded.** Both go through `remedy.call` now: typeahead with `MAX_ROLLS` and cause `:roundtime`, the trap with `MAX_REMEDIES` plus a comment that it should not recur anyway.
4. **`last_failure` thread safety.** Concurrent movers are realistic (go2 plus a follower script on one character), so it is thread-local now, keyed on the calling thread, and the mutex is gone. Spec covers it.

27 move examples green, rubocop clean.

mrhoribu (approved):
# PR Review (round 2): fix(all): bound move()'s remedy retries, record why a move failed, lift it into Lich::Common::Move (elanthia-online/lich-5#1622)

**Reviewed:** delta since round 1 — commit `3d6eb190` (2026-09-14), on top of the previously-reviewed `eca0919`. 2 files, +56/-21 in `lib/common/move.rb`, +31 in `spec/lib/common/move_spec.rb`.
**Verdict:** Approve
**Summary:** All four round-1 findings are fixed, correctly and precisely — I re-traced each one against the new code rather than the commit message, and independently hand-traced the new "does not let a one-shot fix overwrite what the game said to the stand" spec to confirm it exercises the actual bug (it does). No new issues found in the diff.

## Findings from round 1 — verified fixed

1. **Cause misclassification on the "drop this exit" branch** (was Major). `lib/common/move.rb:289` now calls `finish.call(false, :map)` explicitly instead of relying on `classify(last_line)`; the exhausted drag fix at `lib/common/move.rb:337` now passes `:drag` the same way. Confirmed the specific contradiction from round 1 is closed: `"You may not pass."` now records `:map`, not `:denied`. The new spec (`spec/lib/common/move_spec.rb:120-126`) checks this plus two of the other previously-`:unknown` phrasings directly.

2. **Stale `current_kind` leaking into an unrelated remedy's saved reply** (was Minor). `command` now takes the kind as an explicit, defaultable argument (`|cmd, kind = current_kind|`, `lib/common/move.rb:228`), and the three call sites that used to bypass `remedy.call` now name their own kind explicitly: drag fix → `:drag` (`:342`), open fix → `:open` (`:360`), boltstone stand → `:trap` (`:420`). I hand-traced the new spec at `spec/lib/common/move_spec.rb:191-198` (stand fails once and its reply is saved, a closed-gate open-fix reply comes next, then stand exhausts its budget) against the new code: `remedy_replies[:stand]` and `remedy_replies[:open]` land in separate slots, and the give-up path correctly reports the stand's own "overburdened" reply, not the gate's. This is exactly the P2-class bug from the PR's own earlier rounds, now closed for these three sites too.

3. **Typeahead and boltstone-trap branches left unbounded** (was Nit). Both now go through `remedy.call`: typeahead at `lib/common/move.rb:392-394` (`MAX_ROLLS`, cause `:roundtime`), the trap at `:413-422` (`MAX_REMEDIES`, cause `:unknown`, with a comment noting it shouldn't recur in practice but is bounded anyway). Matches the PR's original "bound every branch" claim without exception now.

4. **`Move.last_failure` as unscoped shared state** (was an open question). Replaced the mutex-guarded class ivar with `Thread.current[LAST_FAILURE_KEY]` (`lib/common/move.rb:76, 83, 131, 145`) — per-thread rather than per-process. I checked this isn't a fiber-local/thread-local mismatch trap (a real Ruby gotcha, since `Thread#[]` is fiber-local, not thread-local): grepped the repo for `Fiber.new`/`Fiber.yield`/`.resume` and found none — scripts run on genuine native threads (`Thread.new` in `lib/common/script.rb`), never Fibers, so fiber-local and thread-local coincide here and the fix is sound. The new spec (`spec/lib/common/move_spec.rb:129-133`) confirms a second thread sees `nil` rather than the first thread's failure.

## New findings

None. The diff is narrowly scoped to the four items above, with a test for each; I didn't find scope creep or a new issue introduced while fixing these.

## Coverage notes

- Re-cloned fresh and checked out the PR branch at its new head (`3d6eb190`) rather than reusing the round-1 checkout (which didn't survive between sessions).
- `bundle exec rspec spec/lib/common/move_spec.rb`: 27 examples, 0 failures (up from 23). `bundle exec rspec` (full suite): 7598 examples, 0 failures (up from 7594, matching the 4 new specs). `bundle exec rubocop` on the two changed files: no offenses.
- Did not re-run CodeRabbit (still rate-limited from round 1, per the author's own reply on the thread; not checked again).
- Only `lib/common/move.rb` and `spec/lib/common/move_spec.rb` changed in this round — `lib/global_defs.rb` (the shim) is untouched, so the round-1 verification of that file still stands.

PR #1625, opened 2026-09-14
fix(dr): halt script when too distracted to turn recipe book

MahtraDR:
## Summary

`DRCC.find_recipe` and `DRCC.find_recipe2` handle the "too distracted to turn the book" response (i.e. the character is in combat) by printing `DRCC: Cannot turn book, assuming engaged in combat.` and then calling `fput('exit')`.

`fput('exit')` sends the literal game command `exit` to DragonRealms — it does **not** stop the script. The intent (confirmed) is to stop the script. Because the book was never turned, execution then fell through to `read my #{book}`, so the branch neither exited nor prevented the failing read.

## Fix

Replace `fput('exit')` with Ruby `exit` in both methods, so the script halts before the read.

## Tests

Added specs for `.find_recipe` and `.find_recipe2` asserting each raises `SystemExit` on the distracted response and never reaches the `read my book` call. Both specs were verified to fail against the old `fput('exit')` (execution fell through to the read) and pass after. Full `common_crafting_spec` (26 examples) and RuboCop are green.

## Notes

- Long-standing behavior (added in `880934d3`, predates the #1231 refactor); not introduced by that PR.
- The preceding `fput('look')` is left as-is — out of scope for this fix.
- Independent of #1624 (the `get_crafting_item` halt fix); no overlap.


mrhoribu (commented):
Should this be exit or return a `:failed` or similar that the caller then can determine what to do instead of always EXIT'ing. Should be fine in current version as long as nothing but a script calls this method.

MahtraDR:
> Should this be exit or return a `:failed` or similar that the caller then can determine what to do instead of always EXIT'ing. Should be fine in current version as long as nothing but a script calls this method.

yeah it should... but then the scripts would need to be changed. :D

The current usage is this, and someday I'll get around to fixing it...

mrhoribu (approved):
Reviewed in an isolated worktree against the actual PR head (ffc5fd0), not just the diff.

- Reverted the fix locally and confirmed both new specs fail exactly as described (execution falls through to `read my #{book}`); re-applying the fix makes them pass. Full suite (7583 examples) and RuboCop are clean.
- Traced the halt mechanism through `lib/common/script.rb`: each script runs in its own thread, `Script.__execute` rescues `SystemExit` as a normal completion, and the thread's `ensure` calls `kill` regardless of how the block ended — so bare `exit` here converges on the exact same cleanup path as `Script#exit`. No risk of tearing down the whole Lich process.
- Consistent with the same-author, same-file, currently-open sibling fix in #1624 (same idiom, same test pattern).
- Checked blast radius: `DRCC.find_recipe2` is called from 7 dr-scripts scripts, none of which use its return value — a hard halt on the distracted case is strictly safer than the old silent fall-through.

No blockers/majors/minors. One heads-up: #1624 and this PR both append specs to the tail of `common_crafting_spec.rb`, so whichever merges second will need a trivial conflict resolution there — not a defect in either PR.

PR #1626, opened 2026-09-14
refactor(all): frontend markup translation moves to Lich::Common::Markup

Nisugi:
## What

`fb_to_sf`, `sf_to_wiz`, `strip_xml` and `monsterbold_start`/`_end` move out of `global_defs.rb` into a new `lib/common/markup.rb`. The six global names stay, as thin delegators — scripts in the wild call them unqualified.

Net: `global_defs.rb` loses 148 lines; the new module and its spec add 468.

## Why

Testability. None of these five had **any** coverage, for the reason `spec/lib/global_defs_spec.rb` documents at the top of the file:

> We intentionally do NOT `require 'global_defs.rb'` here. Loading the entire file redefines global methods (respond, get, put, etc.) with production implementations that depend on game infrastructure.

A standalone module can be required on its own, so the new spec exercises all five directly — 29 examples where there were none.

`global_defs.rb`'s own header has asked for this since 2022 (*"this needs to be broken up even more"*). This is the self-contained piece: pure string transformation, no entanglement with script state.

## Behavior

Unchanged. Verified by diffing old against new over 21 representative lines × 5 functions, plus 4 split-element sequences: **byte-identical** output, including the `nil` returns and the trailing CRLF the GSL channel wrappers leave behind.

`sf_to_wiz` was ~75 lines of straight-line rewriting; it's now a pipeline of named translators (thoughts, voln, familiar, death, room, bold). The four stream/tag regexes that `sf_to_wiz` and `strip_xml` both used are named constants shared between them rather than duplicated literals.

## Two hazards the move introduced

Both were caught by the output diff, not by the tests — worth calling out for reviewers, since neither would have been obvious in review:

**1. `$~` clobbering in the thoughts translator.** Rewriting it to non-mutating `gsub` broke it: `$1.gsub(...)` runs a match internally, which resets `$~` and blanks `$2`. Reading `$2` afterwards raised `NoMethodError` — swallowed by `sf_to_wiz`'s `rescue` into a silently dropped line. Every multi-word ESP channel would have vanished from Wizard frontends with no error anywhere. Both captures are now bound before either is rewritten, with a regression test.

**2. Constant resolution.** `monsterbold_*` now call `::Frontend` explicitly. The bare name meant the top-level constant while this code was top-level; nested in `Lich::Common` it resolves to `Lich::Common::Frontend` instead — usually the same object (`frontend.rb` aliases one to the other at load), but not guaranteed, and divergence is invisible at the call site.

## Not changed

The multiline carry stays in its process-globals (`$sftowiz_multiline`, `$strip_xml_multiline`) rather than becoming module state: `games.rb` clears `$strip_xml_multiline` directly on reconnect (`games.rb:435`), and that remains a supported reach-in.

`strip_xml_simple` / `strip_xml_multiline` were already marked `private` and have no external callers, so they became private module methods.

## Testing

- 29 new examples, passing isolated and in the full suite across several seeds
- Full-suite failure sets diffed line-by-line against clean upstream at the same seed: **identical — zero new, zero fixed** (compared sets, not counts; counts fluctuate across seeds and initially hid hazard 2 above)
- Rubocop clean


MahtraDR (approved):
**Approve.** Reviewed at `7ed80e1`. No blockers, no majors. CI is green, and I verified the "byte-identical behavior" claim by **execution** rather than reading.

## How I verified
Ran the base `global_defs.rb` implementations against the new `Lich::Common::Markup` side-by-side over a differential harness: **336 single-call cases + 6,033 sequence cases** (including 2,000 randomized split-element fuzz sequences exercising the `$sftowiz_multiline` / `$strip_xml_multiline` buffering across 1–4 reads). **Zero mismatches** in output, `$_CLIENT_.puts` side-effects, and buffer-global state at every step. Also: `rubocop` clean on both files, `markup_spec` 29 examples pass, and a correctly-scoped CodeRabbit pass returned no findings on any of the three PR files.

## One thing worth a note (Minor, non-blocking)
`markup.rb` adds `# frozen_string_literal: true`, which `global_defs.rb` never had — so `monsterbold_start` / `monsterbold_end` now return **frozen** strings. Values are byte-identical, but in-place mutation now raises:

```
OLD monsterbold_start << "X"  ->  ok
NEW monsterbold_start << "X"  ->  raises FrozenError
```

Every in-repo caller (`messaging.rb`, `release_installer.rb`) combines the result with `+` and is safe. The only exposure is a wild script that mutates the return in place — an uncommon idiom, but the PR's contract is that the six globals behave as before for unqualified callers, so it's the one seam. `.dup` in the two delegators closes it if you want zero risk; documenting it is also defensible.

## Verified benign (looked risky, isn't)
- The `$~`-clobber fix in `translate_thoughts` is correct — both captures are bound before either `gsub`, and the regression test genuinely fails on the naive rewrite.
- `::Frontend` preserves the original top-level resolution now that the code is nested in `Lich::Common`.
- The voln/thoughts-stream reorder produces identical output (the substitutions target disjoint tag structures — checked a line carrying both).
- The shared regex constants (`SUPPRESSED_STREAMS`, `SPELLS_STREAM`, `DATA_ELEMENTS`, `ANY_TAG`) match the old inline literals byte-for-byte.

## Also good
- `markup.rb` explicitly `require`s `xml_entities.rb` rather than relying on ambient load order (strictly safer than the old `strip_xml_simple`).
- No constant-name collisions in `Lich::Common`; the privatized helpers had no external callers.

## Coverage
Precision and recall stated separately: what I found is real and the equivalence is proven by execution — that does not prove there is *nothing* left, but the surface is small, pure, and I exercised its full state axis by differential fuzzing. Not exercised: live-reconnect clearing of `$strip_xml_multiline` (unchanged by this PR) and the `rescue` error path (would require forcing an exception in a pure string op). Third-party scripts are unknowable from here — their only bearing is the frozen-string Minor above.

<sub>🤖 Review assisted by [Claude Code](https://claude.com/claude-code)</sub>

mrhoribu (approved):
Reviewed at 7ed80e16. Traced every transformation in `sf_to_wiz`/`strip_xml`/`strip_simple` against the pre-PR inline code in `global_defs.rb` — the four shared regex constants match the original patterns character-for-character, and the one reordering I found (the plain `<stream id="thoughts">` branch now running before `translate_voln` instead of after, since both got folded into `translate_thoughts`) is between two regexes that can never match overlapping text, so it can't change output.

The two hazards called out in the PR body check out:
- `translate_thoughts` binds `channel, msg = $1, $2` in one multiple assignment before either is touched by `.gsub`, avoiding the `$~`-clobbering bug — pinned by the multi-word-channel regression test.
- The rescue's return value is unchanged: `Lich.log` is `$stderr.puts(...)`, which is always `nil`, so the explicit `nil` after `report_error` matches the old fall-through behavior exactly.

Grepped every caller of the six global names across `lib/` (`main.rb`, `games.rb`, `messaging.rb`, `update/release_installer.rb`) — all call the unqualified delegators, signatures intact including `sf_to_wiz`'s `bypass_multiline:`. `strip_xml_simple`/`strip_xml_multiline` had no external callers to begin with.

Ran the full suite (7610 examples, 0 failures) and rubocop on the changed files (0 offenses on `markup.rb` and `markup_spec.rb`; the offenses rubocop reports elsewhere in `global_defs.rb` are pre-existing and outside this diff's hunk).

Two non-blocking nits, no blockers:
1. `lib/common/markup.rb:115-120` — the comment justifying the explicit `::Frontend` qualifier overstates the risk: since `Markup` is lexically nested inside `Lich::Common`, which directly owns `Frontend` as a submodule, a bare `Frontend` reference would resolve to `Lich::Common::Frontend` via lexical scope regardless of the top-level aliasing in `frontend.rb:928` — all three forms (old bare, new bare, explicit `::`) resolve to the same object in this codebase today. Harmless as written, just worth tightening the comment.
2. No test exercises the `rescue` branch in `fb_to_sf`/`sf_to_wiz` — pre-existing untested behavior being relocated, not a regression, but since the whole point of this PR is closing coverage gaps, a quick example pinning the rescue's `nil` return would close it.

mrhoribu (approved):

PR #1627, opened 2026-09-14
test(all): characterize do_client's command dispatch chain

Nisugi:
## What

60 characterization tests for `do_client`'s `;command` dispatch chain. **No production code changes.**

This is groundwork: `do_client` is 462 lines and 42 branches of ordered `if/elsif`, and only three existing examples touch it. A follow-up PR proposes refactoring that chain into an ordered command registry; these tests are what that refactor gets checked against.

## Characterization, not specification

The assertions pin what the chain does **today** — exact response strings, exact call arguments — rather than what it arguably ought to do. That's deliberate. A refactor that changes either intended or unintended behavior then shows up as a failure and gets a decision, instead of passing silently.

## Ordering is the load-bearing part

The chain is first-match-wins, and the final `else` treats anything unmatched as a script name. So order isn't cosmetic — several branches are only reachable because they sit ahead of a broader one. The `branch ordering` group pins the overlaps a reordering would break:

| must come first | or else |
|---|---|
| `^debuglogs?$` | the `\b` invalid-argument catch swallows it |
| `l5u\s+(.*)` | the bare `l5u` form answers with `--help` |
| `force <name> <args>` | `force <name>` drops the args |
| *every built-in* | the catch-all starts it as a script |

Two rows that earlier revisions of this description claimed — `execname`/`en` ahead of `exec`/`e`, and `lt` ahead of `list` — turned out to be wrong, and the `branch ordering` group deliberately does not pin them. `/^(?:exec|e)(q)? (.+)$/` requires a space straight after `e`/`exec`, so it cannot match `en job …` or `execname job …`; likewise `/^list\s?(?:all)?$|^l(?:a)?$/i` is anchored and cannot match `lt` or `list trusted`. Neither pair overlaps, so neither ordering is load-bearing.

## How it runs

Each example runs a probe in a child process. That's not a stylistic choice — it's the technique `global_defs_spec.rb` already uses for its own `do_client` tests, for the reason documented at the top of that file:

> We intentionally do NOT `require 'global_defs.rb'` here. Loading the entire file redefines global methods (respond, get, put, etc.) with production implementations that depend on game infrastructure.

The probe helper stubs `XMLData.game` (which gates 16 of the 42 branches), `UpstreamHook`, `respond`, and the `Script`/`ExecScript` entry points each branch reaches for.

## Verification

Green tests prove little on their own for a suite like this, so I mutation-tested the parts that matter:

- **Breaking the debuglogs ordering** (so the `\b` catch wins) → the ordering example fails
- **Changing one display message** → only that toggle's example fails
- **Crossing two display accessors** (`roommono` writing `room_links`) → only that toggle's example fails

That last pair matters because the six uniform display toggles share a single probe rather than spawning six subprocesses (each costs ~1s). The mutations confirm the sharing doesn't mask a per-toggle failure.

Full-suite failure sets diffed line-by-line against the same seed without this file: **identical, zero new**. (Compared sets, not counts — raw counts drift across seeds because of pre-existing flaky specs, which is exactly how a real regression hid from me once already.)

Runtime is ~34s for the file, from the subprocess-per-example model.


MahtraDR:
# PR Review: test(all): characterize do_client's command dispatch chain (lich-5#1627)

**Reviewed:** 1 file, +518/-0 at `6f5beb99` on 2026-09-15. Based on: full local clone (`elanthia-online/lich-5`) + PR head worktree; suite executed; ordering guards mutation-tested.
**Summary:** Adds 44 characterization tests for `do_client`'s `;command` dispatch, run via subprocess probes, as groundwork for a later refactor into an ordered command registry. The assertions faithfully pin current behavior; the only issues are two "ordering" tests that guard orderings which aren't actually load-bearing, and a scope claim ("every branch group") the suite doesn't meet.

## Blockers
None. Test-only PR; no production code changes; CI green (rubocop + tests pass); base `main` green. Suite runs clean here: **44 examples, 0 failures** (~5.2s on this machine).

## Major
None.

## Minor

### 1. Two "branch ordering" tests guard an overlap that does not exist — `spec/lib/do_client_spec.rb:461`, `:481`
The PR's central claim is that ordering is load-bearing and the `branch ordering` group pins the overlaps a reorder would break. For two of the six rows that is false: the earlier branch's regex cannot match the later branch's input at all, so order between them is irrelevant and the test cannot fail from a reorder.

- **execname/exec** (`spec/lib/do_client_spec.rb:461`, "prefers ;execname over ;exec"). The exec branch is `/^(?:exec|e)(q)? (.+)$/` — after `e`/`exec` it requires `q`-then-space or an immediate space. `;en …` has `n` after `e`; `;execname …` has `n` after `exec`. Neither matches. Verified:
  ```
  exec_re vs "en job puts 1":        no
  exec_re vs "execname job puts 1":  no
  ```
  So the exec branch can never swallow `en`/`execname`, regardless of order. (Note the source already has exec *before* execname at `lib/global_defs.rb:2553`/`:2556` and `en` still routes correctly — the ordering table's "or else exec/e matches first" is inaccurate here.)

- **lt/list** (`spec/lib/do_client_spec.rb:481`, "prefers ;lt over ;list"). The list branch is `/^list\s?(?:all)?$|^l(?:a)?$/i`. `lt` matches neither alternative (`^l(?:a)?$` needs `l`, optional `a`, end — `lt` fails). Verified: `list_re vs "lt": no`. `lt` never reaches the list branch, so order is irrelevant.

By contrast the other three ordering rows are genuine overlaps and their tests are real guards — I confirmed the overlaps (`l5u`, `force`, `debuglogs` bare-vs-`\b` all double-match) and mutation-tested the l5u one: swapping the two `l5u` branches in `global_defs.rb` turns "prefers ;l5u with an argument over the bare form" red (1 example, 1 failure), green again on restore.

These two tests still pin real behavior (`en`→name path, `lt`→its branch), so they aren't useless — they're mislabeled. The risk is false confidence: the stated purpose is to catch a refactor that breaks ordering, and for these two pairs the suite would stay green through any reordering because there was never an overlap to protect. Recommend either dropping the "ordering" framing for these two or replacing them with a test that actually pins each command's distinct output.

### 2. The `lt` test asserts almost nothing — `spec/lib/do_client_spec.rb:481-489`
```ruby
Script.define_singleton_method(:trusted) { [] }
do_client(';lt')
expect(out).not_to include('no active scripts')
```
On modern Ruby, `;lt` hits `lib/global_defs.rb:2586`, whose body is gated by `RUBY_VERSION =~ /^2\.[012]\./` and otherwise responds `"--- Lich: this feature isn't available in this version of Ruby"`. The test never asserts that message — only that the string `no active scripts` is absent, which is guaranteed for any input that misses the list branch. It also stubs `Script.trusted`, but the code path calls `Script.list_trusted` (and only on old Ruby), so the stub is dead. A positive assertion on the actual output would make this a real pin.

### 3. "Every branch group" is overstated; several groups have zero coverage — coverage claim in PR body / commit message
The body says the tests "pin what it does today — every branch group." They don't. Uncovered branch groups in `do_client` (`lib/global_defs.rb`): `trust`/`distrust` (2560-2585), the old-Ruby `list trusted` path (only the modern "unavailable" path is reachable), `set <var> on|off` (2598), `hmr` (2612), all four GS `infomon` branches (2625-2647), `sk` (2646), and the DR-only branches — `display flaguid` (2648), `display roomid` (2679), `display expgains`/`inlineexp`/`exp-status` (2728-2773), and the DR `banks all|reset|reset all` variants (2798). For a suite whose value is protecting a refactor, those branches will refactor unprotected. Not a blocker for groundwork, but the scope claim should match reality (e.g. "covers the game-agnostic and GS branch groups; DR-specific and trust/set/hmr/infomon/sk branches are follow-up").

## Nits
- `spec/lib/do_client_spec.rb:481` — `Script.trusted` stub is unused (see Minor 2).

## Open questions
- The display-toggle probe stubs each accessor's getter to a constant `false`, so only the toggle-from-`false` transition is pinned; the toggle-from-`true` case is never exercised. **Settles it:** is pinning only the false→true direction sufficient for the planned registry refactor, or should the tests also assert the true→false read-negate path? (Reasoned, not executed: I did not add a true-start case.)

## What this PR gets right
- **Subprocess isolation is correct and load-bearing.** `probe` raises unless the child exits `0`, so a branch that blows up under the stubs fails the test rather than passing silently — this is a genuine strength for a characterization suite.
- **The assertions match the source.** I traced every example against `do_client` (`lib/global_defs.rb:2456-2917`): kill/pause/unpause (bare, `all`, targeted, prefix-match, not-found), list/listall, force with/without args, send/send-to, exec/eq/execname, the six uniform display toggles (messages and accessors verified verbatim), debuglogs (show/set/reject), l5u, GS `banks`/`magic` gating, offline mode, `<c>`-wrapped input, and help. All accurate.
- **The kwargs assertion is right for Ruby 3** — `Script.start($1, :force => true)` collects into a trailing positional hash under a `*a` stub, and the test accepts both `{force: true}` and `{:force=>true}` renderings.
- **Honest, useful mutation testing** already documented for the display-toggle sharing and debuglogs ordering — exactly the right instinct for this kind of suite.

## Coverage notes
- **Access:** full local clone + PR-head worktree at `6f5beb99` (re-verified as current head before delivery).
- **Executed:** `bundle exec rspec spec/lib/do_client_spec.rb` → 44/44 green. Mutation-tested the l5u ordering guard (fails on reorder, passes on restore). Verified regex overlap/non-overlap for all six ordering rows directly in Ruby.
- **Independent reviewer:** CodeRabbit CLI ran against base `main`. It surfaced 7 findings, **all in unrelated pre-existing files** (`setup_files.rb`, `currency_spec.rb`, `qstrike.rb`, `injured.rb`, `stash.rb`, `bank.rb`, `remedy.lic`) — **zero on the PR's changed file**. Those are out of scope for this PR (the review ran repo-wide, not diff-scoped).
- **Precision vs recall:** what I report is grounded/executed. On recall: I checked assertion-accuracy for all 44 examples and the ordering-guard validity for all six rows, but I did not exhaustively mutation-test every non-ordering example (the author reports having done so for the display toggles). The uncovered `do_client` branches listed in Minor 3 are the recall gap in the *suite*, not in my review.
- **Rubocop:** passes in CI; not re-run locally.

## Repo conventions I did not apply
No `CONTRIBUTING.md`/style rule was consulted as an enforcement source. The repo's PR-title lint requires the `type(scope):` form and this PR satisfies it (`test(all):`). Nothing in repo guidance would change the findings above; the requester decides whether the "every branch group" scope claim is worth editing in the PR body.

Nisugi:
Thanks — this caught a real problem, and it had already propagated further than this PR.

## Minor 1: the two non-overlapping "ordering" tests — you're right

Verified both directly:

```
exec_re vs "en job puts 1":        false
exec_re vs "execname job puts 1":  false
list_re vs "lt":                   false
list_re vs "list trusted":         false
```

Worse than a mislabeled test: I'd carried the same two claims into the follow-up refactor branch as authoritative comments, and acted on one of them — registering `;execname` *ahead of* `;exec`, reversing the original chain, to "fix" a shadowing that never existed. That's now reverted to the original order, with both comments rewritten to say these are explicitly **not** ordering constraints.

Both tests are out of the `branch ordering` group and replaced with ones that assert the output each command actually produces. The group comment now names the three genuine overlaps (`debuglogs` bare vs `\b`, `l5u` with args vs bare, `force` with args vs bare) *and* names these two as the pairs that look like constraints and aren't — so the distinction survives the next reader.

## Minor 2: the `;lt` test asserted nothing — correct

`not_to include('no active scripts')` holds for any input that misses the list branch, and the `Script.trusted` stub was dead (the path calls `Script.list_trusted`, and only on old Ruby). It now asserts the message the branch really emits on a modern Ruby, and covers `;list trusted` alongside `;lt`.

## Minor 3: the scope claim — correct, and now actually true

Rather than soften the wording I closed the gap. Newly covered: `trust`/`distrust`/`untrust`, `list trusted`, `set`, `hmr`, all four GS `infomon` branches, `sk`, and the DR-only branches (`display flaguid`, `roomid`, `expgains`, `inlineexp`, `exp-status`, and the four `banks` variants). **44 → 60 examples.**

One of those is pinned as broken on purpose: **`;infomon show full` does not work.** Its capture is `" full"` (leading space) but the source compares against `'full'`, so the full listing never prints. The test asserts `SHOW full=false` twice and comments why. I hit this while writing the refactor — my first draft silently "fixed" it — so it seemed worth nailing down before it gets fixed by accident. Happy to send a separate one-line fix if you want it.

## Open question: toggle direction — good catch

The getters were stubbed to a constant `false`, so only false→true was ever exercised. Added a second shared probe reading `true`, so the read-negate path is now pinned in both directions.

## New tests mutation-tested

Swapping `SK.main`'s two arguments, and crossing two `DRBanking` calls, each fail exactly the one affected example.

---

Two notes on your review that I want to flag as *useful*, not just agreed-with: the observation that `probe` raising on non-zero exit is load-bearing is exactly right and I'd not stated it; and checking the regex overlap for all six rows rather than trusting the PR body's table is what found the bug. The follow-up refactor PR is ready and inherits all of this.

Nisugi:
The follow-up refactor is now open as #1628. It stacks on this PR — the two test commits here appear in its diff until this one merges, then it collapses to just the refactor.

Worth noting for whoever reviews it: the 60 tests here caught five real regressions while I was writing that refactor (the `;force` and `;send` handlers used `match[0]`, which is the matched span only, so `force foo bar` lost its arguments). That's the argument for landing this one first.

The one behavior change over there is the `;infomon show full` typo discussed above — isolated in its own `fix(gs):` commit, and the dispatch-trace diff against main comes to exactly one line.

mrhoribu (approved):
# PR Review: test(all): characterize do_client's command dispatch chain (elanthia-online/lich-5#1627)

**Reviewed:** 1 file, +701/-0 at `62e355b8` on 2026-09-15. Based on: full local clone + PR-head worktree; suite executed; targeted mutation testing.
**Summary:** 60 characterization tests (grown from an initial 44) for `do_client`'s `;command` dispatch chain, run via subprocess probes, as groundwork for a later refactor into an ordered command registry. This is a second pass — a prior review on this PR (`6f5beb99`) flagged two mislabeled "ordering" tests, a weak `;lt` assertion, and an overstated scope claim; the follow-up commit (`62e355b8`) fixes all three, and I independently re-verified each fix rather than trusting the commit message. One new gap survives: the two-argument `;force <name> <args>` test doesn't actually pin that `force: true` is still passed, which I confirmed by mutation.

## Blockers
None. Test-only PR, no production code changes. CI is green (rubocop, both Ruby jobs, PR-title lint). Full suite here: **7641 examples, 0 failures** — no regressions from this file. `spec/lib/do_client_spec.rb` alone: **60 examples, 0 failures** in ~14s.

## Major
None.

## Minor

### 1. The two-argument `;force <name> <args>` test doesn't pin that `force: true` is still passed — `spec/lib/do_client_spec.rb:226-234`

```ruby
it 'passes force through for ;force <name> and ;force <name> <args>' do
  out = probe(<<~RUBY)
    Script.define_singleton_method(:start) { |*a| puts "START \#{a.inspect}" }
    do_client(';force foo')
    do_client(';force foo bar baz')
  RUBY
  expect(out).to include('START ["foo", {force: true}]').or include('START ["foo", {:force=>true}]')
  expect(out).to include('"foo", "bar baz"')
end
```

The bare-`;force foo` case is asserted fully (name plus the `force: true` kwarg). The `;force foo bar baz` case only checks the substring `"foo", "bar baz"` — it never checks that `{force: true}` (or `{:force=>true}`) trails it, even though the production branch is `Script.start($1, $2, :force => true)` (`lib/global_defs.rb:195`), a three-argument call.

I mutation-tested this: removing `:force => true` from that branch (`Script.start($1, $2, :force => true)` → `Script.start($1, $2)`) leaves the full suite green — **60 examples, 0 failures**, including this test — because `START ["foo", "bar baz"]` still contains the substring `"foo", "bar baz"`. Restored afterward; suite confirmed clean again.

Since this suite's entire purpose is to catch a refactor that silently drops behavior, this is exactly the kind of change it's meant to catch and currently doesn't, for this one branch. Fix: assert the full array, e.g. `expect(out).to include('START ["foo", "bar baz", {force: true}]').or include('...{:force=>true}]')`, mirroring the bare-form assertion above it.

### 2. PR body still makes claims the code disproves and the test file itself removed — PR description

The description's "Ordering is the load-bearing part" table still lists these two rows:

| must come first | or else |
|---|---|
| `execname` / `en` | `exec` / `e` matches first |
| `lt` | `list` matches first |

Both are now known to be false — the second commit's own message says so ("the reviewer was right: `/^(?:exec|e)(q)? (.+)$/` requires a space … straight after e/exec, so it cannot match 'en job …' or 'execname job …'") — and I independently re-verified both regexes don't overlap (`exec_re` doesn't match `"en job …"`/`"execname job …"`; `list_re` doesn't match `"lt"`/`"list trusted"`). The test file's own `branch ordering` `describe` block comment now explicitly says these two are "Deliberately NOT here" for exactly this reason (`spec/lib/do_client_spec.rb:655-662`). The PR body was never updated to match, so a reader of just the description gets the disproven claim while the code and commit history disagree with it. Also stale: the opening line still says "44 characterization tests" though the suite is now 60 (confirmed by running it). Not code, so not blocking, but worth a body edit before merge since the table is the PR's central argument.

## Nits
None beyond the PR-body staleness folded into Minor 2.

## Open questions
None — the one open question from the prior review round (whether the true→false toggle direction needed its own pin) was answered by this commit's `toggle_from_true` addition, and I mutation-verified it's load-bearing: flipping the negation in the `display_lichid` toggle (`new_value = !(Lich.display_lichid)` → `new_value = (Lich.display_lichid)`) fails exactly the two `lichid` toggle examples, restored clean afterward.

## What this PR gets right
- **Both fixes from the prior review are correct, not just claimed.** I re-derived the `exec`/`execname` and `lt`/`list` regex non-overlaps myself in Ruby rather than trusting the commit message, and both check out.
- **The `;lt` fix is a real pin now.** `spec/lib/do_client_spec.rb:482-491` asserts the actual "unavailable" message for both `;lt` and `;list trusted` (two occurrences), rather than the previous "doesn't include an unrelated string" non-assertion, and drops the dead `Script.trusted` stub.
- **The scope claim is now accurate.** I walked every branch in `do_client` (`lib/global_defs.rb:2456-2921`) against the spec and everything is covered: trust/distrust/list-trusted, `set`, `hmr`, all four GS infomon branches, `sk`, and every DR-only branch (flaguid, roomid, expgains, inlineexp, exp-status, banks variants) that were previously missing.
- **The `;infomon show full` bug pin is correct.** I independently confirmed `Regexp.last_match(1)` for `"infomon show full"` is `" full"` (leading space) against a `case ... when 'full'` that never matches — the test pins the fall-through to the short listing, which is genuinely how the code behaves today.
- **Subprocess isolation still holds and the true→true-negation direction is now load-bearing**, per the mutation test above.
- **No production regressions**: full 7641-example suite green, rubocop clean on Ruby 4.0.5 (the repo's pinned target).

## Coverage notes
- **Access:** full local clone (fresh, separate from the reviewer's own dirty `lich-5` checkout) + `gh pr checkout` of the PR head at `62e355b8`.
- **Executed:** `bundle exec rspec spec/lib/do_client_spec.rb` (60/60 green) and the full suite (`bundle exec rspec`, 7641/7641 green, no new failures). `bundle exec rubocop spec/lib/do_client_spec.rb` clean under Ruby 4.0.5.
- **Mutation testing performed independently** (not reused from the PR author's or prior reviewer's claims): (1) removed `:force => true` from the two-arg force branch — not caught, see Minor 1; (2) flipped the `display_lichid` read-negation — caught by both `lichid` toggle examples, confirming the new true→false pin is real.
- **Independent regex verification**: re-checked `exec_re`/`list_re` overlap claims and the `infomon show full` capture bug directly in Ruby rather than accepting the commit message's word for it.
- **CodeRabbit**: the automated bot comment on the PR shows "Review skipped: automatic reviews are disabled" for this repo — no CodeRabbit output was available to cross-check.
- **Prior review**: this PR already carries a full review from `MahtraDR` against `6f5beb99`. I treated its three findings as leads to re-verify rather than facts to restate, and only report what I confirmed myself; all three were fixed correctly by `62e355b8` except that its own remaining loose end (PR-body staleness) wasn't in that review's scope.

mrhoribu (approved):
# PR Review: test(all): characterize do_client's command dispatch chain (elanthia-online/lich-5#1627)

**Reviewed:** 1 file, +701/-0 at `e758948c` on 2026-09-15 (re-review after `62e355b8`). Based on: full local clone + PR-head worktree; suite executed; targeted mutation testing.
**Verdict:** Approve
**Summary:** 60 characterization tests for `do_client`'s `;command` dispatch chain, groundwork for a later registry refactor. This is a third pass. My previous review (against `62e355b8`) found one remaining gap — the two-argument `;force <name> <args>` test didn't actually pin the trailing `force: true` kwarg — and noted the PR body was stale relative to the code. Commit `e758948c` fixes the test gap directly, and the PR body has been updated to match (test count and the ordering table both corrected). I independently re-verified both rather than trusting the commit message.

## Blockers
None. Test-only PR, no production code changes. CI green (rubocop, both Ruby jobs, PR-title lint). Full suite: **7641 examples, 0 failures**. `spec/lib/do_client_spec.rb` alone: **60 examples, 0 failures**.

## Major
None.

## Minor
None outstanding — both items from the prior review round are resolved (see below).

## Resolved since last review

### 1. `;force <name> <args>` now pins `force: true` — `spec/lib/do_client_spec.rb:233`

```ruby
-      expect(out).to include('"foo", "bar baz"')
+      expect(out).to include('START ["foo", "bar baz", {force: true}]').or include('START ["foo", "bar baz", {:force=>true}]')
```

I re-ran the same mutation as before — removing `:force => true` from `Script.start($1, $2, :force => true)` (`lib/global_defs.rb:195`) — and it now fails exactly this example:

```
expected "START [\"foo\", {force: true}]\nSTART [\"foo\", \"bar baz\"]\n" to include "START [\"foo\", \"bar baz\", {force: true}]"
1 example, 1 failure
```

Restored the source afterward and confirmed the full file is green again (60/60). The gap I flagged is closed.

### 2. PR body updated to match the code — PR description

The description's opening line now reads "60 characterization tests" (was 44), and the "Ordering is the load-bearing part" table has dropped the `execname`/`en` and `lt` rows that the code and the test file's own comments already disproved. The description's central claim now matches what's actually pinned.

## Open questions
None.

## What this PR gets right
- **The force-arg gap is closed correctly**, not just addressed superficially — the fix mirrors the bare-form assertion directly above it and I confirmed by mutation that it's load-bearing.
- **The PR body was brought in line with the code** rather than left to drift further from an already-corrected test file.
- **No regressions**: full 7641-example suite green, rubocop clean, all CI checks passing.

## Coverage notes
- **Access:** full local clone, checked out at PR head `e758948c` (fetched via `gh`).
- **Executed:** `bundle exec rspec spec/lib/do_client_spec.rb` (60/60), full suite `bundle exec rspec` (7641/7641), `bundle exec rubocop spec/lib/do_client_spec.rb` (clean, Ruby 4.0.5).
- **Mutation testing performed independently**: re-removed `:force => true` from the two-arg force branch; now caught (previously wasn't, per my prior review). Restored and re-confirmed clean.
- **This is a delta review** on top of two prior rounds (an initial review by another reviewer against `6f5beb99`, and my own full review against `62e355b8`); I re-verified the specific fix rather than re-deriving the entire suite from scratch, since nothing else changed in this diff (`git show e758948c` touches exactly one line plus the PR body).
- **CodeRabbit**: still disabled for this repo; no output available.

PR #1628, opened 2026-09-15
refactor(all): do_client's command chain becomes an ordered registry

Nisugi:
> **Stacks on #1627.** That PR's two test commits appear in this diff until it merges; please merge #1627 first and this collapses to just the refactor. The tests are what this change is verified against, so they're worth landing in that order.

## What

`do_client` was **462 lines**: one `if/elsif` chain of 43 branches, 16 of them repeating an `XMLData.game` gate inline. It's now **30 lines** — run the upstream hook, match the lich char, dispatch, fall back to starting a script. The branches live in `Lich::Common::ClientCommands` as registered entries of pattern, optional game gate, and handler.

## Order is load-bearing, and stays that way

Registration order is match order and the first hit wins. **This is not a hash lookup and must not become one** — three pairs genuinely overlap, and the narrower one is only reachable because it's registered first:

| must come first | or else |
|---|---|
| `^debuglogs?$` | the `\b` invalid-argument catch swallows it |
| `l5u\s+(.*)` | the bare form answers `--help` |
| `force <name> <args>` | `force <name>` drops the args |

Both new files say so, and `spec/lib/do_client_spec.rb` pins them.

Equally documented: `;execname`/`;exec` and `;lt`/`;list` **look** like ordering constraints and are not — the `exec` pattern requires a space (or `q`+space) straight after `e`/`exec`, so it can't match `en job ...`; the `list` pattern can't match `lt`. I originally believed otherwise and reordered the exec pair on that basis; @MahtraDR caught it reviewing #1627. Reverted to the original order, and both files now name these as the pairs that aren't constraints, so the real ones don't get diluted by imagined ones.

## Handlers take their MatchData

Rather than reading `$1`/`$2`. Those globals are reset by any later match — including one run inside a call as innocuous as `String#gsub` — which makes them a standing hazard in a file this size. This isn't hypothetical: the same trap bit this refactor's sibling in `lib/common/markup.rb` (#1626), where it turned into a `NoMethodError` swallowed by a `rescue` into a silently dropped line. Passing the match explicitly removes the class of bug rather than avoiding it case by case.

Handlers whose pattern is only a prefix test (`;force`, `;send`, `;hmr`) also receive the full command, because `match[0]` is the matched span alone — `force foo bar` matches only `force foo`.

## The six uniform display toggles become a table

`lichid`, `uid`, `exits`, `stringprocs`, `roomlinks`, `roommono` differed only in accessor and message: ~95 lines of near-identical bodies → a table plus one handler. The non-uniform display branches (DR's `flaguid`, `roomid`, `expgains`, `inlineexp`, `exp-status`) keep their own handlers and original positions.

## Two behavior changes, each deliberate and isolated

**`;infomon show full` didn't work.** `/^infomon show( full)?/i` captured `" full"` and compared it against `'full'`, so it never matched and the command silently ignored its only argument (`Infomon.show(full)` is what keeps zero-valued rows). Moving the capture inside the group fixes it; the comparison is also downcased, since the pattern is case-insensitive and `;infomon show FULL` would otherwise hit the same bug one level down. The argument was undocumented too, so help now lists it.

It's a separate `fix(gs):` commit, not folded into the refactor.

**`;display <toggle> TRUE` did the opposite of what it said.** Every toggle pattern carries `/i`, so an uppercase argument was captured — and then compared case-sensitively against lowercase literals. Matching none of them, it fell through to negating the current value. `;display lichid TRUE` turned the flag **off** when it was already on; `;display lichid FALSE` turned it **on** when it was already off. An explicit argument inverted the flag exactly when it was already satisfied.

The refactor's `toggle_value` transcribes this faithfully, bug included, so the registry stays byte-identical to the chain it replaced. The downcase is a separate `fix(all):` commit, same as the `;infomon` one. It affects all five toggle sites — the six uniform display toggles, DR's `flaguid`, `expgains` and `inlineexp`, and `infomon show_durations` — since they share `toggle_value`.

Found by @MahtraDR in review.

## Registration is serialized

The atomic-publish design was right for the single-threaded reload it was built to fix, but `@staging` was one class-level slot with no lock, and `command` re-reads it on every call. `main.rb:939` runs a thread per attached detachable client, each reading straight into `do_client`, so two frontends running `;hmr client_commands` close together put two threads in `define` at once. `DetachableClientRegistry` — same directory, same threading model — already guards its state with a `Mutex`.

Two outcomes, both reproduced against the unlocked code:

- The second `define`'s `@staging = []` redirected **both** threads' `command` calls onto its array, and whichever published last discarded the other's registrations — silently, no exception.
- One interleaving published an **empty table**: the inner `ensure` restored `@staging` before the outer thread read it, so `@commands` became `[]`. Every `;` command then falls through to `Script.start` as a script name until the next reload.

Serialized with a reentrant guard rather than a plain `Mutex.synchronize`, because `define` nests (that is what `previous` and the `ensure` are for) and Ruby's `Mutex` is not reentrant — the obvious fix deadlocks. `dispatch` is untouched; it only reads the frozen published array. Stress-tested at 800 concurrent defines: zero corrupt or empty tables.

The regression test deliberately avoids a queue handshake pinning one thread inside `define` while another enters — the lock makes that interleaving impossible, so such a test deadlocks against the fix instead of passing. (That was my first attempt at it.) It runs four threads that each register a table only they could have produced and asserts every publish is one thread's work entire; verified failing against the pre-lock code, `expected: 4, got: 1`.

Found by @mrhoribu in review, with an independent reproduction. The empty-table consequence was flagged there as plausible rather than verified — it reproduces.

## Verification

A dispatch probe drove **97 commands through the real `do_client` under both GSIV and DR**, every collaborator stubbed to announce itself, comparing call traces against `origin/main`.

- At the refactor commit: **byte-identical** — same handler, same order, same arguments.
- After the `infomon` fix: differs in **exactly one line**, which is the fix.
- After the toggle fix: differs in the toggle lines, which are that fix.

```
'infomon show full'
  old: Infomon.show:[false]
  new: Infomon.show:[true]
```

That's the useful property — because the refactor itself was byte-identical, this isolates the behavior change to precisely the intended command.

The probe drove **lowercase** arguments throughout, which is why it did not surface the toggle-case bug; @MahtraDR found that by executing both versions directly. The byte-identical claim holds as stated for lowercase input, and the refactor commit is byte-identical for uppercase too now that the downcase has moved out of it.

### The case-insensitivity spec was vacuous

Making the refactor commit case-sensitive should have failed `client_commands_spec.rb`'s "is case-insensitive" example. It passed. Both assertions used `current` equal to `!expected`, so the case-sensitive fall-through to `!current` coincided with the expected answer — the example went green against *either* implementation, including a build where `toggle_value` did not downcase at all.

Rewritten to pass `current` equal to the value being requested, where the two implementations must disagree, and confirmed failing against the pre-fix lib before committing.

The reload spec also stopped writing its edited fixture over the tracked `builtins.rb`; it goes to a `Tempfile` now, so a hard kill mid-test can't leave the tracked file corrupted and parallel runners can't race on it. Flagged by @MahtraDR and CodeRabbit independently.

### The toggle fixture is pinned rather than derived

`do_client_spec`'s `UNIFORM_DISPLAY_TOGGLES` is a hand-typed second copy of `DISPLAY_TOGGLES`, which @mrhoribu flagged as a drift risk. It is not derived from the production constant on purpose: it characterizes what the original if/elsif chain did, so it has to be written independently — deriving it would make the spec agree with production by construction and stop testing anything.

The drift risk is real all the same, so the two are now asserted equal. The characterization stays independent, and an unmirrored change on either side fails loudly instead of silently narrowing coverage. Verified failing when a production message string is altered.

The 60 characterization tests from #1627 pass. **They caught five genuine regressions during this work**: the `;force` and `;send` handlers had used `match[0]`, so `force foo bar` lost its arguments and `send to x hi` broke entirely. That's the case for landing #1627 first.

Full-suite failure sets diffed line-by-line at the same seed: **zero new**. Rubocop clean.

Diffed three ways at seed 1234, after the concurrency fix: this branch is byte-identical to the same branch with the fix commits stashed. Two order-dependent examples (`frontend_spec:213`, `script_lifecycle_spec:819`) fail on this branch but not at the merge base; both pass in isolation and both fail the same way with these commits removed, so they are pre-existing ordering flakes on the branch rather than anything these changes introduced.


Nisugi:
Confirmed and fixed in `fa2b4e99`. Thank you — this is a real bug I introduced, and the reproduction was exact.

## Reproduced

```
entries after initial load: 43
entries after one reload:   86
;k dispatches to: <original handler>   # the appended copy never runs
```

Your diagnosis is precisely right, and it's reachable in normal use rather than only in theory: `HMR.reload` calls `load()`, which re-executes the file body, and **both `;hmr client_commands` and the broader `;hmr common` match these files**. So the failure mode is "edit a built-in, hmr it, watch nothing change" — and the doubling is silent.

## The fix

Took your suggested shape — build and publish atomically. Registrations now happen inside `ClientCommands.define`, which accumulates into a staging list and assigns the published table only once the block completes:

```ruby
def define
  previous = @staging
  @staging = []
  yield
  @commands = @staging.freeze
ensure
  @staging = previous
end
```

Three properties follow:

1. **Reload replaces rather than grows** — 43 entries stay 43, and the reloaded handler is the one that dispatches.
2. **A raise mid-registration leaves the previous table intact**, not a truncated one. This is what "publish only after registration completes" buys beyond fixing the doubling, and it seemed worth having explicitly.
3. **`command` outside a `define` block raises** instead of silently writing somewhere that never gets published.

`builtins.rb` is wrapped in a single `define` block — ignoring indentation, `git diff -w` there is **exactly 2 lines**.

## Tests

New `spec/lib/common/client_commands_spec.rb`, 11 examples covering the registry directly: the three reload behaviors, the guard, first-match ordering, game gating, the whole-command second argument, and `toggle_value`.

The reload examples include one that edits `builtins.rb` on disk, reloads it, and asserts the *edited* handler is what runs — the scenario you described, rather than a proxy for it.

**Verified they're real guards**: I reverted `command` to the old appending behavior and re-ran. All four reload examples fail. A regression test that passes against the bug it's meant to catch is worse than none, so this seemed worth checking before claiming coverage.

## No drift

The dispatch probe across 97 commands under both GSIV and DR still differs from `origin/main` only in the `;infomon show full` fix and its help line — the reload change moved nothing else. Full suite unchanged, rubocop clean.

---

Worth saying: my verification for this PR was a trace diff of a *single* load, which is structurally blind to anything that only manifests on a second one. State that accumulates across loads wasn't something I thought to test.

MahtraDR (approved):
**Reviewed at `fa2b4e99`.** Approving — the refactor is faithful and CI is green. Two minor, non-blocking notes below.

I diffed all 43 original `if/elsif` predicates against the 43 registered patterns and walked registration order end to end: every branch, every `/^GS/`→`:gs` / `/^DR/`→`:dr` gate, and the full relative order are preserved (including the three real overlaps — `debuglogs` numeric→bare→`\b`, `l5u` args→bare, `force` args-inside-handler). Passing `MatchData` instead of `$1`/`$2`, the atomic `define`/`@commands` publish, and the isolated `;infomon show full` fix are all right. Ran `spec/lib/do_client_spec.rb` + `spec/lib/common/client_commands_spec.rb` locally: 71 examples, 0 failures.

### Minor 1 — `toggle_value` changes behavior for uppercase toggle args (a second, undisclosed behavior change in the refactor commit)

`lib/common/client_commands.rb` — `toggle_value` downcases its argument before matching. The original toggles compared the capture **case-sensitively** against lowercase literals, but every pattern carries `/i`, so an uppercase arg used to fall through to `!current` (toggle). Confirmed by executing both versions:

```
arg="FALSE" current=false  old=true  new=false  <-- DIVERGES   (;display lichid FALSE)
arg="OFF"   current=false  old=true  new=false  <-- DIVERGES   (;display expgains OFF)
```

`;display lichid FALSE` when lichid is already off: old code turned it **on**, new code leaves it **off**. The new behavior is arguably the correct one, so I'm not asking you to revert — but it lands in the **refactor** commit, not a labeled `fix(...)`, and it makes the "byte-identical / same arguments" claim false for these inputs. The probe missed it because it drove lowercase; `do_client_spec` only exercises `true`/`false` on the display path, and `client_commands_spec` asserts the *new* `toggle_value` semantics in isolation. Suggest either calling it out in the description as a deliberate second fix or splitting it out.

### Minor 2 — reload spec mutates the tracked `builtins.rb` on disk

`spec/lib/common/client_commands_spec.rb` (~L78-84) overwrites the real tracked `lib/common/client_commands/builtins.rb`, `load`s it, and restores it in `ensure`. A hard kill of the child process (CI OOM/SIGKILL, or a mistimed Ctrl-C) between the two `File.write`s leaves the tracked file corrupted with the injected `EDITED HANDLER`, and parallel runners would race on the shared file. Writing the edited body to a `Tempfile` and `load`-ing that exercises the reload path just as well (`builtins.rb`'s `require File.join(LIB_DIR, ...)` resolves from a temp location). Also flagged independently by CodeRabbit.

_Reviewed with local clone + worktree at `fa2b4e99`; the `toggle_value` divergence was proven by execution, not inference._

mrhoribu (approved):
# PR Review: refactor(all): do_client's command chain becomes an ordered registry (elanthia-online/lich-5#1628)

**Reviewed:** 5 files, +1581/-436 at `4f2f7295` on 2026-09-15. Based on: full clone (base + PR branch + 8-commit history), full test suite run, rubocop run, and a standalone Ruby reproduction of the concurrency claim below.
**Summary:** This does what it says: `do_client`'s 462-line, 43-branch if/elsif chain becomes a 24-line dispatcher over `Lich::Common::ClientCommands`, a small ordered-registry module (`lib/common/client_commands.rb`) plus a 404-line transcription of the original branches (`lib/common/client_commands/builtins.rb`). I walked every one of the original 43 branches against its new registry entry (pattern, game gate, handler body) and found the mapping faithful, including the two intentional, isolated bugfixes and the collapse of the six uniform display toggles into a table. The one real problem is architectural rather than in the transcription: the new `ClientCommands.define`/`command` staging mechanism that makes HMR-reload replace the table atomically is not thread-safe, and the codebase's own runtime model (one thread per attached detachable client) makes a concurrent `;hmr` collision reachable, not just theoretical — I reproduced silent data loss from it directly.

## Major

### 1. `ClientCommands.define`'s staging is shared, unsynchronized class state — two concurrent reloads silently corrupt the table — `lib/common/client_commands.rb:100-107`

```ruby
def define
  previous = @staging
  @staging = []
  yield
  @commands = @staging.freeze
ensure
  @staging = previous
end
```

and `command`:

```ruby
def command(pattern, game: nil, &handler)
  raise 'ClientCommands.command must be called inside ClientCommands.define' if @staging.nil?

  @staging << Command.new(pattern, game, handler).freeze
end
```

`@staging` is a single class-level ivar with no `Mutex`. `command` doesn't cache the array it's appending to — it re-reads `@staging` on every call — so if a second thread calls `define` while a first thread's `yield` is still in flight, the second thread's `@staging = []` silently redirects every subsequent `command` call from *either* thread onto the new array, and whichever thread's `@commands = @staging.freeze` runs last wins, discarding the other's registrations with no exception and no log line.

This isn't hypothetical for this codebase: `;hmr` is dispatched through `do_client`, and `lib/main/main.rb:939` spins up `Thread.new(client) { |attached_client| handle_detachable_client(attached_client) }` for every attached detachable frontend, each running its own `while (client_string = client.gets)` loop straight into `do_client` (`lib/global_defs.rb:2591-2602`). Two frontends attached to the same Lich session (a normal use of the detachable-client feature) each running `;hmr client_commands` or `;hmr common` around the same moment is exactly this race. Notably, `lib/common/detachable_client_registry.rb:5-10` — the sibling registry for that very same multi-thread client model, required by this PR's own test probe — is explicitly documented "Thread-safe registry for clients sharing one detachable listener" and guards its state with a `Mutex`. `ClientCommands` lives in the same directory, serves the same threading model, and doesn't follow that precedent.

I reproduced the data loss directly, outside the test suite, with two threads racing `define` (queue-synchronized to force thread B's `define` to fully complete — publish and restore — while thread A is still mid-`yield`):

```
commands after race: 2
commands: ["^a1$", "^a2$"]
```

Thread B's `command(/^b1$/)` was registered, published (`@commands` briefly held B's one-entry table), and then silently overwritten when A's later publish reasserted its own two entries — B's registration vanishes with no error raised on either thread. A different interleaving (three overlapping `define` calls, or two where the inner `ensure` resolves `@staging` to `nil` at the moment the outer thread reads it for `@commands = @staging.freeze`) can push `@commands` to `nil` outright; the `commands` accessor's `@commands ||= [].freeze` (`lib/common/client_commands.rb:79-81`) would then silently coerce that to an empty, frozen table on the very next dispatch — every `;` command would route to `Script.start` as if it were a script name until the next successful reload. I did not additionally reproduce that specific three-way interleaving, so treat the "collapses to empty" extension as a plausible consequence of the same unguarded-state pattern rather than as separately verified — the two-thread data-loss result above is directly demonstrated.

Neither `spec/lib/common/client_commands_spec.rb`'s reload tests nor `spec/lib/do_client_spec.rb` exercise concurrent `define` calls — all four reload examples are single-threaded. A `Mutex` around `define` (wrapping `previous`/`@staging`/`yield`/`@commands =` in one `synchronize` block, matching `DetachableClientRegistry`'s pattern) closes this; it wouldn't need to touch `dispatch`, since reads of the frozen `@commands` array are already atomic.

## Minor

### 2. The display-toggle test fixture duplicates the production table by hand — `spec/lib/do_client_spec.rb:33-40`

```ruby
UNIFORM_DISPLAY_TOGGLES = {
  'lichid'      => ['display_lichid',      'Changing Lich to display Lich ID#s to'],
  'uid'         => ['display_uid',         'Changing Lich to display RealID#s to'],
  ...
}.freeze
```

This is a second, hand-typed copy of `Lich::Common::ClientCommands::DISPLAY_TOGGLES` (`lib/common/client_commands.rb:62-69`), which is a public constant the spec could have read directly. As written, if a future change edits an accessor or message string in the production table (or adds a seventh toggle) without updating this fixture, the spec doesn't fail loudly with "you forgot the fixture" — it just silently stops testing the real behavior for the edited entry, or never tests the new entry at all, while continuing to pass. Not a bug today; a drift risk for the next person to touch `DISPLAY_TOGGLES`.

## Open questions

- `lib/common/client_commands.rb:100-107`: does anything in Lich's runtime ever invoke `HMR.reload` (and therefore `ClientCommands.define`) from more than one thread in practice today, or is the detachable-multi-client scenario above currently rare enough that this has simply never been hit? **Settles it:** has `;hmr` (on `client_commands`/`common`, specifically) ever been observed run from two attached frontends close together, or is multi-frontend attachment itself an uncommon deployment? This doesn't change whether the bug exists, only how urgently the fix belongs in this PR versus a fast-follow.

## What this PR gets right

- The branch-by-branch mapping is accurate. I independently reconstructed the original 43-branch chain from `origin/main:lib/global_defs.rb` and matched every entry (including all 13 game-gated branches, the six-toggle table's exact position for DR's `roomid` between `uid` and `exits`, and the `force`/`send`/`hmr` handlers that need the whole command string) against `lib/common/client_commands/builtins.rb`, and found nothing dropped or reordered beyond the two documented, isolated fixes.
- The three claimed genuine ordering overlaps (`;debuglogs` exact vs. its `\b` catch, `;l5u <args>` vs. bare `;l5u`, `;force <name> <args>` vs. `;force <name>`) and the two claimed non-constraints (`;exec`/`;execname`, `;list`/`;lt`) all check out — I verified each with direct `Regexp#match` tests against representative inputs (`ruby -e`), not just by reading the patterns.
- Both intentional behavior changes are real fixes, isolated in their own commits, and match what the code does: `;infomon show full` (`lib/common/client_commands/builtins.rb:255-257`) now correctly captures `'full'` instead of `' full'`, and `toggle_value` (`lib/common/client_commands.rb:262-268`) downcases before comparing, fixing the case-sensitive-against-lowercase-literal bug at all five toggle sites (the six-entry table plus DR's `flaguid`, `expgains`, `inlineexp`, and infomon's `show_durations`) — I confirmed all ten call sites route through the shared, fixed helper.
- The atomic-publish design in `ClientCommands.define` (stage into a fresh array, publish only on successful completion, restore `@staging` in an `ensure` so a mid-registration raise doesn't leave a truncated table live) is the right shape for the single-threaded case it was built to fix (HMR reload no longer doubles the table), and is directly tested (`spec/lib/common/client_commands_spec.rb:57-113`) with tests that were verified to fail against the pre-fix appending design.
- `spec/lib/common/client_commands_spec.rb`'s reload test no longer writes over the tracked `builtins.rb` — confirmed at lines 69-94, it edits a copy and loads a `Tempfile`, so a hard kill mid-test can't leave the repo's own file corrupted.
- Handlers taking an explicit `MatchData` (and, for `;force`/`;send`/`;hmr`, the whole command string as a second argument) is a real fix, not just a style change — the PR's claim that `;force foo bar` and `;send to x hi` were broken on `main` via a `match[0]`-only read is consistent with what `git show 6face0bf` describes and what the current handlers now do correctly (`lib/common/client_commands/builtins.rb:105-111`, `115-138`, `222-233`).
- Full test suite (`bundle exec rspec`) passes: 7652 examples, 0 failures. `rubocop` is clean on all three touched production files.

## Coverage notes

Full clone with complete history (all 8 commits on the branch, including #1627's two stacked test commits, reviewed as they'd land if #1628 merges — **#1627 is not yet merged**, so this diff currently includes `6f5beb99` and `62e355b8` from that PR). Ran the full spec suite (7652 examples, 0 failures) and `rubocop` (clean) locally against the PR branch. Verified the five ordering/overlap regex claims and the regex-interpolation-with-embedded-slash detail (`HMR.reload(/#{m[:pattern]}/)`) with standalone `ruby -e` tests rather than by inspection alone. Reproduced the `ClientCommands.define` race directly with a synchronized two-thread script outside the test suite (see Major finding #1) to move it from suspected to confirmed.

`coderabbit review --agent` was attempted but hit the org's usage-based rate limit immediately (`"Rate limit exceeded" ... "waitTime":"39 minutes"`) and returned no findings; this review did not have a second automated pass to cross-check against.

Not independently re-verified: the PR description's claim of a 97-command dispatch probe comparing byte-identical call traces against `origin/main` under both GSIV and DR — I did not re-run that probe myself, but the claim is consistent with the line-by-line branch mapping I did perform, and with the full local test suite passing.

Nisugi:
> does anything in Lich's runtime ever invoke `HMR.reload` (and therefore `ClientCommands.define`) from more than one thread in practice today, or is the detachable-multi-client scenario above currently rare enough that this has simply never been hit?

Answering the deployment half from @Nisugi: multi-frontend attachment is **rare** — the capability is fairly new — but not unused; he has run it himself while testing.

That combination is why the fix landed here rather than as a fast-follow. Rare means usage volume won't surface it. New means adoption is ahead of it rather than behind. And the two populations overlap badly: `;hmr` is a development command, and the people most likely to have two frontends attached are the people doing development — so the race concentrates on exactly the users who run the command that triggers it, rather than being diluted across the whole player base.

The failure mode also doesn't report itself accurately. No exception, no log line — just `;` commands silently routing to `Script.start` as script names until the next successful reload. That arrives as "lich broke after I ran hmr," which doesn't point anywhere near a registration race.

Both findings are fixed in `0ad34f7a` and `c51060c9`. Three things worth re-reading rather than assuming from the review:

**The empty-table outcome reproduces.** You flagged it as a plausible consequence you hadn't separately verified — it is verified now, and by a different route than theorized. `@commands` lands as `[]` directly, because the outer thread reads `@staging` *after* the inner `ensure` restored it; the `@commands ||= [].freeze` coercion never enters into it.

**It is a reentrant guard, not `Mutex#synchronize`.** `define` nests — that is what `previous` and the `ensure` are for — and Ruby's `Mutex` is not reentrant, so wrapping `define` as suggested deadlocks on the nested case. There is a test pinning that.

**The regression test deliberately is not your reproduction.** I wrote it as a queue handshake first and it hung, because the lock makes that interleaving impossible — the test was encoding the bug as a requirement. It now runs four threads each registering a table only they could have produced, asserting every publish is one thread's work entire. Verified failing against the pre-lock code (`expected: 4, got: 1`), and stress-tested at 800 concurrent defines.

On the fixture: I did not derive `UNIFORM_DISPLAY_TOGGLES` from `DISPLAY_TOGGLES`. The duplication is load-bearing — it characterizes the original if/elsif chain, so deriving it would make the spec agree with production by construction and stop testing anything. Instead the two are now asserted equal, which keeps the independence and still fails loudly on unmirrored drift.

One correction to the PR description while I was there: the old "identical, zero new" full-suite claim now names two order-dependent flakes (`frontend_spec:213`, `script_lifecycle_spec:819`) that fail on the branch but not at the merge base. They are pre-existing — byte-identical results with these commits stashed — but the previous wording would have read as false to anyone running the suite.

mrhoribu (approved):
# PR Re-review: refactor(all): do_client's command chain becomes an ordered registry (elanthia-online/lich-5#1628)

**Reviewed:** delta since `4f2f7295` (my prior approved review) → `c51060c9`, 2 new commits, 3 files touched (+162/-7). Full suite (7656 examples) and rubocop re-run clean at head.
**Verdict:** Approve
**Summary:** Both items from my previous review are resolved, and both fixes are well-targeted rather than superficial patches.

## Resolved: concurrent registration race (was Major #1)

`0ad34f7a fix(all): serialize command-table registration` wraps `ClientCommands.define` in a class-level `Mutex` (`lib/common/client_commands.rb:53-56` for `LOCK`, `123-133` for `define`), with hand-rolled reentrancy: a thread already holding the lock (`@define_owner == Thread.current`) calls the now-private `define_unlocked` directly instead of re-entering `LOCK.synchronize`, which Ruby's `Mutex` doesn't support and would otherwise deadlock nested `define` calls. `command` and `dispatch` are unchanged — correctly so, since `command` only runs while a thread holds the lock (inside `define`'s block), and `dispatch` reads the already-frozen, atomically-published `@commands` array, which never needed the lock.

I didn't just read this — I re-ran my own two/four-thread reproduction from the original review against the new code (`race_check.rb`, 100 iterations of 4 threads racing `define`/`command`): zero data loss, every published table was exactly one thread's complete registration set. Against the pre-fix code this failed reliably. The PR's own new tests (`spec/lib/common/client_commands_spec.rb:126-215`) independently cover the same ground plus the nesting/deadlock case (`Timeout.timeout(5)` around a nested `define`), and the full suite passes at 7656 examples with rubocop clean.

This closes the open question from my last review (whether multi-thread `;hmr` collisions are reachable in practice) as moot — the fix doesn't depend on how often it was hit, it removes the race outright.

## Resolved: toggle fixture drift risk (was Minor #2)

`c51060c9 test(all): pin the toggle fixture against the production table` doesn't just point the spec's hand-typed `UNIFORM_DISPLAY_TOGGLES` fixture at the production `DISPLAY_TOGGLES` constant — that would have collapsed the fixture's value as an *independent* characterization of the original if/elsif chain's behavior, which was the point of hand-typing it. Instead it adds a new example (`spec/lib/do_client_spec.rb:310-327`) that asserts the two agree, keeping the fixture's independence while making drift between them fail loudly instead of silently narrowing coverage. This is the better fix for the concern I raised, not just the minimal one.

## What's still true from the original review

Nothing else changed in the diff since `4f2f7295` — the branch-mapping, ordering-overlap, and the two behavior-fix commits I verified previously are untouched. No new findings.

## Coverage notes

Same local clone (`/tmp/reviewpr-lich-5-1628/repo`), fetched and checked out to `c51060c9`. Verified the mutex fix with an independent standalone reproduction (not just reading the code or trusting the PR's own tests). Ran the full spec suite (7656 examples, 0 failures) and rubocop (clean, 3 changed files) locally against the new head. Did not re-verify the 97-command dispatch-probe claim or re-walk the full 43-branch mapping, since neither `lib/common/client_commands/builtins.rb` nor `lib/global_defs.rb` changed in this delta.

PR #1629, opened 2026-09-15
fix(gs): preserve group state after ambiguous disband reply

therealatari:
## Summary

- treat `You have no group to disband.` as an ambiguous failed-disband reply rather than proof that the character is ungrouped
- preserve the observed roster and leader while marking group state unchecked, so the next authoritative `GROUP` response refreshes it
- retain clearing behavior for authoritative `You are not currently in a group.` responses and successful leader disbands
- add focused coverage for leader, follower, empty-group, join, and cooldown behavior

## Why

The same `DISBAND` reply can be seen by two materially different states:

- a character who is already ungrouped; or
- a follower who remains in another character's group and therefore has no group of their own to disband.

Clearing cached membership on that reply alone falsely removed followers from Lich's group model even though the game still considered them grouped. The ambiguous reply now invalidates confidence without inventing an empty roster; an authoritative group response performs the eventual correction.

## Verification

- focused group empty-reply/join/cooldown suite: **36 examples, 0 failures**
- hosted RSpec, syntax, RuboCop, CodeQL, and Actions analysis checks: green
- `git diff --check`: clean


mrhoribu (commented):
This isn't fullproof. When you're in a group already and not the leader, when you do DISBAND it gives the messaging you added but you are still a member of a group.
```
J>disband
You have no group to disband.
Your group status is now closed.
J>group
You are grouped with Nexushealbot who is leading, Butch, Linus, Heracles, Trixie, and Turon.
Your group status is currently open.  You may be joined or taken into another group by the actions of another player.
Click GROUP CLOSE to set your group status to closed, or click GROUP HELP for a list of other options.
```

therealatari:
Fixed the follower case from review and updated against current `origin/main`.

`You have no group to disband.` is now treated as ambiguous: it preserves the observed roster/leader and marks the cache unchecked so the next authoritative `GROUP` response refreshes it. It no longer falsely clears a follower who remains in someone else’s group. The authoritative `You are not currently in a group.` response and successful leader disband still clear membership and reset the leader as before.

Verification on the merged head:
- 36 group empty-reply/join/cooldown examples, 0 failures
- RuboCop clean
- `git diff --check` clean

Latest head: `6d3eb2bf`.

mrhoribu (changes requested):
Code changes look right — the second commit correctly addresses the follower-membership issue from my earlier review comment (ambiguous `You have no group to disband.` reply now just invalidates the `checked` flag instead of clearing real membership), and it's consistent with how `ADDED_TO_NEW_GROUP`/`JOINED_NEW_GROUP`/`HOLD_*_SECOND` already handle the same kind of uncertainty elsewhere in this file. RuboCop and the full `spec/lib/gemstone/` suite pass clean.

One thing to fix before merge: the PR title and description still describe the *first* commit's approach ("clear cached members and leader when that reply arrives"), which was superseded by the second commit. Since the final behavior deliberately does *not* clear members/leader for this reply, please update the title/description to match what actually shipped — right now they'd misdescribe the change for anyone reading it later (changelog, `git log`, etc).

Otherwise this looks good to me.

therealatari:
@mrhoribu The title and body now describe the corrected follower-safe behavior rather than the original clearing approach. The ambiguous `You have no group to disband.` reply preserves the observed roster while marking it unchecked; an authoritative `GROUP` response performs the refresh.

The fix is at `6d3eb2bf`, with **36 focused examples, 0 failures** and hosted checks green. Could you re-review when convenient?

mrhoribu (approved):

PR #1630, opened 2026-09-15
fix(all): update puts to Lich.log

athias:
Frontend.create_session_file wrote its session-descriptor line with puts, which always goes to $stdout. In headless mode $stdout is never redirected to a log file — only $stderr is, at startup in lib/init.rb ($stderr = File.open(debug_filename, 'w')), and Lich.log writes to that $stderr. So this line was the one place still printing straight to the console instead of into the session's debug log.

The detachable-client server in lib/main/main.rb calls create_session_file every time it (re)establishes the listener — including each reconnect when Lich is run with --reconnect. Each of those calls re-triggered the puts, so a long-running headless/reconnecting session (e.g. --login <name> --headless <port> --reconnect) kept re-emitting the descriptor line to the console indefinitely, even with no terminal attached to read it.

Fix: swap puts for Lich.log, so the message is written to the per-run debug log alongside the rest of Lich's diagnostic output, instead of stdout.


mrhoribu:
This will break existing functionality that I know exists as being used by other FE applications to read this response output. Would argue to put an option that makes this more configurable or do both. I'll have to dig up where this is used currently in real world examples, just need a bit to do that.

mrhoribu:
For the life of me I can't find it, maybe it's a figment of my imagination that they rely on the text output instead of the actual file itself being generated. I can't find evidence in the FE's themselves that they use the stdout messaging.

mrhoribu (approved):
# PR Review: fix(all): update puts to Lich.log (elanthia-online/lich-5#1630)

**Reviewed:** 1 file, +1/-1 at `f46dd6e` on 2026-09-15. Based on: full clone (fetched `main` and PR head, walked callers, blame, and tests).
**Verdict:** Approve
**Summary:** Replaces a stray `puts` in `Frontend.create_session_file` with `Lich.log` so the session-descriptor diagnostic line lands in the per-run debug log instead of `$stdout`, matching the sibling diagnostics in the same call path. Verified against the actual reconnect-loop caller and confirmed no other repo or spec depends on the old stdout text.

## Blockers
None.

## Major
None.

## Minor
None.

## Nits
None.

## What this PR gets right

- **The fix matches the actual bug, precisely.** [lib/main/main.rb:919](https://github.com/elanthia-online/lich-5/blob/main/lib/main/main.rb#L919) calls `create_session_file(char_name, ...)` without `display_session:`, so it defaults to `true`, inside the detachable-client server's reconnect `loop` — exactly the "re-triggered every reconnect" path the PR body describes. I traced this call site directly rather than taking the description on faith.
- **`Lich.log`'s destination claim checks out.** [lib/lich.rb:249](https://github.com/elanthia-online/lich-5/blob/main/lib/lich.rb#L249) is `$stderr.puts "...: #{msg}"`, and [lib/init.rb:494](https://github.com/elanthia-online/lich-5/blob/main/lib/init.rb#L494) redirects `$stderr` to the debug log file unconditionally at startup, before the detachable-client thread that calls `create_session_file` ever runs. `$stdout` is never redirected. So the before/after behavior is exactly as claimed: console spam → log file.
- **Consistent with local idiom in the same method.** The `rescue => e` branch two lines below the call site already uses `Lich.log "warning: failed to create session file: ..."`, and the surrounding reconnect loop uses `Lich.log "info: detachable client server listening..."`. This change makes the three diagnostics in that block consistent instead of one being the odd one out.
- **No blast radius.** `create_session_file` has exactly two callers ([lib/main/main.rb:500](https://github.com/elanthia-online/lich-5/blob/main/lib/main/main.rb#L500) passes `display_session: false` and is unaffected; line 919 is the target). No other file in lich-5, dr-scripts, scripts, or ProfanityFE greps for the `"writing session descriptor"` text, so nothing downstream was relying on it appearing on stdout.
- **Tests and lint are clean and already compatible.** `spec/lib/common/frontend_spec.rb` stubs `Lich.log` at the top of the file specifically "for test output," so this change needed no spec updates. Ran `bundle exec rspec spec/lib/common/frontend_spec.rb` (178 examples, 0 failures) and `bundle exec rubocop lib/common/frontend.rb` (no offenses) against the PR head.
- **History supports "oversight, not design."** The `puts` was introduced in #1558 (a large "GUI add configurable frontend registry" feature PR merged 2026-09-08) with no comment or surrounding intent suggesting the console line was meant to stay visible to an operator — it reads as inconsistency introduced alongside a bigger feature, which this PR quietly cleans up.

## Coverage notes

Full clone of `elanthia-online/lich-5` (`main` + PR head `f46dd6ed`), diffed and inspected in context. Traced both call sites of `create_session_file`, read `Lich.log`'s implementation and the `$stderr` redirection in `lib/init.rb`, checked `git log -S`/blame on the changed line, ran the full `frontend_spec.rb` suite and rubocop on the changed file, and grepped dr-scripts/scripts/ProfanityFE for any dependency on the old stdout text (none found). CodeRabbit CLI was not available in this environment, so no automated second-pass was run — not a concern here given the diff's size and the depth of manual tracing. Changelog: lich-5's `CHANGELOG.md` is release-please/conventional-commit generated, not manually maintained per PR (unlike the `scripts` repo), so no changelog entry is expected from this PR.

PR #1634, opened 2026-09-16
feat(all): vendor the lich-6 WebUI as an opt-in launcher (--webui-dev)

Nisugi:
First of eight stacked PRs bringing a GTK-to-WebUI compatibility shim to lich-5. This one vendors the foundation and changes no behaviour.

## What this is

Doug's WebUI from `Lich5/lich-6` at `d952f2fd` (his PRs #4 and #5): contract 2.5.0, validator, adapter, runtime, dispatcher, loopback HTTP/WebSocket server, browser launcher, and native login launcher. Lands as new files under `lib/webui/`, `lib/webui.rb`, `lib/api/webui.rb` and `lib/common/webui_launcher*`, with their 24 specs.

**GTK is untouched.** lich-6 PR #6 — the core cutover that deleted the GTK login and `gtk.rb` — is deliberately not taken. The WebUI only runs when `lich.rbw` is started with `--webui-dev`; without the flag, startup is byte-for-byte the GTK path it was before.

## The hooks into existing code — 17 lines total

- `lib/main/main.rb`: a `--webui-dev` branch ahead of the GTK login
- `lich.rbw`: skip `Gtk.main` under the flag (the WebUI holds the main thread)
- `lib/main/argv_options.rb`: parse `--webui-dev`
- `lib/common/session_launcher.rb`: `force_path_flags` passthrough, so detached Multi-Launch children inherit the parent data dir

Everything else is additive: 58 files changed, 9490 insertions, 1 deletion.

## Not ported

lich-6's CI-foundation workflow, security manifest, `script/ci` and `script/support`. lich-5 has none of that scaffolding and the WebUI does not need it to run.

## Testing

Ported suite passes: `bundle exec rspec spec/lib/webui` → **150 examples, 0 failures**.

Two specs gain `Gem.win_platform?` guards (symlink creation and POSIX mode bits are not available on NTFS); every other assertion in those examples still runs on Windows.

## What follows

Seven more PRs, each building on this one:

| | |
|---|---|
| 02 | `ScriptScope` + the `Gtk` shim namespace — vars/alias/Builder scripts run unmodified |
| 03 | menus, Pango markup, pointer gestures |
| 04 | box packing, scroll adjustments, window presentation, grid weights |
| 05 | map groundwork (allocation, stubbed-widget reporting) |
| 06 | `Gtk::Image`/`Layout`, Cairo markers, composites — `;map` renders |
| 07 | Paned, Overlay, ProgressBar, ListBox, Dialog |
| 08 | key events, map interaction, Win32 window presentation |

Contract goes 2.5.0 → 2.15.0 across the series, additively; `negotiate!` only checks the major, so 2.x clients keep working.


mrhoribu (approved):
# PR Review: feat(all): vendor the lich-6 WebUI as an opt-in launcher (--webui-dev) (elanthia-online/lich-5#1634)

**Reviewed:** 58 files, +9490/-1 at 01d51c8e on 2026-09-16. Based on: full clone, changed files at head, and full repo for cross-referencing.
**Verdict:** Approve
**Summary:** This is a clean vendor-drop. The 17-line integration surface (`lib/main/main.rb`, `lich.rbw`, `lib/main/argv_options.rb`, `lib/common/session_launcher.rb`) is correctly gated behind `--webui-dev` and, traced line by line, leaves the non-flag path byte-for-byte unchanged as claimed. The security-relevant surface (loopback-only HTTP/WS server, session-token cookie auth, single-use short-lived launch tokens, origin/Sec-Fetch-Site checks, CSP headers, `SensitiveValue` redaction-and-zeroing for credentials, path-traversal-safe `FileService`, argv-array `spawn` for the browser launcher) is well built and matches or exceeds what I'd expect from first-party code. `bundle exec rspec` on all webui-touching specs (206 examples) and `rubocop` on every touched/added file both pass clean. The two Windows-guarded specs skip only the assertions the PR body says they skip. No eager `require` of any `lib/webui*` file exists outside the `--webui-dev` branch, so the "changes no behaviour" claim holds. Findings below are Nit-level only.

## Nits

1. **Vestigial dead condition in `terminal_launch`.** `lib/common/webui_launcher.rb:893-897`:
   ```ruby
   def terminal_launch(launch, origin)
     @mutex.synchronize { @launch_result = launch }
     @on_launch.call(launch, origin)
     close(reason: :launch) unless origin == :saved_entry && @persistent
   end
   ```
   The only call site that passes `origin: :saved_entry` is `perform_saved_launch` (`lib/common/webui_launcher.rb:826-845`), and it's reached only from the `else` branch of `if @persistent ... else ... terminal_launch(launch, :saved_entry) end` — i.e. exactly when `@persistent` is `false`. So `origin == :saved_entry && @persistent` can never be true through any live call path today; `close` always runs when `terminal_launch` runs. Confirmed by reading every `terminal_launch` call site (`grep -n terminal_launch lib/common/webui_launcher.rb` → lines 821, 844, 893, 896 only). Not a functional bug — `close` firing unconditionally is the correct current behavior — just a guard clause that reads as live logic but isn't reachable, which will confuse whoever next touches this method. Confidence: confirmed (dead-code claim), low severity (no behavioral effect).

2. **`spec/lib/main/webui_dev_option_spec.rb` verifies source text, not parsed behavior.** All three examples (`lib/main/webui_dev_option_spec.rb:1-43`) read `main.rb`/`argv_options.rb`/`lich.rbw` as strings and assert on literal substrings/`.index` ordering (e.g. `expect(webui_branch).to be < gtk_branch`), rather than actually invoking the ARGV parser and asserting `@argv_options[:webui_dev] == true`. This is grounded as a reasonable choice, not laziness: the neighboring `spec/lib/main/argv_options_spec.rb:6-11` documents that `argv_options.rb` "auto-executes `ArgvOptions.process_argv` at load time" and "cannot be required in isolation," and that existing spec copes by extracting a method body via regex and `module_eval`-ing it into a harness rather than requiring the file directly. The new spec doesn't reuse that harness pattern for the `--webui-dev` line itself, so a reformat that preserves behavior (e.g. reindenting main.rb, or rewording the branch without changing semantics) would break this test without there being a real regression. A behavioral assertion via the same extract-and-eval technique would be more robust, but given the file's documented load-time side effects this is a minor, not a correctness gap. Confidence: confirmed (as written, these are textual assertions).

## Open questions

- None that rise above the nits above — the integration surface, force_path_flags scoping, and security surface all had a concrete grounding fact in the repo (see Coverage notes) rather than needing to be routed back to the author.

## What this PR gets right

- **`--webui-dev` gating is airtight.** `lich.rbw:163` changes `if defined?(Gtk)` to `if defined?(Gtk) && !@argv_options[:webui_dev]`; when the flag is absent this is `defined?(Gtk) && true`, i.e. unchanged. `lib/main/main.rb:197-208` inserts the webui branch as a sibling `elsif` ahead of the pre-existing `elsif defined?(Gtk) and (ARGV.empty? or @argv_options[:gui])` (`lib/main/main.rb:209`), so neither branch is reachable when the other's condition holds. `lib/main/argv_options.rb:83-84` adds a cleanly anchored, non-overlapping regex next to the existing `--gui`/`--no-gui` cases. Verified no other file loads `lib/webui.rb` or anything under `lib/webui/**`/`lib/api/webui.rb` eagerly (`grep -rn "require.*webui" lib/ lich.rbw` finds only the one gated `require` at `lib/main/main.rb:198`, and `lib/api/webui.rb` — unlike its sibling `lib/api/active_sessions.rb`, which is `require`d unconditionally at `lich.rbw:85` — has no boot-time reference at all).
- **`force_path_flags` is scoped correctly.** `lib/common/session_launcher.rb:191` (`return value if context[:force_path_flags] == true`) only fires inside `overridden_path_value` when `context.key?(context_key)` is already true for that specific path flag (`lib/common/session_launcher.rb:185`). The only place that sets `force_path_flags: true` is `WebUILauncher#launch_context` (`lib/common/webui_launcher.rb:1015-1019`), which sets it alongside `data_dir` only — none of the other seven `OPTIONAL_PATH_FLAGS` entries are present in that context hash, so the force-override cannot leak to `home_dir`/`script_dir`/`temp_dir`/etc. The existing GTK caller (`lib/common/gui_login.rb:655`, unmodified) never sets this key, so its behavior is provably unaffected. A dedicated spec (`spec/lib/common/session_launcher_spec.rb:210-224`, "forces the parent data directory into detached Multi-Launch children") demonstrates the flag causes `--data=<value>` to be emitted even when the value equals the process's own `DATA_DIR` default — exactly the scenario the non-forced path (lines 193-197) would otherwise suppress.
- **Server binding is defensively double-checked.** `lib/webui/server.rb:35` rejects any non-loopback `host` at construction (`LOOPBACK_HOSTS = %w[127.0.0.1 ::1]`), and `lib/webui/server.rb:70-75` re-validates the actually-bound address after `TCPServer.new` and raises + closes the socket if it somehow resolved outside loopback — defense against a surprising OS/resolver behavior, not just a single trust-the-input check.
- **Credential hygiene**: `SensitiveValue` (`lib/webui/sensitive_value.rb`) redacts on every serialization path (`to_s`, `inspect`, `as_json`, `to_json`, `encode_with`, `marshal_dump`) and actively zeroes its backing buffer (`clear_value!`, lines 85-90) after the one permitted `consume` block exits; all `Lich.log`/`@logger.call` sites touching auth/launcher errors log `error.class` only, never `error.message` or raw values (checked across `runtime.rb`, `dispatcher.rb`, `browser_launcher.rb`, `catalog.rb`).
- **`FileService` path handling is correct, not just plausible.** `lib/webui/file_service.rb:107-109`'s containment check (`candidate.start_with?("#{root}#{File::SEPARATOR}")`) avoids the classic prefix bug (`/foo/bar` matching a root of `/foo/barbaz`), and `resolve` (lines 56-73) calls `File.realpath` on the joined path before the containment check, so a symlink inside an allowed root that points outside it is caught — verified by the (Windows-partially-skipped) `escape.png` symlink test in `spec/lib/webui/file_service_spec.rb:16-31`.
- **Browser launch avoids shell injection.** `BrowserLauncher.open` (`lib/webui/browser_launcher.rb:35`) calls `spawn.call(*command, ...)` with `command` built as an argv array (`lib/webui/browser_launcher.rb:62`), so the token-bearing `--app=#{url}` argument is never shell-interpreted even though it contains a `SecureRandom.hex(32)` token and query string.
- **No new runtime dependency and no lint/test regressions.** `Gemfile`/`Gemfile.lock` are untouched by this diff — the WebUI relies only on stdlib (`socket`, `json`, `securerandom`, `uri`, `digest/sha1`, `openssl`). `rubocop` reports zero offenses across all 31 new/touched Ruby files under `lib/webui/`, `lib/common/webui_launcher*`, `lib/main/`, and `lich.rbw`. `bundle exec rspec` on every webui-related spec file (206 examples, superset of the 150 the PR body cites for `spec/lib/webui` alone) passes with 0 failures.
- **Windows-guard claim checked and accurate.** `spec/lib/webui/file_service_spec.rb:19-31` only skips the symlink-escape assertion (`expect(service.resolve('images', 'escape.png')).to be_nil if symlink_created`) and symlink creation itself on Windows; every other assertion in that example (traversal, extension allowlist, directory rejection, unregister/revoke) runs unconditionally. `spec/lib/common/webui_launcher/window_geometry_store_spec.rb:20` only skips the POSIX-mode-bits assertion (`File.stat(...).mode & 0o777`), consistent with the PR body's stated rationale (NTFS has no POSIX mode bits).

## Coverage notes

- Read in full: `lib/main/main.rb` (relevant region), `lich.rbw`, `lib/main/argv_options.rb` (relevant region), `lib/common/session_launcher.rb`, `lib/webui/server.rb`, `lib/webui/websocket.rb`, `lib/webui/file_service.rb`, `lib/webui/sensitive_value.rb`, `lib/webui/browser_launcher.rb`, `lib/webui.rb`, `lib/api/webui.rb`, `lib/webui/protocol.rb` (excerpt), `lib/webui/viewer_store.rb`, `lib/common/webui_launcher/serial_executor.rb`, `spec/lib/main/webui_dev_option_spec.rb`, `spec/lib/webui/file_service_spec.rb`, `spec/lib/common/webui_launcher/window_geometry_store_spec.rb`, `spec/lib/common/session_launcher_spec.rb` (relevant sections).
- Skimmed for structure/red flags but not traced line-by-line: `lib/webui/adapter.rb`, `lib/webui/runtime.rb`, `lib/webui/dispatcher.rb`, `lib/webui/contract.rb`, `lib/webui/validator.rb`, `lib/webui/tree_builder.rb`, `lib/webui/registry.rb`, `lib/webui/page.rb`, `lib/webui/component.rb`, `lib/webui/modal_coordinator.rb`, `lib/webui/future.rb`, `lib/webui/submission.rb`, `lib/webui/errors.rb`, `lib/webui/service.rb`, `lib/webui/assets/app.js`, `lib/common/webui_launcher.rb` (most of the 1042 lines beyond the sections cited above), `lib/common/webui_launcher/catalog.rb`. No red flags surfaced in the parts read (attribution/licensing headers, lich-6-only references, GTK cutover leftovers) — none found.
- Did not run `coderabbit review` (time-boxed against the depth already achieved by manual tracing of the integration surface and security-relevant files, which the task explicitly weighted as the priority); did not attempt a native `bundle exec rspec` run of the *entire* suite (only the webui-relevant subset, which passed 206/206) since gtk3/gdk_pixbuf2-adjacent native gems were flagged as a possible sandbox risk and the webui subset is what this PR's own testing claim is about.
- Did not verify runtime behavior in a real browser (Playwright/manual) — this was a static/behavioral-spec-level review only, consistent with the task's framing of this as a large vendor-drop where the priority is the integration surface and security surface rather than re-deriving every widget's rendering correctness.

mrhoribu:
# PR Review: feat(all): vendor the lich-6 WebUI as an opt-in launcher (--webui-dev) (elanthia-online/lich-5#1634)

**Reviewed:** 58 files, +9490/-1 at `01d51c8e` on 2026-09-16. Based on: full clone of `elanthia-online/lich-5` plus the eight stack branches; no Ruby toolchain available, so nothing was executed or linted.
**Summary:** Vendors Doug's lich-6 WebUI wholesale behind a `--webui-dev` flag, with 17 lines of real integration into existing code. The security surface is the best part of the series and the opt-in gating is genuinely airtight; what needs attention is resource lifetime — a browser process per window, and socket writes with no timeout on paths a script thread reaches.

## Major

### 1. Every WebUI window launches a separate Chrome process with its own temp profile — `lib/webui/browser_launcher.rb:30`

```ruby
profile_dir = Dir.mktmpdir('lich-webui-browser-') if on_exit
command = command_for(url, platform: platform, browser_path: ..., profile_dir: profile_dir)
pid = spawn.call(*command, out: File::NULL, err: File::NULL)
```

`profile_dir` is created whenever `on_exit` is supplied, and every caller in the series supplies one — `Session#open_browser` (`lib/common/script_scope/gtk/session.rb`, added in #1635) passes `on_exit: on_exit || proc { ... }`, which is always truthy. The consequence is that a private profile directory, and therefore a completely separate Chrome instance, is created per script window rather than per session. A player running `;map`, `;bigshot` and `;vars setup` gets three full browser processes plus the launcher's, each with its own disk profile.

The temp directories are cleaned by `monitor_process`'s `ensure`, which only runs if the monitor thread survives. A Lich crash or `kill -9` leaks them, and `Dir.mktmpdir` gives no session prefix to sweep later.

This may well be deliberate — a shared profile would make `--app` windows share a process and lose independent lifetimes. If so it deserves a comment saying so, because the cost is not obvious from the call site. If it is not deliberate, the fix is to create one profile dir per `Service` and pass it down.

### 2. Socket writes have no timeout, on paths a script thread reaches — `lib/webui/server.rb:160`

```ruby
def write(bytes)
  return false unless @alive
  @write_mutex.synchronize { @socket.write(bytes) }
  true
rescue IOError, SystemCallError
  @alive = false
  false
end
```

`Runtime#refresh` calls `send_render` → `Connection#send_text` → this, synchronously, on whatever thread called `refresh`. In the shim that is the session thread; through `Lich::WebUI.refresh` it can be a script thread directly. If a viewer's browser stops draining its socket — a suspended laptop, a tab the OS froze, a wedged renderer — the loopback send buffer fills and `IO#write` blocks with no deadline. The script's UI thread stops, and `Session#sync` callers block behind it.

`Runtime#shutdown` has the same shape with no escape:

```ruby
threads = @refresh_mutex.synchronize { @refresh_state.values.filter_map { |state| state[:thread] } }
threads.each(&:join)
```

An unbounded `join` on a thread that is itself blocked in `write` hangs shutdown. `Server#stop` does the right thing (`join_or_kill` with a 0.5s budget); `Runtime#shutdown` should too, and the write itself wants a `IO.select([], [socket], nil, timeout)` guard that marks the connection dead rather than waiting forever.

I have not reproduced this — it needs a stalled browser — so treat the trigger as `suspected`. The absence of any timeout on the write path is `confirmed` by reading.

### 3. `Adapter#handle_for` is a linear scan with deep struct comparison, on the commit path — `lib/webui/adapter.rb:399`

```ruby
def handle_for(node)
  @nodes.key(node)
end
```

`Hash#key` walks every entry comparing values with `==`. `Node` is a `Struct`, so `==` is memberwise and recurses into `props` (a Hash) and `children` (an Array). `flush!` calls `handle_for` twice per dirty root, and `adapter_page_id(root)` calls it again from `ensure_page!`. Every widget mutation in the shim ends in a `flush!`, so for a page the size of `;map` or `;bigshot` — several hundred nodes, some with large `layers` arrays — each commit does hundreds of deep hash comparisons to answer a question the adapter could answer in O(1) by keeping the reverse map it already has the information for.

There is a latent correctness edge here too: `Hash#key` returns the *first* value that compares equal. Two nodes with identical type, props, children, parent, slot, bindings and page would be indistinguishable. Today the shim always stamps a unique `key` prop into every node's props (`Widget#common_props`), so this cannot happen — but that invariant is load-bearing and undocumented, and a caller using `Lich::API.webui_adapter` directly is not bound by it.

**Fix:** maintain `@handles = {}.compare_by_identity` alongside `@nodes`, or make `Node` a plain class so `==` is identity.

## Minor

- **`lib/api/webui.rb` is entirely dead code.** 50 lines, ten public API methods, no `require` anywhere in the tree, no spec. Its sibling `lib/api/active_sessions.rb` is wired up at `lich.rbw:85`; this one never was. It is also the only documented entry point for the "native WebUI page" story, so shipping it unreachable is worse than shipping it not at all — the next person will use it and find nothing loads it.

- **`Server#websocket_upgrade?` (`lib/webui/server.rb:447`) has no callers.** `handle_client` routes on `request[:path] == '/ws'` directly.

- **`Dispatcher#await` cannot return** — `lib/webui/dispatcher.rb:73`. Both branches raise `ReentryError`; the first `if` only changes the message. If the intent is "there is no synchronous wait, ever", say that in one unconditional raise. If a re-entry check is meant to matter, the second raise makes it dead.

- **`Dispatcher#enforce_bounds!` is quadratic at the boundary** — `lib/webui/dispatcher.rb:137`. `viewer_count`/`page_count` are each `O(n)` and both are recomputed on every iteration of the eviction loop, which can run up to `PAGE_LIMIT` (1024) times. Only reachable under queue pressure, but that is exactly when you least want it.

- **`handle_file` reads the whole file into memory with no cap** — `lib/webui/server.rb:299`. `FileService` restricts extensions to images and confines paths by realpath, so the worst case is a large PNG under `MAP_DIR`, but the read is unbounded and the response is built in one string.

- **`Validator#validate_properties` checks a `:forced` key nothing sets** — `lib/webui/validator.rb:105`. `Contract.property` accepts `shape`, `required`, `scope`, `default` only; `grep -n forced` finds no producer. Dead branch.

- **`ModalCoordinator#open` can leak a page on failure** — `lib/webui/modal_coordinator.rb:55`. The rescue cancels the future but does not unregister the page, so a failure between `@registry.register(page)` and the end of the method leaves the page in the registry with nobody owning it.

- **`WebSocket.read_frame` blocks after `IO.select` says readable** — `lib/webui/websocket.rb:172`. `read_exact` uses `io.read(n)`, which blocks until `n` bytes arrive or EOF. A client that sends two header bytes and stalls parks the client thread until `stop` shuts the socket down. Loopback-only and bounded by `Server#stop`, so low impact, but the select gives a false sense of non-blocking.

- **The launch token travels in the browser's command line** — `lib/webui/server.rb:92`, consumed at `browser_launcher.rb:39` as `--app=#{url}`. Any other local user can read it from `/proc` or `ps` for its 60-second lifetime and exchange it for the session cookie, which is full control of the WebUI. Single-use and short-lived, and a shared multi-user desktop is already outside Lich's threat model — but it is the one place the otherwise-careful token design touches a world-readable channel, and it is worth a line in the docs.

## Nits

- `lib/main/main.rb:205` uses `next unless @launch_data` to abort startup. It works — the enclosing construct is the `@main_thread = Thread.new {` block at `main.rb:53`, so `next` terminates the thread and `lich.rbw`'s `@main_thread.join` returns — but `next` a thousand lines inside a thread block reads like a loop control. A comment naming what it exits would save the next reader the hunt.

- `lib/common/session_launcher.rb:191`: `return value if context[:force_path_flags] == true` sits between the empty-value opt-out and the default-comparison suppression. The placement is correct for the stated intent, but the method's doc comment ("Returns an explicit per-launch path override **when it differs from default**") no longer describes what it does.

- `SensitiveValue#clear_value!` (`lib/webui/sensitive_value.rb:88`) zeroes `@value`, which the constructor already made as `value.dup`. The caller's original string is untouched, and the scrub does nothing about copies the VM may have made. That is fine as defence in depth; it is not fine if anyone reads it as a guarantee. One sentence in the class doc saying "best-effort, not a guarantee" would prevent that.

## Open questions

- `lib/common/webui_launcher.rb:527` — `manual_connect` calls `transfer_secret(...)` *before* `begin_operation`, and `begin_operation` raises `Lich::WebUI::Error` when an operation of that kind is already active. A double-click on Connect therefore consumes the submitted password into a `SensitiveValue` that is then orphaned by the raise, and the dispatcher swallows the error (`dispatcher.rb:112`) so the viewer sees nothing at all. **Settles it:** is the button disabled client-side during `:authenticating`? If it is, this is unreachable; if not, it is a silent dead end on a path users will hit.

- The four `Gem.win_platform?` guards added to the ported specs — which assertions do they skip, and is the Windows path therefore untested for symlink containment in `FileService`? That is the one place where a platform difference could turn a containment check into a no-op.

## What this PR gets right

The security design is the load-bearing good decision and should be protected through the rest of the series:

- Loopback-only bind enforced twice — at construction (`LOOPBACK_HOSTS`) and again against the resolved `addr` after bind.
- Single-use, 60-second launch tokens exchanged for a `SecureRandom.hex(32)` session cookie compared with `Protocol.secure_compare`; `HttpOnly; SameSite=Strict`.
- Host allowlist, Origin allowlist, and `Sec-Fetch-Site`/`Sec-Fetch-Mode` checks layered rather than alternatives.
- `Protocol.parse_client_message` rejects unknown fields, duplicate headers, request bodies, and oversized frames before anything else looks at the message, then deep-freezes what survives.
- `FileService` does realpath containment *after* expansion (so an absolute path in the relative slot is caught), an extension allowlist, null-byte rejection, and owner-scoped revocation.
- The `--webui-dev` gating really is complete: without the flag, `lich.rbw` and `main.rb` take byte-identical paths to before.

## Coverage notes

Reviewed from a full clone with all eight branches fetched. **No Ruby is installed in this environment**, so `bundle exec rspec` and `rubocop` were not run and the "150 examples, 0 failures" claim is unverified — I read specs as documentation of intent, not as evidence. `lib/webui/assets/app.js` was reviewed for injection sinks and reconnect behaviour but not the full 1403 lines. `lib/common/webui_launcher/catalog.rb` (460 lines, credential storage) was only spot-checked around the master-password paths.

Nisugi:

Reviewed head for #1634: `01d51c8e`.



No additional standalone finding is assigned to this vendor layer in this pass. The runtime close/detach interaction is reported as F3 on #1648, where the modal-close behavior makes it observable. The full-stack release recommendation is on #1655. This does not constitute approval of the complete stack or a claim that every imported line was exhaustively verified.

### Validation and scope

PR #1635, opened 2026-09-16
feat(all): GTK compatibility shim - ScriptScope and the widget vocabulary

Nisugi:
**Stacked on #1634.** Review that first; this diff is against `webui/01-vendor`.

## What this is

The compatibility shim itself. Scripts written against ruby-gnome GTK 3 run **unmodified**, rendering through the WebUI contract instead of GTK.

The mechanism: scripts are eval'd in a `Lich::Common::ScriptScope` binding, so constant lookup goes `ScriptScope -> Lich::Common -> Lich -> Object`. A bare `Gtk` in a script resolves to `ScriptScope::Gtk` — the shim — while core code still sees the real `::Gtk`. Nothing outside `lib/common/script_scope/gtk/` mentions a GTK constant.

Covered here: `Window`, `Box`, `Table`, `Label`, `Entry`, `Button`, `CheckButton`, `ScrolledWindow`, `Viewport`, `MessageDialog`, `Gtk.queue`, the `AttachOptions`/`ResponseType` constants, `Gdk::Screen`, GLib timers — plus `Gtk::Builder` and the data widgets (`Notebook`, `SpinButton`, `ComboBox`, the `TreeView` family, `TextView`, `Expander`, `RadioButton`).

`;vars setup` and `;alias setup` run in a browser; all nine Glade-built scripts render headlessly.

## The one core change worth your attention

`lib/common/script.rb` gains `Script.__trusted_binding`, used at two call sites:

```ruby
def Script.__trusted_binding
  if defined?(Lich::Common::ScriptScope) && Lich::Common::ScriptScope.active?
    Lich::Common::ScriptScope.script_binding
  else
    TRUSTED_SCRIPT_BINDING.call
  end
end
```

Without `--webui-dev`, `ScriptScope.active?` is false and the historical binding is used unchanged. **This is the only place the whole series touches core.** Everything else lives under `lib/webui/` or `lib/common/script_scope/`.

## Design notes

- **One session thread per owning script**, created lazily from the script's own thread so it inherits the ThreadGroup and dies with it. Every signal handler, `Gtk.queue` block and timer callback runs there in order — the concurrency model GTK scripts were written against.
- **Class identity is real.** Scripts do `.class == Gtk::Entry` and subclass `Gtk::Window`, so shim classes are the actual constants scripts see; never wrapped.
- **Mutators return self** — 642 call sites chain them.
- **An unimplemented constant degrades** rather than killing the script, and says so where the player can see it.

Contract 2.5.0 → 2.6.0 (focus/blur events).

## Testing

`bundle exec rspec spec/lib/webui spec/lib/common/script_scope` → **176 examples, 0 failures**.


mrhoribu (changes requested):
# PR Review: feat(all): GTK compatibility shim - ScriptScope and the widget vocabulary (elanthia-online/lich-5#1635)

**Reviewed:** 23 files, +5560/-13 at d068306b on 2026-09-16. Based on: full clone (shared `.git`, all 8 stack refs + main), changed files at head, full repo checkout for cross-referencing, and live execution of the shim's code (`bundle exec ruby`) to confirm two findings by direct reproduction rather than static reading alone.
**Verdict:** Request changes
**Summary:** The core-touching change (`Script.__trusted_binding`) is sound and correctly gated — confirmed safe for the non-`--webui-dev` path, and the constant-lookup mechanism (`ScriptScope -> Lich::Common -> Lich -> Object` via lexical nesting) is implemented exactly as claimed, with no leakage into core. However, two of the PR's own headline design claims are demonstrably false at this commit, both confirmed by actually running the code: (1) "mutators return self" is false for dozens of `set_*`-style setters across nearly every widget class, which return `nil` instead, and (2) "an unimplemented constant degrades rather than killing the script" is false — there is no `const_missing` hook anywhere in the shim, so referencing any widget class not yet implemented (menus, dialogs, etc. — deferred to PRs #1636-#1641) raises a plain `NameError`. Separately, `TreeIter#next!` never terminates and corrupts the underlying `ListStore`/`TreeStore` on the very first walk past two rows — a real, currently-used script (`jinx.lic`) hits this by the later commit's own admission; that fix lands in a subsequent PR in the stack, not this one, so at this PR's head the bug is live and unguarded by any test.

## Blockers

### 1. `TreeIter#next!` never terminates and corrupts the model it walks
**File:** `lib/common/script_scope/gtk/widgets_data.rb:788-796` (also `:872-891`)
**Confidence:** confirmed (reproduced by execution)

```ruby
def next!
  following = @model.iter_after(self)
  return false unless following

  @key = following.key
  @values = following.values
  @parent_key = following.parent_key
  true
end
```
```ruby
def iter_first
  @rows.first
end
...
def iter_after(iter)
  index = @rows.index(iter)
  index && @rows[index + 1]
end
```

`ListStore#iter_first`/`#get_iter`/`#iter_after` hand back the model's own row objects, not copies. `TreeIter#==` compares by `@key`. `next!` mutates `self`'s `@key` in place to the successor's key — but `self` IS a row stored in `@rows`, so this rewrites that row's own key/values. On the next call, `@rows.index(iter)` (via `Array#index`, which uses `==`) matches `self` against itself (its key having just changed to the successor's), so `iter_after` returns the *same* successor forever, and the row that was overwritten is now a duplicate of its successor.

Reproduced directly:
```ruby
store = ListStore.new(String)
store.append.tap { |i| i[0] = 'alpha' }
store.append.tap { |i| i[0] = 'beta' }
store.append.tap { |i| i[0] = 'gamma' }
iter = store.iter_first
11.times { visited << iter[0]; break unless iter.next! }
# visited => ["alpha", "beta", "beta", "beta", ...] (never advances past "beta")
# store afterward => ["beta", "beta", "gamma"]  (row 0 overwritten in place)
```
A script that does `iter = model.iter_first; while iter; ...; break unless iter.next!; end` (the standard ruby-gnome walk idiom, and exactly how many `.lic` inventory/list scripts iterate a `ListStore`) hangs forever on any store with more than 2 rows, silently corrupting the row data as it goes.

**Grounding:** commit `18706d93` ("fix(all): the map renders again, and walking a list store terminates"), which exists in this repo's shared history but is **not** part of this PR's commit range (`git log --oneline origin/webui/01-vendor..HEAD` — 8 commits, `18706d93` is not among them; it belongs to a later PR in the stack), fixes exactly this bug and states in its own commit message: *"TreeIter#next! never terminated and corrupted the model... jinx.lic walks a store exactly this way at two sites."* That fix is not present at this PR's head (`d068306b`); `git diff origin/webui/01-vendor...HEAD -- lib/common/script_scope/gtk/widgets_data.rb` shows `iter_first`/`get_iter`/`iter_after` still returning live rows, no `dup_row`. No spec in `spec/lib/common/script_scope/gtk_builder_spec.rb` or `gtk_shim_spec.rb` exercises `next!`, `iter_after`, or `iter_first` (`grep -n "next!\|iter_after\|iter_first" spec/lib/common/script_scope/*.rb` → no hits), so nothing in this PR's own test suite would catch it.

This is flagged per the task brief's own instruction to check `git log -S` on `TreeIter`/`next!` history — doing so surfaces that the bug is real, already diagnosed by the author, and fixed one PR later, but is genuinely present and untested at this PR's head. If this PR is merged/released standalone (e.g., cherry-picked, or the stack lands out of order), it ships a data-corrupting infinite loop in one of the nine data widgets this PR advertises as covered.

## Major

### 2. "Mutators return self" is false for the entire `set_*` setter family
**Files:** `lib/common/script_scope/gtk/widgets.rb` (e.g. `:275-279`, `:309-313`, `:341-351`, `:353-363`) and `lib/common/script_scope/gtk/widgets_data.rb` (e.g. `:95-96`, `:184`, `:437`, `:1228`) — pattern repeats across ~40+ methods.
**Confidence:** confirmed (reproduced by execution)

```ruby
def sensitive=(value)
  @sensitive = value ? true : false
  changed!          # <- changed! is the method's last expression
end
alias set_sensitive sensitive=
```
```ruby
def changed!
  @dirty = true
  window = window_root
  return unless window&.handle && @session.on_session_thread? == false
  @session.enqueue { @session.commit }
end
```
`changed!`'s only paths return `nil` (the bare `return unless ...`, or `Session#enqueue`, which itself ends in an explicit `nil`, `session.rb:226-232`). So every setter whose body ends in a bare `changed!` call — `sensitive=`, `visible=`, `tooltip_text=`, `halign=`, `valign=`, `margin_*=`, `xalign=`, `label=`, `title=`, `text=`, `editable=`, `spacing=`, `orientation=`, `active=`, `expanded=`, `page=`, `headers_visible=`, and their `set_*` aliases — returns `nil`, not `self`.

For the `x=` assignment form this is invisible (Ruby always evaluates `obj.x = v` to `v`, ignoring the method's actual return). It is **not** invisible for the `set_*` aliases, which are ordinary method calls and are the idiom ported GTK/ruby-gnome scripts actually use (`entry.set_text("foo")`, `checkbox.set_active(true)`, `treeview.set_headers_visible(false)`, etc.). Reproduced directly against a live `Label`:
```
set_sensitive     -> nil (self? false)
set_visible       -> nil (self? false)
set_tooltip_text  -> nil (self? false)
set_halign        -> nil (self? false)
set_valign        -> nil (self? false)
set_margin_top    -> nil (self? false)
set_xalign        -> nil (self? false)
set_border_width  -> #<Label...> (self? true)   # correctly implemented, for contrast
set_name          -> "x" (self? false)          # returns the arg, not self either
```
Compare with `Container#add`/`Box#pack_start`/`#pack_end` (`widgets.rb:598-604`, `:695-705`), which explicitly end in `self` and are correct — so the PR clearly intends the "returns self" contract, and mostly delivers it for the widget-composition methods, but not for the attribute setters.

**Why this matters here specifically:** `NilClass#method_missing` (`lib/common/class_exts/nilclass.rb`, unconditionally required) makes `nil.anything` silently return `nil` rather than raise. So `label.set_text("hi").show_all` does not error — it silently no-ops the `.show_all` and everything chained after the setter, with no log line, no exception, nothing visible to the player or the script author. This is exactly the failure mode this PR's own body warns about elsewhere in the stack (`nil.width / 2` producing a confusing downstream error instead of a clear one) — here it's worse, because the downstream call vanishes with no error signal at all.

The existing spec (`spec/lib/common/script_scope/gtk_shim_spec.rb:263-268`) only covers the `method_missing`-based degrade path (`entry.set_icon_from_icon_name(...)`, a genuinely undefined method, which correctly returns `self` because `Widget#method_missing` — `widgets.rb:447-452` — special-cases `set_`-prefixed names). It does not cover any of the ~40 *explicitly defined* `set_*` methods above, which is exactly why this gap wasn't caught: they don't go through `method_missing` at all.

**Settles it:** the PR body claims "642 call sites chain them" — I can't verify that count against the actual `dr-scripts`/`scripts` corpus from this checkout, so I can't say how many of those 642 sites are `set_*`-on-an-attribute (broken) versus `add`/`pack_start`/`show`/`hide` (correct, verified above). Worth asking the author whether the count was audited against real `.lic` files or is a source-side static count of "methods that look like mutators."

### 3. "An unimplemented constant degrades" is false — no `const_missing`, referencing a not-yet-shimmed widget class raises `NameError`
**Files:** `lib/common/script_scope/gtk/widgets.rb` (whole file — no `const_missing` defined anywhere in `module Gtk` or its nested namespaces), `lib/common/script_scope/gtk/widgets_data.rb`, `lib/common/script_scope/gtk/builder.rb`
**Confidence:** confirmed (reproduced by execution)

`grep -rn "const_missing" lib/common/script_scope/` returns nothing. The degrade mechanism that exists — `Widget#method_missing` (`widgets.rb:447-452`) and `Gtk.log_unsupported` (`widgets.rb:123-131`) — only intercepts calls to *undefined methods on an already-instantiated shim object*. It does nothing for a bare constant reference to a class the shim hasn't implemented at all. The PR body's own list of covered widgets and its own text ("Later PRs (#1636-#1641) add menus, layout, images/map rendering, and more widgets on top of what this PR establishes") confirms that `Gtk::Menu`, `Gtk::MenuBar`, `Gtk::MenuItem`, `Gtk::Toolbar`, `Gtk::AboutDialog`, `Gtk::FileChooserDialog`, `Gtk::IconTheme`, `Gtk::Image`, `Gtk::Statusbar`, `Gtk::Paned`, `Gtk::Fixed`, `Gtk::Calendar`, etc. do not exist yet — confirmed by the class list in `widgets.rb`/`widgets_data.rb` (grepped every `class`/`module` declaration; none of the above appear).

Reproduced directly:
```ruby
Lich::Common::ScriptScope::Gtk::Menu
# => NameError: uninitialized constant Lich::Common::ScriptScope::Gtk::Menu
```
Any script that does `Gtk::Menu.new` (or any other widget outside this slice) directly in Ruby code (as opposed to via `Gtk::Builder`/Glade XML, which *does* degrade cleanly — see "What this PR gets right") gets a raw `NameError`, which crashes the script exactly the same way it always would have without any shim — i.e., the "degrades rather than killing the script, and says so where the player can see it" property is simply not implemented for the constant axis. This isn't a regression (a script referencing a genuinely absent GTK class would have failed before the shim too), but it directly contradicts an explicit, prominently-stated design guarantee of this PR, and the only existing test for the degrade path (`gtk_shim_spec.rb:263-268`) exercises only the method case, not the constant case, so the gap is untested as well as unimplemented.

**Why it's Major and not just a documentation nit:** later PRs in the stack (#1636-#1641) are being built and reviewed on top of this one; if their authors or reviewers believe "unimplemented constant degrades" is already a solved, general mechanism (as the PR body states), they may not re-verify it for whatever widgets they still don't cover, propagating the false assumption forward.

## Minor

### 4. `Session.for`'s `@null_owner` memoization races outside the mutex that is supposed to guard session identity
**File:** `lib/common/script_scope/gtk/session.rb:139-144`
**Confidence:** confirmed (by reading; not exercised concurrently)

```ruby
def for(owner)
  owner ||= (@null_owner ||= NullOwner.new('gtk-shim'))
  @registry_mutex.synchronize do
    @sessions[owner] ||= new(owner).tap { |session| session.send(:hook_owner_exit) }
  end
end
```
`@null_owner ||= NullOwner.new('gtk-shim')` runs *before* acquiring `@registry_mutex`, while the mutex is specifically there to make `@sessions[owner] ||=` a safe check-then-act. Two threads calling `Session.for(nil)` "at the same time" (both resolving to `current_script.nil?` — i.e. widget code running outside any script's own thread) can each observe `@null_owner` as `nil`, each construct a distinct `NullOwner` instance, and each then create and register a *separate* session under a *different* key — defeating the "one shared session for widgets created outside any script" invariant the class comment (`session.rb:104-111`) describes. Impact is limited (no crash; it just means two independent, uncoordinated shared sessions could coexist instead of one), and the window requires genuinely concurrent calls with no owning script, which is a narrow, edge-case path — hence Minor rather than Major.

**Settles it:** does anything besides `Session.current` (`session.rb:135-137`, called from `Widget#initialize`) call `Session.for(nil)` from more than one thread in practice? If GLib timer/idle callbacks (`widgets.rb:1852-1886`) or a `Gtk.queue` call can plausibly run with `current_script` nil from two different OS threads close together, this is reachable; otherwise it's theoretical.

### 5. `Script.__trusted_binding` — the one core-touching symbol — has no dedicated spec
**File:** `lib/common/script.rb:1120-1132`
**Confidence:** confirmed (absence verified by grep)

`grep -rn "__trusted_binding" spec/` finds no matches anywhere in the spec suite; the only coverage is indirect, through `gtk_shim_spec.rb`'s use of `ScriptScope.script_binding` directly (bypassing `Script.__trusted_binding` and its `active?` branch entirely) and through `session_launcher_spec.rb`'s coverage of the CLI-flag plumbing that leads up to `activate!`. Given the PR's own framing — "the only place the whole series touches core... treat any bug here as maximum severity" — the absence of a spec that pins both branches (`ScriptScope.active?` false → returns `TRUSTED_SCRIPT_BINDING.call` unchanged; `ScriptScope.active?` true → returns `ScriptScope.script_binding`) is a gap worth closing, even though I could not find an actual bug in the method itself (see "What this PR gets right").

### 6. `Table#attach`'s GTK2 `xoptions`/`yoptions` flags are silently discarded, with no `log_unsupported` call
**File:** `lib/common/script_scope/gtk/widgets.rb:896` (`def attach(child, left, right, top, bottom, _xoptions = nil, _yoptions = nil, _xpadding = 0, _ypadding = 0)`)
**Confidence:** confirmed (by reading)

`Gtk::AttachOptions`/`EXPAND`/`SHRINK`/`FILL` constants are defined (`widgets.rb:25-32`) specifically so scripts can pass them here (and the PR's own spec does exactly that — `gtk_shim_spec.rb:65-70`, `gtk::AttachOptions::FILL`/`EXPAND`/`SHRINK`), but the parameters are prefixed `_` and never read — the flags have no effect on layout. This is a reasonable simplification given the contract-based layout model, and it's not a truthiness bug (nothing here tests these ints for truthiness, so the "0 means false" GTK2 foot-gun the task description asked about does not apply — the values are dropped outright, not misinterpreted). It is, however, inconsistent with this PR's stated "unimplemented ... says so where the player can see it" principle: every other unsupported feature in this file routes through `Gtk.log_unsupported`; this one is silently absorbed.

## Nits

### 7. `GLib::Timeout.add`'s source-id registration has a narrow startup race
**File:** `lib/common/script_scope/gtk/widgets.rb:1852-1867`
**Confidence:** suspected

```ruby
def self.add(interval, &block)
  session = Gtk::Session.current
  id = nil
  thread = Thread.new do
    loop do
      sleep(interval.to_f / 1000.0)
      keep = session.sync { block.call }
      break unless keep
    end
  rescue StandardError
    nil
  ensure
    GLib.remove_source(id) if id
  end
  id = GLib.register_source(thread)
end
```
The spawned thread's `ensure` reads the closure variable `id`, which is only assigned *after* `Thread.new` returns. If the new thread reaches its `ensure` (e.g. `interval` is 0 and `block.call` returns falsy immediately) before the calling thread executes `id = GLib.register_source(thread)`, `id` is still `nil` in the thread's closure, so `GLib.remove_source(id) if id` is skipped, and the entry in `@sources` (keyed by the id assigned moments later) is never removed. The dead thread reference just sits in `@sources` until the process exits or `GLib::Source.remove` happens to be called with that id later — not a live-thread leak, just a leaked hash entry pointing at an already-terminated thread. **Settles it:** is any script's `GLib::Timeout.add` interval realistically `0` or small enough for the new thread to win this race against a same-process thread creation? If not, this is theoretical.

## Open questions

- **On finding #2:** Was "642 call sites chain them" audited against the actual `dr-scripts`/`scripts` `.lic` corpus, and if so, how many of those 642 are `set_*`-style property setters (broken) versus container/composition methods like `add`/`pack_start` (correct)? This determines whether finding #2 is a live regression risk for real scripts or mostly theoretical.
- **On finding #3:** Is a `Gtk.const_missing` (returning a inert stub widget class, logging once via `Gtk.log_unsupported`) planned for this PR, or deliberately deferred to a later PR in the stack? If deferred, the PR body's phrasing ("An unimplemented constant degrades") should be scoped to "an unimplemented *method*" until the constant case is actually covered, so later reviewers in the stack don't build on a false premise.
- **On finding #1:** Given the fix already exists one commit later in the shared history (`18706d93`), was it deliberately held back to a separate PR for stacking/review-size reasons, or did it land after this PR's head by accident (e.g. a rebase/cherry-pick boundary)? If deliberate, is there a plan to prevent this PR from being merged/released independently of the fix, since it ships a data-corrupting hang in a widget class this PR itself introduces?

## What this PR gets right

- **The core change is exactly as advertised and safe.** `Script.__trusted_binding` (`lib/common/script.rb:1120-1132`) is a clean two-line branch; with `ScriptScope.active?` false (the default, and the only state possible without `--webui-dev`), it calls `TRUSTED_SCRIPT_BINDING.call` unchanged — verified by reading both call sites (`script.rb:172`, `:3383`) and confirming neither the proc nor its behavior changed from before this PR (`git diff` shows only the call-site substitution, `TRUSTED_SCRIPT_BINDING` itself is untouched).
- **`ScriptScope.active?` cannot flip on unintentionally.** It's set once, under a mutex, from a single call site in `lib/main/main.rb:219-222`, itself gated by `@argv_options[:webui_dev]` — which requires the `--webui-dev` CLI flag. `activate!` is idempotent (`return true if @active`) and there is no code path that calls it without that flag.
- **The constant-resolution mechanism works as described and doesn't leak.** Traced the actual mechanics: `ScriptScope.script_binding` (`script_scope.rb:49-51`) creates `Proc.new {}.binding` from inside a method lexically nested in `module Lich; module Common; module ScriptScope`, so `Module.nesting` at that point is `[ScriptScope, Lich::Common, Lich]` — exactly the claimed `ScriptScope -> Lich::Common -> Lich -> Object` bare-constant lookup order, and verified live: `eval('Gtk', scope.script_binding)` resolves to the shim (`gtk_shim_spec.rb:104`). Repo-wide grep confirms nothing outside `lib/common/script_scope/` references `ScriptScope::Gtk` or includes `ScriptScope` anywhere, so core code's own `::Gtk` (real gem, when present) is untouched.
- **The already-in-stack bugfix commits are real, targeted fixes**, not padding: `2989bdc2` (child-session `--webui-dev` propagation via `SessionLauncher.optional_spawn_flags`, with its own regression test at `spec/lib/common/session_launcher_spec.rb:210-224`) and `b1fecfbd`/`1a7afa98`/`d068306b` each read as a real bug caught and fixed during the PR's own development, with commit messages that explain the mechanism rather than just the symptom.
- **`Gtk::Builder`'s degrade path for unrecognized Glade widget classes is correctly implemented** (unlike the direct-constant-reference case in Major finding #3): `Builder#build_object` (`builder.rb:87-93`) looks up the Glade class name in a `CLASSES` hash and returns `nil` with a `log_unsupported` call *before* ever touching `Gtk.const_get`, so a Glade file with an unsupported widget type degrades that one element cleanly rather than raising during parse.
- **Test suite passes as claimed:** `bundle exec rspec spec/lib/webui spec/lib/common/script_scope` → 176 examples, 0 failures, reproduced in this environment.

## Coverage notes

- Ran the full test suite for the paths the PR body names (`spec/lib/webui`, `spec/lib/common/script_scope`) — 176 examples, 0 failures, matching the PR's claim.
- Read `lib/common/script.rb`, `lib/common/script_scope.rb`, `lib/common/script_scope/gtk/{boot,session,builder,widgets,widgets_data}.rb`, `lib/common/session_launcher.rb`, `lib/common/webui_launcher.rb`, and the relevant slice of `lib/main/main.rb` in full; read `spec/lib/common/script_scope/gtk_shim_spec.rb` and skimmed `gtk_builder_spec.rb`/`session_launcher_spec.rb` for what they actually assert.
- Did not deeply review `lib/webui/contract.rb`, `lib/webui/server.rb`, `lib/webui/assets/app.js`/`app.css`, or `docs/webui-gtk-shim-plan.md` beyond their diff stats — these are either base-adapter (`lib/webui/*`, largely pre-existing from PR #1634 with only small additions here) or a planning doc, both lower-value for a "does the shim actually work" review than the widget/session/builder code that is this PR's substance.
- Did not run `coderabbit review` — skipped in the interest of time; static reading plus live reproduction (two of the findings above were confirmed by actually executing the shim's code, not just read) covered the highest-value ground the task asked for.
- Two findings (Blocker #1, Major #2 and #3) were verified by writing and running small standalone Ruby scripts against the actual checked-out code (`bundle exec ruby` against `lib/common/script_scope/gtk/boot.rb` and its dependents) rather than relying on static reading alone; the repro scripts were deleted after use and are not part of the diff.
- Did not attempt to independently verify the "642 call sites chain them" count against the `dr-scripts`/`scripts` repos — those checkouts were not in scope for this task and the claim is about scripts outside this repo.

mrhoribu:
# PR Review: feat(all): GTK compatibility shim - ScriptScope and the widget vocabulary (elanthia-online/lich-5#1635)

**Reviewed:** 23 files, +5560/-13 at `d068306b` (base `webui/01-vendor`) on 2026-09-16. Based on: full clone; no Ruby toolchain, nothing executed or linted.
**Verdict:** Request changes
**Summary:** The mechanism — evaluating trusted scripts in a binding whose cref is `ScriptScope`, so a bare `Gtk` resolves to the shim while core keeps the real one — is elegant and the single core change is genuinely small. The cost nobody has priced is that the same cref change moves every script-defined *constant* into a new namespace, and Lich marshals arbitrary script objects into `Vars`.

## Major

### 1. Script-defined classes change their fully-qualified name under `--webui-dev`, and `Vars` marshals them — `lib/common/script_scope.rb:50`

```ruby
def self.script_binding
  Proc.new {}.binding
end
```

The binding is created inside `module ScriptScope`, so its cref is `Lich::Common::ScriptScope`. That is exactly the point for constant *lookup*. But cref also decides constant *definition*: a script's top-level `class Foo` now defines `Lich::Common::ScriptScope::Foo`, where the historical `TRUSTED_SCRIPT_BINDING` (`lib/common/script.rb:23`, `def _script` inside `module Common`) defined `Lich::Common::Foo`.

`Vars` persists with `Marshal.dump(@@vars)` into sqlite (`lib/common/vars.rb:83`), and Marshal records the class name. So an object of a script-defined class stored in `Vars` while running `--webui-dev` is written as `Lich::Common::ScriptScope::Foo`, and the next normal-mode session cannot load it. The failure is not loud:

```ruby
hash = Marshal.load(h)
hash.each { |k, v| @@vars[normalize_key(k)] = v }
@@md5 = Digest::MD5.hexdigest(hash.to_s)
rescue StandardError => e
  respond "--- Lich: error: #{e}"
```

`@@vars` stays empty, `@@load_state` still becomes `LOADED`, and the save proc then computes a new md5 for the empty hash and writes it back — **the user's `Vars` are silently replaced with nothing.** That amplification is pre-existing lich-5 behaviour, not this PR's doing, but this PR introduces a brand-new way to reach it, and one that toggles with a command-line flag.

The same reasoning applies anywhere a script-defined constant is resolved by name from outside the ScriptScope binding: `Object.const_get`, YAML `!ruby/object:` tags in profiles, anything core does with `Lich::Common.const_defined?`.

Marked `suspected` only because I cannot check the script corpus from here — `dr-scripts` and `scripts` are not checked out in this environment. **Settles it:** does any script in either repo put an instance of a class it defines into `Vars` or `Settings`? A `grep -l 'Vars\[' ` cross-referenced against top-level `class` definitions answers it in a minute. If the answer is yes, this is a blocker until `ScriptScope` either aliases constants into `Lich::Common` or the mode is made sticky per character.

### 2. `Widget#signal_connect` hands out ids that collide after a disconnect — `lib/common/script_scope/gtk/widgets.rb:496`

```ruby
def signal_connect(signal, *_args, &block)
  name = Gtk.normalize_signal(signal)
  @handlers[name] << block
  id = @handler_ids.length + 1
  @handler_ids[id] = [name, block]
  id
end
```

`@handler_ids.length + 1` is not monotonic. Connect A (id 1), connect B (id 2), `signal_handler_disconnect(1)` — length drops to 1, so the next `signal_connect` returns 2 and overwrites B's entry. B is now permanently unremovable, and a later `signal_handler_disconnect(2)` detaches the newest handler instead of B.

The disconnect-then-reconnect pattern is not exotic in GTK scripts: it is the standard way to suppress a `changed` handler while programmatically setting a value. Nothing inside the shim calls `signal_handler_disconnect`, so this is only reachable from script code — which is precisely the code the shim exists to run unmodified.

**Fix:** a monotonic counter (`@next_handler_id ||= 0; id = (@next_handler_id += 1)`).

### 3. `GLib::Idle.add` with a repeating block spins the session thread and leaks a source entry per iteration — `lib/common/script_scope/gtk/widgets.rb:3163`

```ruby
def self.add(&block)
  session = Gtk::Session.current
  session.enqueue do
    keep = block.call
    GLib::Idle.add(&block) if keep
  end
  GLib.register_source(nil)
end
```

Two problems compound. A block that returns true re-enqueues itself with no delay, and `Session#run_loop` runs `commit` after every job — so a repeating idle source becomes a tight loop that fully re-materializes every window in the session on each pass. Under GTK, an idle source that returns true also reschedules, but it competes with a real main loop that has other priorities; here it is a bare `Queue` and it wins every time.

And each pass calls `GLib.register_source(nil)`, which does `@sources[id] = nil` under a mutex and returns an id the caller discards. `remove_source` is the only thing that deletes, and nothing can call it with an id nobody kept. `@sources` therefore grows by one entry per iteration, forever.

Even the one-shot case leaks one entry. **Fix:** do not register a source for idle at all (the doc comment already says the id maps to nil and `Source.remove` is a no-op for it), or return a real id and make the re-enqueue reuse it.

### 4. `Widget#respond_to_missing?` answers true for everything, and `method_missing` answers `nil` — `lib/common/script_scope/gtk/widgets.rb:800`

```ruby
def method_missing(name, *args, &block)
  return super if PROTOCOL_METHODS.include?(name)
  Gtk.log_unsupported(short_class_name, name)
  return self if name.end_with?('=') || name.start_with?('set_')
  nil
end
```

This is a deliberate design decision and the `PROTOCOL_METHODS` escape hatch shows real thought — letting `coerce`, `to_int`, `to_ary` and the comparison operators fall through to `NoMethodError` is exactly right, and #1638's commit message explains why in terms of a real debugging session.

The residual cost is still worth stating, because it interacts with a lich-5 patch that is easy to forget: `NilClass#method_missing` (`lib/common/class_exts/nilclass.rb`) already returns `nil` for anything called on `nil`. So a typo'd widget method returns `nil`, and every subsequent call on that `nil` also returns `nil`, until the value finally reaches something that cares — often many frames and one `changed!` later, in a different thread. There is no point in the chain where anything raises.

`respond_to_missing?` returning `true` is the sharper edge, because it makes duck-typing checks lie. The shim itself has already been bitten: `Window#collect_scrollers` carries the comment *"respond_to? answers true for everything on a Widget, so the test has to be what the widget IS"*. Any script — or any future shim code — that reaches for `respond_to?` gets a wrong answer. Consider narrowing `respond_to_missing?` to the setter/`set_*` shapes `method_missing` actually handles, and returning `super` otherwise.

## Minor

- **Every off-thread property write enqueues a full commit** — `widgets.rb:857`. `changed!` ends with `@session.enqueue { @session.commit }` whenever the caller is not on the session thread. A script thread doing `10.times { entry.text = ... }` queues ten full re-materializations of every window in the session. Coalescing (drop the enqueue if a commit is already pending) would be a few lines and would remove a whole class of surprise.

- **`Box#render_children` is a query with side effects, called three times per materialize** — `widgets.rb:1138`. It assigns `child.placement` as a side effect of being asked what the children are. `Box#common_props` calls it (via `all_packed_end?`), `Box#node_props` calls it, and `Container#materialize!` calls it again for `desired`. Three passes plus three rounds of placement assignment on every commit.

- **`GridLayout#render_children` is O(rows × cols × children)** — `widgets.rb:1355`. The inner `row_children.find` rescans, and `covered.values.any?` runs per row. `vars`/`alias` tables are small enough not to care; a large `Gtk::Table` would.

- **Grid fillers are memoized and never reclaimed** — `widgets.rb:1414`. `filler_for` caches by `[column, row]` in `@fillers` and `filler_children` returns all of them, which means `Container#materialize!`'s "gone child" sweep (`@synced_children - ordered_children - filler_children`) can never see a stale filler. A table that shrinks keeps the nodes for cells it no longer has.

- **`hexpand=` / `vexpand=` omit `changed!`, and `@vexpand` is never read** — `widgets.rb:647-655`. This was raised as finding 9 in `docs/webui-review-2026-09-16.md` (vendored in #1639) and is still unfixed at `webui/08-polish`. `@hexpand` is consumed by `Box#node_props` and `Grid#expanding_columns`, so setting it after the first render does take effect on the *next* commit — but nothing schedules one. `@vexpand` is written by the setter and read nowhere in `lib/`, so vertical expansion is silently dropped, including from Glade `vexpand="True"`.

- **`Session#serve_file` alias derivation can collide across owners** — `session.rb:298`. `"gtk-#{owner_label.downcase.gsub(/[^a-z0-9]+/, '-')}-#{@file_roots.size}"` maps `my_map` and `my-map` to the same prefix, and `FileService#register` overwrites an existing alias with no owner check (`file_service.rb:37`) while `unregister` correctly verifies ownership. The prior review flagged the asymmetry as "not currently exploitable"; the alias derivation is what makes that true, and it is one `gsub` away from not being.

- **`@file_roots ||= {}` is lazily initialized outside the mutex** — `session.rb:272`. `serve_file` is reachable from both the session thread and a script thread.

- **`@viewers` is a default-proc Hash, so reading creates entries** — `session.rb:166`. `viewers_for(page)` does `@viewers[page].keys`, which inserts a key for any page asked about. Only `close_window` deletes. Small, but it holds `Page` objects alive.

## Nits

- `widgets.rb:203` — the "dropped child" log message is two adjacent string literals separated by 23 spaces on one line:

```ruby
message = "webui-gtk-shim: dropped #{child.short_class_name}"                       "#{" key=#{label}" if label} from its parent: #{error.message}"
```

Ruby concatenates adjacent literals so the output is correct, but this is plainly a line continuation whose `\` was lost, and `rubocop -A` under Ruby 4.0 will have an opinion.

- `docs/webui-gtk-shim-plan.md` is 538 lines of design narrative in the repo. Useful, but it will rot — the vendored review doc already records that the plan's contract version was two minors stale by #1639. Consider a header line saying which commit it was last true for.

## Open questions

- `Script.__trusted_binding` is also used by `;exec` (`lib/common/script.rb:3383`). An `;e` one-liner now defines its constants into `ScriptScope` too. Is that intended, or should `;exec` keep the historical binding so ad-hoc debugging sees what core sees? **Settles it:** one sentence on whether `;e Foo = 1` is expected to be visible to a normal script afterwards.

- `ScriptScope::Nesting#const_added` (added in #1636) includes `ScriptScope` into every module and class a script defines, so every script's top-level `def` becomes an instance method of every other script's classes. Under the old binding the same methods landed on `Lich::Common`, which is included into `Object`, so the reachability is roughly equivalent — but the *shadowing* precedence is not, since an included module sits above `Object` in the ancestor chain. **Settles it:** is there a script in the corpus that defines a top-level method named after something `Object` provides? If so, its classes now resolve to the script's version where they previously resolved to Ruby's.

## What this PR gets right

- **The cref trick is the right mechanism.** Shadowing a toolkit for scripts without core ever seeing it, with zero conditionals in the widget code, is a much better answer than a `if webui?` fork of the GTK login.
- **`Script.__trusted_binding` is genuinely the only core change**, and it degrades to the exact historical expression when `ScriptScope.active?` is false. That claim in the PR body holds up under reading.
- **One session thread per owning script, created lazily from the script's own thread so it inherits the ThreadGroup**, is the correct concurrency model and is documented where it matters (`Session.start_service`'s comment about the accept loop dying with the first script is exactly the kind of thing that would otherwise be rediscovered painfully).
- **Class identity is real** — shim classes are the actual constants scripts see, not wrappers — which means `.class == Gtk::Entry` and `< Gtk::Window` both work. That decision is why the whole approach is viable and should not be traded away for convenience later.

## Coverage notes

Read in full: `script_scope.rb`, `gtk/session.rb`, `gtk/widgets.rb` (all 3210 lines at head). Read partially: `gtk/widgets_data.rb`, `gtk/builder.rb`. The `spec/` additions were read as intent, not run — **no Ruby is installed here**, so the "176 examples, 0 failures" claim is unverified. The cross-mode `Vars` concern in Major 1 could not be confirmed against the script corpus because `dr-scripts` and `scripts` are not checked out in this environment.

Nisugi:
## Review findings addressed at the tip of the stack (#1641)

Three findings from this PR are fixed in `webui/08-polish`, not here — they span slices (the `set_*` fix touches files from slices 2, 3, 5, 6 and 8), so fixing them in place would have meant force-pushing six already-reviewed branches.

**`TreeIter#next!` still corrupted the store — `2faa9091`.** The stack-level blocker. `iter_first`/`get_iter`/`iter_after` were taught to hand out copies; `append`/`prepend`/`insert` and `#each` were not, and those are how a script actually gets an iter. Reproduced by execution:

```
before = ["x", "y", "z"]
after  = ["y", "y", "z"]   CORRUPTED=true    # .next! on an each-yielded iter
```

`dup_row` already existed and already shares the row's `values` array, so writes still reach the model — it just needed applying at the four missed entry points. Eight regression specs now cover them, confirmed to fail against the unfixed code.

**"Mutators return self" — `1e31db0f`.** Reported as false for ~40 setters; the real figure is **829 of 1109**. `alias set_foo foo=` inherits Ruby's rule that an assignment method evaluates to its argument, so `Gtk::Entry.new.set_text(v)` handed back a String — and `.text` on it raised `NoMethodError`. That is line 28 of `perfume.lic`, one of the smallest GTK scripts in the census. Now declared through a `Setters.def_setter` helper, with a sweep spec that fails if any future setter regresses.

**`signal_connect` id collision — `f22205f4`.** `id = @handler_ids.length + 1` reused a live id after any disconnect:

```
first=1 second=2 third_after_disconnect=2   COLLIDES=true
```

So disconnecting the handler a script meant to drop killed a later one instead. The counter is monotonic now.

**`GLib::Idle.add` — `f22205f4`.** Re-enqueued itself with no delay (spinning the session thread) and registered a fresh source per pass, while the id it returned mapped to `nil` so `Source.remove` could never stop it. Rebuilt on the shape `Timeout.add` already had. Verified: repeats and stops, one source rather than one per pass, and `remove` now returns `true` and actually halts it.

**The `Vars` marshal question — answered, not fixed.** The mechanism is real: a script-defined class is `Lich::Common::ScriptScope::Foo` under `--webui-dev`, and loading that blob without the shim raises `ArgumentError: undefined class/module`. The reviewer named the deciding fact but could not check it without the corpus. Checked here across all 230 scripts: every `Vars`/`Settings`/`CharSettings` assignment constructing an object builds a core type (`Hash.new`, `Array.new`), and no script stores an instance of a class it defined itself. Not reachable today; recorded in the plan doc as a known limitation with the fix named for when it becomes real.

Nisugi:
Superseded by the re-layered stack: this work now lives in webui/shim-core (L6a). See docs/webui-rebuild-plan.md on webui/core-corrections for the map from the old PRs to the new layers. The branch is left in place for reference.

PR #1636, opened 2026-09-16
feat(all): GTK shim slice three - menus, markup, and pointer gestures

Nisugi:
**Stacked on #1635.** Diff is against `webui/02-shim-slice-one`.

## What this adds

Contract 2.6.0 → **2.7.0**, three additions:

- **`menu` / `menu_item` types.** `Gtk::Menu`, `MenuBar`, `MenuItem`, `CheckMenuItem`, `RadioMenuItem`, `SeparatorMenuItem`, `ImageMenuItem`. A popup menu is not in any window until `popup` attaches it to the one that received the gesture.
- **`context_menu` property + `press`/`release` events** on the surfaces scripts hang popups from (`EventBox`, `Label`).
- **`text.markup`** — the Pango subset scripts actually use (`b i u s tt big small span`, with a bounded span-attribute list). The validator parses it; the client builds nodes from the parse, **never from HTML**.

Written up in `docs/webui-contract-2.7-menus-markup.md`.

## Two fixes that came out of live testing

- A missing `Gtk` constant now **degrades instead of killing the script** — an enum member becomes the symbol it was named, a widget class becomes an empty container that says so where the player can see it. Previously a single unimplemented name took the whole window down.
- Script helpers reach the modules and classes the script defines. `ScriptScope` eval'd top-level `def`s into the wrong place, so a script that defined a helper module could not call into it.

## Testing

`bundle exec rspec spec/lib/webui spec/lib/common/script_scope` → **190 examples, 0 failures**.


mrhoribu (changes requested):
# PR Review: feat(all): GTK shim slice three - menus, markup, and pointer gestures (elanthia-online/lich-5#1636)

**Reviewed:** 19 files, +1637/-36 at 86b4fdbc on 2026-09-16. Based on: full clone, changed files at head, and full repo (adapter.rb, viewer_store.rb, tree_builder.rb, widgets_data.rb) for call-site context.
**Summary:** This is a careful, well-tested PR. The headline security claim ("client builds nodes from the parse, never from HTML") holds up end-to-end: the server validator (REXML, raise-on-violation) rejects disallowed tags/attributes/malformed XML before anything reaches the wire, the client independently re-parses with `DOMParser` and only ever calls `createElement`/`createTextNode`/scoped `.style.*` property assignment, and there's a dedicated `security_id: 'sec-escaping'` test exercising `<script>`, `onclick`, a bogus `href`, and CSS-injection-flavored values. I could not find an XSS gap. The one concrete bug I found undercuts one of the PR's own claims: the new `Gtk.const_missing` heuristic that is supposed to make "a missing Gtk constant degrade instead of killing the script" misclassifies any not-yet-implemented class name that starts with two consecutive uppercase letters (e.g. `Gtk::UIManager`, a real, menu-adjacent GTK class) as an enum member, producing the same category of crash the fix exists to prevent. Everything else is Minor/Nit-level: a stale `@pointer_window` that's never invalidated on window close, a doc/behavior mismatch around when the native context menu gets suppressed, and a narrow `method_missing` precedence subtlety in the new `InheritedHelpers` module. Tests (190 examples) and rubocop both pass clean on the diff.

## Major

### `Gtk.const_missing`'s widget-vs-enum heuristic fails on acronym-prefixed class names, reproducing the exact crash the fix claims to prevent

**Location:** `lib/common/script_scope/gtk/widgets.rb:250-258` (new in this PR):

```ruby
def const_missing(name)
  value = if name.to_s.match?(/\A[A-Z][a-z]/)
            unimplemented_widget(name)
          else
            name.to_s.downcase.to_sym
          end
  log_unsupported('Gtk', name, note: 'constant is not implemented')
  const_set(name, value)
end
```

The regex `/\A[A-Z][a-z]/` is used to distinguish "this is a widget class" (→ empty-container degrade) from "this is an enum/flag" (→ symbol). It only matches names whose *second* character is lowercase. Any not-yet-implemented GTK class name that starts with two consecutive uppercase letters — the common acronym-prefix pattern (`UIManager`, `IMContext`, `RGBA`, `GLArea`) — falls into the "enum member" branch instead, producing a bare `Symbol`:

```
$ ruby -e '
names = %w[UIManager IMContext CssProvider VBox TreeView Layout]
names.each { |n| puts "#{n}: widget-like? #{n.match?(/\A[A-Z][a-z]/)}" }
:uimanager.new rescue puts "$!"
'
UIManager: widget-like? false
IMContext: widget-like? false
CssProvider: widget-like? true
VBox: widget-like? false
TreeView: widget-like? true
Layout: widget-like? true
undefined method 'new' for an instance of Symbol
```

**Concrete trigger:** a script calling `Gtk::UIManager.new` — GTK2/3's XML-driven menu/toolbar builder, plausible in exactly the menu-heavy scripts this PR's own census names (map, xnarost, orbuculum) since `UIManager` predates hand-built `Gtk::Menu` as the idiomatic way to wire up menus. `Gtk::UIManager` is not defined anywhere in this shim (`grep -rn UIManager lib/common/script_scope/gtk/` returns nothing), so it goes through `const_missing`, gets classified as an enum member, becomes `:uimanager`, and `Gtk::UIManager.new` raises `NoMethodError: undefined method 'new' for an instance of Symbol` — an uncaught exception, taking the script down exactly as it would have before this PR (just with a different exception class/message).

**Grounding:** `VBox`/`HBox`/`HSeparator`/`VSeparator` — the obvious other two-cap-letter-prefix widget names — are safe only because they're already explicitly defined as real classes elsewhere in `widgets.rb` (lines 898, 904, 1717, 1723) and never reach `const_missing` at all. That's survivorship, not evidence the heuristic is sound; it just means nobody has hit it yet with a name that isn't already implemented.

**Confidence:** confirmed (the regex behavior and resulting `NoMethodError` are reproduced above); suspected for real-world frequency (I don't have a script corpus to confirm `Gtk::UIManager` specifically is referenced by name in the field, only that it's a standard, plausible GTK API for the class of scripts this PR is compatibility-shimming).

**Fix direction:** classify by "is this constant already a known enum/flag name" (allowlist, or check if the name is `SCREAMING_SNAKE_CASE`) rather than by the widget-class heuristic, or default unrecognized-shape names to `unimplemented_widget` (safer failure: an inert container beats a `NoMethodError` on `.new`) and reserve the symbol path for names that are unambiguously flag-shaped (all-uppercase with underscores).

## Minor

### `@pointer_window` is never invalidated when its window closes, so a `popup` racing a `window.destroy` silently attaches to a dead window

**Location:** `lib/common/script_scope/gtk/session.rb:321-329` (new in this PR):

```ruby
def note_pointer(window)
  @mutex.synchronize { @pointer_window = window } if window
end

def popup_window
  @mutex.synchronize { @pointer_window || @windows.find(&:handle) || @windows.first }
end
```

`close_window` (`lib/common/script_scope/gtk/session.rb:274-289`, pre-existing) does `@windows.delete(window)` but never touches `@pointer_window`. If the window that last received a pointer gesture is subsequently destroyed, and a script later calls `menu.popup` (e.g. a persistent `Menu` built once in setup and reused across window rebuilds — a normal GTK idiom), `popup_window` still returns the stale, destroyed window (the `||` fallbacks only trigger when `@pointer_window` is `nil`, not when it's stale). `Window#attach_popup` (`widgets.rb:1483-1492`) then attaches the popup to that dead `Window` instance and calls `changed!`.

**Trace of the actual impact:** `commit` (`session.rb:291-298`) iterates `@windows.dup`, which no longer contains the destroyed window (it was deleted in `close_window`), so the popup attached to the dead window is simply never materialized/rendered — a silent no-op (the menu never appears) rather than a crash, since `commit` also rescues `Lich::WebUI::Error`. Low severity, but worth a one-line fix (clear `@pointer_window` in `close_window` when it matches, or have `popup_window` check `@windows.include?(@pointer_window)`).

**Confidence:** confirmed for the mechanism (no invalidation exists); suspected for exact user-visible symptom (traced to "silently swallowed," not independently executed against a live session).

### Native context menu is suppressed whenever `press`/`release` is wired, not only when `context_menu` is set — contradicts the doc

**Location:** `lib/webui/assets/app.js:462-472` (`wireSurface`, new in this PR):

```js
if (!press && !release && !menuKey) return;
element.addEventListener("contextmenu", (event) => {
  event.preventDefault();
  event.stopPropagation();
  if (menuKey) {
    const menu = findByKey(page, menuKey);
    if (menu) openMenu(page, menu, event.clientX, event.clientY, menu);
  }
});
```

`docs/webui-contract-2.7-menus-markup.md:58-60` describes the suppression as a consequence of setting `context_menu`: "The client opens that menu on the secondary-button gesture ... and suppresses the browser's own context menu." In the implementation, the `contextmenu` listener (and its unconditional `preventDefault()`) is installed whenever `press`, `release`, *or* `menuKey` is truthy — so a component that only binds `press`/`release` (no `context_menu`) also loses its native right-click menu, silently, with no menu opening to replace it. A script author reading only the doc would not expect that. Likely intentional (GTK widgets with a button-press handler don't get a browser chrome menu either), but the doc should say so, or the code should gate the `preventDefault()` on `menuKey` specifically.

**Confidence:** confirmed (code reads as described); this is a documentation-accuracy gap, not a functional bug.

## Nits

### `InheritedHelpers#method_missing` sits ahead of the shim's own `method_missing` in the MRO, so a same-named script helper intercepts before the superclass "has its say"

**Location:** `lib/common/script_scope.rb:52-58` (new in this PR):

```ruby
module InheritedHelpers
  def method_missing(name, *args, &block)
    return ScriptScope.public_send(name, *args, &block) if ScriptScope.respond_to?(name)

    super
  end
  ...
```

The commit message and the code comment above it both claim helpers are reached "only after the whole superclass chain has had its say, so a helper can neither shadow an inherited method nor pre-empt the shim's degrade-and-log `method_missing`." That holds for *real* methods (Ruby resolves those before ever consulting `method_missing`, so nothing here can shadow an actually-defined method on the shim widget). It does not hold for names that exist *only* via the shim's own `method_missing` fallback (`widgets.rb:581`, unchanged by this PR): Ruby's `method_missing` lookup follows the same ancestor order as ordinary methods, and `include` places a module directly above the including class, so `InheritedHelpers#method_missing` is consulted before the superclass's. If a script's own top-level helper happens to share a name with an unimplemented Gtk method that only the shim's `method_missing` would otherwise log-and-degrade, the script's helper runs instead — the shim's `method_missing` is reached only via the explicit `super` when `ScriptScope.respond_to?(name)` is false. In practice this needs a naming collision between a script's own helper and a Gtk method name that's never internally invoked outside a script's own call sites, so the blast radius is small; flagging because the code comment overstates the guarantee.

**Settles it:** is there a case where the shim itself (not the script) calls a method on a script-subclassed widget by a name that could collide with a common script helper name (`render`, `update`, `refresh`, etc.)? If not, this is inert in practice and the comment is just slightly imprecise.

**Confidence:** confirmed for the MRO mechanics (verified against Ruby's `include` semantics); suspected for practical impact.

## Open questions

1. **`Gtk::UIManager` and similar** — does the shim's target script census (`docs/webui-contract-2.7-menus-markup.md`'s table of 28 GTK-menu scripts) actually reference `Gtk::UIManager`, `Gtk::ActionGroup`, or another acronym-prefixed, not-yet-implemented class by name? That would upgrade the Major finding above from "plausible" to "definitely hits a real script."
2. Double-`popup()` and pre-`popup()` method calls were traced by reading (`Menu#open!` guards re-entrancy with `return self if @open`, and calling item/menu accessors before any window exists just sets local state harmlessly) and look safe, but there's no test exercising rapid double-click or "call a menu accessor before any window is shown." Worth a regression test given how load-bearing the attach lifecycle is, even though I found no bug in it.

## What this PR gets right

- The Pango markup pipeline is defense-in-depth done properly: `Label#set_markup` pre-strips `<a>` tags to plain text before anything reaches the validator; `Label.markup_allowed?` (`widgets.rb:1565-1583`) validates through the real `Lich::WebUI::Validator` and caches the verdict, rescuing `Lich::WebUI::Error` to degrade to plain text rather than propagating; the server validator (`lib/webui/validator.rb:311-347`) parses with REXML and enforces both an element allowlist and a per-attribute allowlist+value-shape via `violation!`, which *raises* (confirmed: `SchemaViolationError`), so a bad markup value aborts the whole component validation rather than partially applying; and the client (`app.js`) independently re-parses with `DOMParser` and builds nodes exclusively via `createElement`/`createTextNode`/scoped `style.*` assignment — never `innerHTML`. I verified with REXML directly that the `<m>...</m>` wrapping structurally blocks the classic `</m><script>...` breakout and DOCTYPE/XXE injection (malformed XML, caught).
- There's a real, targeted security regression test (`spec/lib/webui/validator_spec.rb`, `security_id: 'sec-escaping'`) covering `<script>`, an `onclick` attribute, a disallowed `href` attribute, and a `font_desc` value containing `;` — exactly the vectors a reviewer would probe for.
- The two "fixes that came out of live testing" are each real, correctly scoped fixes for genuine pre-existing gaps, not just claimed ones: `const_missing` did not exist on `Gtk` before this PR (confirmed via diff against `origin/webui/02-shim-slice-one`), while `method_missing` already existed and already degraded gracefully — so the PR's framing ("a missing Gtk constant" specifically, not "any unimplemented name") is accurate, even though the mechanism itself has the gap noted above. The `ScriptScope::Nesting`/`InheritedHelpers` fix is verified against a real reported crash (`;armor died on undefined method 'message' for module ScriptScope::Armorinfo`) and has a passing regression test reproducing that exact shape (module + nested class calling a top-level helper).
- Menu attachment/lifecycle guards re-entrancy correctly (`Menu#open!`'s `return self if @open`, `Window#attach_popup`'s `return self if @popups.include?(menu)`), and pointer-gesture routing (`press`/`release` stop propagation only when the element itself is wired; `contextmenu` stops propagation so the innermost wired surface wins) matches its own documentation.
- Tests: `bundle exec rspec spec/lib/webui spec/lib/common/script_scope` → 190 examples, 0 failures (reproduced locally). `rubocop` on every changed `.rb` file → no offenses (reproduced locally).

## Coverage notes

- `coderabbit review --agent` was started but did not complete within the time budget (backgrounded, then reported failed/timed-out) — no CodeRabbit findings to fold in. GitHub's own CodeRabbit check on the PR shows "Review skipped: automatic reviews are disabled on this repository," so there's no upstream CodeRabbit pass to cross-reference either.
- `gh pr view 1636 --repo elanthia-online/lich-5 --comments` shows only the automated CodeRabbit skip notice; no human review comments to reconcile with.
- CI (`gh pr checks 1636`) is green: Rubocop pass, both Ruby test jobs pass.
- I read `lib/webui/adapter.rb`, `lib/webui/viewer_store.rb`, and the `ShimAdapter#update`/`#set` wrapper in `session.rb` (pre-existing, not part of this diff) only as call-site context for the popup-lifecycle and validator-enforcement questions above — did not review them for their own bugs, since they're outside this PR's diff.
- Did not attempt to run a browser against `app.js`/`app.css`; the JS review is static reading plus the DOM-API semantics (`Element.style.<prop> = value` always being scoped to that single CSS property, never able to inject a second declaration) rather than an executed test.

mrhoribu:
# PR Review: feat(all): GTK shim slice three - menus, markup, and pointer gestures (elanthia-online/lich-5#1636)

**Reviewed:** 19 files, +1637/-36 at `86b4fdbc` (base `webui/02-shim-slice-one`) on 2026-09-16. Based on: full clone; no Ruby toolchain.
**Summary:** Contract 2.6→2.7 adds menus, a bounded Pango subset, and pointer gestures. The markup path is the security-critical part of this PR and it is built correctly — parse server-side, build nodes client-side from the parse, never from HTML. Two smaller things: mnemonic stripping mangles legitimate underscores, and `const_missing`'s permanent `const_set` is a sharper edge than the guard acknowledges.

## Minor

### 1. `MenuItem#mnemonic_free` strips every underscore, not the mnemonic — `lib/common/script_scope/gtk/menus.rb:82`

```ruby
def mnemonic_free(label)
  label.to_s.gsub(/_(\S)/, '\1')
end
```

Three divergences from GTK in one line. The `gsub` is global, so *every* `_x` pair is stripped rather than the first; `__` (GTK's escape for a literal underscore) is not handled, so it collapses to a single character rather than to `_`; and it is applied unconditionally even though `MenuItem.new`'s second parameter is named `_use_underline` and defaults to `true` — real GTK only interprets mnemonics in `new_with_mnemonic`.

A menu item labelled `"Go to room_id"` renders as `"Go to roomid"`. `"Auto_save _mode"` becomes `"Autosave mode"`. Scripts that build menu labels from game data — room names, spell names, script names — will hit this.

**Fix:** handle `__` first, then strip at most one remaining `_`:

```ruby
label.to_s.sub(/(?<!_)_(?!_)(\S)/, '\1').gsub('__', '_')
```

### 2. `Gtk.const_missing` installs its stub permanently, and the guard only covers names the shim already knows — `lib/common/script_scope/gtk/widgets.rb:290`

```ruby
def const_missing(name)
  if OWN_DEFINITIONS.include?(name)
    raise NameError, "Gtk::#{name} is defined by the shim but not loaded yet; ..."
  end
  ...
  const_set(name, value)
end
```

The `OWN_DEFINITIONS` list and the `webui_stub?` marker are both good instincts, and the comment explaining *why* (`const_missing` const_sets its answer, so a later definition reopens the stub instead of replacing it) is exactly right. But the list is hand-maintained and enumerates only the classes the shim defines in files loaded after `widgets.rb`. Every future slice has to remember to add to it, and the failure mode for forgetting is silent: the new class is shadowed by an empty box for the rest of the process, which is precisely the bug the list exists to prevent.

`boot.rb` requires the files in a fixed order, so a cheaper invariant is available: have `boot.rb` assert that every name in `OWN_DEFINITIONS` is defined and not a `webui_stub?` once loading finishes, and have each slice file raise if it finds a stub under its own name. That turns "someone forgot" from a rendering mystery into a load-time error.

### 3. `NAMESPACE_SUFFIXES` decides widget-versus-enum by name shape — `lib/common/script_scope/gtk/widgets.rb:322`

```ruby
NAMESPACE_SUFFIXES = /(?:Flags|Type|Types|Mode|Modes|Action|Actions|Mask|State|Direction|Priority|Options|Defaults|Style|Policy|Position|Order|Level|Role|Hint|Format|Class|Kind|Target)\z/
```

This is a reasonable heuristic and the commit message documents the real bug it fixes (`Gtk::TargetFlags::SAME_APP` raising `NameError` out of a stub class and killing `ewaggle`). The risk is the other direction: a real GTK widget class whose name ends in one of these suffixes degrades to an inert module rather than an empty container, so a script that calls `.new` on it gets `NoMethodError` instead of a box. `Gtk::CellRendererText`, `Gtk::FileChooserAction`, `Gtk::ShortcutsShortcutType` — the first is safe, the others are genuinely enums. It probably holds for the corpus; it is worth a one-line comment saying the list is empirical and where it came from, so the next person knows it can be extended.

## Nits

- `ScriptScope::Nesting#const_added` calls `const_get(name)` and then `value.singleton_class.include(Nesting)` on every module a script defines. The recursion terminates because `include`/`extend` add no constants, but that is not obvious from reading and the method has no test for the depth-2 case. A spec asserting `Foo::Bar::Baz` reaches a top-level script helper would pin the behaviour the comment promises.

- The `InheritedHelpers#method_missing` fallback (`script_scope.rb:69`) forwards to `ScriptScope.public_send(name, ...)` after the superclass chain has had its say. That is the right precedence, but it means a script class subclassing a shim widget reaches helpers *through* `Widget#method_missing` first — and `Widget#method_missing` returns `nil` for anything it does not recognise rather than calling `super`. So for a script class that inherits from a shim widget, `InheritedHelpers` is unreachable: `Widget#method_missing` swallows the call before the singleton's `method_missing` ever runs. Worth checking; if I have the ancestor order wrong this is nothing, but if I have it right the whole `InheritedHelpers` branch is dead for exactly the case it was written for. **Settles it:** a spec where a script defines `class MyWindow < Gtk::Window` and calls a top-level helper from an instance method.

## What this PR gets right

- **The markup path has no XSS surface and is built so it cannot grow one.** The server parses and validates (`validator.rb`, +55), and the client builds from the parse with `createElement`/`createTextNode` against an allowlist (`app.js:1069`). An unknown tag recurses into its children rather than being dropped or passed through, so `<script>alert(1)</script>` arrives as the literal text `alert(1)`. `spanStyle` writes through `element.style.<property>`, which parses a single value and ignores anything invalid — there is no way to break into a second declaration or smuggle a `url()`. No `innerHTML` anywhere in the file.

- **"A missing constant degrades instead of killing the script"** is the right call for this codebase, and splitting the two cases — an enum member costs nothing, a missing *widget class* costs the script everything it was going to put there — is a distinction most people would not have drawn. Telling the player once per class per session, rather than only writing a `warning:` line to a debug file, is what makes the degradation honest.

- **A popup menu is not in any window until `popup` attaches it** to the window that received the gesture. That is the correct model and it is why `Session#note_pointer` exists.

## Coverage notes

Read in full: `gtk/menus.rb`, the `script_scope.rb` diff, the `const_missing` region of `widgets.rb`, and `app.js`'s markup renderer. `docs/webui-contract-2.7-menus-markup.md` was skimmed. Specs read as intent only — **no Ruby in this environment**, so "190 examples, 0 failures" is unverified. The `InheritedHelpers` concern in the Nits is reasoned from ancestor order, not executed.

Nisugi:
## `const_missing`'s acronym blind spot is closed (at the tip, #1641)

The review noted that this PR's `const_missing` closes most of the gap #1635 left, but that its classification heuristic — `/\A[A-Z][a-z]/` — misclassifies acronym-style class names as enums, reproducing the same category of uncaught crash the fix exists to prevent. Confirmed and fixed in `e2bbb5a1`.

```
before:  Gtk::UIManager => :uimanager
         Gtk::UIManager.new => NoMethodError: undefined method 'new' for an instance of Symbol

after:   Gtk::UIManager => Lich::Common::ScriptScope::Gtk::UIManager
         Gtk::UIManager.new => OK (a stub Container, as intended)
```

Requiring a lowercase *second* letter sent `UIManager`, `IMContext` and friends to the enum-member fallback, where they became a bare symbol. An enum **member** is the thing actually being distinguished, and those are `SCREAMING_SNAKE_CASE` — so the test is now "starts uppercase and is not all-caps" rather than "looks CamelCase".

Everything else is unchanged and covered: members read through a flags namespace still answer as symbols (`TargetFlags::SAME_APP => :same_app`), the census's bare `Gtk::EXPAND`/`Gtk::FILL` still return their real integers (1 and 4), and real widget classes still resolve to real classes. `RGBA` stays a symbol, correctly — it is all-caps and indistinguishable from an enum member by name alone.

**Latent rather than live.** Every acronym-style class the 49 GTK scripts actually name — `HBox` (38 uses), `VBox` (24), `HSeparator`, `VSeparator`, `HPaned`, `VPaned` — is explicitly implemented and never reaches `const_missing`, which is why this had not bitten. The review read the guarantee as unconditional, which is the right reading of how it is documented.

### Still open from this PR

**`mnemonic_free` strips every underscore**, mangling legitimate labels. Real, and narrower than it looks — but the fix wants a corpus check on which labels actually contain underscores before choosing between "strip only a single leading underscore" and "strip \_ only when followed by a letter". Left for its own change rather than guessed at here.

Nisugi:
## `mnemonic_free` — the review's reading does not survive the real gem

This PR's `mnemonic_free` was reported as stripping every underscore and so mangling legitimate labels (`lootsack_2`, `snake_case_name`). Half right, and the conclusion inverts once you ask GTK.

I wrote the "fix" first — preserve underscores past the first, since only the first is the accelerator — then measured against gtk3 3.24.52:

```
Gtk::Label#set_text_with_mnemonic
  "E_xit"           -> "Exit"
  "snake_case_name" -> "snakecasename"
  "lootsack_2"      -> "lootsack2"
  "a__b"            -> "a_b"
```

GTK strips **every** single underscore, not just the first. So the blanket `gsub(/_(\S)/, '\1')` was faithful, and my fix would have made the shim diverge from the toolkit it emulates — a worse outcome than the reported bug. Backed out.

**The one real gap** is the last line: a doubled underscore is a literal `_` in GTK, and the old gsub dropped it entirely. That is what shipped in `5d88ab67`, with the other four cases pinned as specs so the next reader does not have to re-derive this.

For grounding on reachability: every literal underscore in the 230-script corpus is a genuine mnemonic — `E_xit`, `_Save & Close`, `_Opacity`, `_Borderless`, `_Hide scrollbars`. Interpolated labels do reach this path and are affected exactly as they would be under GTK, which is the point.

**`report_stubbed_widget`'s mismatched guard — `180b6a92`.** Also fixed, and narrower than reported. It tested `defined?(::Lich::Messaging) || Kernel.respond_to?(:respond, true)` and then called `Kernel#respond` unconditionally, so with Messaging loaded but respond absent the call raised into the method's own `rescue StandardError` and the script was never told. It now guards on the method it actually calls. Verified with `DEBUG=1` that both the log warning and the script-facing notice do reach their destinations today, so this was the guard being wrong rather than the notice being broken — the two conditions travel together in every configuration checked.

Nisugi:
Superseded by the re-layered stack: this work now lives in webui/shim-data (L6b; menus are not shimmed, see the plan). See docs/webui-rebuild-plan.md on webui/core-corrections for the map from the old PRs to the new layers. The branch is left in place for reference.

PR #1637, opened 2026-09-16
feat(all): GTK shim slice four - packing, scrolling, presentation, grid weights

Nisugi:
**Stacked on #1636.** Diff is against `webui/03-slice-three-menus`.

## What this adds

Contract 2.7.0 → **2.11.0**, four related pieces, each the answer to something rendering wrong in a browser:

- **Box packing (2.8).** GTK's `pack_start(child, expand, fill, padding)` in every spelling scripts use, including the GTK 2 positional form where the flags are integers and `0` means false — though `0` is truthy in Ruby. Along a vertical box packing becomes child placement (`grow`/`pad`); along a horizontal one it becomes the `columns` weights.
- **Scroll adjustments (2.9).** Scripts scroll by writing pixels to an `Adjustment`, but they derive the number from an extent only the viewer knows — `upper - page_size` is the scroll-to-bottom idiom in vars, alias and localchat. The viewer now reports its real extent back, and a write that lands on it travels as the *intent* ("the bottom") rather than a pixel guess.
- **Window presentation.** `keep_above`, `decorated`, `opacity`, `resizable` kept as readable shadow state and declared through a `presentation` facility; whatever the host cannot honour is recorded as a degradation rather than silently dropped.
- **Grid column weights (2.10).** GTK's default is *not* to expand, so a table with no `EXPAND` anywhere keeps every column at natural width. Only columns a child asked to expand into get a share of the leftover.

## The layout fixes underneath

Roughly a dozen commits of live-testing corrections, each with its own message: a one-sided GTK margin is no longer applied to all four sides (bigshot has 518 of them); an unweighted grid hugs its content like a GTK table; a long tooltip no longer costs a grid cell; `pack_end` alignment; wrapped labels; checkbox columns; stale events.

Also two that are not layout:

- **`flush!` refreshes outside its lock**, so a render cannot deadlock it.
- **A background render keeps an unsent edit, its focus and its caret.** The tree is rebuilt on every commit, so without this the viewer loses what they were typing whenever a script repaints on a game event.

## Testing

`bundle exec rspec spec/lib/webui spec/lib/common/script_scope` → **257 examples, 0 failures**.


mrhoribu (changes requested):
# PR Review: feat(all): GTK shim slice four - packing, scrolling, presentation, grid weights (elanthia-online/lich-5#1637)

**Reviewed:** 18 files, +1954/-92 at b030edee on 2026-09-16. Based on: full clone, changed files at head, and full repo for cross-referencing.
**Verdict:** Request changes
**Summary:** The four headline mechanisms (0-vs-truthy packing flags, scroll-to-bottom-as-intent, `flush!` outside its lock, grid column weights) are all implemented correctly and are individually well-grounded — I traced each one to the actual read sites and, where relevant, to the browser-native CSS Grid/flex behavior they delegate to, and found no bug in any of the four. The fifth thing this PR touches, the client-side stale-generation retry added in `648a0d40`, has a real bug: the "replayed once so a genuinely rejected event cannot loop" guarantee the commit message states is not actually enforced by the code, and a second, independent issue in the same block means the wrong event can be replayed under concurrent UI activity. CodeRabbit's automated pass, run independently, converged on the same loop-guard defect. Everything else (margins, headers, method_missing protocol methods, presentation facility) checked out against its stated rationale.

## Major

### The stale-generation retry can loop forever, contradicting its own "replayed once" guarantee
**File:** `lib/webui/assets/app.js:893-905`
**Confidence:** confirmed (independently corroborated by CodeRabbit's automated review of the same diff, which flagged the identical root cause as "major")

```js
if (message.reason === "stale_generation" && pendingEvent) {
  const retry = pendingEvent;
  pendingEvent = null;
  const page = pages.get(retry.address);
  const component = page && findComponent(page.tree, retry.cid);
  if (page && component && !retriedEvents.has(retry.cid + retry.event)) {
    retriedEvents.add(retry.cid + retry.event);
    window.setTimeout(() => {
      retriedEvents.delete(retry.cid + retry.event);
      emit(page, component, retry.event, retry.payload);
    }, 0);
    return;
  }
}
```

The commit message (`648a0d40`) states: "Replayed once per component and event, so a genuinely rejected event cannot loop." But `retriedEvents.delete(...)` runs *inside* the `setTimeout` callback, immediately before the replayed `emit()` call — i.e., the guard is removed at the exact moment the retry is sent, not after its outcome is known. If the replayed event is *also* refused with `stale_generation` (plausible for a widget on a frequently-repainting window — exactly the creaturebar/bigshot scripts this PR's own body says repaint "on a game event, not on anything the viewer did," and this same PR increases render frequency via the `flush!`-outside-lock change), `retriedEvents.has(key)` is false again by the time the second refusal arrives, so the code retries a second time, a third time, and so on — an unbounded retry loop for that cid+event, hammering the websocket. **Trigger:** any component/event pair that keeps landing in stale-generation refusals across two or more separate round trips (a busy repaint loop is sufficient; no unusual timing required beyond "the retry itself lands stale too").

### The retry replays whatever is in `pendingEvent`, not the event the refusal actually names
**File:** `lib/webui/assets/app.js:98-109`, `888-907`
**Confidence:** confirmed

```js
function emit(page, component, event, payload = {}) {
  ...
  pendingEvent = { address: page.address, cid: component.cid, event, payload };
  socket.send(JSON.stringify(message));
}
```
`pendingEvent` is a single, last-write-wins global slot overwritten by *every* call to `emit()` — not just clicks. Scroll events emit on every native `scroll` tick with no throttling (`{ passive: true }`, `lib/webui/assets/app.js:195-201`), so dragging a scrollbar, typing (`change`), or any other bound event will clobber `pendingEvent` while an earlier event's response is still in flight over the websocket (there is no request/response gating — `emit()` fires and returns immediately). When a `refusal` with `reason: "stale_generation"` arrives, the handler at `app.js:893` never checks it against `message.cid` (which the server does send — it's used two lines later at `app.js:910` for the generic-error path) before deciding to replay `pendingEvent`. **Concrete trigger:** user clicks button A (sets `pendingEvent = A`, message in flight); before A's response returns, the user scrolls or types, emitting event B (`pendingEvent` now `= B`); A comes back refused as stale — the code replays B (already-sent and possibly already-succeeded) a second time instead of A, while A's action is silently dropped. **Settles further:** does the server ever interleave responses out of order on one connection, or is send-then-immediately-send-again by the client (without waiting) actually possible in the real browser event loop? I traced the client code and confirmed nothing gates a second `emit()` on the first's response, so the race is real regardless of server ordering.

Both issues sit in the same 20-line block and share a fix: correlate the retry against `message.cid`/`message.page` (which the wire already carries) instead of trusting a single mutable slot, and drop the guard only once the retried message's own outcome is known (success or a second genuine refusal) rather than the instant it is re-sent. There is no test coverage for either: `spec/lib/webui/assets_spec.rb` (the file this PR added JS-source assertions to for a different feature) has no `stale_generation`/`pendingEvent`/`retriedEvents` references, and `spec/lib/webui/runtime_spec.rb` only tests the server's own stale-generation raise, not the client's retry.

## Minor

### `fill` is parsed but has no effect on rendering
**File:** `lib/common/script_scope/gtk/widgets.rb:1037-1044` (parsing), `960-1003` (only consumer)
**Confidence:** confirmed

```ruby
def packing_from(positional, options)
  expand, fill, padding = positional
  {
    expand: packing_flag(options.fetch(:expand, expand), default: true),
    fill: packing_flag(options.fetch(:fill, fill), default: true),
    padding: options.fetch(:padding, padding || 0).to_i,
  }
end
```
`packing[:fill]` is computed (correctly, via `packing_flag`, avoiding the 0-truthy trap) but is never read anywhere downstream — `render_children`/`node_props` only consult `packing[:expand]` and `packing[:padding]` (confirmed by grepping every `[:fill]`/`:fill)` occurrence in `widgets.rb`: the only two are in `packing_from` itself). `spec/lib/common/script_scope/gtk_packing_spec.rb:27-37` only asserts that `fill` is parsed into the hash correctly, never that it changes rendered output. Concretely, GTK's `pack_start(child, expand: true, fill: false, padding: 0)` — "grow the child's allocated space but keep it at natural size, centered in the extra room" — renders identically to `fill: true` here (both just get `grow: 1`). This is a real (if narrow) behavioral gap for any script using that exact combination, and the PR body's claim to handle packing "in every spelling scripts use" slightly overstates what's actually wired up. Not blocking since it may be a deliberate, documented-nowhere simplification of the flex/grid model this PR uses, and `expand`+`fill` disagreeing is a less common spelling in practice.

### PR body's version claim is stale relative to the actual head
**File:** `lib/webui/contract.rb:9`
**Confidence:** confirmed

The PR description says "Contract 2.7.0 → **2.11.0**." The actual `VERSION` at head (`b030edee`) is `'2.12.0'` — bumped by `53a1cb79` ("hide internal tree view headers, and open a window GTK's size"), which isn't mentioned in the "four related pieces" list at all. Not a code bug, but worth fixing in the description before merge since a reviewer skimming the body would look for the wrong final version.

## Nits

- `lib/webui/assets/app.js:898-901`: `retriedEvents` is keyed by `retry.cid + retry.event` (bare string concatenation, no separator). Given real `cid` values always contain `/`/`:` delimiters and `event` names are a fixed contract enum, an actual collision is not realistically reachable today — but it's a fragile compound-key pattern for the next person who copies it with less constrained inputs.
- `spec/lib/webui/adapter_spec.rb:89-129`: the new deadlock regression test (`'refreshes outside its own lock...'`) has no `ensure` to reap `flusher`/`attacher` if the timeout assertions fail — if the deadlock this test guards against is ever reintroduced, the failing run leaks a permanently stuck thread for the rest of the process. Doesn't affect correctness of the assertion itself, just cleanup hygiene on the failure path. (Also flagged independently by CodeRabbit.)

## Open questions

- **Multi-viewer scroll extent:** `ScrolledWindow#receive_event` (`lib/common/script_scope/gtk/widgets.rb:1343-1353`) writes the *shared*, per-widget `@vadjustment` from whichever viewer's browser last fired a `:scrolled` event. If a session's page is ever attached by two simultaneous viewers with different window/viewport sizes, one viewer's scroll extent (`upper`/`page_size`) would silently overwrite the other's, and the next `scroll_position` computed from it would be reporting a size that's wrong for at least one of them. I could not determine from this diff or the surrounding code whether multi-viewer-per-page is an actually-supported scenario for this shim (the `Session#viewer_write` doc comment talks about push-per-viewer for *values*, but scroll adjustment travels through the ordinary `node_props`/props path, not `viewer_write`). **Settles it:** is a single GTK-shim window ever expected to be open in more than one browser tab/viewer at once? If not, this is moot.

## What this PR gets right

- **The headline "0 is truthy in Ruby" fix is real and correctly applied everywhere it needs to be.** The diff shows the *old* code did have exactly the naive-truthiness bug (`expand: options.fetch(:expand, expand) ? true : false`), and this PR's `packing_flag` (`widgets.rb:1046-1052`, `case value; when Integer then !value.zero?; ...`) is used consistently at every downstream read site I could find (`render_children`, `node_props`, `all_packed_end?`) as well as in the separate `Table#expand?` (`widgets.rb:1213-1215`, proper bitmask `& AttachOptions::EXPAND`). No missed spot.
- **The scroll "intent, not pixel guess" design is genuinely race-free by construction.** `Adjustment#at_extent?` compares a script's write against the *same* `@upper`/`@page_size` instance variables the script itself read to compute it (not an independently-tracked copy), so staleness in the shim's own view of the extent cancels out. And critically, the wire payload sends `{bottom: true}` rather than a computed pixel value; the browser applies it at render time as `element.scrollTop = element.scrollHeight` (`app.js:856`) against the *live* DOM, so even a fully stale Ruby-side `@upper` still produces a correct scroll-to-bottom. I traced this end to end and could not find the race the review brief hypothesized.
- **`flush!`'s lock-order fix is correct and has a real regression test.** `adapter.rb`'s new `flush!` computes the dirty page list and calls `refresh_definition` (which only touches `Page`'s own separate `@mutex`, not the adapter's) while still holding the adapter's lock, then calls `@service.refresh` (which re-enters `render_children`/the adapter's lock via `Page#render`) only after releasing it — exactly matching the lock-order-inversion story in the comment. `spec/lib/webui/adapter_spec.rb:89-129` reproduces the actual inversion with two real threads and a paused render block, not just an isolated unit assertion.
- **Widget identity (`cid`) is stable per-instance, which makes the new focus/caret-preservation feature safe.** Every GTK-shim widget gets `@key = Gtk.next_key` (a monotonic global counter) once at construction, always passed through as the `TreeBuilder#component` `key:` parameter via `common_props`/`**props`. This means `cid` tracks the widget instance, not its position among siblings, so the "stale caret state lands on the wrong widget after a rebuild" scenario the review brief asked about does not occur even when a script adds/removes/reorders siblings dynamically. I verified this against `tree_builder.rb`'s `cid = "#{parent.cid}/#{normalized_type}:#{segment}"` construction and confirmed `segment` resolves to the stable key, not `position`, whenever a widget's own `common_props[:key]` is present (which is unconditional for every GTK-shim widget).
- **Grid/columns weight distribution correctly isolates expanding from non-expanding tracks**, because it delegates to native CSS Grid (`Nfr` for expanding columns, `auto` for the rest) rather than hand-rolling pixel math — there is no rounding-remainder-leaks-into-natural-width-columns bug to find because the browser's grid algorithm sizes `auto` tracks to content before distributing `fr` space, by spec.
- **The per-side margin fix (`a4387b82`) is scoped correctly on both axes** — `@margins`/`@padding` are always fully-populated 4-key hashes (defaulted to 0 at construction), so the "all equal → send a plain integer" collapse path can never fire on a partially-specified margin and silently reintroduce the box-on-all-sides bug it's fixing.
- Test coverage for the parts of this PR that got it (packing, margins, presentation, hexpand-in-a-box, use-markup, headers) is thorough and each one pins the specific live-testing symptom described in its commit message.

## Coverage notes

- Ran `coderabbit review --agent --base origin/webui/03-slice-three-menus --committed`; it completed and returned 2 findings (a minor test-thread-cleanup nit and the same stale-generation loop-guard defect reported above as Major), both cross-checked against the actual code rather than cited directly.
- Did not run `bundle exec rspec` — not attempted given the native-gem-dependency caveat and that the PR body already reports a clean 257/0 run; no reason to doubt that claim from reading the specs.
- Read all 16 commits' messages and diffs individually; spot-checked `d6bf3694`, `f8185b66`, `dd22b793`, `82a37514`, `2d2853af`, `0f272c86`, `a4387b82`, `84225e69`, `a574627e`, `648a0d40`, `a38e9436`, `53a1cb79`, `0e4ebe21`, `de6f639f`, `b030edee` in full; `f7b81e7e` is docs-only and was skimmed rather than deeply reviewed since it only records already-shipped behavior.
- Did not independently verify the GTK-real-world behavior claims (e.g. exact pixel heights bigshot/eloot render at) against a live GTK install — these are taken on the author's word as reported from live testing, consistent with how the commit messages describe them.

mrhoribu:
# PR Review: feat(all): GTK shim slice four - packing, scrolling, presentation, grid weights (elanthia-online/lich-5#1637)

**Reviewed:** 18 files, +1954/-92 at `b030edee` (base `webui/03-slice-three-menus`) on 2026-09-16. Based on: full clone; no Ruby toolchain.
**Summary:** Sixteen commits of layout corrections plus two that are not layout at all — and those two (`flush!` outside its lock, and preserving an unsent edit across a background render) are the most valuable things in the PR. The layout work is sound; my notes are about bounds that are silently clamped rather than reported.

## Minor

### 1. A horizontal box silently drops everything past twelve children — `lib/common/script_scope/gtk/widgets.rb:1163`

```ruby
props = { count: children.length.clamp(1, 12), gap: [@spacing, 64].min }
weights = children.first(12).map do |child| ... end
```

`count` is clamped to 12 and `weights` is truncated to the first 12, but `render_children` still returns every child and `Container#materialize!` still attaches all of them. So a horizontal `Gtk::Box` with fifteen children declares twelve columns, weights twelve, and hands the client fifteen nodes. Whatever the client does with the overflow, it is not what the script asked for, and nothing is logged.

The rest of this file is careful to *report* what it cannot honour — `Gtk.log_render_failure` exists precisely for that, and `Gtk::Layout#layers` reports a non-image child rather than dropping it. This clamp should do the same: one `log_unsupported` when `children.length > 12` would turn a mystery layout into a one-line answer.

Same shape at `Table#column_count` and `Grid#column_count`, both `clamp(1, 24)`, and at `contract_margin`'s `clamp(0, 512)`.

### 2. `size_request_axes` decides by parent class, which changes after the fact — `lib/common/script_scope/gtk/widgets.rb:1667`

```ruby
def size_request_axes
  parent.is_a?(Window) ? [] : [:height]
end
```

This is read during `common_props`, i.e. at commit time, and `parent` is set by `Container#add`. A `ScrolledWindow` built, sized with `set_size_request`, and *then* added to a window reports `[:height]` on any commit that happens in between and `[]` afterwards — so `props[:height]` appears and then disappears, which `Widget#materialize!`'s diff turns into an `update` removing the property. Probably harmless given that scripts normally build bottom-up and show once, but it means the node's props depend on construction order rather than on final state.

### 3. The scroll-intent heuristic has a documented blind spot that is also the common case — `lib/common/script_scope/gtk/widgets.rb:1852`

`Adjustment#at_extent?` decides "the script means the bottom" by testing whether the written value lands exactly on `upper - page_size`. The commit message explains at length why magnitude alone was wrong and why this is better, and it is. But before the viewer has reported an extent, `upper` and `page_size` are the constructor's `100.0` and `0.0` — so *any* write of exactly `100.0` is read as "scroll to the bottom". The guard at `scroll_position`'s tail (`return nil if !viewport_known? && position[:bottom] && position[:x]`) catches the centring case by noticing that a log window never asks for a horizontal offset, which is clever, but it does not catch a script that scrolls a single axis to a computed `100`.

This is narrow enough that I would not block on it, but the assumption deserves a comment at `Adjustment#initialize` noting that the constructor defaults are load-bearing for `at_extent?` and should not be changed casually.

## Nits

- `ScrolledWindow#receive_event` sets `@centre_request = nil unless first_extent` *after* calling `replay_centre_request` (which already nils it). One of the two is redundant; keeping both makes the reader check whether the ordering matters.

- The `presentation` facility is declared through `ShimAdapter#presentation_source`, keyed by handle, with the window supplying its own reader. The indirection is well explained — handles are opaque, so the adapter cannot walk back to the widget — but `@presentation_sources` is never pruned when a handle is destroyed (`destroy_node!` clears `@placements` but not `@presentation_sources`). One stale closure per destroyed window.

## Open questions

- Sixteen commits, several described as "live-testing corrections", and the PR body says the fixes came from watching real scripts render. Were `bigshot`'s 518 one-sided margins and `vars`'s pack_end label rows checked *after* the grid-weight change (2.10) as well as after the margin fix, or were they verified at different points in the stack? **Settles it:** a note saying which scripts were re-checked at `b030edee` rather than at the commit that fixed each symptom. Layout fixes interact, and this is the PR where four of them land together.

## What this PR gets right

Two commits here are worth more than the layout work and should not get lost in the stack:

- **`flush!` refreshes outside its lock** (`lib/webui/adapter.rb:240`). The comment is the clearest thing in the series: `@service.refresh` reaches `Page#render`, which takes the page's render mutex and then calls back into `render_children`, which takes the adapter mutex — while a viewer attaching on the connection thread walks the same two locks in the other order. Holding both across the refresh is a lock-order inversion between the script thread and the connection thread. Identifying that from symptoms, and fixing it by moving only the refresh out while everything that *reads* adapter state stays in, is exactly right.

- **A background render keeps an unsent edit, its focus and its caret.** The tree is rebuilt on every commit, and scripts repaint on game events — so without this, typing into a `;vars setup` field while anything at all happened in the game would lose the keystrokes. This is the difference between a demo and something a player will actually use.

- **GTK's default is not to expand, so an unweighted grid hugs its content.** Getting this backwards is the single most common way a GTK-to-web port looks wrong, and the fix is correctly scoped to "only columns a child asked to expand into get a share".

- The `pack_start(child, 0, 0, 0)` GTK-2 positional form, where the flags are integers and `0` means false despite being truthy in Ruby, is handled explicitly in `packing_flag`. That is the kind of thing that is invisible until a decade-old script renders wrong.

## Coverage notes

Read in full: the `widgets.rb` diff (+493), `adapter.rb` diff, `gtk_packing_spec.rb` as intent. `app.js` (+204) reviewed around the scroll/extent reporting and the columns renderer, not line by line. **No Ruby in this environment** — "257 examples, 0 failures" is unverified, and the layout behaviour in particular is the kind that only a running browser can really confirm.

Nisugi:
## Rubocop backlog cleared; the retry loop was already fixed

Two items from the reviews land on this PR.

**The rubocop backlog (`35b45529`, at the tip).** The review traced the red CI to somewhere inside this PR's 16 commits, with every subsequent PR adding to the pile. Cleared: across the **101 Ruby files this branch touches, rubocop reports no offenses**.

Most were layout and autocorrectable. The ones that were not:

- Two `Lint/Void` in `images.rb` were *created* by the `set_*` mutator fix — `value` and `path` were trailing so the old aliases returned something useful, and `def_setter` ignores the writer's return.
- `PixbufSources` keyed on `pixbuf.object_id`, which is what `compare_by_identity` spells properly; it also no longer leaves an integer key behind once the pixbuf is collected.
- `rescue StandardError, ArgumentError` shadowed the second — `ArgumentError` *is* a `StandardError`.
- Non-ASCII split three ways rather than one. Typographic ellipses in comments became `...`. PNG/JPEG magic bytes and a deliberately non-ASCII tooltip fixture are load-bearing, so they carry scoped disables with the reason — as the review suggested, rather than `-A`. Two unused keyword arguments (`Dialog`'s `parent`/`flags`, presentation's `borderless`) are the documented shape of APIs callers still pass to, so they are disabled in place; renaming them would break those callers.

I did *not* blanket-run `rubocop -A` across the stack. On two map-interaction assertions autocorrect produced a column-42 continuation that is compliant and materially less readable, so those were restructured by hand (the payload is a local now). Worth knowing before anyone runs `-A` on the rest of the repo's 838 remaining offenses.

**The stale-generation retry loop — already fixed.** The review lists this as "fix the client-side stale-generation retry loop (unbounded retry + wrong-event replay)", corroborated by CodeRabbit. Checking `app.js` at head, it is closed: `pendingEvent` is cleared *before* replay, `UNSOLICITED` keeps a scroll report from stealing a click's retry slot, and a `retriedEvents` set keyed on `cid + event` bounds it to one attempt. That came in with `648a0d40` ("...stale events"), which post-dates the state the review appears to have read. No change made — flagging it so the item can be closed rather than re-fixed.

### Still open from this PR

**A horizontal box silently drops weights past 12 children**, and grid columns past 24. Both are part of the second review's "degrade honestly is applied inconsistently" theme — six places that clamp or drop without a `log_unsupported`. Worth doing as one batch across all six rather than piecemeal.

Nisugi:
## The silent clamps from this PR now report themselves

Fixed at the tip in `f2d581ba`, as part of the second review's "degrade honestly is applied inconsistently" theme.

- **A horizontal box past twelve children** silently lost the excess weights.
- **A grid past twenty-four columns** — both `column_count` implementations.

The clamps themselves are correct: 12 and 24 are the contract's own maxima (`count: property(integer(min: 1, max: 12))`), not arbitrary limits. What was missing was saying so, since a dropped column turns a script bug into a rendering mystery. Each now reports once per class and property through a new `Gtk.log_clamped`, deduped like the shim's existing signals so a commit loop does not flood the log, and there is a spec asserting **silence when the box fits** — a reporter that fires on every render is worse than none.

Went in alongside the same treatment for margins past 512px and `Paned#add` on an occupied pane.

**`Adapter#handle_for` — `66f198df`.** Also relevant here, since the review named it as one of the two highest-leverage fixes for the commit-path performance theme. It was `Hash#key(node)` — a scan over values where each candidate is a memberwise `Struct` `==` recursing into `props` and `children`:

```
401-node tree, 200 calls:  32.4ms  ->  ~0ms through a reverse identity map
```

The other half — **dirty-gated commits** — I attempted and backed out. A session-level flag set by `Widget#changed!` skipped commits that had real work to do (44 failures), because `changed!` marks the widget and not its ancestors, and structural mutations do not route through it at all. Recorded rather than guessed at: **1.3ms per no-op job on a 200-row window**, and it is the same root cause as #1641's "every key press re-materializes every window". Worth its own PR with every mutation path audited.

The rubocop backlog this PR started remains clear — **102 files, no offenses**.

Nisugi:
Superseded by the re-layered stack: this work now lives in webui/shim-core (L6a). See docs/webui-rebuild-plan.md on webui/core-corrections for the map from the old PRs to the new layers. The branch is left in place for reference.

PR #1638, opened 2026-09-16
fix(all): what ;map needed before it could render

Nisugi:
**Stacked on #1637.** Diff is against `webui/04-slice-four-layout`. Three commits.

Small, but they are the difference between `;map` raising on load and `;map` getting far enough to be debugged — so they are separated out rather than buried in the images PR.

- **Every widget reports an allocation.** A script reads `allocation.width` and does arithmetic on it. Only `Window` answered, so anything else returned nil — and Lich's `NilClass#method_missing` patch turns `nil.width / 2` into `coerce must return [x, y]`, raised from inside `Integer#+`, naming neither the widget nor the method. `;map` died there. Every widget now answers with the size it asked for, falling back to its window's default, which is what GTK reports before the first allocation anyway.

- **A stubbed widget class says so.** `Gtk.const_missing` degrades an unimplemented widget to an empty container, which is right — but an empty box on screen is indistinguishable from a layout bug, and the only trace was one `warning:` line in a debug file. `map`'s `Gtk::Image` *is* the map. It now tells the player, once per class per session.

- **The launcher's entry rows lose the gap** the bare-page rule had re-added.

## Testing

`bundle exec rspec spec/lib/webui spec/lib/common/script_scope` → **271 examples, 0 failures**.


mrhoribu (changes requested):
# PR Review: fix(all): what ;map needed before it could render (elanthia-online/lich-5#1638)

**Reviewed:** 5 files, +256/-11 at 0477c2bc on 2026-09-16. Based on: full clone, changed files at head, and full repo for cross-referencing.
**Summary:** The three commits do what the PR body says: `Widget#allocation` now answers with numbers on every widget (verified complete across the class hierarchy, and by hand-tracing the parentless-widget and no-default-size cases), `Gtk.const_missing` now tells the player once a stubbed widget class is actually used, and the CSS specificity fix for launcher entry rows is correctly computed and doesn't regress the bare-page group rule it was fighting. Locally reproduced the PR's own test claim (`bundle exec rspec spec/lib/webui spec/lib/common/script_scope` → 271 examples, 0 failures). Two real issues found: the "once per class per session" warning is actually once per class per **process** (contradicts the PR's own stated design and quietly reopens the exact bug this PR fixes, just delayed to the second occurrence), and this PR's own new comments introduce non-ASCII characters that are currently failing the required Rubocop CI check on the live PR.

## Major

### 1. The stubbed-widget warning is "once per class per process," not "once per class per session" as the PR claims

`lib/common/script_scope/gtk/widgets.rb:169-170`:
```ruby
@unsupported = {}
@dropped = {}
```
This is a module-level instance variable on the `Gtk` module itself, assigned exactly once when the file is loaded (Ruby process boot), and it is **never reset anywhere** — confirmed by `grep -rn "@unsupported\|@dropped" lib/common/script_scope/gtk/*.rb`, which shows only the init and the three guard/set sites, no clear/reset call. `report_stubbed_widget` (`widgets.rb:291-307`) uses this same hash:
```ruby
def report_stubbed_widget(name)
  key = "Gtk::#{name}"
  return if @unsupported[key]

  @unsupported[key] = true
  ...
  Kernel.send(:respond, "[#{script || 'gtk'}: Gtk::#{name} is not supported yet -- ...")
```

Meanwhile, "session" is a defined, specific concept in this same codebase: `Session.for(owner)` creates one session per script owner and `Session.release(owner)` tears it down, hooked to the script's own `at_exit` (`lib/common/script_scope/gtk/session.rb:177-186, 234-247`). A session is a single script run, not the life of the Lich process.

**Concrete trigger:** Player runs `;map`, it hits `Gtk::Image` via `const_missing`, `report_stubbed_widget` fires and the player sees `[map: Gtk::Image is not supported yet -- that part of the window will be blank]`. Player exits `;map` (session released) and, ten minutes later in the same still-running Lich process, runs `;map` again — a brand-new session by this codebase's own definition. `@unsupported["Gtk::Image"]` is still `true` from the first run, so `report_stubbed_widget` returns at the guard on line 293 and never calls `respond`. The player gets a silent empty box — exactly the "indistinguishable from a layout bug" state this PR's second bullet exists to eliminate — with no warning at all, because the warning already fired once, for a different session, possibly hours or days earlier.

This is squarely on-theme for a PR whose entire second bullet is about making a silent failure mode loud: the fix helps the *first* session in a given Lich process and silently reverts to the old behavior for every session after that. Grounded, confirmed (not speculative) — the scoping mismatch is directly readable from the code, not inferred.

**Fix shape:** key the dedup by the current session/owner (e.g. `Session.current.owner_label` or similar) in addition to the class name, or reset the relevant subset of `@unsupported` when a new session starts, rather than reusing the log-only, process-lifetime dedup hash for a player-facing message that the PR body itself says should be per-session.

## Minor

### 2. This PR's own new comments fail the repo's Rubocop CI check that is running on this exact PR right now

`bundle exec rubocop` (the repo's own canonical command, run from `/tmp/reviewpr-lich-5-1634-1641/wt-1638`) flags, among others:
```
lib/common/script_scope/gtk/session.rb:516:32: C: Style/AsciiComments / Custom/AsciiOnlySource
            # "map:2466", not "…/map.lic:2466". Matching only ".lic:" found
lib/common/script_scope/gtk/session.rb:539:14: C: Style/AsciiComments / Custom/AsciiOnlySource
          # "…/scripts/map.lic:2462:in 'block'" -> "map.lic:2462".
spec/lib/common/script_scope/gtk_packing_spec.rb:777:22: C: Style/AsciiComments / Custom/AsciiOnlySource
  # "map:2466", not "…/map.lic:2466". Matching only ".lic:" found no frame at
```
Both cops are `Enabled: true` in `.rubocop.yml:78-82`. These three lines are new in this PR's diff (the "…" ellipsis character in the `script_origin`/`report` explanatory comments added by commit `517fa10c`, and its mirror in the new spec). Confirmed live: `gh pr checks 1638 --repo elanthia-online/lich-5` shows `Run Rubocop on Ruby  fail`, and the CI log (`gh run view 35137784592 --log`) shows exactly these six offense entries (three lines × two cops each) plus 9 unrelated pre-existing offenses.

To be precise about attribution: I checked out `origin/webui/04-slice-four-layout` (this PR's base) into the working tree and ran `bundle exec rubocop` there — the base branch already has 11 offenses (`contract.rb`, `gtk_shim_spec.rb`, and a pre-existing `widgets.rb:215` implicit-concatenation warning that this PR doesn't touch), so the Rubocop check was **already red before this PR**. This PR adds 4 more offense entries on top of that (the two new ellipsis lines in `session.rb`, each flagged by two cops), bringing the total from 11 to 15, matching the CI log's `1258 files inspected, 15 offenses detected`.

Net: this PR does not break a previously-green required check, but it does add new, self-authored lint violations to an already-failing one — the workflow's own header comment (`.github/workflows/rubocop.yaml:1-14`) states its purpose is to "keep the RuboCop check present and green for branch protection," which it currently is not, partly because of this PR. Trivial one-line-per-site fix: replace `…` with `...` in the three comments (all three are `[Correctable]`, i.e. `rubocop -A` fixes them automatically).

## Open questions

### 3. `MessageDialog` doesn't participate in the allocation/protocol contract this PR completes for everything else
`lib/common/script_scope/gtk/widgets.rb:2413` — `class MessageDialog` (no superclass, not a `Widget`/`Container`) has no `allocation`, no `PROTOCOL_METHODS` guard, and no `method_missing` at all; calling an undefined method on it raises a plain Ruby `NoMethodError` naming the class directly (not the confusing coerce failure this PR is about, since MessageDialog never silently returns `nil`). This is pre-existing and untouched by this diff, and `MessageDialog` is a modal shown via `@session.modal(...)` rather than something embedded in a widget tree a script would query `.allocation` on, so it's likely legitimately out of scope rather than a gap in "every widget now answers." **Settles it:** does any real `.lic` script call `.allocation`/`width`/`height` (or otherwise do arithmetic) on a `Gtk::MessageDialog` instance? If not, this is correctly out of scope and can be dropped.

### 4. `allocation`'s fallback mirrors the *requested* size, never the browser's actual rendered size — acknowledged in the PR's own comment, but worth confirming intent for the next PR in the stack
`widgets.rb:588-591` says outright: "Only the browser knows the real one, so this answers with the size the widget asked for." Since there is no feedback path from the browser back into `@width_request`/`@height_request` in this diff, `allocation.width` will always equal what the script itself requested (or the window default), never the true post-layout size — even after a user resizes the browser window. This is self-disclosed, not hidden, so I'm not raising it as a defect, but per the task's own framing (PR #1639 builds real image/map rendering next): **settles it** — does #1639 wire actual browser-reported geometry back into `allocation`, or is the requested-size approximation intended to remain permanent? If the latter, `center_viewport_on`-style math will be systematically off by however much real layout differs from the request, which would be worth flagging explicitly to whoever picks up #1639 rather than assumed away.

## What this PR gets right

- **Allocation completeness is real, not just claimed.** Traced every widget class definition in `widgets.rb`/`widgets_data.rb`: every on-screen widget (`Box`, `Table`, `Grid`, `ScrolledWindow`, `Entry`, `Button`, etc.) descends from `Widget` and calls `super()` in `initialize` (spot-checked `Box`, `Table`, `Entry`, `ComboBox`), so `@width_request`/`@height_request` are always initialized and `allocation` (widgets.rb:599-606) resolves correctly. The only classes without it (`CellRenderer*`, `TreeViewColumn`, `Adjustment`, model/iterator classes) are non-widgets in real GTK too, so the omission is correct, not accidental.
- **The parentless-widget case is handled, not just asserted.** Hand-traced `window_root` → `toplevel`: a widget with `@parent == nil` returns itself from `toplevel`, fails `is_a?(Window)`, so `window_root` returns `nil`, and `@width_request || root&.default_width || 640` still resolves to the literal `640`/`480` fallback rather than raising or silently going through `NilClass#method_missing`. Confirmed by the PR's own new spec ("still answers on a parentless widget") and by re-running it locally.
- **The CSS specificity fix is correctly computed, not guessed.** `.webui-page .webui-group[data-cid*="group:entry-"]` is 2 classes + 1 attribute = specificity (0,3,0); the bare-page variant `.webui-page.bare .webui-group[data-cid*="group:entry-"]` is 3 classes + 1 attribute = (0,4,0), which beats the competing `.webui-page.bare .webui-group` rule at (0,3,0) (`app.css:235`) that was re-adding the margin/padding on bare pages. Grepped for other `.webui-page.*` modifier classes — only `.bare` exists — so there's no third page-variant this fix could be blindsided by.
- **The `coerce` fix is verified to produce the right error, not just "no longer nil."** `Widget::PROTOCOL_METHODS` lets `coerce`/`+`/`-`/etc. fall through to `super` (real `NoMethodError`) instead of the old catch-all `method_missing` that answered every unknown method with `nil`. Ran the PR's own test for this (`100 - renderer` raises `TypeError, /CellRendererText can't be coerced/`, not the old unhelpful "coerce must return [x, y]") and it passes, matching real Ruby coercion semantics (verified directly: `renderer.coerce` undefined → `NoMethodError` internally caught by `Integer#-`'s C implementation → the class-naming `TypeError`, reproduced with a standalone Ruby snippet).
- **`script_origin`'s bare-script-name matching is correct for the cases exercised**, including the boundary case that motivated it (Lich evals scripts under their bare name, e.g. `"map:2466"`, not `".../map.lic:2466"`), and falls back correctly to the old `.lic:`-matching behavior when the owner name doesn't appear in the backtrace.
- Ran the PR's exact testing claim myself: `bundle exec rspec spec/lib/webui spec/lib/common/script_scope` → **271 examples, 0 failures**, matching the PR body precisely.

## Coverage notes

- `coderabbit review --agent --base origin/webui/04-slice-four-layout --committed` completed cleanly (not stalled) and returned **0 findings** across all 5 changed files. Treated as corroboration only; it did not surface either of the two issues above, both of which came from tracing session/process lifetime and running the repo's own Rubocop command.
- GitHub's native CodeRabbit auto-review is disabled for this repo (confirmed via the PR's own comment thread), so there was no separate bot review to cross-check against.
- No human review comments exist on the PR yet (`reviews: []`).
- Ran the full claimed test scope and the single spec file directly (74 examples in `gtk_packing_spec.rb` alone, 271 across the claimed scope) — both matched the PR body's numbers exactly.
- Ran `bundle exec rubocop` against both HEAD (`0477c2bc`) and the PR's base (`origin/webui/04-slice-four-layout`) to separate pre-existing offenses from ones this PR introduces, rather than asserting from the CI log alone.
- Did not attempt to verify the PR body's claim that GTK "reports [the requested size] before the first allocation anyway" against real GTK behavior — no GTK runtime available in this sandbox to check against; flagged as an open question rather than asserted true or false.
- CSS specificity computation was done by hand against the CSS Selectors spec's standard (id, class-or-attribute-or-pseudoclass, element) counting rule; not independently re-verified in a real browser, but the rule itself is unambiguous and not GTK-runtime-dependent.

mrhoribu:
# PR Review: fix(all): what ;map needed before it could render (elanthia-online/lich-5#1638)

**Reviewed:** 5 files, +256/-11 at `0477c2bc` (base `webui/04-slice-four-layout`) on 2026-09-16. Based on: full clone; no Ruby toolchain.
**Verdict:** Approve
**Summary:** Three small, well-motivated fixes, correctly separated from the images PR. The allocation change in particular is the right answer to a genuinely nasty diagnostic problem. Nothing here blocks; two notes below are about honesty of the fallbacks, not correctness.

## Minor

### 1. `allocation` fabricates 640×480 when there is no window root — `lib/common/script_scope/gtk/widgets.rb:766`

```ruby
def allocation
  root = window_root
  Allocation.new(0, 0,
    @width_request || root&.default_width || 640,
    @height_request || root&.default_height || 480)
end
```

The reasoning in the commit message is sound and I agree with the call: a widget that answers `nil` here becomes `nil.width / 2`, which lich-5's `NilClass#method_missing` patch turns into `coerce must return [x, y]` raised from inside `Integer#+`, naming neither the widget nor the method. Answering with the requested size — which is what GTK reports before the first allocation anyway — is strictly better.

The residual cost is that a widget with no window at all now returns a confident 640×480 rather than anything a script could detect. A script doing `allocation.width / 2` on an unparented widget gets a plausible number and centres on the wrong point, silently. That is a better failure than the old one, but it is still silent.

Since the shim already has a channel for exactly this (`Gtk.log_unsupported`, deduplicated per class), one line when `window_root` is nil would cost nothing and would turn "the map centred somewhere odd" into a log entry. `ScrolledWindow#allocation` already does the honest thing by preferring the viewer's reported extent; the base case could at least say when it is guessing.

### 2. `report_stubbed_widget`'s guard and its call do not test the same thing — `lib/common/script_scope/gtk/widgets.rb:355`

```ruby
return unless defined?(::Lich::Messaging) || Kernel.respond_to?(:respond, true)
Kernel.send(:respond, "[#{script || 'gtk'}: Gtk::#{name} is not supported yet -- ...]")
rescue StandardError
  nil
```

The guard passes if *either* `Lich::Messaging` is defined *or* `Kernel` responds to `respond`, but the call only uses the second. In a context where `Lich::Messaging` exists but `respond` has not been mixed into `Kernel`, the guard lets the call through, `NoMethodError` is raised, and the bare `rescue StandardError` swallows it — so the player is never told, which is the exact outcome this method exists to prevent.

Either drop the `defined?(::Lich::Messaging)` clause (it is not used) or route through `Lich::Messaging` when that is what is available.

This matters more than it looks because the whole point of the change is that *an empty box on screen is indistinguishable from a layout bug*. If the one channel that reaches the player fails open into a `rescue nil`, the fix is only half-landed on whichever platform hits that combination.

## Nits

- `@unsupported` is now doing double duty: `log_unsupported` keys it as `"#{klass}##{method}"` and `report_stubbed_widget` keys it as `"Gtk::#{name}"`. No collision is possible given the shapes, but two different dedupe policies (once per class per session for a stubbed widget; once per method for an unsupported call) sharing one hash is worth a comment.

- The launcher CSS fix (`app.css`, -12/+12) is unrelated to `;map` and is the kind of thing that gets lost in a PR titled after something else. Not worth splitting at this size, but the changelog entry should mention it.

## What this PR gets right

- **Separating these three from the images PR was the correct call**, and the PR body says why in one sentence: they are the difference between `;map` raising on load and `;map` getting far enough to be debugged. Reviewing 256 lines that make a thing debuggable is much easier than finding them inside 2465 lines that make it render.

- **The diagnosis in the first commit is the valuable artifact.** "`nil.width / 2` raises `coerce must return [x, y]` from inside `Integer#+`, naming neither the widget nor the method" is precisely the kind of ecosystem-specific trap that costs someone a day and is never written down. It is written down now, in a commit message, next to the fix.

- **"A stubbed widget class says so"** correctly identifies that the previous behaviour was *right in kind but wrong in volume* — degrading to an empty container is the correct degradation, and the bug was that nobody could tell it had happened.

## Coverage notes

All five changed files read in full. `gtk_packing_spec.rb` (+162) read as intent. **No Ruby in this environment** — "271 examples, 0 failures" is unverified. The `Lich::Messaging`/`Kernel#respond` combination in Minor 2 is reasoned from the guard, not reproduced; whether it is reachable depends on load order I did not trace.

Nisugi:
Superseded by the re-layered stack: this work now lives in webui/contract-primitives (L3) and webui/shim-core (L6a). See docs/webui-rebuild-plan.md on webui/core-corrections for the map from the old PRs to the new layers. The branch is left in place for reference.

PR #1640, opened 2026-09-16
feat(all): slice five's long tail - Paned, Overlay, ProgressBar, ListBox, Dialog

Nisugi:
**Stacked on #1639.** Diff is against `webui/06-images-and-map`. One commit.

Five widget classes that each fell through `Gtk.const_missing` into an empty box, plus the `Stock` labels a dialog's OK button names:

- **`Paned` / `HPaned` / `VPaned`** → contract `split`, with named first/second slots and pixel-to-percent conversion against the window axis.
- **`Overlay`** → contract `overlay`, base plus stacked children in document order.
- **`ProgressBar`** → contract `progress`, including `pulse` as indeterminate.
- **`ListBox` / `ListBoxRow`** → a stack of group rows.
- **`Dialog`** (the custom kind, not `MessageDialog`) with `content_area`, `add_button`, `respond` and a blocking `run`.

`Dialog#run` is the interesting one. `gtk_dialog_run` blocks the main thread while still servicing events. Here `run` is called from a queued job — a button handler — and the response arrives as *another* queued job, so it pumps the session queue on-thread rather than deadlocking against itself. It returns the exact response object the script passed to `add_button`, and `DELETE_EVENT` when the viewer closes it.

`split` and `overlay` had contract types but no client renderer, so spellson and jinx rendered blank panes; those renderers land here too.

## Testing

`bundle exec rspec spec/lib/webui spec/lib/common/script_scope` → **353 examples, 0 failures**.


mrhoribu (changes requested):
# PR Review: feat(all): slice five's long tail - Paned, Overlay, ProgressBar, ListBox, Dialog (elanthia-online/lich-5#1640)

**Reviewed:** 8 files, +738/-5 at 7aeb8541 on 2026-09-16. Based on: full clone, changed files at head, full repo for cross-referencing, and live reproduction scripts run against the actual `Session`/`Dialog` classes in this worktree (not just static reading).
**Verdict:** Request changes
**Summary:** The five new widget classes (`Paned`, `Overlay`, `ListBox`/`ListBoxRow`, `ProgressBar`) are solid — boundary cases for the pixel-to-percent conversion are handled correctly and the existing test suite passes (25/25, confirmed). `Dialog#run`'s blocking semantics, however, do not deliver on the PR's own headline claim. I built and ran three small reproductions against the actual code in this worktree: (1) closing the browser tab the ordinary way (no script `:delete_event` handler — the common case for a plain confirmation dialog) never unblocks `run`; (2) a script exiting/being killed while a dialog is showing leaves the session thread spinning forever instead of returning `DELETE_EVENT`; (3) two threads both calling `run` on the same `Dialog` leaves the second one hung permanently. All three are confirmed hangs, not just theoretical races. There is also a confirmed (if narrow) `ProgressBar#show_text` logic bug and two new rubocop violations introduced by this diff.

## Blockers

### 1. `Dialog#run` never unblocks when the viewer closes the tab and the script has not connected `:delete_event` — contradicts the PR's own stated contract
**File:** `lib/common/script_scope/gtk/widgets.rb:2023-2032` (inherited, unmodified `Window#viewer_closed`/`#browser_exited`) and `widgets.rb:2195-2202` (`Dialog#run`), confirmed against `lib/common/script_scope/gtk/session.rb:552-560` (`bind_lifecycle`) and `lib/webui/assets/app.js:1161-1174` (`detachPages`/`pagehide`).

The PR body explicitly promises: *"It returns the exact response object the script passed to `add_button`, and `DELETE_EVENT` when the viewer closes it."* The mechanism for "the viewer closes it" (browser tab X, navigate away, refresh) is the client's `pagehide` handler sending an explicit `detach` message (`app.js:1172-1174`), which the server routes to `enqueue_lifecycle(attachment, :close, reason: :user)` (`lib/webui/runtime.rb:199`), which `bind_lifecycle` wires to `window.viewer_closed` (`session.rb:558`):

```ruby
adapter.bind(handle, :close, proc { |_context| enqueue { window.viewer_closed } })
```

`Window#viewer_closed` (unchanged by this PR, `widgets.rb:2023-2028`) does only this:

```ruby
def viewer_closed
  return if @delete_emitted || @destroyed
  @delete_emitted = true
  emit(:delete_event, Event.new(:delete))
end
```

It emits a signal — it does **not** call `destroy`, does **not** set `@response`, and does **not** push to `@responses`. Real GTK's default `delete-event` handler destroys the window when nothing vetoes it; nothing in this shim reproduces that default. `Dialog` doesn't override `viewer_closed` to hook in the response mechanism the way it overrides `destroy` and `browser_exited` (`widgets.rb:2208-2216`) — those two paths *do* push `DELETE_EVENT` to `@responses`, but the ordinary "user closed the tab" path does not go through either of them.

I reproduced this directly against the real classes (script at `/tmp/claude-1000/-home-ryan/eaa68042-daf2-42b8-b84f-12e7cfa6a726/scratchpad/repro_viewer_closed_spec.rb`): built a `Dialog.new(title: 'CloseMe', buttons: [['OK', :ok]])` (the exact pattern the PR body itself uses for creaturebar/map — no `:delete_event` handler connected, which is the common case since a script relying on `run`'s documented contract has no reason to connect one), called `run` from a separate thread, then invoked `dialog.viewer_closed` (exactly what `bind_lifecycle`'s `:close` binding does) to simulate the tab being closed:

```
outcome after viewer_closed with no delete_event handler: :HANG
```

`run` hung past a 2-second timeout with no sign of resolving — it will hang forever, since nothing else will ever touch `@response`/`@responses` for that dialog. A script written against the documented contract ("it always returns DELETE_EVENT when the viewer closes it") will freeze the session thread the first time a player closes the dialog's tab instead of clicking a button — a worse failure than the "custom dialogs are not rendered yet" stub this PR replaces, which at least returned immediately.
**Confidence:** confirmed.

### 2. `Dialog#run` spins forever instead of returning after `session.shutdown` (script exit/kill) while a dialog is showing
**File:** `lib/common/script_scope/gtk/session.rb:357-372` (`pump`), `session.rb:391-407` (`close_window`), `session.rb:498-511` (`shutdown`), `widgets.rb:2197-2201` (`Dialog#run`'s on-session-thread branch).

`pump`'s own doc comment states its purpose precisely: *"Returns false when nothing was waiting within +timeout+ seconds, so the caller can re-check its own exit condition. A :stop is put back for run_loop."* But `Dialog#run`'s caller does not re-check anything using that return value:

```ruby
@session.pump(0.05) until @response || destroyed?
```

`pump`'s return value is discarded by `until`; the loop's only exit conditions are `@response` and `destroyed?`. When `session.shutdown` runs (wired to the owning script's `at_exit`, `session.rb:610-616`), it calls `close_window` for every registered window — including a showing `Dialog`, since `Dialog < Window` and `Window#initialize` self-registers (`widgets.rb: register_window` call in `Window#initialize`). `close_window` destroys the *adapter node* (`adapter.destroy(handle)`) but never calls the widget's own `destroy` — so `@destroyed` is never set, and nothing sets `@response`. `shutdown` then pushes `@queue << :stop` (`session.rb:509`). Every time the pumping `run` call happens to pop that `:stop` sentinel, `pump` puts it right back (`@queue << :stop`, `session.rb:362`) so `run_loop` can still see it — but `Dialog#run`'s own loop just calls `pump` again immediately, forever, since it never inspects the sentinel or `@closed`.

This is a real regression relative to the codebase's own established pattern for exactly this problem: the pre-existing modal system's `ModalCoordinator#terminate_owner` (`lib/webui/modal_coordinator.rb:63-68`) explicitly cancels every pending `Future` on owner termination (`futures.each { |future| future.cancel(reason: :terminated) }`). `Dialog#run` invents an entirely separate blocking mechanism (`@response`/`@responses`) that has no equivalent hook into `terminate_owner`/`shutdown`.

Reproduced directly (`/tmp/claude-1000/-home-ryan/eaa68042-daf2-42b8-b84f-12e7cfa6a726/scratchpad/repro_dialog_shutdown_spec.rb`): enqueued a job that calls `dialog.run` on the session thread, let it start pumping, then called `session.shutdown`:

```
HANG CONFIRMED: dialog.run() never returned within 3s after session.shutdown
```

Practical effect: the session thread never returns from this call, so `run_loop` never gets back to see its own `:stop` and exit cleanly. Whether the script's thread group ultimately force-kills this thread (per the comment at `session.rb:203-206`) depends on Lich's own script-kill timing, which is outside this diff — but the shim's own graceful-shutdown path does not resolve `run`, which is the guarantee the PR claims.
**Confidence:** confirmed.

## Major

### 3. Two threads calling `Dialog#run` on the same dialog off the session thread: the second one hangs forever
**File:** `widgets.rb:2195-2202` (`run`), `widgets.rb:2187-2192` (`respond`), `widgets.rb:2142` (`@responses = Queue.new`).

The off-session-thread branch is `@responses.pop` — a single-consumer wait on a `Queue`. `respond` (called once per click) and `destroy`/`browser_exited` (called once per close) each push exactly **one** item. If two different threads call `dialog.run` concurrently off the session thread (plausible: a script's timer/event thread and its main flow both awaiting the same confirmation dialog), only one `.pop` ever receives the pushed value; the other blocks on an empty queue forever.

Reproduced (`/tmp/claude-1000/-home-ryan/eaa68042-daf2-42b8-b84f-12e7cfa6a726/scratchpad/repro_double_run_spec.rb`): two threads each call `dialog.run`, then the OK button is activated once:

```
waiter1: :ok, waiter2: :TIMEOUT
```

The second waiter timed out at 2s with no resolution — it hangs indefinitely, not merely "resolves out of order." Nothing in `run`, `respond`, `destroy`, or `browser_exited` guards against or documents this as unsupported.
**Confidence:** confirmed.

## Minor

### 4. `ProgressBar#show_text` is dead code — setting it to `false` never hides a set `text`
**File:** `lib/common/script_scope/gtk/containers.rb:74-79`.

```ruby
def node_props
  props = { value: @fraction }
  props[:label] = @text if @text && !@text.empty? && (@show_text || !@text.empty?)
  props[:indeterminate] = true if @pulsing
  props
end
```

The guard is `A && B && (C || B)` where `B` is `!@text.empty?`. Since `B` is already required true by the preceding `&&`, `(C || B)` is always true regardless of `C` (`@show_text`) — `@show_text` has no observable effect. Verified directly:

```
$ ruby -e 'show_text=false; text="Loading"; puts !!(text && !text.empty? && (show_text || !text.empty?))'
true
```

A script that sets `bar.text = 'Spirit Warding I'` then `bar.show_text = false` (to temporarily hide the label while still tracking it, matching real `Gtk::ProgressBar#show_text` semantics) will still see the label rendered. The existing spec (`spec/lib/common/script_scope/gtk_slice5_spec.rb:668-685`) only exercises `text` presence and `pulse`, never `show_text = false` with a set `text`, so this shipped untested. Likely intended: `props[:label] = @text if @show_text && @text && !@text.empty?`.
**Confidence:** confirmed.

### 5. Re-adding the same child to the same `Paned` slot duplicates it in the render list
**File:** `containers.rb:271-279` (`place`).

```ruby
def place(child, slot)
  existing = @slotted.key(slot)
  super_remove(existing) if existing && !existing.equal?(child)
  @slotted[child] = slot
  Container.instance_method(:add).bind_call(self, child)
  self
end
```

The guard `!existing.equal?(child)` correctly skips removing the *existing occupant* when re-adding the *same* object to its own slot, but the unconditional `Container.instance_method(:add).bind_call(self, child)` on the next line still appends `child` to `@children` again — `Container#add` has no dedup guard. Verified at the unit level (`/tmp/claude-1000/-home-ryan/eaa68042-daf2-42b8-b84f-12e7cfa6a726/scratchpad/repro_paned_dup_spec.rb`): calling `paned.add1(child); paned.add1(child)` on the same object leaves 2 occurrences of `child` in `paned.send(:render_children)`. I was not able to cleanly reproduce the full render-pipeline consequence within this review's time budget (my attempts to observe it end-to-end through `materialize!`/`commit` gave inconsistent results, likely a test-harness timing artifact rather than evidence the bug is harmless) — flagging as suspected downstream impact, confirmed at the data-model level. Trigger: a script calling `add1`/`add2`/`pack1`/`pack2` a second time on a widget already in that slot (e.g. idempotent "ensure this widget is placed" code), instead of first removing it.
**Confidence:** confirmed (duplicate entry in `@children`/`render_children`); suspected (visible rendering consequence).

## Nits

### 6. New rubocop violations introduced by this diff
Run: `rubocop lib/common/script_scope/gtk/containers.rb lib/common/script_scope/gtk/session.rb lib/common/script_scope/gtk/widgets.rb spec/lib/common/script_scope/gtk_slice5_spec.rb` (repo's own configured cop set, Ruby 4.0.5).

- `widgets.rb:2135` — `Lint/UnusedMethodArgument` on both `parent:` and `flags:` in `Dialog#initialize(title: nil, parent: nil, flags: nil, buttons: nil, **_options)`. Every other ignored parameter in this file follows the `_name` convention (e.g. `_options`, `_args` elsewhere in the same class); these two should be `_parent:`/`_flags:` to match.
- `spec/lib/common/script_scope/gtk_slice5_spec.rb:136` — `Lint/UselessAssignment` on `dialog` in the `'renders its content and buttons and validates'` example; the local is assigned from `session.sync` but never read afterward.

(The other four offenses rubocop reports in these files — two `AsciiOnlySource`/`AsciiComments` at `session.rb:617,640` and one `ImplicitStringConcatenation` plus one `EmptyLineBetweenDefs` at `widgets.rb:215,2791` — are all outside this diff's changed hunks; pre-existing, not introduced here.)
**Confidence:** confirmed.

## Open questions

- **Job-ordering during a nested pump (review point 1b).** Traced through: since `run` blocks synchronously, a script's own code after `dialog.run` cannot enqueue anything until `run` returns, so the scenario "script queues A, user's click queues B, script's own code queues C" cannot actually produce C-before-B — C is only reachable after `run` (and therefore B) has resolved. Any *unrelated* third job queued by something else entirely (a timer, another window's handler) is processed by `pump` in strict FIFO arrival order, same as `run_loop` would — this matches nested-mainloop semantics in real GTK and isn't a bug I could find. Flagging as an open question rather than a finding since I couldn't construct a concrete script pattern where this diverges from GTK's own recursive-mainloop behavior.
- **Recursion depth on legitimately nested `run` calls (review point 1c, same-thread case).** A button handler that opens and `run`s a second dialog from within the first dialog's pump loop adds one Ruby stack frame per nesting level (`run` → `pump` → `job.call` → `run` → ...) — normal, bounded nesting (a handful of levels) is fine, but nothing caps it. Not flagging as a finding since I have no evidence of a realistic script pattern that nests deeply enough to matter; worth asking the author whether any corpus script is known to chain dialogs this way.
- Does Lich's script-kill path force-`Thread#kill` the GTK session thread promptly enough that Blocker #2's spin loop is harmless in practice (thread dies moments after `shutdown` regardless)? That's outside this diff (it's in Lich's script-lifecycle code, not `lib/common/script_scope/gtk/`), so I could not verify it from this repo alone. Settles it either way: if yes, #2 is lower severity (still worth fixing since it defeats `pump`'s own documented contract, but not a hang beyond a moment); if no, it's exactly the blocker described above.

## What this PR gets right

- The `Paned` pixel↔percent conversion (`containers.rb:286-300`, `apply_event` at `261-269`) correctly guards the zero/absent-window-axis case (`extent&.positive?`) rather than dividing by zero, and the round-trip (percent sent to the viewer, percent received back from a drag converted through the same `axis_extent`) is consistent both ways. `Session#commit` unconditionally re-materializes every window and all descendants each pass (`Container#materialize!`'s children loop, `widgets.rb:959` area), so a `Paned` created before its window's `set_default_size` correctly picks up the real extent on the very next commit — the boundary case flagged in the review brief does not reproduce as a bug.
- `ListBox`/`ListBoxRow` avoid the PR #1639 `TreeIter` bug's shape: `ListBoxRow#index` recomputes `parent.children.index(self)` fresh on every call against the live `@children` array rather than caching an identity-based key that could go stale after a mutation. `reorder`/`insert` mutate the same backing array `Container#remove`/`#add` use, so there's no parallel index structure to fall out of sync.
- `ProgressBar#pulse`/`.fraction` not diverging (aside from finding #4's `show_text` bug) matches real GTK's own ambiguity here — real `Gtk::ProgressBar` doesn't meaningfully update `fraction` from `pulse` either, so a script reading `.fraction` after `.pulse` gets the same kind of stale-but-not-wrong value it would against real GTK.
- The new JS `split`/`overlay` renderers (`lib/webui/assets/app.js`) correctly release pointer capture on `pointerup`, so a drag gesture doesn't leave a dangling document-level listener, and both are covered by the new `assets_spec.rb` checks.
- Full existing + new spec suite passes as claimed: `bundle exec rspec spec/lib/common/script_scope/gtk_slice5_spec.rb spec/lib/webui/assets_spec.rb` → 25 examples, 0 failures, confirmed in this worktree.

## Coverage notes

- Ran `ruby -c` on all four changed Ruby files — syntax OK.
- Ran the repo's own `rubocop` against the four changed Ruby files (see Nits #6).
- Ran `bundle exec rspec spec/lib/common/script_scope/gtk_slice5_spec.rb spec/lib/webui/assets_spec.rb` — 25 examples, 0 failures.
- Did not run the full `spec/lib/webui spec/lib/common/script_scope` suite the PR body cites (353 examples) — time budget went to building targeted reproductions for the `Dialog#run` concurrency claims instead, which is where this review's value is; no reason to doubt the broader suite given the two targeted files passed cleanly and the diff outside `Dialog`/`Paned`/session.rb is small and additive.
- Did not run `coderabbit` (not invoked in this session; time was spent on hands-on reproduction of the concurrency claims instead, which is the higher-value activity for this specific PR per its own stated risk area).
- Three scratch repro scripts written to `/tmp/claude-1000/-home-ryan/eaa68042-daf2-42b8-b84f-12e7cfa6a726/scratchpad/` (`repro_viewer_closed_spec.rb`, `repro_dialog_shutdown_spec.rb`, `repro_double_run_spec.rb`, plus `repro_paned_dup_spec.rb` for finding #5) exercise the actual shipped classes against the actual `Lich::WebUI::Service`/`Session` — not mocks of the code under test — and are safe to hand to the author to reproduce independently; none modify the worktree.

mrhoribu:
# PR Review: feat(all): slice five's long tail - Paned, Overlay, ProgressBar, ListBox, Dialog (elanthia-online/lich-5#1640)

**Reviewed:** 8 files, +738/-5 at `7aeb8541` (base `webui/06-images-and-map`) on 2026-09-16, one commit. Based on: full clone; no Ruby toolchain.
**Summary:** Five widget classes that were each degrading to an empty box, plus the `Stock` labels. `Dialog#run` is the one with real machinery in it and it is broadly right — the nested-pump model matches what `gtk_dialog_run` actually does — but the pump loop has no sleep and the response queue is never drained between runs.

## Minor

### 1. `Dialog#run`'s pump loop has no sleep once `:stop` is queued — `lib/common/script_scope/gtk/widgets.rb:2447`

```ruby
if @session.on_session_thread?
  @session.commit
  @session.pump(0.05) until @response || destroyed?
  @response || ResponseType::DELETE_EVENT
```

`Session#pump` returns `false` immediately when it finds `:stop`, after pushing it back:

```ruby
job = @queue.pop(timeout: timeout)
return false if job.nil?
if job == :stop
  @queue << :stop
  return false
end
```

So once `Session#shutdown` has enqueued `:stop`, `pop(timeout:)` returns instantly on every iteration and the `until` loop spins at 100% CPU with no yield. `shutdown` does not set `@destroyed` on the dialog widget (it goes through `close_window`, which touches the adapter and the session's own bookkeeping, not the widget), so neither exit condition ever becomes true.

In practice this terminates, because the session thread is created from the script's own thread and therefore belongs to the script's ThreadGroup, which Lich kills shortly after `at_exit` fires. So the window is short and the symptom is a CPU spike at script exit rather than a hang. That is why this is Minor rather than Major — but relying on an external thread kill to break a spin loop is not a property you want to depend on, and it will not hold if a session thread ever ends up outside the script's group. A `@session.pump(0.05)` that returned `false` should at least be followed by a short sleep, or `pump` should not consume-and-replace `:stop` at all.

### 2. `@responses` is never drained, so a second `run` can return the first one's answer — `lib/common/script_scope/gtk/widgets.rb:2443`

```ruby
def respond(response)
  @response = response
  @responses << response
  emit(:response, response)
  self
end

def run
  show unless @shown
  @response = nil            # resets the ivar, not the queue
  ...
  else
    @session.enqueue { @session.commit }
    @responses.pop
  end
end
```

`run` clears `@response` but leaves `@responses` alone. Every `respond` pushes unconditionally. So a viewer who clicks twice before the script's `run` returns leaves an entry behind, and the *next* `run` on the same dialog returns it immediately without ever showing anything. The on-session-thread branch has the mirror problem: it never pops at all, so `@responses` grows by one per button press for the dialog's lifetime.

`@responses.clear` at the top of `run` fixes both.

### 3. `run` exits on `@response` being truthy, so `nil` and `false` responses hang — `lib/common/script_scope/gtk/widgets.rb:2449`

`add_button(label, response)` stores whatever the script passed and `respond` assigns it to `@response`; the loop condition is `until @response || destroyed?`. A script calling `add_button('Cancel', nil)` or `add_button('No', false)` — both legal, since the PR's stated contract is that `run` returns *the very object the script gave `add_button`* — produces a button that can never end the loop. Use a sentinel (`@response = NO_RESPONSE` on entry, compare with `equal?`) rather than truthiness.

### 4. `ProgressBar#show_text` is ignored — `lib/common/script_scope/gtk/containers.rb:57`

```ruby
props[:label] = @text if @text && !@text.empty? && (@show_text || !@text.empty?)
```

The third clause is implied by the second, so the condition reduces to `@text && !@text.empty?` and `@show_text` never affects anything. A script calling `set_text('50%')` followed by `set_show_text(false)` still shows the text. Either honour the flag (`@show_text && @text && !@text.empty?`) or drop `@show_text` and log it as unsupported — carrying a setter that does nothing is the worse of the three options.

### 5. `Paned#add` silently replaces the second child when both slots are full — `lib/common/script_scope/gtk/containers.rb:196`

```ruby
def add(child)
  place(child, @slotted.value?('first') ? 'second' : 'first')
end
```

`place` removes whatever occupies the target slot. GTK's `gtk_paned_add` warns and does nothing when the paned is full. A script that adds three children loses the second without any trace — no `log_unsupported`, no `log_render_failure`. Given how carefully the rest of the shim reports what it cannot honour, this one should too.

## Nits

- `Paned#position=` hand-writes `changed!` followed by `@session.viewer_write(window_root, self, :position, position_percent)`. That is exactly the pair `Widget#viewer_push` was introduced in #1639 to stop anyone writing by hand — the comment there says the idiom "was hand-copied at ten sites and forgotten at five". This is the eleventh. `viewer_push(:position, position_percent) if position_percent` says the same thing and cannot drift.

- `ListBox#select_row` / `selected_row` / `unselect_all` maintain `@selected`, which nothing reads and which never reaches the contract. The class comment says selection is deliberately not modelled because "creaturebar's does not act on it" — in which case these three should log as unsupported rather than quietly maintaining state a script might read back and act on.

- `ListBoxRow#index` uses `parent.respond_to?(:children)`, which is always true given `Widget#respond_to_missing?` returns true for everything. It happens to work because `Container#children` is real, but the guard is not doing what it looks like it is doing — the same trap `Window#collect_scrollers` has a comment about.

## What this PR gets right

- **`Dialog#run`'s nested main loop is the correct model and the comment explains it properly.** `gtk_dialog_run` blocks the main thread *while still servicing events*, which is how the button it is waiting on can ever be pressed; pumping the session queue on-thread rather than blocking against itself is the only way to reproduce that. Detecting which thread you are on and picking between pumping and waiting is right.

- **Returning the exact response object the script passed to `add_button`** — so `creaturebar`'s `:ok` and `map`'s `Gtk::ResponseType::OK` both compare equal to what they handed in — is the kind of fidelity that makes "scripts run unmodified" true rather than nearly true.

- **`Gtk::Stock`** as a plain constants module, with a comment naming the exact failure it fixes (`const_missing` turned `Stock` into a widget class, so `Gtk::Stock::OK` raised `NameError` and killed map's room-list dialog). One of the clearest bug-to-fix explanations in the series.

- **Pixel-to-percent conversion for `Paned` against the window axis** is the honest answer to a real impedance mismatch, and `axis_extent` returning `nil` when there is no window (so the prop is simply absent) is better than guessing.

## Coverage notes

All eight changed files read in full, including `gtk/containers.rb` (+329) and the `widgets.rb` Dialog/Stock additions (+114). `app.js` (+51) reviewed for the `split` and `overlay` renderers. `gtk_slice5_spec.rb` read as intent — **no Ruby in this environment**, so "353 examples, 0 failures" is unverified, and the `Dialog#run` pump behaviour in particular is something only a running session would confirm.

Nisugi:
## `Dialog#run`'s hangs are fixed (at the tip, #1641)

The stack-level blocker from this PR. All three scenarios were reproduced by execution before the fix and after — `2faa9091`.

```
                                    before          after
viewer closes the tab               never returned  -4 (DELETE_EVENT)
browser process exits               -4              -4
two threads call run                second hung     both -4
destroy while waiting               hung            -4
script's own response               7               7   (no regression)
```

**Root cause.** `Dialog` overrode `#browser_exited` — the whole browser dying — but not `#viewer_closed`, which is what a viewer closing the *tab* arrives as (`session.rb:562` binds `:close` to it). The base implementation only emits `:delete_event`, so `run` sat on a `Queue` nobody would ever push to. A plain confirmation dialog with no `:delete_event` handler hung the script forever, which is the common case rather than an edge one.

Releasing waiters is now a single method called from all three teardown paths, and it pushes once per *waiting* thread rather than once: a single push wakes a single consumer, which is why the second concurrent `run` stayed parked. Six regression specs cover the paths nothing in the previous 413 examples exercised — including the existing "unblocks with DELETE\_EVENT when the viewer closes it", which closed the dialog with `#destroy` and so never touched the failing path. That naming gap is why the bug survived a test suite that looked like it covered it.

**`ProgressBar#show_text` — `6519646c`.** Also fixed. The guard was tautological (`@text && !@text.empty? && (@show_text || !@text.empty?)`), so `show_text` decided nothing and a bar whose text a script had switched off still showed it. Checked against real GTK 3.24.52: `bar.text = "x"` leaves `show-text` false and the text does not render. `calibrate_creaturebar.lic` drives this from a "Text" checkbox, so it was user-visible. The existing spec asserted the buggy behaviour and now sets `show_text` as a script must, with a second example for the withheld case.

### Still open from this PR

- **`Paned#add` on a full paned silently drops** — part of the second review's "degrade honestly is applied inconsistently" theme. One `log_unsupported` call each; worth doing as a batch across the six places rather than piecemeal here.
- **`Dialog#run` never drains `@responses`** — a second `run` can still return the first's answer. Narrower than the hangs and not reproduced against a real script yet.

Nisugi:
## Two more from this PR are fixed (at the tip, #1641)

**`Dialog#run` never drained `@responses` — `a1cbb6af`.** Reproduced: answer one run twice (a double click, or a `respond` racing a close), then start a second run.

```
before:  second run returned :ok immediately   STALE=true
after:   second run waits for its own answer   STALE=false
```

The leftover answer was popped and returned before the viewer had seen the dialog at all. Draining on entry is the whole fix — the queue is single-consumer per run by design; what it must not do is carry an answer across runs.

**`Paned#add` on an occupied pane — `f2d581ba`.** GTK refuses a second child; the shim evicts the first instead, which is survivable but was silent, so the script's earlier widget simply vanished. It now reports through `log_unsupported` naming the pane.

That went in alongside three other silent clamps from the second review's "degrade honestly is applied inconsistently" theme: a horizontal box past twelve children, a grid past twenty-four columns, and a margin past 512px. The clamps themselves are right — they are the contract's own limits — so what was missing was saying so. All four now report once per class and property, deduped like the shim's existing signals, with a spec asserting silence when the box *fits* (a reporter that cries wolf on every render is worse than none).

**`ProgressBar#show_text`** was fixed in the first round (`6519646c`).

Nisugi:
Superseded by the re-layered stack: this work now lives in webui/shim-data (L6b). See docs/webui-rebuild-plan.md on webui/core-corrections for the map from the old PRs to the new layers. The branch is left in place for reference.

PR #1641, opened 2026-09-16
feat(all): key events, map interaction, and real window presentation

Nisugi:
**Stacked on #1640.** Diff is against `webui/07-slice-five`. Nine commits — the last in the series.

Contract 2.13.0 → **2.15.0**. Mostly live-testing fixes, several of which were one bug wearing different clothes.

## Key events (2.14)

A script connects `key-press-event` on its window and reads `event.keyval` against `Gdk::Keyval::KEY_Left`. Nothing delivered it. A page root turns out to have **no per-cid binding channel** — its bindings are routed wholesale to the lifecycle validator and never reach `render.bindings`, which is the only place the runtime looks for a callback. So `key` is a page lifecycle event, which also makes it non-coalescable: two different keys pressed quickly would otherwise fold into one.

## Map interaction

Four symptoms, one root cause between them: **a viewer-scoped property whose overlay nothing ever clears.**

- The marker only centred **once**. `scroll_position` is viewer-scoped, and `ViewerStore` seeds such a property into the viewer's overlay exactly once — after that the viewer's stale copy wins forever, so every later walk was silently discarded.
- The **menu reopened on every step**, because `open` is viewer-scoped too and nothing cleared it on close. Harmless while the page rarely redrew; the centring fix made every room change re-render, which surfaced it.
- **Toggles never toggled** — same shape again, for `active`.
- A **click found no room**: the gesture now carries the scroller's live offset, so `(adjustment.value + pointer - offset) / scale` is exact at any scroll position.

Also: left-drag pans the map (the browser scroller does it; the script is told nothing), and the page no longer carries a second scrollbar from a `calc(100vh)` that ignored its own padding.

## Real window presentation

`keep_above` and `opacity` now reach the **actual OS window** on native Windows, via user32 — following the Fiddle/`EnumWindows` pattern `lib/common/frontend.rb` already uses. Nine scripts ask for `keep_above`; `map` offers an opacity menu, and a CSS fade cannot make a window translucent because what shows through is the browser's own background.

Two things this had to get right: `SetWindowPos`'s `hWndInsertAfter` **must** be a pointer, not a `long` — as a `long` on LLP64 the sentinel arrives truncated, and the call does nothing while reporting no error. And turning a property *off* is invisible, since `Window#presentation` omits false values, so the session resolves the absence into an explicit default.

Everything else keeps degrading honestly: `xdotool windowstate --add ABOVE` is a no-op under Wayland, so claiming success there would be worse than admitting the gap. **Borderless is refused** — Chromium draws its title bar inside the client area, measured at 8px of non-client region and no caption, so clearing `WS_CAPTION` removes a frame that was never there.

## Testing

`bundle exec rspec spec/lib/webui spec/lib/common/script_scope` → **413 examples, 0 failures, 0 pending**.

(The image specs had been silently skipping on a `GdkPixbuf` guard that nothing satisfied — `gdk_pixbuf2` is a separate gem from `gtk3`. Requiring it turned five pending examples back on, including the one covering this PR's marker-scaling fix.)


mrhoribu (changes requested):
# PR Review: feat(all): key events, map interaction, and real window presentation (elanthia-online/lich-5#1641)

**Reviewed:** 24 files, +2111/-58 at 493c5bc6 on 2026-09-16. Based on: full clone with all 9 commits (`origin/webui/07-slice-five..HEAD`), the base branch worktree (`wt-1640`, base commit `7aeb8541`) for before/after diffing, `gh pr checks`/`gh run view` CI logs, and a local Fiddle/Ruby 4.0.5 repl to verify the Win32 pointer claim empirically rather than by inspection alone.

**Summary:** This is a well-reasoned, unusually well-grounded PR — nearly every claim in the body checks out against the code. The "one root cause, four symptoms" narrative is real: all four viewer-scoped-property bugs (`scroll_position`, menu `open`, `active`, and the earlier-fixed shadowing pattern) resolve through one method, `Widget#viewer_push` (`lib/common/script_scope/gtk/widgets.rb:824`), not four one-off patches. The `SetWindowPos`/`hWndInsertAfter` pointer fix is correct — verified with a live Fiddle repl that `Fiddle::Pointer.new(-1)` produces the exact 64-bit all-ones bit pattern Win32's `HWND_TOPMOST` sentinel requires, which a 4-byte `long` on this LLP64 target would not. The click-to-room coordinate math is directionally correct and no longer double-counts scroll. The `gdk_pixbuf2` test-skip fix is genuinely committed (a `require` added to the spec file itself), not just observed locally. The one real, verifiable problem is process, not logic: the PR's own new code adds 5 new Rubocop offenses on top of an already-broken `Run Rubocop on Ruby` CI check, which is visibly failing on the PR right now.

## Major

- **CI's `Run Rubocop on Ruby` check is failing on this PR, and this PR's own new code contributes 5 of the offenses.** `gh pr checks 1641 --repo elanthia-online/lich-5` shows `Run Rubocop on Ruby  fail`. Pulling the failed run's log (`gh run view 35137866007 --log-failed`) shows `1271 files inspected, 37 offenses detected` with `Process completed with exit code 1`. Running `bundle exec rubocop` against the PR's base (`wt-1640` at `7aeb8541`, i.e. `origin/webui/07-slice-five`) gives `1266 files inspected, 32 offenses` — so the check was *already* red before this PR (inherited debt, not this PR's fault), but diffing the two offense sets by file shows this PR adds exactly 5 new ones, all traceable to its own new lines via `git blame`:
  - `lib/webui/contract.rb:609:11` — `Layout/FirstArgumentIndentation` (`[Correctable]`) on the `keyval: property(...)` block, added by this PR's own commit `521e6c22d`.
  - `lib/webui/window_presentation.rb:122:51` — `Lint/UnusedMethodArgument: borderless` on `def apply(hwnd, always_on_top:, opacity:, borderless: false)` — the whole file is new in this PR.
  - `spec/lib/common/script_scope/gtk_map_interaction_spec.rb:109:11`, `:124:11` — `Layout/FirstArgumentIndentation` (`[Correctable]`), and `:191:41` — `Layout/SpaceAroundMethodCallOperator` (`[Correctable]`) — the whole file is new in this PR.
  4 of the 5 are auto-correctable by `rubocop -A`, which CLAUDE.md documents as the pre-commit step for this repo ("rubocop -A before committing"). **Confirmed.** Settles it: run `bundle exec rubocop -A lib/webui/contract.rb lib/webui/window_presentation.rb spec/lib/common/script_scope/gtk_map_interaction_spec.rb` and commit the result; the `UnusedMethodArgument` one needs a manual `_borderless` rename or an inline disable since `borderless` is deliberately unread (see `window_presentation.rb:122`'s own comment — the argument is accepted and ignored on purpose, but Rubocop doesn't know that).

## Minor

- **The single-mechanism protection against a fifth instance of this bug is a comment, not a guard.** `lib/common/script_scope/gtk/widgets.rb:816-828` documents exactly the risk this PR fixes ("That pairing was hand-copied at ten sites and forgotten at five... One method, so it cannot be half-copied again") and `viewer_push` is genuinely the single place every fix in this PR routes through (`scroll_position` at `widgets.rb:1738`, `open` via `menus.rb:298,349`, `active` via `menus.rb:118,228`). That's real and good. But nothing *structural* ties a `scope: :viewer` contract property (`lib/webui/contract.rb`, ~29 properties declared this way) to a requirement that its writer call `viewer_push` — there's no lint rule, no test asserting the invariant, no base-class enforcement. A sixth viewer-scoped property added later by a contributor who writes `@foo = value; changed!` directly (as the five forgotten sites originally did) would silently reproduce this exact bug family with nothing catching it before live testing does. **Confirmed** (mechanism), **suspected** (no future contributor will get this wrong — there's no way to verify that in advance). Settles it: a cheap regression guard would be a spec that walks `Contract` schemas for every `scope: :viewer` property and asserts some structural invariant, or at minimum a comment on `ViewerStore`/`Contract.property` cross-referencing `viewer_push`.

- **`event.repeat` filtering silently drops OS-level key auto-repeat.** `lib/webui/assets/app.js:196`: `if (event.isComposing || event.repeat) return;` in the page-root `keydown` listener means a script connecting `key-press-event` and expecting the browser's native auto-repeat while a key is held (e.g. holding an arrow key to keep panning/scrolling) will only ever see the single initial press, never the repeats a real GTK window would deliver. This is a genuine behavioral gap from real GTK, not a bug in the coalescing sense the PR describes — the PR's own design intentionally treats a key as a single discrete lifecycle event. **Confirmed** the filtering exists and changes behavior from native GTK; **can't verify** whether any of the 9 scripts that use `keep_above` (or any other consumer) actually relies on held-key repeat, since consuming `.lic` scripts live outside this repo (dr-scripts/scripts). Worth a one-line question to the author rather than a blocking finding.

## Nits

- **Click coordinate math rounds twice before subtracting, which can introduce a ±1px discrepancy.** `lib/webui/assets/app.js:791-803` (`surfacePayload`): `absX = Math.round((event.clientX - rect.left) / s)`, `sx = Math.round(scroller.scrollLeft / s)` (via `clampGeometry`), then `payload.x = absX - sx`. Rounding each term independently before subtracting can differ from rounding the combined expression once by up to 1px in either direction (e.g. `absX` rounds up 0.5px while `sx` rounds down 0.5px at certain scroll/scale combinations). Given map rooms are almost certainly many pixels wide, this is very unlikely to land a click on an adjacent room in practice, but it's exactly the kind of fractional-pixel seam the PR's own review focus flags, so noting it rather than silently passing over it. Not a blocker.

- **A dangling paragraph in the doc update reads oddly.** `docs/webui-gtk-shim-plan.md:651-655`: the sentence "`always_on_top` and `opacity` stay degraded on every other host, deliberately: ..." sits as an unindented, bullet-less paragraph between the "Hide Scrollbars: **Done**" bullet and the "Changing Scale... **Fixed**" bullet, with no bullet marker of its own. Purely cosmetic — an internal planning doc, not user-facing — but worth a formatting pass next time the file is touched.

- **`window_presentation.rb`'s `EnumWindows` declares `lParam` as `long` where Win32 defines it as pointer-sized `LPARAM`.** `lib/webui/window_presentation.rb:230`: `extern 'int EnumWindows(void*, long)'`. On Win64, `LPARAM` is `LONG_PTR` (8 bytes), so this is technically the same *class* of mismatch the PR fixed for `hWndInsertAfter`. It's inert here only because every call site passes the literal `0` (`win32.EnumWindows(callback, 0)`, `window_presentation.rb:102`), and 0 zero-extends and sign-extends identically — so there's no observable bug today, but it's a footgun for whoever next threads a real value through that second argument. **Confirmed** the declaration mismatch exists; **confirmed** it's inert today because the only call site passes 0.

## Open questions

- **`lib/common/frontend.rb`'s own `EnumWindows`/`refocus_windows` (lines ~774-880) has what looks like the identical pointer-truncation bug class this PR fixed, unfixed.** `frontend.rb:860,868`: `hwnd_buf[0, Fiddle::SIZEOF_VOIDP] = [hwnd].pack('L!')` and `hwnd_buf[0, Fiddle::SIZEOF_VOIDP].unpack1('L!')` — `'L!'` is Ruby's *native* `long` pack directive, which on Windows (LLP64) is 4 bytes, packed into an 8-byte (`Fiddle::SIZEOF_VOIDP`) buffer. That looks like it could truncate a real 64-bit HWND the same way the PR's own commit message describes ("arrives... not a sentinel but an invalid handle... does nothing while reporting no error"). **This file is untouched by this PR** (`git log -1 -- lib/common/frontend.rb` shows the last change is an unrelated commit, `66e25811`, well before this PR's base) — it predates the stack and is out of scope for this diff's review, but since the review brief specifically asked to check sibling Win32 call sites for the same bug class: this one looks worth a follow-up issue. Settles it: exercise `Frontend.refocus_windows` on a real multi-monitor / high-handle-value Windows box and see if it silently fails to refocus, the same way the PR's own bug silently failed to raise the window.
- **Ordering/latency of `scrolled` events relative to a subsequent click after a drag-pan** — `pointermove` in `lib/webui/assets/app.js:500-509` writes `scroller.scrollLeft`/`scrollTop` directly, which triggers native `scroll` events that asynchronously push `scrolled` back to the Ruby side to update `@hadjustment`/`@vadjustment`. A click immediately following the *same* drag gesture is explicitly suppressed (`dataset.dragged`), so that race is closed. It's not fully verifiable from static reading whether a *separate*, fast follow-up click after a drag could ever race ahead of its own `scrolled` update over the wire (WebSocket ordering is FIFO per-connection, so in-order delivery is likely guaranteed, but network/dispatcher scheduling wasn't traced end-to-end). Given human reaction times versus websocket round-trip, this is very unlikely to matter — flagging only because it's adjacent to the exact math this PR reworked.

## What this PR gets right

- **The "one root cause, four symptoms" claim is true, not narrative.** `viewer_push` (`widgets.rb:824`) really is the single mechanism every one of the four map-interaction fixes routes through, and the surrounding comment explicitly documents the failure mode it closes (five call sites that previously hand-copied "mutate + push" and forgot the push half). This is exactly the kind of structural fix the review brief was skeptical of getting for free.
- **The `SetWindowPos`/`hWndInsertAfter` fix is correct, verified empirically.** `Fiddle::Pointer.new(-1)` on Ruby 4.0.5 was confirmed live to produce the 64-bit all-ones bit pattern (`0xFFFFFFFFFFFFFFFF`) that Win32's `HWND_TOPMOST` sentinel requires; the `extern` declaration (`void*` for the second argument) and the call site (`Fiddle::Pointer.new(...)`) agree, and the sibling calls in the same file (`GetWindowLongW`/`SetWindowLongW`/`SetLayeredWindowAttributes`) don't have the analogous bug because none of them carry a pointer-typed sentinel through a narrower slot — they're all genuinely 32-bit Win32 values.
- **`Window#presentation`'s absence-vs-default resolution is correct, not just plausible.** Traced `set_keep_above(false)` and "never called" through `presentation` (`widgets.rb:2173`) into `apply_window_presentation` (`session.rb:632`): both states produce an omitted `:always_on_top` key, which `requested[:always_on_top] ? true : false` resolves to `false` in both cases — the semantically correct outcome, since GTK itself treats "never set" and "explicitly false" identically. No bug here despite it being flagged as worth scrutinizing.
- **The non-coalescable key-event path doesn't reorder or delay relative to ordinary events.** `Dispatcher#enqueue` (`lib/webui/dispatcher.rb`) puts both lifecycle and ordinary component events through the *same* per-owner FIFO queue and thread; `coalescable: false` only disables the fold-into-last-event optimization in `coalesce_last!`, it doesn't route key events through a separate or slower path. Verified by reading `run_owner`'s single-queue loop.
- **The `gdk_pixbuf2` test-skip fix is a real, committed fix, not a local-only observation.** `spec/lib/common/script_scope/gtk_images_spec.rb` now `require`s `gdk_pixbuf2` and `cairo` directly (guarded by `rescue LoadError`), and `Gemfile.lock` confirms `gdk_pixbuf2 (= 4.3.6)` is a hard-pinned transitive dependency of the `gtk3` gem already in `DEPENDENCIES` — so any CI environment with `gtk3` installed will now actually run these specs instead of silently skipping them.
- **The scrollbar/`calc(100vh)` fix removes the magic number rather than relocating it.** The old `.webui-page.bare > .webui-scroll { height: calc(100vh - 12px); }` is replaced with a flex layout (`app.css`) that has no hardcoded padding constant to drift out of sync at all — better than the DRY concern the review brief anticipated.
- **The doc file (`docs/webui-gtk-shim-plan.md`) is kept in sync with the code across the PR's own commits.** An early commit in this same PR (`dd84bc64`) recorded three items as deferred/unresolved; later commits in the same PR fix two of them and correctly retroactively strike them through as "Done"/"Fixed" rather than leaving a stale doc contradicting the shipped code.
- **Full local test run matches the PR's own claim.** `bundle exec rspec spec/lib/webui spec/lib/common/script_scope` → `413 examples, 0 failures` at HEAD, reproducing the PR body's stated result exactly.

## Coverage notes

- `coderabbit review --agent` was not run for this pass (time budget went to the deep verification the review brief specifically requested — empirical Fiddle testing, CI log diffing, and tracing the viewer-overlay mechanism end-to-end — rather than a broad automated sweep); no CodeRabbit-sourced leads are included above. `gh pr checks` shows CodeRabbit's own automated check as `pass` with "Review skipped: automatic reviews are disabled" for this PR, so a CodeRabbit pass would not have added anything here regardless.
- Ran the actual spec suite (`bundle exec rspec spec/lib/webui spec/lib/common/script_scope`): 413 examples, 0 failures, reproducing the PR's claim.
- Ran `bundle exec rubocop` locally on both HEAD and the base branch worktree to isolate exactly which offenses this PR adds versus inherits; cross-checked against the actual failing CI run's log rather than relying on a local rubocop version potentially differing from CI's.
- Did not have a Windows box to exercise `WindowPresentation` end-to-end against a real `user32.dll`; the pointer-representation claim was instead verified at the Ruby/Fiddle semantics level (confirmed `Fiddle::Pointer.new(-1)` yields the correct 64-bit bit pattern), which is the part actually under this PR's control — the OS-side behavior of `SetWindowPos` given that bit pattern is documented Win32 behavior, not something this review can independently exercise.
- `lib/common/frontend.rb` (the sibling Win32 code the review brief asked to check) is untouched by this PR and was reviewed only far enough to identify it as a plausible but out-of-scope open question, not to file it as a finding against this diff.

mrhoribu:
# PR Review: feat(all): key events, map interaction, and real window presentation (elanthia-online/lich-5#1641)

**Reviewed:** 24 files, +2111/-58 at `493c5bc6` (base `webui/07-slice-five`) on 2026-09-16, nine commits. Based on: full clone; no Ruby toolchain, and **no Windows host** — the user32 path could not be exercised at all.
**Summary:** The strongest reasoning in the series is in this PR. The four map symptoms really do reduce to one cause (a viewer-scoped property whose overlay nothing clears), the `SetWindowPos` pointer-versus-long analysis is correct, and refusing to fake `borderless` rather than reporting a no-op as success is the right instinct. My notes are about cost-per-event and about a Windows-only feature that no CI can check.

## Minor

### 1. Every key press triggers a full re-materialization of every window in the session — `lib/common/script_scope/gtk/session.rb:470`

```ruby
if window.key_wanted?
  adapter.bind(handle, :key, proc { |context| enqueue { window.receive_key(context) } })
end
```

The dispatcher callback correctly does nothing but enqueue, so the owner's dispatcher thread never blocks — that part is well designed. But `Session#run_loop` runs `commit` after every job, and `Session#commit` calls `window.materialize!(adapter)` for every window, which walks the whole widget tree, recomputes `common_props`/`node_props` for every node, diffs them, and ends in `flush!` (which itself does an O(nodes) `handle_for` scan — see the #1634 review).

So holding an arrow key in `;map` — which is exactly the intended use, since the PR adds keys so a script can read `Gdk::Keyval::KEY_Left` — produces one full tree walk and one full render per auto-repeat, at whatever rate the OS repeats. The `Queue` is unbounded, so if a repeat outpaces a commit the backlog grows rather than dropping frames.

`Session#pump` and `run_loop` both commit unconditionally. A cheap improvement: commit only when the adapter is actually dirty, or coalesce — drain the queue and commit once at the end rather than once per job. The latter is a few lines and would help every event type, not just keys.

### 2. Key events are non-coalescable by design, which removes the queue's only pressure valve — `lib/webui/contract.rb:599`, `lib/webui/dispatcher.rb:137`

The reasoning for `lifecycle: true` (and therefore `coalescable: false`) is correct and stated in the contract comment: `coalesce_last!` matches on `[page_id, cid, event]` and ignores the payload, so coalescing would fold two *different* keys into one.

The consequence is that `enforce_bounds!` can only evict coalescable events, and key events are not. If the owner dispatcher ever backs up past `VIEWER_LIMIT` (256), `Dispatcher::OverflowError` is raised, and `Runtime#event` handles it by closing the viewer's attachment *and the connection*:

```ruby
rescue Dispatcher::OverflowError
  @viewers.close(connection_id: connection.viewer_id, address: message[:page])
  connection.close
  raise Protocol::Refusal.new(:overflow, 'viewer event queue overflow')
```

I do not think this is reachable today — the key callback returns immediately and nothing else blocks that dispatcher — so I am raising it as a bound to be aware of rather than a bug. **Settles it:** is there any path on which an owner's dispatcher callback can block for longer than 256 key repeats? If a future slice makes a lifecycle callback do real work, this becomes "holding a key closes the window" and the failure will look nothing like its cause.

### 3. `WindowPresentation.find_window` allocates a Fiddle closure per poll — `lib/webui/window_presentation.rb:88`

`discover` polls every 100ms for up to 15 seconds, and each `find_window` builds a fresh `Fiddle::Closure::BlockCaller`. That is up to 150 closures per browser window, each of which allocates executable memory through libffi and is reclaimed only by GC. Hoisting the closure and the `matches` array out of the poll loop (rebuilding `matches` per call, reusing the closure) is straightforward.

### 4. `presentation_support` allocates and freezes a hash on every render — `lib/webui/runtime.rb:41`

```ruby
def presentation_support(_page = nil)
  host = WindowPresentation.support
  return PRESENTATION_SUPPORT if host.empty?
  PRESENTATION_SUPPORT.merge(host).freeze
end
```

`record_presentation_degradations` calls this from `validated_render`, which runs on every attach and every refresh. `WindowPresentation.support` is constant for the life of the process (it depends only on `native_windows_runtime?` and whether `dlload` succeeded), so this can be memoized once.

### 5. `find_window` returns nil unless there is *exactly* one match, and the reason it holds is in a different file

The comment justifies the strictness by noting that the shim always spawns with a private profile directory, so the honest case is one window. That is true — but the fact it depends on is `Session#open_browser` always passing `on_exit`, which makes `BrowserLauncher.open` create a `Dir.mktmpdir`. Three files away, in a different PR. If anyone ever makes profile directories shared to reduce the process count (which is a reasonable thing to want — see the #1634 review), `find_window` silently stops working and reports nothing, because returning nil is its success-shaped failure. A `log` on the `matches.length > 1` branch would make that discoverable.

## Nits

- `apply` takes `borderless:` and ignores it. The comment explains exactly why, at length and convincingly. But an ignored keyword argument in a method whose whole job is applying state is the sort of thing a future reader will "fix". Either drop the parameter and let `Session#apply_window_presentation` not pass it, or keep it and add `# rubocop:disable Lint/UnusedMethodArgument` so the intent survives a cleanup pass.

- `owning_pid` passes a Ruby `String` from `[0].pack('L')` as a `void*` out-parameter. This works — Fiddle hands the string's buffer through and the string is not frozen despite the file's `frozen_string_literal` magic comment (which only freezes literals) — but it is subtle enough to deserve the one-line comment the `'L'`-vs-handle choice already got.

- The `INLINE_IMAGE` regex (`runtime.rb:431`) is anchored and character-class-bounded, so it is linear and safe. Worth noting that it accepts `={0,2}` padding without checking that the length is a multiple of four; a malformed base64 string passes validation and fails in the browser instead. Harmless, since the only producer is `PixbufSources.data_uri`.

## Open questions

- The PR states `xdotool windowstate --add ABOVE` is a no-op under Wayland and therefore Linux/macOS keep degrading honestly. Agreed on the reasoning. But `Gtk::Window#set_keep_above` is used by nine scripts per the PR body, and on Linux those now get a recorded degradation the player may never see. **Settles it:** does anything surface `Page#degradations` to the player, or is it only readable from a script? If it is only readable from a script, nine scripts silently lose a feature on the platform where most testing happens.

- `MAX_DATA_URI` is 8192 bytes and the marker-scaling fix in the last commit makes the marker's size depend on zoom. At what zoom does map's marker stop fitting, and what does the player see when it does? `PixbufSources.data_uri` returns nil and `Image#node_props` falls back to `{ src: '', alt: 'no image' }`, so the marker just vanishes. **Settles it:** one measurement of the marker's encoded size at maximum zoom.

## What this PR gets right

- **The `SetWindowPos` analysis is correct and is the kind of thing that costs days.** `HWND_TOPMOST` is `(HWND)-1`, which as a 64-bit pointer is all ones; declared as `long` on LLP64 it arrives as `0x00000000FFFFFFFF`, an invalid handle, and the call does nothing while returning no error. Getting there from "keep_above doesn't work" is real debugging, and writing it into the code so nobody re-derives it is the right thing to do with the answer.

- **"Turning a property *off* is invisible, so the session resolves the absence into an explicit default."** `Window#presentation` omits false values and returns nil once nothing is set, so a toggle could set but never unset. Spotting that the *absence* carries meaning, and deciding where to give it one, is a genuinely subtle fix.

- **Four symptoms, one cause.** The marker centring once, the menu reopening on every step, toggles not toggling, and the click finding no room — all traced to a viewer-scoped property whose overlay nothing ever clears. That is the second time in this series the same shape has been found (see #1639's finding 2), which suggests the underlying `ViewerStore` seeding rule deserves a design note of its own rather than a third round of spot fixes.

- **Refusing borderless rather than faking it.** Measuring the actual non-client region (8px of resize border, no caption) and concluding that clearing `WS_CAPTION` removes a frame that was never there — then *not shipping* the no-op — is the right call, and the opposite of what most people do under deadline.

- **Requiring `gdk_pixbuf2` so five silently-pending image specs actually run.** Finding that a guard nothing satisfied had been quietly skipping tests, including the one covering this PR's own fix, is the sort of thing that never gets noticed.

## Coverage notes

Read in full: `lib/webui/window_presentation.rb`, the `runtime.rb`/`contract.rb`/`validator.rb`/`viewer_store.rb` diffs, the `widgets.rb` key-event and presentation additions, `gtk/images.rb`'s scaling changes. `app.js` (+263) reviewed around the key listener, the composite gesture offset, and the scroller changes. **This environment has no Ruby and no Windows**, so the entire user32 path — `WinPresentation`'s `dlload`/`extern` declarations, `find_window`, `set_always_on_top`, `set_opacity` — was reviewed by reading against the Win32 API contract only. `window_presentation_spec.rb` (+205) drives it through the `win32` test seam, which is the right design, but a seam cannot catch a wrong `extern` signature. Somebody with a Windows box should confirm keep-above and opacity actually take effect before this merges.

Nisugi:
## Review fixes pushed

Both combined reviews of this stack have landed fixes here at the tip rather than in the PR that introduced each bug. They span slices — the `set_*` fix alone touches files from slices 2, 3, 5, 6 and 8 — so splitting them per-origin would have meant force-pushing six already-reviewed branches. Each fix is commented on its originating PR too.

Every one was reproduced by execution before the fix and re-run after. **445 examples, 0 failures**; rubocop clean across all 101 Ruby files this branch touches.

| Commit | Fix | Severity |
| --- | --- | --- |
| `2faa9091` | `TreeIter#next!` corrupted the store via `append`/`prepend`/`insert`/`each`; `Dialog#run` hung on tab close and on concurrent `run` | **Blocker** ×2 |
| `1e31db0f` | 829 of 1109 `set_*` mutators returned the assigned value, not the widget | **Major** |
| `62ea172b` | `disabled` never reached `select`/`table`; WebSocket reconnect could close the live socket; `page_refresh_lock` leak; `hexpand=`/`vexpand=` missing `changed!` | **Major** ×4 |
| `f22205f4` | `signal_connect` ids collided after a disconnect; `GLib::Idle.add` spun the session thread and leaked a source per pass | **Major** ×2 |
| `6519646c` | `ProgressBar#show_text` was tautological and decided nothing | Minor |
| `e2bbb5a1` | `const_missing` sent acronym-led class names (`UIManager`) to the enum fallback, so `.new` raised on Symbol | Minor |
| `a24c414d` | `EnumWindows` declared its pointer-sized LPARAM as `long` | Minor |
| `35b45529` | Rubocop backlog cleared | — |

### The two blockers, reproduced

```
# TreeIter, before:  after: ["y", "y", "z"]   CORRUPTED=true
# TreeIter, after:   after: ["x", "y", "z"]   CORRUPTED=false

# Dialog#run viewer_closed (tab closed):  before: never returned
#                                          after: -4 (DELETE_EVENT)
```

`dup_row` already existed and already shares the row's values array, so writes still reach the model; it simply was not applied at the four entry points a script actually uses. For `Dialog`, releasing waiters is now one method called from all three teardown paths, pushing once per waiting thread — which also closes the concurrent-`run` hang, since one push wakes one consumer.

### The `Vars` question, answered

The second review asked for one grep of the script corpus. Done, across all 230 scripts: every `Vars`/`Settings`/`CharSettings` assignment that constructs an object builds a core type (`Hash.new`, `Array.new`), and **no script stores an instance of a class it defined itself**. So the marshal-namespace break is real at the mechanism level (confirmed: `ArgumentError: undefined class/module Lich::Common::ScriptScope::Foo`) but not reachable today. Recorded in the plan doc as a known limitation with the fix named for when it becomes real.

### Not fixed, deliberately

- **`#1637`'s stale-generation retry loop** — already fixed in `648a0d40`, which added the `retriedEvents` guard and clears `pendingEvent` before replay. The review appears to have read an earlier state.
- **`lib/api/webui.rb`** — dead as reported (zero references), but it reads as an intended public API for script authors rather than an accident. Deleting it is the author's call, so I have left it.
- **Theme 1 (commit-path performance)** and the remaining Majors from the second review — `handle_for`'s O(nodes) scan, dirty-gated commits, `webui_data_uri` memoisation, browser-process-per-window, socket write timeouts. Real, but each is a design change rather than a bug fix, and they deserve their own PR rather than riding in on a review-response commit.

One correction to the second review's coverage note: it flagged that `WinPresentation` could not be verified without a Windows host. It has been — `EnumWindows` enumerates 409 windows through the corrected signature on a live x64-mingw-ucrt box, and keep-above and opacity were confirmed visually earlier in development.

Nisugi:
## Second round of review fixes

Five more commits, closing the remaining items from both reviews that are bug fixes rather than design changes. **457 examples, 0 failures** across three seeds; **102 files rubocop-clean**.

| Commit | Fix |
| --- | --- |
| `66f198df` | `Adapter#handle_for` scanned every node with a deep Struct `==` |
| `f2d581ba` | Four silent clamps now report what they could not carry |
| `5d88ab67` | A doubled underscore renders as a literal one |
| `a1cbb6af` | `Dialog#run` answered a second run with the first's leftover |
| `180b6a92` | The stubbed-widget notice guarded on the wrong method |

### `handle_for` — measured, not assumed

`Hash#key(node)` scans *values*, and `Node` is a `Struct`, so each candidate is a memberwise `==` recursing into `props` and `children`. `flush!` calls it twice per dirty root, on every commit, and the shim commits on every property write.

```
401-node tree, 200 calls:   32.4ms  (0.162ms each)
same through a reverse map:  ~0ms
```

The map is written where `@nodes` is written and deleted where it is deleted, including the cascade through `destroy_node!`. Three specs assert the two stay in step — a reverse map that drifts makes `handle_for` lie, which is worse than slow.

### One I tried and backed out

**Dirty-gated commits**, the other half of the performance theme. A session-level dirty flag set by `changed!` skipped commits that had real work to do — **44 failures** — because `changed!` marks the widget and not its ancestors, and structural mutations (`add`, `pack_start`, `show_all`) do not route through it at all. It needs every mutation path to signal, which is wider than a review response should carry. The measurement is recorded for whoever picks it up: **1.3ms per no-op job on a 200-row window**, and it is the same root cause as "every key press re-materializes every window" (#1641's own Minor).

### One where the review was wrong, and I was wrong first

`mnemonic_free` was reported as mangling legitimate labels. I wrote a fix that preserved underscores past the first — then checked the real gem:

```
gtk3 3.24.52, Gtk::Label#set_text_with_mnemonic
  "E_xit"           -> "Exit"
  "snake_case_name" -> "snakecasename"
  "lootsack_2"      -> "lootsack2"
  "a__b"            -> "a_b"
```

GTK mangles them identically, so the blanket strip was faithful and my fix would have made the shim diverge from the toolkit it emulates. Backed out. The one real gap is the last line — a doubled underscore is a literal `_` and the old `gsub` dropped it — which is what shipped, with the other cases pinned as specs so nobody re-derives this.

### Deliberately not done

- **`lib/api/webui.rb`** — dead as reported, but it reads as an intended public API for script authors rather than an accident. Deleting it is the author's call.
- **`webui_data_uri` memoisation**, **`install_pixbuf_tracking!` scoping**, **browser-process-per-window**, **socket write timeouts**, **`Runtime#shutdown`'s unbounded join** — all real, all design decisions rather than defects. Each deserves its own PR.
- **`Widget#allocation` fabricating 640×480** — the existing comment argues the fabrication is deliberate (nil turns a script's arithmetic into a `TypeError` several frames away). Left alone rather than second-guessed.

Nisugi:
## Two more, closing the last of the review items

`f67a2c17..29f28fd6`. **458 examples, 0 failures**; 99 files rubocop-clean.

**`webui_data_uri` memoisation — `3b1179d0`.** The memo is gone rather than made clever. `@webui_data_uri ||=` served a pixbuf's first frame forever once a script redrew it in place, and `fill!` plus a writable `#pixels` make that reachable. Not reachable in the corpus today — map.lic builds a fresh Pixbuf from a new Cairo surface per draw rather than mutating one, which is the question the review left open.

I tried keying the memo on a fingerprint of the pixel content first, then measured it:

```
200x200   encode 1.226ms   pixel-hash 1.975ms   (1.6x WORSE)
 24x24    encode 0.042ms   pixel-hash 0.031ms
 12x12    encode 0.033ms   pixel-hash 0.010ms
```

The fingerprint has to read `#pixels`, and fetching that string is most of what the encode costs — so the guard costs more than the thing it guards at any size where a guard would matter. At the size markers actually are (map draws an X of a dozen pixels) the encode is 0.033ms and not worth guarding at all. If a large inline image ever lands on a hot path, the fix is a memo the *mutator* invalidates, not one that re-derives the content to check it.

**`lib/api/webui.rb` deleted — `c139cf10`.** Fifty lines, ten methods, every one a single-line delegation to `Lich::WebUI`, with no require and no caller in `lib`, `spec`, or the 230-script corpus. Its sibling settles that it was never wired up rather than merely unused: `lich.rbw:85` explicitly requires `api/active_sessions.rb` and nothing requires this one — `lib/api` is an explicit-require directory, not a glob, so the file has never been loaded. Verified `lib/webui` still loads standalone with `Lich::API` correctly absent.

### One pre-existing thing worth a follow-up issue

Running the whole `spec/lib/common` tree in one process fails to load `gtk_images_spec.rb`:

```
TypeError: GLib::Timeout is not a module (Class)
  glib2-4.3.6/lib/glib2.rb:117
```

The shim defines `GLib::Timeout` as a **class** and the real `glib2` expects a **module**, so the two cannot coexist once anything pulls in `gdk_pixbuf2`. It reproduces at the previous tip, so it is not from these commits, and the targeted gate (`spec/lib/webui spec/lib/common/script_scope`) is unaffected — but it means the image specs are silently unrunnable in a full-tree run. Worth fixing as `module Timeout` with the class methods on it, separately from this stack.

That closes every review item except the ones called out as design changes rather than defects: dirty-gated commits (attempted, backed out, measured), `install_pixbuf_tracking!` scoping, browser-process-per-window, socket write timeouts, and `Runtime#shutdown`'s unbounded join.

Nisugi:
## Third review: F1, F2, F4, F6, F7, F9, F11, F12, F13 fixed

`29f28fd6..52a62791`. **477 examples, 0 failures**; 101 files rubocop-clean. Every fix was reproduced by execution before and after, and this review's own probes were the starting point rather than its prose.

| Finding | Sev | Commit | Reproduced |
| --- | --- | --- | --- |
| F1 retry correlation + unlimited retries | P1 | `140c4681` | refusing A sent `A, B, B`; one event sent 5× |
| F2 selection iterators alias the model | P1 | `8669d39b` | walk `[alpha, beta, beta…]`, store `[beta, beta, gamma]` |
| F4 shutdown abandons dialog waiters | P1 | `52c677a9` | both caller forms blocked, `destroyed?` false |
| F6 refused viewer write evicts the viewer | P2 | `2f6e510a` | viewer list emptied while attachment live |
| F7 password draft lost on rerender | P2 | `3cd1c629` | any unrelated render cleared the field |
| F9 pixbuf source map retains every bitmap | P2 | `f39ce2aa` | WeakRef alive after 3 GCs |
| F11 second pane gets the `first` slot | P2 | `f39ce2aa` | `add2` alone → `slot: "first"` |
| F12 attach races refresh | P2 | `92fac8b7` | generations delivered `[2, 1]` |
| F13 duplicate attach on every broadcast | P2 | `92fac8b7` | refusal per modal open |

### F9 is my regression, and thank you for catching it

`PixbufSources` was keyed on `pixbuf.object_id` — no reference held, but an integer key left behind after collection. Rubocop reads that as a hand-rolled `compare_by_identity`, and I took the advice in the lint-cleanup commit `35b45529`. That turned every key into a **strong reference**, so the module-global map owned every bitmap the tracking patch ever saw, for the life of the process, with no production caller for `forget_all`. An integer leak traded for native image buffers — and, as the review says, hidden in a style-only commit. It is an `ObjectSpace::WeakKeyMap` now, which is what both versions were reaching for.

### F1, in detail

Outstanding events are a Map keyed on page + cid + event, so a refusal is matched to the event it names. Each record carries the payload **and the submission it was sent with** — a replay used to rebuild the submission from whatever the controls held by then — plus an attempt count that travels with the retry instead of being released just before it. A render for a page clears everything outstanding against it, so the map cannot grow. Verified by executing the logic: refusing A replays A, B stays sent exactly once, and repeated refusals drop after the first retry.

### Two where I took the narrow half

- **F6**: the rescue is fixed — a `SchemaViolationError` is reported and the viewer kept; only real "not attached" errors forget it. The underlying **ordering** (a structural property and a viewer value changed in one job commit separately, so the push validates against the old definition) is a sequencing change in `commit`, and belongs with the dirty-gating work.
- **F4**: fixed for both caller forms through one `session_terminated` hook. The nested form needed no separate change — `run` loops `until @response || destroyed?`, and marking the dialog destroyed is what lets it leave.

### Still open, and why

- **F3 (password value transport)** and **F8 (table row_activate / cell editors)** are missing *features* rather than broken code: F3 needs submission scopes registered through the shim adapter plus a sensitive-value handoff across `dispatch_proc`'s async hop; F8 needs `row_activate` and contract cell editors implemented client-side. Both deserve their own PR with the browser tests you describe, not a squeeze into a review response. Worth noting for F8 that `ewaggle`, `repository` and `jinx` connect `row-activated`, so it has real corpus reach.
- **F5 (launch after close during auth)** is in `webui_launcher.rb`, which is **not wired into startup** — nothing references it yet. Real, and worth fixing before that code goes live, but it cannot affect a user today.
- **F10 (TreeStore hierarchy)** — agreed that a bare `ListStore` subclass should not claim TreeStore support. Implementing hierarchy-aware paths, traversal and removal is a feature.
- **Theme 1 (commit path)**: `handle_for` is done (`66f198df`, 32.4ms → ~0ms at 401 nodes). Dirty-gated commits I attempted and backed out — a session-level flag set by `changed!` skipped commits that had real work, 44 failures, because `changed!` marks the widget and not its ancestors and structural mutations do not route through it. Measured at 1.3ms per no-op job on a 200-row window.

### On the test-gap point

Taken. The retry spec I replaced asserted the old source strings *in order* — it could not have caught either F1 flaw, and it is exactly the shape you describe. Where a string assertion is all a change warrants I have kept one, but F1, F2, F4, F6, F9 and F11 are now asserted against executed behaviour: real stores, a real session shutdown, a real `WeakKeyMap`, real contract slots. A browser suite is the right next step and I have not built one.

One correction: the `SessionLauncher` `home_dir` failure you saw also reproduces at `main@ed662ca7`, which matches your finding — it is one of ~55 known environment failures on this Windows box, not a stack regression.

Nisugi:
Superseded by the re-layered stack: this work now lives in webui/core-corrections (L1) and webui/contract-primitives (L3). See docs/webui-rebuild-plan.md on webui/core-corrections for the map from the old PRs to the new layers. The branch is left in place for reference.

PR #1643, opened 2026-09-16
The client renders everything the contract declares

Nisugi:
Stacked on #1641.

A declared property that renders nothing is worse than an absent one: an author reads the schema and builds against it. These three commits close the gap between what the contract promises and what the browser client does.

## `tabs.vertical` was gated on a hardcoded cid

The contract has declared it all along and `app.css` has always implemented the layout. The client applied it like this:

```js
if (component.cid.includes("saved-account-tabs")) tabs.classList.add("account-tabs-left");
```

So the property rendered nothing for anyone but the accounts page, and the class was named after that page rather than the behaviour. It is keyed on `props.vertical` now, with the cid check kept as a fallback because the accounts page relies on the old behaviour without setting the property. `.account-tabs-left` becomes `.webui-tabs.vertical` across all six selectors including the max-width query.

## An audit of all 31 types and 420 properties

`tabs.vertical` was not alone.

**Three node types had no renderer at all.** A page using one got a visible `Renderer not implemented` box:

| type | now |
| --- | --- |
| `markdown` | rich_text through the same markup pipeline `text` uses |
| `log` | lines, honouring `max_lines` retention and `follow` |
| `slider` | range input over min/max/step, emitting `change` |

The GTK shim never emits those three, which is why nothing noticed. A client written against the contract finds them immediately.

**Five properties were declared and never read:**

- **`table.sortable` / `table.sort`** — the validator *enforces* both. It refuses a sort on a column that is not sortable, and refuses the `sort_change` event unless the table is sortable, while the header was plain text that could not be clicked. Sortable columns are real controls now, with `aria-sort`, keyboard activation and a direction indicator.
- **`table.max_height`** — a table grew to full height regardless. The wrapper already had `overflow: auto` and simply had no height to overflow against.
- **`scroll.scroll_to`** — names a cid to bring into view. The validator checked it named a declared region; nothing ever scrolled.
- **`columns.compact`** — packs a row rather than spacing it at the page gap.
- **`group.collapsible`** — becomes the same `<details>`/`<summary>` an `expander` uses, so the two agree instead of one silently doing nothing.

Short table cells also stop wrapping: `Built-in` breaking into `Built-` / `in` doubled the height of every row. Long values ellipsize instead.

## No contract version bump

Every one of these was already declared, defaulted and documented. This is the client catching up to what was promised.

The JSON export commit is unchanged in substance -- it only loses a few non-ASCII characters that fail the repo's own cop.

**224 examples, 0 failures** in `spec/lib/webui`; rubocop clean across all 23 files.


Nisugi:
Superseded by the re-layered stack: this work now lives in webui/contract-primitives (L3). See docs/webui-rebuild-plan.md on webui/core-corrections for the map from the old PRs to the new layers. The branch is left in place for reference.

PR #1644, opened 2026-09-16
The WebUI launcher gets the Frontends tab

Nisugi:
Stacked on #1643.

PR #1558 added a fourth tab to the GTK launcher: a frontend catalog with detection status, and an editor for built-in launch overrides and custom frontends. The WebUI launcher had no counterpart, so under `--webui-dev` a player could *select* a configured frontend but had no way to *configure* one.

## Two halves of one gap

**It could not offer what you configured.** `discover_frontends` listed only `FrontendLocator#available`, which is the set of frontends whose executable discovery could find. A custom frontend has no registry entry, no macOS bundle id and no conventional path -- it is reachable only because the player gave it a launch command -- so discovery can *never* return one. The launcher could not offer it however carefully it had been configured. Vellum is exactly this case.

The GTK selector has always done the opposite: list the catalog, let discovery annotate each entry as configured, detected or unavailable. That logic moves to `Lich::Common::FrontendChoices`.

**It had no way to configure one.** That is this tab.

## The tab

A catalog table over an editor for the selected row, with Add Custom / Save / Delete Custom / Reload in one row underneath, matching GTK's order.

The editing rules move to `Lich::Common::FrontendEditor` -- which frontends a list shows, what a built-in may override, what makes an id acceptable, how a shell-quoted argument string parses. They were private methods wrapped around `Gtk::Entry` instances in `GUI::FrontendManagerTab`, so reaching them meant either instantiating GTK or writing a second copy of the validation, and a second copy is how the two launchers drift. Nothing in the module reads or writes a file: callers hand in the document and persist the result.

A built-in keeps its identity -- id, label, working directory and capabilities are locked, only the launch override is editable, with the detected path shown beside it. Its protocol capabilities come from the catalog, so the editor says what the frontend speaks without letting it be claimed otherwise.

**Saves re-read `frontends.yml` immediately before writing.** The GTK launcher and any other Lich share that file, and a stale document would drop whatever they had added.

`GUI::FrontendManagerTab` is deliberately not refactored onto the shared module. It works, and rewriting the launcher people actually use for no behavioural gain is the wrong trade while this is opt-in. The two must change together until the GTK launcher retires, which `frontend_editor_spec` pins.

## A launcher bug found by review

Closing the launcher during authentication launched anyway. `SerialExecutor#stop(wait: false)` does not interrupt work already running, and the persistent path called `SessionLauncher#launch` *before* consulting `complete`, while both terminal paths called `terminal_launch` without checking whether the completion was accepted.

```
before:  session_launcher called 1 time after close
after:   0 times, and 1 time when the launcher stays open
```

The existing persistent-launch spec uses an immediate executor, which cannot show this -- the launch has to be in flight while close runs. The new specs block inside `authenticate`, and the closed-launcher one was confirmed to fail with the guard removed.

## Fidelity

Several rounds against the GTK tab side by side: labels beside fields rather than stacked above them, capabilities three across, the catalog bounded so the editor stays on screen, the first row selected on load so the editor is populated rather than showing a placeholder.

**559 examples, 0 failures**; rubocop clean across 105 files.


Nisugi:
Superseded by the re-layered stack: this work now lives in webui/launcher (L5). See docs/webui-rebuild-plan.md on webui/core-corrections for the map from the old PRs to the new layers. The branch is left in place for reference.

PR #1648, opened 2026-09-17
fix(webui): core corrections to the vendored WebUI runtime, server and coordinator

Nisugi:
First of eight stacked PRs that re-layer the WebUI stack (#1635–#1644) by architecture rather than by the order things happened. Base is #1634, Doug's vendored core, untouched. This one is **server-side corrections to that core only**: no client, no launcher, no shim. Read the plan first: `docs/webui-rebuild-plan.md` (this PR carries it, with the review ledger and the handoff).

Fixes to the vendored runtime, dispatcher, server, adapter and modal coordinator, each with a spec that fails without it:

- **Dispatcher**: refuses work for an owner that has been shut down, checks the tombstone and the state lookup under one lock (a concurrent enqueue used to revive the owner), and bounds its queue in one pass.
- **Runtime**: answers a stale event with the refusal first and then the render, names the event it refused, echoes the client's optional request id, and clears a sensitive field on request. Shutdown is bounded.
- **Server**: bounds what it reads and serves; a socket write has one monotonic deadline (the lock wait included) so a browser that stops reading, or drains a byte at a time, cannot park a script; the session cookie is named after the port so two Lich sessions in one browser profile do not evict each other; `stop` closes its socket.
- **Adapter**: `handle_for` answers from a reverse map; refresh runs outside the adapter's own lock.
- **Modal coordinator**: a modal page knows it is one; a failed open does not leave it registered; a modal whose window is closed is dismissed at once, a dropped socket after a five-second grace, instead of waiting out its hour.
- **Window presentation** (Windows): keep-above and opacity applied to the real browser window through its Win32 handle, found by pid and, when a shared profile handed the page to an already-running Chrome, by title. Honest degradation elsewhere. `PresentedWindow` gives any page that path, not only the shim.
- **Specs**: no spec can open a real browser (`LICH_WEBUI_NO_BROWSER`), and the GLib stubs let the real gem load.

## Review items closed

Doug's W01–W13 that concern the core (the ledger has the table), and from the outside review of 2026-09-17: R3, R5 (server half), R6, R9, R12. `docs/webui-review-ledger.md` part 4 answers every finding.

## Acceptance

`bundle exec rspec spec/lib/webui` at 0 failures. On the whole chain: 1382 examples across every layer's surface, 1 failure, the known Windows environment case (`launch_data_spec:166`, fails on `main`).


Nisugi:

Reviewed head for #1648: `dba90bf2`.

### F3 — P2: Modal close races destruction of the attachment used for the next lifecycle callback

**Layer:** #1648, interacting with the foundational runtime. **Code:** [runtime.rb:292](https://github.com/elanthia-online/lich-5/blob/28cd856f1f971d8b64de2bebb29eafc9ad352497/lib/webui/runtime.rb#L292), [runtime.rb:572](https://github.com/elanthia-online/lich-5/blob/28cd856f1f971d8b64de2bebb29eafc9ad352497/lib/webui/runtime.rb#L572).

`detach` enqueues `:close`, then separately constructs and enqueues `:detach`. The owner dispatcher is free to run the first callback immediately. The new modal close callback resolves the Future; completion closes the page; `ViewerStore#destroy_locked!` clears `attachment.render`. The second `enqueue_lifecycle` then dereferences `attachment.render.tree` and raises `NoMethodError`.

This exception is outside the runtime's normal contract/protocol refusal handling. The modal may already be resolved, but the detach message fails and the detach lifecycle callback is lost.

**Reproduction:** Use the real modal coordinator, runtime and viewer store with a dispatcher seam that executes each enqueued callback immediately. Attach a modal and detach it. The result is `undefined method 'tree' for nil` at `runtime.rb:572`. This forces a legal ordering of the actual independent worker; it does not replace the modal implementation.

**Fix direction:** Capture the necessary lifecycle contexts before dispatching either callback, or serialize the complete detach transition. Ensure destruction is idempotent and does not invalidate data still needed by the close path.

### F7 — P2: Title fallback can apply a new window's settings to an existing window

**Layer:** #1648. **Code:** [window_presentation.rb:169](https://github.com/elanthia-online/lich-5/blob/28cd856f1f971d8b64de2bebb29eafc9ad352497/lib/webui/window_presentation.rb#L169), [presented_window.rb:79](https://github.com/elanthia-online/lich-5/blob/28cd856f1f971d8b64de2bebb29eafc9ad352497/lib/webui/presented_window.rb#L79).

Native presentation discovery falls back to a title-prefix search on its very first poll if the spawned PID has no window. With a shared Chromium profile, the process can hand off asynchronously. If one older window already matches the same page title and the new window has not appeared yet, that older window is the unique match and is adopted immediately.

Refusing multiple matches does not cover this ordering: there is only one match at discovery time. Presentation settings intended for the new window then change the old one; the new window is never discovered. The PID check in `adopt` compares the requested PID with the stored requested PID, not the actual ownership of the discovered HWND.

**Reproduction:** Run the real discovery/presentation orchestration with controlled finder results: new PID has no HWND yet; title search returns existing HWND 31. `PresentedWindow.open` immediately applies the new settings to 31. No actual desktop window is modified by the probe.

**Fix direction:** Use an identity unique to the particular browser window/open attempt. At minimum exclude preexisting candidates and resolve ambiguity after the newly opened page identifies itself. A reusable title prefix is not sufficient ownership evidence.

### Validation and scope

mrhoribu (commented):
# PR Review: fix(webui): core corrections to the vendored WebUI runtime, server and coordinator (lich-5#1648)

**Reviewed:** 32 files, +3,411/−91 at `dba90bf2` on 2026-09-17. Based on: full clone.
**Summary:** Server-side corrections to the vendored core, each paired with a spec. The corrections are real and I found no new defect in them. Two findings from the second outside review land in this layer (F3, F7); both are fixed in #1657 and I verified them there, with one sub-point of F7 left open.

## Findings

None new at Blocker, Major, or Minor severity in this PR's own diff.

**F3 — modal close raced the detach context.** Fixed in #1657 (`lib/webui/runtime.rb:290`). `enqueue_lifecycle` is split into a `lifecycle_job` that captures `attachment.render` at build time and returns a lambda holding only the dispatch; `detach` now builds both jobs before calling either. The `return unless callback && render` guard also means a nil render yields no job rather than a `NoMethodError`. This is the tidiest of the seven fixes — smaller than "serialize the detach transition" and it removes the class of problem rather than the one ordering.

**F7 — title discovery adopted a pre-existing window.** Fixed for the reported reproduction in #1657 (`lib/webui/window_presentation.rb:160`, `presented_window.rb:29`). `PresentedWindow.open` now enumerates windows already carrying the page title *before* asking the browser for a new one and passes them as `exclude:`, which `enumeration` filters out of every `EnumWindows` pass. The closure is still built on the thread that calls it, so the Fiddle constraint the surrounding comment documents is respected.

## Open questions

- **`adopt` still compares requested pid to requested pid** — `lib/webui/presented_window.rb:79`. The second review's F7 made two points; the exclusion list answers the first. The second — "the PID check in `adopt` compares the requested PID with the stored requested PID, not the actual ownership of the discovered HWND" — is untouched by the fix. **Settles it:** was this deliberately left as covered-in-practice by the exclusion list, or missed? One line from you closes it either way. I could not test any of this: it is Windows-only and this is a Linux box.

## What this PR gets right

The port-derived cookie name (`server.rb:15`) is the correct fix for R3 rather than a workaround — it solves the shared-browser-profile collision without giving up the shared profile, which is what the shim needs. Bounding the socket write on one monotonic deadline *including the lock wait* is likewise the stronger reading of R12 rather than the minimal one.

## Coverage notes

Windows-only code (`window_presentation.rb`, `presented_window.rb`, the Fiddle/Win32 handle work) is unexecutable on this box, so F7's fix is reviewed as code only. The full suite on the stack tip (`e945baf0`, #1658) is green: 8,717 examples, 0 failures.

PR #1649, opened 2026-09-17
feat(all): one launcher choice, and ScriptScope for a script's toolkit

Nisugi:
Second of the stack, on top of `webui/core-corrections`. **How Lich chooses a launcher, and how a script's toolkit is resolved.** Small and new: nothing here is curated from the old stack.

- **One launcher choice** (`lib/common/launcher_choice.rb`, `Lich.launcher`): the flag (`--webui`, its `--webui-dev` alias, `--gtk`), then the persisted `launcher` setting, then `DEFAULT`. Every site reads that one answer: `init.rb` skips `require 'gtk3'` when the launcher is the WebUI, `main.rb` branches on it, `lich.rbw` runs the GTK main loop only when GTK was chosen. `DEFAULT` stays `:gtk` here; the flip is the last PR.
- **Child sessions** inherit an explicit launcher flag, and the WebUI launcher names its own.
- **ScriptScope**: trusted scripts resolve bare constants (`Gtk`, `Gdk`, `GLib`) through a scope core never sees. This is the seam the shim (later PRs) plugs into; core stays toolkit-free.
- `--gtk` alone opens the native launcher. The GTK branch used to require an empty ARGV or `--gui`, so the advertised way back started Lich headless (review 2026-09-17, R1). The spec now evaluates both branch predicates against the real parser for `--gtk`, `--webui`, no arguments under each persisted choice, `--gui`, `--no-gui`, and a gtk3 that failed to load.

## Acceptance

`spec/lib/main`, `launcher_choice_spec`, `script_scope_spec`, `session_launcher_spec` at 0 failures (plus the one known Windows environment case in `session_launcher_spec`, which fails on `main`).


Nisugi:

Reviewed head for #1649: `eab7a799`.



No additional standalone finding was identified in the wiring layer in this pass. The explicit GTK launcher-route correction is present. Outstanding downstream launcher and shim findings are on #1652 and #1653; the stack release recommendation is on #1655.

### Validation and scope

mrhoribu (commented):
# PR Review: feat(all): one launcher choice, and ScriptScope for a script's toolkit (lich-5#1649)

**Reviewed:** 14 files, +671/−55 at `eab7a799` on 2026-09-17. Based on: full clone.
**Verdict:** Approve
**Summary:** Collapses launcher selection to one answer (`Lich.launcher`) that every site reads, and introduces `ScriptScope` as the seam the later shim plugs into. Small, well-scoped, and correctly tested. No findings.

## Findings

None at Blocker, Major, or Minor severity.

## What this PR gets right

The R1 fix is the notable part. The previous GTK branch required `ARGV.empty?` or `--gui`, so `--gtk` — the advertised escape hatch — started Lich headless. The new predicate at `lib/main/main.rb:213` admits `@argv_options[:launcher] == :gtk`, and the replacement spec (`spec/lib/main/launcher_option_spec.rb`, 161 lines, superseding the 43-line `webui_dev_option_spec.rb`) evaluates *both* branch predicates against the real parser across `--gtk`, `--webui`, bare ARGV under each persisted setting, `--gui`, `--no-gui`, and a failed gtk3 load. That is the right shape of test for a routing bug: the original defect was invisible to source-presence assertions, and this spec would have caught it.

Keeping `DEFAULT = :gtk` here and deferring the flip to #1655 is the right sequencing — it makes the whole series reviewable without any behaviour change until one line moves.

`FLAGS` retaining `'--webui-dev' => :webui` as an alias means #1634's flag keeps working for anyone already using it. Small thing, easy to forget, not forgotten.

## Coverage notes

`spec/lib/main`, `launcher_choice_spec`, `script_scope_spec` and `session_launcher_spec` all pass on this box with no environment failures — the Windows `session_launcher_spec` case the description mentions does not fail on Linux.

PR #1650, opened 2026-09-17
feat(webui): the contract at 2.19 and a client that renders all of it

Nisugi:
Third of the stack, on top of `webui/wiring`. **The contract and the whole browser client.** Largest PR by lines because `app.js` could not be split by layer without surgery; it is one thing to approve: the vocabulary a native app or a shim renders against, and a client that renders all of it.

- **Contract 2.17 → 2.19**: every type and property the stack declared, published as `lib/webui/contract.json` (the suite fails when the committed copy drifts). 2.18 lets a password say it changed without saying what. 2.19 adds `surface_zoom` (ctrl+wheel over a composite, with the pointer, for a map to zoom around) and an optional request id on an event that a refusal echoes.
- **The client renders everything declared**: markdown, log, slider, composite layers (image, region, bar, label, and the drawn line/rect/ellipse shapes so a map's markers are data, not pixbufs), menus with submenus, tables with editors and multi-select, tree rows as a tree.
- **Window fitting**: a window sizes itself to its page on first render (the larger of declared geometry and natural content), scrolls when smaller, closes with its last page; a window-level scroller is not the page's natural size; geometry stops restyling a fitted page.
- **Correctness under refresh**: a stale refusal replays exactly the event it names, once, into the render that follows; records survive unrelated renders (a click no longer vanishes when a refresh lands before its refusal); half-typed inputs, passwords and table cells survive a render, and a value the script changed still wins; a committed cell edit is the script's to answer.
- **The jsdom harness** (`spec/webui_client/`, 22 cases, run by `client_harness_spec`): the real `app.js` in jsdom against a fake WebSocket, driven through DOM events. This is what the audit asked for: a client claim somebody executed. A fresh checkout needs `npm ci` in that directory; without node the example is pending, loudly.

## Review items closed

Doug's W12 and the client items in the ledger; Astra's client comments; from the 2026-09-17 review: R5 (client half), R10, R14, and the client side of R7.

## Acceptance

`bundle exec rspec spec/lib/webui` at 0 failures with the harness at 22/22.


Nisugi:

Reviewed head for #1650: `4251f279`.

### F5 — P2: Retry records have no effective idle expiry and retain cleared passwords

**Layer:** #1650. **Code:** [app.js:171](https://github.com/elanthia-online/lich-5/blob/28cd856f1f971d8b64de2bebb29eafc9ad352497/lib/webui/assets/app.js#L171), [app.js:1868](https://github.com/elanthia-online/lich-5/blob/28cd856f1f971d8b64de2bebb29eafc9ad352497/lib/webui/assets/app.js#L1868), `page_closed` at 1972.

Every eligible event adds a request record, including its submission. Successful records are retained for possible delayed stale refusals. The stated 30-second TTL is checked only when another render of that same page arrives. There is no timer, count bound, or page-close cleanup of these records. An idle or closed page can therefore retain them indefinitely; `clear_sensitive` empties controls but leaves the submitted password in this map.

Additionally, the render handler checks `record.replay` before checking age. A stale refusal arriving after the nominal expiry marks the old record for replay and bypasses expiration entirely.

**Reproduction:** Submit a synthetic password, receive `clear_sensitive`, advance the browser clock by an hour, then deliver a stale refusal naming the original request and a render. The visible field remains empty, but the client sends a second event containing the original password. This demonstrates retention and expired replay, not credential disclosure to an outside party.

**Fix direction:** Enforce expiry before accepting a refusal or replay, sweep idle records, and remove records for closed pages. Define a bounded lifecycle for sensitive submissions that still permits legitimate in-flight stale recovery; an acknowledgment protocol is another option.

Cross-layer finding F4 is posted on #1654: the native hierarchy fixture binds `row_toggle`, but the shim does not, so expansion disappears on refresh. Please cover the shim-generated contract as well as the native fixture.

### Validation and scope

mrhoribu (commented):
# PR Review: feat(webui): the contract at 2.19 and a client that renders all of it (lich-5#1650)

**Reviewed:** 35 files, +13,949/−129 at `4251f279` on 2026-09-17. Based on: full clone; **the client harness could not be executed here** — see coverage notes.
**Summary:** The contract published as `contract.json` plus the whole browser client. Largest PR in the stack by lines and the one I can say least about with confidence, because its central artifact is JavaScript I could read but not run. F5 lands in this layer and is fixed in #1657, with one narrow gap left open.

## Findings

None new at Blocker, Major, or Minor severity that I can state from reading alone.

**F5 — retry records had no effective expiry and retained cleared passwords.** Fixed in #1657 (`lib/webui/assets/app.js:28`). The fix is thorough and addresses every limb of the finding: `sweepPending()` runs on every send *and* every received message rather than only on a render of the same page; the render handler now sweeps before testing `record.replay`, which closes the "refusal bypasses expiry" path the review specifically called out; `page_closed` and `clear_sensitive` drop the matching records; a dropped socket calls `pendingEvents.clear()`; and the set is capped at 256 with oldest-first eviction. The decision not to use a timer — stated in the comment, because a timer would hold the page's event loop and the harness with it — is the right trade.

## Open questions

- **A replayed record can carry a submission outside its own `scope`** — `lib/webui/assets/app.js:171`. `emit` stores `scope: scope || []` from `page.submissions[component.cid]`, recomputed at send time. On a replay, `message.submission` is the *original* submission, but `scope` is whatever the current render says. If the component was rebuilt so that `page.submissions` no longer has an entry for that cid, the new record carries the original password with `scope: []`, and the `clear_sensitive` sweep (`dropPending`) cannot match it. The 30-second TTL still bounds it, so this is a shortened retention rather than the unbounded one F5 described — which is why it is a question and not a finding. **Settles it:** can a page re-render between a send and its stale refusal in a way that drops that cid's `submissions` entry? One harness case either way.

## What this PR gets right

Publishing `contract.json` with a suite that fails when the committed copy drifts from `contract.rb` is the right way to make a contract a contract rather than a comment — it is the mechanism that lets `;map` and other native rewrites target a version instead of a commit.

The jsdom harness is the most valuable thing here and the description is right that it is what the earlier audit asked for. Running the real `app.js` against a fake WebSocket, driven through DOM events, is the difference between a client claim and a client assertion. Equally important: when `npm` is absent the example goes **pending, loudly**, rather than silently passing — which is exactly what happened on my box, and I knew it.

## Coverage notes

**This is the largest gap in my review of the whole stack.** `npm` is not installed here, so the 23-case jsdom harness did not run, and every client-side behavioural claim in this PR — window fitting, draft preservation, replay correlation, tree rendering, the client halves of F4 and F5 — rests on my reading of `app.js` rather than on execution. `lib/webui/assets/app.js` is the single biggest artifact in the stack and I read it selectively, around the areas the prior reviews flagged, not exhaustively.

Anyone relying on this review should treat the client as verified by the author's harness and the two prior reviews (which did run it: 22 and 23 cases passing respectively), not by me.

PR #1651, opened 2026-09-17
refactor(all): authentication and compaction out of the GTK stack; the core/GTK boundary check

Nisugi:
Fourth of the stack, on top of `webui/contract-primitives`. **The core/GTK boundary**: what moves out of `lib/common/gui` so that core has no GTK in it, and the check that says how far that has got. Mostly moves; the diff is large in lines and small in logic.

- **Authentication out of the GUI**: the saved-login helpers live in `lib/common/authentication` and core never reaches into `lib/common/gui`. The GTK launcher keeps compatibility constants that alias the moved modules; it does not keep copies.
- **Heap compaction** goes through `HeapCompaction`; only the GTK stack knows about `GtkCompaction`.
- **Saved-entry keys are identities, not positions**: derived from account, character, game, frontend and custom launch, so removing one entry no longer renames its neighbours and a stale edit or delete cannot land on the wrong configuration (review 2026-09-17, R8).
- **The boundary check** (`script/ci/check_core_gtk_boundary.rb`) names every core file that still reaches the GTK runtime. It is a warning until Lich 6 deletes GTK, when it becomes a gate; the 13 files it lists are the ones that must still load GTK for `--gtk` to work.

Known and stated: the frontend mutation rules are still duplicated between the two launchers (`FrontendEditor` is shared, the GTK frontend tab still has its own save flow). Consolidating means editing the launcher Lich 6 deletes, so it waits for that deletion.

## Acceptance

`spec/lib/common/authentication`, `cli`, `gui`, the boundary spec and `spec/lib/common/webui_launcher` at 0 failures (plus one known Windows environment case, `launch_data_spec:166`, which fails on `main`).


Nisugi:

Reviewed head for #1651: `4e86df0a`.



No additional standalone finding was identified in the services-boundary layer in this pass. The helper extraction reduces structural duplication. The remaining identity mismatch is in the launcher consumer (F6 on #1652), rather than being a new finding against this extraction. See #1655 for the stack release recommendation.

### Validation and scope

mrhoribu (commented):
# PR Review: refactor(all): authentication and compaction out of the GTK stack; the core/GTK boundary check (lich-5#1651)

**Reviewed:** 39 files, +2,291/−1,519 at `4e86df0a` on 2026-09-17. Based on: full clone; suite and RuboCop executed.
**Summary:** Mostly file moves: authentication out of `lib/common/gui` into `lib/common/authentication`, heap compaction behind `HeapCompaction`, plus a boundary checker that inventories what still reaches GTK. Large in lines, small in logic, and it holds together. One Minor.

## Minor

### RuboCop offenses introduced here — `spec/lib/common/webui_launcher/catalog_integration_spec.rb:126` and `:132`

```ruby
  entries
   .to_h { |entry| [[entry.frontend, entry.custom_launch], entry.key] }
```

Two `Layout/MultilineMethodCallIndentation` offenses, both autocorrectable with `rubocop -A`. Both were added by commit `b80dcfda`, titled *"style(launcher): align a multiline call in the catalog spec"* — the commit fixed one instance of this cop and introduced two more.

They are still present at the stack tip (`e945baf0`, #1658), where both #1657 and #1658 claim "RuboCop clean." Practical impact is limited: the CI job runs `bundle exec rubocop` repo-wide and is **already red on `main`** (73 offenses there, 77 at the tip), so these do not newly break CI. The finding is about the claim more than the cop.

## What this PR gets right

The R8 fix is the substantive change buried in the moves, and it is the right one. Deriving saved-entry keys from the full identity — account, character, game, frontend *and* custom launch — rather than from a hash plus an enumeration ordinal means removing one entry no longer renames its neighbour, so a stale edit or delete cannot land on a different configuration. That was a data-corruption-class bug and the fix removes the mechanism rather than guarding it.

The compatibility constants in the GTK launcher alias the moved modules instead of keeping copies. That distinction is what makes this a real boundary rather than a fork, and the first outside review verified it independently.

The boundary checker being **a warning, not a gate**, with the 13 remaining files named rather than hidden, is honest engineering. It is inventory, and the PR describes it as inventory.

The PR is also explicit about what it did *not* consolidate — the frontend mutation rules still duplicated between launchers, deferred because fixing them means editing the launcher Lich 6 deletes. Stating a known duplication with its reason is worth more than a silent partial fix.

## Coverage notes

The authentication, CLI, GUI, boundary and `webui_launcher` suites pass here with no failures; the Windows `launch_data_spec:166` case the description mentions does not fail on Linux. I reviewed this diff as a refactor — checking that moved modules are aliased rather than copied and that the R8 key derivation is sound — rather than re-verifying every moved line of credential handling against its original.

PR #1652, opened 2026-09-17
feat(all): the WebUI launcher is a complete login, Frontends tab included

Nisugi:
Fifth of the stack, on top of `webui/services-boundary`. **The WebUI launcher as a complete login**: every workflow the GTK launcher has, in the browser.

- **Every GTK workflow**: saved entries with favourites, sorting and deletion; manual entry with save; account add and removal; encryption mode and master password changes with the unlock modal; multi-launch in persistent mode; the launcher choice toggle (labelled "native launcher", because the boundary check flags any identifier containing `gtk` and the launcher stays on the clean side of it).
- **The Frontends tab**: toolkit-free frontend editing (`FrontendEditor`, `FrontendChoices`) that both launchers call. Save reads the submission by the cid each field was declared under; it used to treat the submission as an Array in declaration order, which the runtime never sends, so a custom frontend could never be saved (review 2026-09-17, R2). The spec now drives the editor with a real `Submission`.
- **Cancellation on every path**: `commit` is the one arbiter between a close and an irreversible step. Account save, encryption change, catalog mutations and the session launch run under the launcher's lock only while the operation is live and the launcher is not closing; a terminal launch accepts the launch and begins the close as one locked step. A close accepted first refuses the step; a close arriving during it waits (review 2026-09-17, R4).

## Acceptance

`spec/lib/common/webui_launcher*` at 0 failures.


Nisugi:

Reviewed head for #1652: `ded07b48`.

### F2 — P1: Master-password changes still commit after launcher cancellation

**Layer:** #1652. **Code:** [webui_launcher.rb:929](https://github.com/elanthia-online/lich-5/blob/28cd856f1f971d8b64de2bebb29eafc9ad352497/lib/common/webui_launcher.rb#L929).

`change_master_password` transfers the three submitted secrets, creates an operation, and queues a worker. That worker calls `@catalog.change_master_password(current, replacement)` directly. It never uses the new `commit(operation)` arbiter.

`close` clears active operations, but `SerialExecutor#stop(wait: false)` places a stop marker behind already queued work. A password change waiting behind another operation therefore still runs after close. Only its subsequent UI completion is rejected. The underlying method changes credential storage, so rejecting the completion is too late.

**Reproduction:** Queue the production handler using real `Submission` and `SensitiveValue` objects, close the launcher, then release the queued work. A recording catalog receives the password-change call while `launcher.lifecycle == :closed`.

**Fix direction:** Arbitrate the mutation itself under `commit(operation)`, as the corrected account-save and persistent-launch paths do. Cover close and viewer cancellation before the queued job starts. Do not treat a rejected `complete` as cancellation of an earlier write.

### F6 — P2: Favorite lookup can choose a different custom launch entry

**Layer:** #1652. **Code:** [webui_launcher.rb:1463](https://github.com/elanthia-online/lich-5/blob/28cd856f1f971d8b64de2bebb29eafc9ad352497/lib/common/webui_launcher.rb#L1463), caller at 1092.

The catalog now correctly distinguishes entries by custom launch command, but `find_entry_key` compares only account, character, game and frontend. Manual Play with Favorite enabled saves the requested command, then finds the first entry sharing those four fields and toggles its favorite. If the same character/frontend has two commands, that can be the other entry.

**Reproduction:** Create an actual temporary catalog with `command-one` and `command-two` for the same account/character/game/frontend. Request the key for `command-one` through the production lookup; it resolves the entry whose command is `command-two`.

**Fix direction:** Return the exact saved key from the upsert operation or share the catalog's complete identity comparison. Avoid duplicating a weaker identity definition in the launcher. Also review whether the Manual Entry checkbox should set favorite true rather than toggle an existing favorite off.

### Validation and scope

mrhoribu (commented):
# PR Review: feat(all): the WebUI launcher is a complete login, Frontends tab included (lich-5#1652)

**Reviewed:** 9 files, +1,624/−62 at `ded07b48` on 2026-09-17. Based on: full clone.
**Summary:** The WebUI launcher reaches parity with the GTK one. Two of the second review's findings (F2, F6) belong to this layer; I verified both fixes in #1657 and both are correct. No new findings.

## Findings

None new at Blocker, Major, or Minor severity.

**F2 — master-password changes committed after cancellation.** Fixed in #1657 (`lib/common/webui_launcher.rb:942`). The catalog call now runs inside `commit(operation)` like every other irreversible step, rather than running unguarded with only its UI completion rejected afterwards. I traced the return-value plumbing, because the fix depends on it: `commit` returns `nil` when it refuses and `[true, yield]` when it commits (`webui_launcher.rb:1207`), `consume_three` propagates its block's value, so `if committed` correctly distinguishes "the write happened" from "the close won." `complete` and `set_notice` now run only on the committed path.

Worth stating explicitly since it looks alarming: `commit` holds `@mutex` while yielding, so a master-password change — which re-encrypts stored credentials — blocks `close` for its duration. That is deliberate and matches this PR's own stated design ("a close arriving during it waits"). It is the same shape as the already-shipped account-save and launch paths, so it is consistent rather than novel.

**F6 — the favourite could land on a sibling entry.** Fixed in #1657 across `webui_launcher.rb:1114` and `catalog.rb:135`. The weaker four-field `find_entry_key` lookup is **deleted** rather than strengthened, and `upsert_manual_entry` now answers with the key it actually wrote, matched on the full identity including `custom_launch`. `toggle_favorite` is refactored onto a shared `update_favorite`, and the save path calls the new `set_favorite(key, true)` — so re-saving an entry that is already a favourite keeps it one instead of toggling it off, which is the secondary point the review raised.

I checked the identity comparison for the nil-vs-empty-string trap, since `character['custom_launch']` can be nil: `character.merge!` coerces `game_code` and `frontend` with `.to_s` before the comparison, and `custom_launch` is compared `.to_s` on both sides. No asymmetry. I also grepped every caller of `upsert_manual_entry` and `toggle_favorite` to confirm nothing relied on the old boolean return — nothing does.

## What this PR gets right

Deleting `find_entry_key` rather than fixing it is the right instinct, and it is what the second review actually asked for ("avoid duplicating a weaker identity definition in the launcher"). A duplicated identity that has been corrected once will drift again; a deleted one cannot.

`commit` as the single arbiter between a close and an irreversible step is a good primitive — the F2 fix was a three-line change precisely because the primitive already existed and had been applied everywhere else.

The R2 fix (reading the submission by cid rather than treating it as an Array) with a spec that drives the editor with a *real* `Submission` is the right correction: the original defect existed only because the passing test supplied an Array the runtime never sends.

PR #1653, opened 2026-09-17
feat(all): GTK compatibility shim, core -- the session and the widget vocabulary

Nisugi:
Sixth of the stack, on top of `webui/launcher`. **The GTK compatibility shim, core half**: the script scope and the widget vocabulary the supported-script list uses, with its declared limits. The admission rule (plan, "The shim's admission rule") is what keeps this finite: an elanthia-online script is never a shim candidate, it gets rewritten natively (`;map` already is); a third-party script enters by naming what it needs.

- **The session**: each script's GTK runs on one thread, the queue is drained and committed once per batch (D3), viewers are written only after a commit. A shim window opens the way lich-6 opens a script page: no private profile, no process monitor (D1); a window closed with its X tells the script through a detach nobody returns from.
- **The vocabulary**: windows, boxes, tables, grids, entries, labels, dialogs; GLib timeouts and idles on the session thread; page keys for a window that connects `key-press-event`; a window's default size as the geometry facility (a size request is only a minimum).
- **What it says it cannot do**: `const_missing`, the stubbed-widget notice, clamp reporting, and a per-script ledger (`Gtk.unsupported_report`) that names every degradation once, including the 640x480 allocation fallback (D5) and presentation the runtime refused (D16). No images, no drawing, no menus, no `Gtk::Layout`: those scripts are rewritten.
- **Decisions from the ledger**, implemented here: D1, D3, D4 (`respond_to?` answers only for the setter shapes `method_missing` handles), D5, D13 (three narrow adapter hooks replace the shim's copy of the traversal), D15 (`Dialog#run` waits on a Future the session tracks, cancelled at shutdown with `MessageDialog`'s), D16, D26 (a shim page refuses a second viewer).
- A password entry's `changed` handler fires as a notification without the value (review 2026-09-17, R13).

## Acceptance

`spec/lib/common/script_scope` (core subset) at 0 failures. In-game: `;eloot`, `;bigshot`, `;armor popup`, `;vars` open, edit, save and close on this box.


Nisugi:

Reviewed head for #1653: `26b6b46b`.

### F1 — P1: A commit exception can kill the session thread and strand synchronous callers

**Layer:** #1653. **Code:** [session.rb:436](https://github.com/elanthia-online/lich-5/blob/28cd856f1f971d8b64de2bebb29eafc9ad352497/lib/common/script_scope/gtk/session.rb#L436), also `run_batch` at 413, `commit` at 480, and `sync` at 358.

`run_batch` rescues errors from `job.call`, but its final `commit unless @closed` executes outside that rescue. `commit` itself rescues only `Lich::WebUI::Error`. A normal `NoMethodError` or other script/materialization exception therefore terminates the worker.

There is a consequential race with `sync`: enqueue a synchronous job while the worker is still alive inside the failing commit. `ensure_thread` sees a live worker and does not start another. When that commit raises, the new job remains queued and its caller waits indefinitely on `done.pop`. A future enqueue might restart the worker, but nothing guarantees another enqueue will arrive.

**Reproduction:** A barrier holds a real session inside a synthetic failing window's `materialize!`; a second thread calls `session.sync`; releasing the barrier raises `NoMethodError`. The observed result is `worker_alive: false, sync_still_waiting: true`. The probe terminates its own waiting thread afterward.

The broad suite stalled with the same commit stack. Interestingly, the existing “survives a script error raised during commit” example passed when isolated while printing worker-death traces: its later call can restart a dead worker. That assertion does not prove the worker survived or cover the enqueue-before-death ordering.

**Fix direction:** Put batch commit inside the worker's exception boundary. Define how a failed session completes or rejects outstanding synchronous jobs, and test the barrier-controlled ordering rather than relying on sleeps and a later successful `sync`.

### Validation and scope

mrhoribu (commented):
# PR Review: feat(all): GTK compatibility shim, core -- the session and the widget vocabulary (lich-5#1653)

**Reviewed:** 18 files, +8,377/−49 at `26b6b46b` on 2026-09-17. Based on: full clone; suite, RuboCop and a targeted concurrency probe executed.
**Verdict:** **Request changes**
**Summary:** The shim's session and widget vocabulary. This layer owns `Session`, and the second review's F1 — the P1 that killed the session thread and stranded synchronous callers — belongs here. #1657's fix for it is a large improvement but **does not close the hole**, and I have a deterministic reproduction.

## Major

### 1. `sync` can still park for ever when shutdown interleaves with `enqueue` — `lib/common/script_scope/gtk/session.rb:346`

```ruby
  job ||= block
  raise ArgumentError, 'block required' unless job
  return nil if @closed

  @queue << job
```

The `@closed` check (line 349) and the push (line 351) are still two separate, unsynchronized steps — the exact structure the fix's own comment at line 385 claims to have replaced: *"The closed check and the enqueue used to be two steps, and a shutdown between them queued a job nobody would ever run."* `sync` was changed to trust `enqueue`'s return value, but `enqueue` itself was not made atomic. `shutdown` (line 753) sets `@closed` and pushes `:stop` without taking `@thread_mutex`, and `run_loop`'s new `ensure reject_queued_jobs` (line 818) drains the queue exactly once, on the way out.

That leaves this interleaving:

1. A script thread enters `enqueue` and reads `@closed` → false.
2. `shutdown` runs: `@closed = true`, `@queue << :stop`.
3. The session thread pops `:stop`, leaves the loop, and `reject_queued_jobs` drains an **empty** queue. The thread ends.
4. The script thread pushes its `SyncJob` and calls `ensure_thread`, which returns immediately because `@closed` is now true (line 787).
5. `enqueue` returns `true` — asserting it took the job — and `sync` blocks on `done.pop` (line 391), which has no timeout.

Reproduced deterministically against a real `Session`, pausing only inside the production `enqueue` between its two steps and driving `shutdown` from another thread:

```
closed=true
worker_alive=false
queue_size=1
sync_result=:STILL_WAITING
```

The new spec at `spec/lib/common/script_scope/gtk_review_fixes_spec.rb:228` does not cover this ordering — it kills the thread *after* the job is already queued, which is precisely the case `reject_queued_jobs` was written for.

This is reachable in production, not just in a probe. `GLib::Timeout.add` spawns a source thread that loops `sleep(interval)` then `session.sync { block.call }` (`lib/common/script_scope/gtk/glib.rb:47`), and script exit fires `@owner.at_exit { session.shutdown }` (`session.rb:833`) on another thread — so any script with a repeating timer that is stopped mid-tick can hit it. The parked thread never reaches its `ensure`, so `GLib.remove_source(id)` never runs and the source entry leaks until Lich tears the script's threads down.

I have graded it Major rather than Blocker because it is strictly narrower than the original F1, which also killed a *live* session's worker and could strand callers indefinitely thereafter. But the fix is advertised as closing F1, and it does not.

**Fix:** put the `@closed` read and the `@queue <<` under one lock that `shutdown` and `reject_queued_jobs` also take, so "took the job" and "a thread will drain it" are decided together. Then `enqueue`'s `true` means what the F1 fix says it means. A `done.pop` timeout would be a useful second line of defence but is not a substitute.

## Minor

### 2. RuboCop offense introduced here — `spec/lib/common/script_scope/gtk_review_fixes_spec.rb:354`

One `Layout/MultilineMethodCallIndentation`, autocorrectable with `rubocop -A`, still present at the stack tip (`e945baf0`, #1658), where both #1657 and #1658 claim "RuboCop clean." CI runs RuboCop repo-wide but is already red on `main`, so this does not newly break CI.

## What this PR gets right

Most of the F1 work is genuinely good and should be kept as-is: moving the batch `commit` inside the job rescue (line 452), wrapping `run_batch` in `run_loop`'s own boundary, and introducing `SyncJob` with a `reject` so a refusal can be delivered to a waiting caller at all. The remaining hole is one missing lock, not a wrong design — `SyncJob#reject` is exactly the primitive needed to close it.

The admission rule stated in the description — an elanthia-online script is never a shim candidate, it gets rewritten natively — is what keeps this layer finite, and `Gtk.unsupported_report` naming each degradation once per script is the right way to make the limits visible instead of mysterious.

D4 (`respond_to?` answers only for the setter shapes `method_missing` actually handles) is a detail that is very easy to get wrong and is got right here.

This PR adds the only non-ASCII in the entire stack — an `'e-acute' * 600` tooltip in `spec/lib/common/script_scope/gtk_shim_spec.rb:171`, where multi-byte clamping is the point of the test — and suppresses the repo's `Custom/AsciiOnlySource` cop narrowly, with a stated reason, re-enabling it 18 lines later. That is how a justified exception should look.

## Coverage notes

The `spec/lib/common/script_scope` suite passes here (part of 752 examples, 0 failures across the WebUI/launcher/shim/main surfaces; 8,717 examples, 0 failures for the whole repo at the stack tip `e945baf0`). The in-game acceptance the description cites — `;eloot`, `;bigshot`, `;armor popup`, `;vars` — I could not attempt: no game session, no browser, no GTK.

PR #1654, opened 2026-09-17
feat(all): GTK compatibility shim, data -- stores, tree views and Builder

Nisugi:
Seventh of the stack, on top of `webui/shim-core`. **The shim's data widgets and Builder**: the rest of the declared surface.

- **Stores and tree views**: `ListStore`, `TreeStore`, iter-as-cursor, editors, row activation, multi-select. The script owns the model on a cell edit, as under GTK: the shim emits only the renderer's own signal (`edited` with the text, `toggled` with the path) and the handler decides what the cell holds; it used to write the value first, so a conventional toggle handler inverted it straight back (review 2026-09-17, R7).
- **`Gtk::Builder`** reads the Glade files the supported scripts ship and degrades any class the shim does not admit. Renderer properties from Glade (`editable`, `activatable`) reach the renderers; a guard on `method_missing` had made that impossible (review 2026-09-17, R11).
- **Paned, Overlay, ProgressBar, ListBox** join, and the examples parked out of the core half come home.
- **Text buffers** have a tag table; a tag on a range, and markup, are reported or reduced rather than fatal (`;armor popup` found both).

## Acceptance

`spec/lib/common/script_scope` at 0 failures. Every script on the supported list that uses a data widget runs unmodified: `jinx`, `ewaggle`, `repository`, `vars`.


Nisugi:

Reviewed head for #1654: `d11525ee`.

### F4 — P2: Shim tree expansion disappears on the next server render

**Layers:** #1654 and #1650 integration. **Code:** [widgets_data.rb:1558](https://github.com/elanthia-online/lich-5/blob/28cd856f1f971d8b64de2bebb29eafc9ad352497/lib/common/script_scope/gtk/widgets_data.rb#L1558), [app.js:1061](https://github.com/elanthia-online/lich-5/blob/28cd856f1f971d8b64de2bebb29eafc9ad352497/lib/webui/assets/app.js#L1061).

The new client renders a hierarchy and changes `row.expanded` locally when its arrow is clicked. It then tries to emit `row_toggle`. Shim `TreeView#always_bound_events` binds selection and cell editing, but not `row_toggle`; the client's `emit` drops unbound events. Consequently the runtime never stores expansion for shim tables. Their row properties also supply no expanded state.

Clicking the arrow works briefly, but any server render reconstructs the rows as collapsed. A selection or periodic script update can therefore collapse the tree the player just opened. `expand_all`, `collapse_all`, and `expand_row` are also still no-ops.

**Reproduction:** Run the real client with a hierarchical table using the shim's bindings and row shape. Visible rows go from 3 to 5 on expansion, no event is sent, and a fresh render returns the visible count to 3. The native client test passes because its fixture binds `row_toggle`; it does not cover the shim combination.

**Fix direction:** Bind expansion events for shim trees so the viewer store receives them, and implement the programmatic expansion methods against the same state model. Test expand → select/update → rerender through the shim-generated contract.

This is an integration finding with the client in #1650. The native client hierarchy test passes, but its bindings differ from those produced by this shim.

### Validation and scope

mrhoribu (commented):
# PR Review: feat(all): GTK compatibility shim, data -- stores, tree views and Builder (lich-5#1654)

**Reviewed:** 10 files, +4,069/−1 at `d11525ee` on 2026-09-17. Based on: full clone; suite and RuboCop executed. Client half not executable — see coverage notes.
**Summary:** Stores, tree views, Builder and the remaining data widgets. F4 belongs to this layer (jointly with #1650) and its fix in #1657 is complete and well-constructed on the server and shim sides. One Minor.

## Minor

### RuboCop offense introduced here — `lib/common/script_scope/gtk/widgets_data.rb:1598`

```ruby
    .rows
   .map { |iter| @model.send(:dup_row, iter) }
```

One `Layout/MultilineMethodCallIndentation`, autocorrectable, still present at the stack tip (`e945baf0`, #1658), where both #1657 and #1658 claim "RuboCop clean." This is the only one of the four stack-introduced offenses that is in `lib/` rather than `spec/`. CI runs RuboCop repo-wide but is already red on `main`, so it does not newly break CI.

## Verification of the finding assigned to this layer

**F4 — shim tree expansion vanished on the next render.** Fixed in #1657, and the fix is more complete than the finding strictly required. It spans three places and they agree with each other:

- **The shim** (`widgets_data.rb:1499`) gains real `expand_all`, `collapse_all`, `expand_row`, `collapse_row` and `row_expanded?` where four of those were `self`-returning no-ops, keeps `@expanded_keys`, adds `:row_toggle` to `always_bound_events` when the model is a `TreeStore` (line 1622) so the client's event is no longer dropped as unbound, emits `row[:expanded]` in `node_props` for rows that can actually open, and handles the inbound `:row_toggle` in `apply_event` without pushing back state the viewer already holds.
- **The runtime** (`lib/webui/runtime.rb:612`) grows `write_row_expansion`, which validates the row exists against `component.props[:rows]` before storing `expanded:<row>` per viewer.
- **Lifecycle bookkeeping** is handled: `model_changed!` (line 1556) now intersects `@expanded_keys` with live row keys alongside `@selected_keys`, so a row removed from the model does not leave a stale expansion behind.

I checked the ordering hazard that this kind of fix usually trips over — a script calling `expand_all` before the window is materialized, when `component.props[:rows]` would be empty and the runtime write would raise. It is handled: `viewer_push` only reaches the session `if @handle` (`widgets.rb:704`), and viewer writes are flushed after commit, so a pre-materialization `expand_all` updates `@expanded_keys` locally and arrives through the render instead.

## Open question

- **Shim expansion is per-page, not per-viewer** — `widgets_data.rb:1434`. The runtime keeps expansion per viewer, but the shim keeps one `@expanded_keys` list and emits `row[:expanded]` as the shared base prop, so a second viewer would inherit the first's expansion. I believe this is unreachable because D26 makes a shim page refuse a second viewer, which is why it is a question and not a finding. **Settles it:** does D26 hold for every shim page, including one re-attached after a reconnect?

## What this PR gets right

The R7 fix is the important behavioural correction in this PR and it is right for the right reason: the shim now emits only the renderer's own signal (`edited` with the text, `toggled` with the path) and lets the script's handler decide what the cell holds, exactly as GTK does. Writing the value first made the conventional `iter[col] = !iter[col]` toggle handler invert it straight back — and the review grounded that against a real installed script (`bsprofiles.lic`). Matching the toolkit's semantics rather than patching the symptom is what makes the shim trustworthy for unmodified scripts.

The R11 fix (checking for an actually-defined setter instead of `!respond_to?(:method_missing)`, which was unconditionally false on a class that publicly defines it) is a small correction of a guard that silently disabled every Builder renderer property.

## Coverage notes

`spec/lib/common/script_scope` passes here. The client half of F4 — that the browser emits `row_toggle` and honours `expanded` on re-render — is covered by the jsdom harness, which **did not run on this box** (`npm` is not installed); the harness example goes pending, loudly. The in-game acceptance cited in the description (`jinx`, `ewaggle`, `repository`, `vars`) was not attempted.

PR #1655, opened 2026-09-17
feat(all): the WebUI is the default launcher; GTK stays behind --gtk

Nisugi:
Last of the stack, on top of `webui/shim-data`. **One constant and one spec line**: `LauncherChoice::DEFAULT` becomes `:webui`.

With this merged, `ruby lich.rbw` opens the WebUI launcher. GTK stays reachable behind `--gtk` and the persisted setting (the launcher's own toggle writes it), and the core/GTK boundary check stays a warning. Lich 6 deletes GTK (plan, L8); this PR does not.

## What it depends on

Everything below it in the stack, in particular the two outside reviews of 2026-09-17: the first's fourteen findings, six of them named as blocking this flip, and the second's seven (two P1: a session thread killed by a raising commit, a master-password change committed after close), fixed in one PR on top of this one, `webui/review-2-fixes`, each pinned by a test that fails without it (`docs/webui-review-ledger.md`, parts 4 and 5).

## Companion: `;map` rewritten

The shim does not cover images or menus (plan, "The shim's admission rule"): org scripts that used them are rewritten against the contract instead. `;map` is the first, as elanthia-online/scripts #2479 (`map.lic` 3.0.0). It needs the contract at 2.19, which this stack ships, and refuses an older Lich with a message rather than failing. `bsprofiles` and `creaturebar` follow.

## Acceptance

`launcher_choice_spec` at 0 failures. Whole chain on this tip: 1382 examples across every layer's surface, 1 failure, the known Windows environment case that fails on `main`. Live on this box: the WebUI launcher, a Multi-Launch child, and the rewritten `;map` (opens at its saved size, follows the room, zooms with ctrl+wheel, keep-above and opacity applied to the real window).


Nisugi:

Reviewed head for #1655: `28cd856f`.



**Recommendation: request changes before enabling WebUI by default.** The default assignment itself is straightforward; the release gate is the unresolved behavior in the layers below.

| Finding | Priority | Owning PR | Result |
| --- | --- | --- | --- |
| F1 | P1 | #1653 | Commit exception kills the shim worker and strands synchronous callers |
| F2 | P1 | #1652 | Master-password mutation can write after launcher close |
| F3 | P2 | #1648 | Modal close races attachment destruction and raises NoMethodError |
| F4 | P2 | #1654 / #1650 | Shim tree expansion disappears on refresh |
| F5 | P2 | #1650 | Retry records retain cleared passwords beyond expiry and can replay them |
| F6 | P2 | #1652 | Favorite lookup chooses another custom launch entry |
| F7 | P2 | #1648 | New-window settings can be applied to an existing title match |

Full reproduction details and fix directions are posted on the owning PRs. The earlier corrections are substantive, but cancellation, modal cleanup, retry lifetime, and shim expansion are not yet fully closed. No additional standalone findings were identified for #1649 or #1651 in this pass.

### Validation and scope

mrhoribu (commented):
# PR Review: feat(all): the WebUI is the default launcher; GTK stays behind --gtk (lich-5#1655)

**Reviewed:** 2 files, +10/−1 at `28cd856f` on 2026-09-17. Based on: full clone.
**Verdict:** Approve the change; **hold the merge** until #1657's Majors are resolved.
**Stack position:** no longer the top of the stack — #1657 (fixes) and #1658 (YARD docs) now sit above it.
**Summary:** One constant and one spec. `LauncherChoice::DEFAULT` goes `:gtk` → `:webui`. There is nothing wrong with the diff; the entire question is whether the ten thousand lines beneath it are ready, and two Major findings say not quite yet.

## Findings

None in this diff.

The change is exactly what it claims:

```ruby
-    DEFAULT = :gtk
+    DEFAULT = :webui
```

`CHOICES` still contains both, `FLAGS` still maps `--gtk`, and the persisted `launcher` setting still overrides the default — so `--gtk` and the launcher's own toggle remain working escape hatches, and reverting is genuinely this one line. The added spec pins all three behaviours (default is `:webui`, a bare `--login` resolves `:webui`, `--gtk` still resolves `:gtk`) rather than only the constant, which is the right level to assert at.

## Why hold the merge

This PR is the gate: below it, a defect affects people who opted in with a flag; above it, a defect affects everyone who runs `ruby lich.rbw`. That changes the severity of two findings elsewhere in the stack, both in #1657:

1. **`sync` can still park for ever** (#1653's layer, `session.rb:346`) — the F1 fix is incomplete and I have a deterministic reproduction. Reachable from any script using `GLib::Timeout.add` that is stopped mid-tick.
2. **`--webui-no-browser` hands out a URL that expires in 60 seconds** (#1657's new remote-play feature, `server.rb:27`) — against a documented workflow that cannot complete in 60 seconds.

Neither is caused by this PR and neither is large. But this PR's own description sets the bar — "it depends on everything below it in the stack, in particular the two outside reviews... fixed in one PR on top of this one" — and one of those seven fixes did not land completely. Merging the flip with F1 still open would make an unbounded hang reachable by default.

The prior review's closing line applies unchanged: *"a green unit/client suite should not be taken as sufficient evidence for the default flip while F1–F7 remain unresolved."* Six of the seven now are.

## What this PR gets right

Keeping the flip to one constant, with the rollback being that same constant, is the correct way to ship a default change of this size. So is leaving the core/GTK boundary check as a warning rather than promoting it to a gate in the same PR — the flip and the deletion of GTK are separate decisions and are kept separate.

Pairing the flip with a native `;map` rewrite (elanthia-online/scripts#2479) that *refuses an older Lich with a message* rather than failing is the right courtesy for users who update the scripts before the engine.

## Coverage notes

`launcher_choice_spec` passes. The whole-repo suite at the stack tip (`e945baf0`, #1658) is green on this box: 8,717 examples, 0 failures, 1 pending (the jsdom harness, skipped because `npm` is absent). The 55 failures the description reports are Windows-specific and do not occur on Linux.

PR #1657, opened 2026-09-17
fix(all): the second outside review of the stack, F1-F7

Nisugi:
On top of `webui/default-flip`: the seven findings of the second outside review of the stack (`docs/webui-pr-review-1634-1648-1655.md`, against #1648 `dba90bf2` through #1655 `28cd856f`), fixed and pinned. One PR rather than a fix per layer, so nothing below is re-pushed; `docs/webui-review-ledger.md` part 5 answers each finding and names the layer it belongs to.

## The findings

| | layer | fix |
|---|---|---|
| F1 P1 a commit that raised killed the session thread and stranded a queued `sync` | L6a | the batch's commit is inside the same rescue as its jobs; the loop reports and continues; a synchronous job is refused if the thread ever ends with it queued; `enqueue` answers whether it took the job |
| F2 P1 a master-password change queued behind other work ran after close | L5 | the catalog call runs under `commit(operation)` like every other irreversible step |
| F3 modal close raced the detach context | L1 | `detach` builds both lifecycle jobs before dispatching either |
| F4 shim tree expansion vanished on the next render | L3 + L6b | a tree store binds `row_toggle`; `expand_all`, `collapse_all`, `expand_row`, `collapse_row` and `row_expanded?` are real, pushed per viewer through `page.set(cid, "expanded:<row>", bool, viewer:)` |
| F5 event records had no effective expiry and kept cleared passwords | L3 | every send and every received message sweep expired records; a refusal is aged before it is honoured; a closed page, a cleared control and a dropped socket drop theirs; capped at 256 |
| F6 the favorite could land on a sibling entry | L5 | the catalog answers an upsert with the key it wrote; `set_favorite` sets rather than toggles |
| F7 title discovery adopted a pre-existing window | L1 | windows already carrying the title are listed before the browser opens and excluded |

Also here, found after the review: **every handler exception in every native WebUI script was discarded**. The dispatcher's rescue logged only the exception class, through a logger Lich never supplied (`Lich::WebUI.service` built the service without one, and the runtime's default dispatcher had none). The log now gets the message, class, script frame and backtrace, a Script owner hears it through `respond` as the shim's handlers already did, and the service, runtime and dispatcher write to `Lich.log`.

## The third review, of the whole stack

`docs/webui-pr-review-stack-combined-2026-09-17.md` re-verified F1–F7 and found two Majors here; ledger part 6 answers it finding by finding.

- **Major 1**, F1 left `enqueue`'s closed check and push as two steps: the check, the push and the thread start are one step under the lock `shutdown` takes; the spec pauses the real push exactly there.
- **Major 2**, a `--webui-no-browser` launch URL died in sixty seconds with a bare 403: a URL the player has to carry lives ten minutes (`REMOTE_LAUNCH_TOKEN_LIFETIME`), and an expired or reused link says so and names `Lich::API.webui_launch_url`.
- The remote-play doc says one tunnel per session, not per script window; a busy `--webui-port` names the port and the way out; the child's `--webui-no-browser` comes through the launch context, not `ARGV`; the three chained continuation lines RuboCop 1.91 flags are gone; `yard` is in the development group.
- Open questions answered in the ledger: the `adopt` pid check is deliberate (a shared-profile window is owned by the Chrome already running; the exclusion list is the evidence); a replay now keeps its record's scope so `clear_sensitive` finds it, with a harness case; shim expansion is per page because D26 holds a shim page to one viewer, resume token included.

**Also here, reported by Tysong and scoped by the reviewer:** a launcher flag beside a session opened the launcher. The R1 fix read `--gtk` / `--webui` as "open the launcher" when it means "select which launcher", so Saga's `<file>.sal --gtk ...`, a `--game=HOST:PORT` proxy and a force mode all opened the launcher instead of connecting. `LauncherChoice.launcher_only?` (empty command line, or launcher flags only) or `--gui` is now the one rule for both branches: `main`'s behaviour plus exactly the bare-flag case R1 asked for. The routing spec covers Saga's real argument list, a proxy, a force mode and a headless port under both toolkits.

Each new spec was run against the pre-fix code and fails there. All three review documents are committed under `docs/`.

## Acceptance

WebUI, shim and launcher surfaces: 648 examples, 0 failures. Client harness: 23 cases, all passing. RuboCop clean. Whole suite on the same tree: 8713 examples, 55 failures, all the known environment cases that fail on `main` on this box.


mrhoribu (commented):
# PR Review: fix(all): the second outside review of the stack, F1-F7 (lich-5#1657)

**Reviewed:** 32 files, +1,346/−83 at `19d2a7be` on 2026-09-17. Based on: full clone; full suite, full RuboCop, and two targeted probes against production code executed on Linux / Ruby 4.0.7.
**Verdict:** **Request changes**
**Stack position:** #1658 (YARD documentation, comment-only) now sits on top of this PR and documents much of the code the fixes below would change.
**Summary:** Six of the seven findings are genuinely fixed, several more thoroughly than required. **F1 is not** — the same "dead worker, stranded `sync`" failure is still reachable through a narrower window, and I have a deterministic reproduction. Separately, this PR carries an entire feature no review has seen (`--webui-port` / `--webui-no-browser` remote play), and that feature ships a 60-second credential against a documented workflow that cannot complete in 60 seconds.

## Fix verification

| | Layer | Verdict |
|---|---|---|
| F1 | #1653 | ⚠️ **Partially fixed** — Major 1 below |
| F2 | #1652 | ✅ Fixed — `commit(operation)` wraps the catalog call; return-value plumbing traced and correct |
| F3 | #1648 | ✅ Fixed — `lifecycle_job` captures context, `detach` builds both before dispatching either |
| F4 | #1654 + #1650 | ✅ Fixed — shim binds `row_toggle`, runtime stores per viewer, `model_changed!` prunes stale keys |
| F5 | #1650 | ✅ Fixed — swept on send *and* receive, aged before honoured, capped, dropped on close/clear/disconnect |
| F6 | #1652 | ✅ Fixed — weaker `find_entry_key` **deleted**; upsert answers with the key it wrote |
| F7 | #1648 | ✅ Fixed for the reported repro — pre-existing titled windows excluded (one sub-point open) |

## Major

### 1. The F1 fix does not close F1 — `sync` can still park for ever — `lib/common/script_scope/gtk/session.rb:346`

```ruby
  job ||= block
  raise ArgumentError, 'block required' unless job
  return nil if @closed

  @queue << job
```

The `@closed` check (349) and the push (351) are still two separate, unsynchronized steps — the exact structure this PR's own comment at line 385 says it replaced: *"The closed check and the enqueue used to be two steps, and a shutdown between them queued a job nobody would ever run."* `sync` was changed to trust `enqueue`'s answer; `enqueue` was not made atomic. `shutdown` (753) sets `@closed` and pushes `:stop` without taking `@thread_mutex`, and the new `ensure reject_queued_jobs` (818) drains the queue exactly once, on the way out.

So:

1. A script thread enters `enqueue`, reads `@closed` → false.
2. `shutdown` runs: `@closed = true`, `@queue << :stop`.
3. The session thread pops `:stop`, leaves the loop, and `reject_queued_jobs` drains an **empty** queue. The thread ends.
4. The script thread pushes its `SyncJob` and calls `ensure_thread`, which returns immediately because `@closed` is true (787).
5. `enqueue` returns `true` — asserting it took the job — and `sync` blocks on `done.pop` (391), which has no timeout.

Reproduced deterministically against a real `Session`, pausing only inside the production `enqueue` between its two steps:

```
closed=true
worker_alive=false
queue_size=1
sync_result=:STILL_WAITING
```

The new spec at `spec/lib/common/script_scope/gtk_review_fixes_spec.rb:228` does not reach this ordering — it kills the thread *after* the job is queued, which is the case `reject_queued_jobs` handles. So the spec passes and the hole remains.

Production reachability: `GLib::Timeout.add` spawns a source thread looping `sleep(interval)` then `session.sync { block.call }` (`glib.rb:47`), and script exit fires `@owner.at_exit { session.shutdown }` (`session.rb:833`) from another thread. Any script with a repeating timer stopped mid-tick can hit it. The parked thread never reaches its `ensure`, so `GLib.remove_source(id)` never runs and the source leaks until Lich tears the script's threads down.

Major rather than Blocker because it is strictly narrower than the original F1, which also killed a live session's worker. But the PR claims F1 closed, and it is not.

**Fix:** one lock around the `@closed` read and the `@queue <<`, taken also by `shutdown` and `reject_queued_jobs`, so "took the job" and "a thread will drain it" are decided together. `SyncJob#reject` — which this PR already added — is the right primitive; it just needs to be unmissable.

### 2. `--webui-no-browser` hands the player a URL that is dead in 60 seconds — `lib/webui/server.rb:27`

```ruby
LAUNCH_TOKEN_LIFETIME = 60
```

The launch URL carries a one-shot token with a 60-second monotonic lifetime; `handle_auth` (352) deletes it on use and answers a bare `403 Forbidden` after expiry, with no body explaining why. Confirmed against the real server over HTTP:

```
TTL constant = 60 seconds
after 61s of wall clock: 403 Forbidden body="Forbidden"
fresh URL:               302 Found -> /
```

Sixty seconds is the right lifetime when Lich opens the browser itself — the URL is consumed in milliseconds. It is the wrong lifetime for the feature this PR adds, whose whole premise is that the player is on another machine. `docs/webui-remote-play.md` instructs:

> Script windows in a game session use their own ephemeral port, so each one needs its own `-L`. Add them to the same SSH command, or **run a second tunnel when the URL appears in the game window.**

Noticing the URL, switching to a terminal, running `ssh -L`, and pasting into a browser will usually exceed 60 seconds. The failure is opaque and the doc names no recovery. There is one, undocumented: `Lich::API.webui_launch_url(page: …)` (`lib/api/webui.rb:30`) mints a fresh URL.

**Fix (any one):** a longer lifetime on the no-browser path than the browser path; a `;` command that re-announces the URL, documented; or at minimum a 403 body that says the link expired and how to get another.

## Minor

### 3. The remote-play doc is wrong about how many tunnels are needed — `docs/webui-remote-play.md`

> Script windows in a game session use their own ephemeral port, so each one needs its own `-L`.

They do not. There is one WebUI server per Lich **process**, and script windows are *pages* on it — `Session#service` resolves to the process-wide singleton (`lib/common/script_scope/gtk/session.rb:326`: `@service ||= Lich::WebUI.service`). Every script window in a session is reachable through that session's single port, so **one `-L` per session** covers all of them. As written, the doc tells a remote player to build N tunnels for N windows.

### 4. "RuboCop clean" is not accurate — 4 offenses added by the stack

Running the repo's own CI command (`bundle exec rubocop`, whole repo): `main` → 73 offenses; this tip → 77. The four new ones are all `Layout/MultilineMethodCallIndentation`, all autocorrectable:

| File:line | From |
|---|---|
| `spec/lib/common/webui_launcher/catalog_integration_spec.rb:126`, `:132` | #1651 |
| `spec/lib/common/script_scope/gtk_review_fixes_spec.rb:354` | #1653 |
| `lib/common/script_scope/gtk/widgets_data.rb:1598` | #1654 |

(A fifth apparent delta, `lib/common/script.rb:2866`, is `main`'s pre-existing offense at `:2852` shifted by added lines — not new.)

CI runs RuboCop repo-wide and exits non-zero, so it is **already red on `main`** and this does not newly break it. `rubocop -A` clears all four. The finding is about the claim.

### 5. A busy `--webui-port` gives a misleading recovery hint — `lib/common/webui_launcher.rb:142`

```ruby
@recovery.call("WebUI launcher unavailable: #{error.class}. Retry with the GTK launcher or abort safely.")
```

`TCPServer.new` on a taken port raises `Errno::EADDRINUSE`, which lands here. The player sees *"WebUI launcher unavailable: Errno::EADDRINUSE. Retry with the GTK launcher or abort safely."* — no port number, and advice that is wrong (the fix is another port, not abandoning the WebUI). A fixed port is the one most likely to collide, since a leftover Lich holds it. Inconsistent with the care right next door: `handle_webui_port` (`lib/main/argv_options.rb:187`) dies with *"--webui-port must be from 1 through 65535, got N"*.

## Nits

### 6. `--webui-no-browser` propagation reads raw `ARGV` — `lib/common/session_launcher.rb:144`

```ruby
flags << '--webui-no-browser' if ARGV.any? { |argument| argument.to_s.casecmp?('--webui-no-browser') }
```

The line directly above resolves its flag through `resolve_launcher(context)`; this one reaches into global `ARGV`. It works — nothing clears `ARGV` wholesale — and it has a spec. But two adjacent flags now have different plumbing, and a caller building a context without a matching `ARGV` gets one and not the other.

## Scope note

`--webui-port`, `--webui-no-browser`, `lib/webui/options.rb`, the help text and `docs/webui-remote-play.md` are **a new feature, not a fix for F1–F7**. That is a fair question to raise on a PR titled "the second outside review of the stack, F1-F7": the feature arrives in the one PR nobody has reviewed against a prior baseline, on top of a stack that is about to become the default, and findings 2, 3, 5 and 6 are all in it. Splitting it into its own PR would cost little and would let the F1–F7 fixes merge on their own evidence.

## Open questions

- **`adopt` still compares requested pid to requested pid** — `lib/webui/presented_window.rb:79`. F7 made two points; the exclusion list answers the first, and the second is untouched. **Settles it:** deliberate, or missed? Windows-only, so I could not test it.
- **A replayed record can carry a submission outside its own `scope`** — `lib/webui/assets/app.js:171`. On a replay `scope` is recomputed from the *current* render; if the component was rebuilt with no `submissions` entry, the record carries the original password with `scope: []` and `clear_sensitive`'s `dropPending` cannot match it. The 30s TTL still bounds it. **Settles it:** can a page re-render between a send and its stale refusal in a way that drops that cid's entry?

## What this PR gets right

**The suite is genuinely green.** `bundle exec rspec` here: **8,717 examples, 0 failures, 1 pending**. The 55 failures the description reports are Windows-specific; on Linux there are none. The claim understates the result.

**Four of the six real fixes remove the mechanism rather than guarding it**, which is the difference between a fix that holds and one that rots: F6 *deletes* the duplicated `find_entry_key` instead of strengthening it; F3 turns `enqueue_lifecycle` into a context-capturing `lifecycle_job` so the ordering cannot recur; F2 routes the last unguarded mutation through the `commit` arbiter every other path already used; F5 sweeps on every send and receive rather than adding a second special case.

**F4 is fixed beyond what the finding required** — `model_changed!` now prunes `@expanded_keys` alongside `@selected_keys`, which the review never asked for but which is the bug you would have filed next.

**Committing both review documents and a ledger that answers each finding** made this review far cheaper and sharper. It is also what let me confirm six fixes quickly instead of re-deriving them, and what made F1's gap findable. Keep doing this.

**The refusal to offer `--bind-address` for the WebUI** (`docs/webui-remote-play.md`, "What is deliberately not offered") is the correct call, correctly reasoned, and written down with its reasoning. Finding 2 is a complaint about a timeout, not about that stance.

## Coverage notes

**Ran:** full `rspec` (8,717 examples, 0 failures) and full `rubocop` on this tip, plus `rubocop` on `main` for a baseline; targeted suites (`spec/lib/webui`, `spec/lib/common/webui_launcher*`, `spec/lib/common/script_scope`, `spec/lib/main`) at 752 examples, 0 failures; two probes against production code — the `enqueue`/`shutdown` race and the launch-token expiry over real HTTP.

**Did not run:** the 23-case jsdom client harness (`npm` not installed), so the client halves of F4 and F5 are reviewed as code only; all Windows-only code, so F7's fix and `window_presentation.rb` are unexecutable here; any live game, browser, or SSH-tunnel acceptance.

Gemfile.lock
```diff
@@ -314,6 +316,7 @@ CHECKSUMS
   webrick (1.9.2) sha256=beb4a15fc474defed24a3bda4ffd88a490d517c9e4e6118c3edce59e45864131
   win32ole (1.9.3) sha256=01f43dc5dc13806e6e58204f538b4a28f3d85968ea89074abc9a3cd118e94d96
   yaml (0.4.0) sha256=240e69d1e6ce3584d6085978719a0faa6218ae426e034d8f9b02fb54d3471942
+  yard (0.9.45)
```
mrhoribu: yard (0.9.45) sha256=52e211493f7cb8a3ebf7e104a25a1e73937a3103092545d34cb88fafebb3dc51

PR #1658, opened 2026-09-17
docs(all): YARD documentation for the WebUI stack

Nisugi:
On top of `webui/review-2-fixes` (#1657): YARD documentation for everything the WebUI stack adds. Comment-only, one commit, 46 files.

## What is documented

Every module, class, public and protected method in:

- `lib/webui/**` and `lib/webui.rb`, `lib/api/webui.rb` (runtime, server, validator, contract, adapter, viewer store, dispatcher, modal coordinator, window presentation, and the rest)
- `lib/common/webui_launcher.rb` and `lib/common/webui_launcher/*` (launcher, catalog, executor, geometry store)
- `lib/common/script_scope.rb`, `lib/common/launcher_choice.rb`
- `lib/common/script_scope/gtk/*` (the GTK compatibility shim: every widget class says which GTK class it stands in for and how it differs)
- `lib/common/authentication/*`, `lib/common/frontend_editor.rb`, `lib/common/frontend_choices.rb`, `lib/util/heap_compaction.rb`

Each block is a one-sentence summary, then the reasoning comment that was already there, kept verbatim (the review references included), then `@param`, `@return`, `@raise`, `@yield` tags; `@example` only where it clarifies. Private methods get a sentence where they had none. Metaprogrammed methods (`def_setter` families, the tree builder's typed components) are covered with `@!method` and `@!macro` directives; `attr_*` with `@!attribute`.

The only non-comment change: two combined `attr_reader` lines are split so each reader carries its own `@return`.

After the third review: every documented private method carries `@api private` as `docs/YARD-STYLE-GUIDE.md` requires (172 of them), `yard` is in the bundle, and `bundle exec yard doc` was run over the tree: the only warnings inside the stack's files were a prose line YARD read as a `@vexpand` tag and an `attr_writer` `@param` named for the wrong argument, both fixed; the `@!method` and `@!macro` directives produce no warning.

## Numbers

`yard stats` over `lib/`, before and after: 56.9% to 68.6% documented, undocumented methods 1959 to 1288. The remaining undocumented objects are outside the stack's files, or private.

## Checks

RuboCop clean over the whole tree. WebUI, shim, launcher, services and frontend specs: 1150 examples, 1 failure, the known Windows quoting case in `launch_data_spec` that fails on `main` on this box.


mrhoribu (commented):
# PR Review: docs(all): YARD documentation for the WebUI stack (lich-5#1658)

**Reviewed:** 46 files, +5,625/−46 at `e945baf0` on 2026-09-17. Based on: full clone; full suite, full RuboCop, and three mechanical doc-integrity checks executed on Linux / Ruby 4.0.7.
**Summary:** A YARD sweep over everything the stack adds. **The "comment-only" claim is true, and I verified it mechanically rather than taking it on trust** — across all 46 changed Ruby files there are exactly two code deltas, both the `attr_reader` splits the description discloses. Documentation quality is high. Three Minors, none of them about the prose: a repeated inaccurate RuboCop claim, a missing tag the repo's own style guide requires, and two new tags that formalize a contract the stack does not currently honour.

## Verification: is it really comment-only?

This is the one claim worth checking hard on a docs PR, because a behaviour change hidden in 5,625 lines of comments is exactly the thing review misses. I compared the Ruby **token stream** of every changed file at `pr-1657` and `pr-1658`, with comment, whitespace and newline tokens removed — a position-independent "same code, whatever the comments say" test.

```
files compared: 46
files whose CODE tokens changed: 2
```

Both are the disclosed split, and both are semantically identical — the token *count* is unchanged in each file (1114 → 1114, 965 → 965), because a `,` was replaced by a repeated `attr_reader`:

```ruby
-          attr_reader :fraction, :text
+          # @return [Float] the current fraction, 0.0 to 1.0
+          attr_reader :fraction
+          # @return [String, nil] the bar's text, or nil when none was set
+          attr_reader :text
```

`lib/common/script_scope/gtk/containers.rb` and `lib/common/script_scope/gtk/degradation.rb`. Nothing else in the PR changes code. The full suite agrees: **8,717 examples, 0 failures, 1 pending** — byte-identical to the result at `pr-1657`.

The probe is attached as `probe-comment-only.rb`; it is worth keeping for future doc sweeps.

## Minor

### 1. "RuboCop clean over the whole tree" is not accurate — and this is the second PR in a row to claim it

Running the repo's own CI command (`bundle exec rubocop`, whole tree) at this head:

```
1302 files inspected, 77 offenses detected, 77 offenses autocorrectable   (exit 1)
```

**Methodology (verified, not assumed):** `bundle exec rubocop` run from the repo root with no flags, which resolves `/home/tysong/claude-stuff/lich-5/.rubocop.yml` (confirmed via `--debug`: *"configuration from .../.rubocop.yml"*), RuboCop 1.91.0 from the bundle, the repo's custom `Custom/AsciiOnlySource` cop registered and enabled (confirmed via `--show-cops`). **No PR in the stack modifies `.rubocop.yml`**, and no `.rubocop_todo.yml` exists, so the `main`-vs-tip comparison is like for like. Per-cop breakdown:

| Cop | `main` | tip |
|---|---|---|
| `Layout/ExtraSpacing` | 68 | 68 |
| `Layout/MultilineMethodCallIndentation` | 4 | **8** |
| `Lint/RedundantCopDisableDirective` | 1 | 1 |
| `Custom/AsciiOnlySource` | 0 | 0 |

The entire delta is four `Layout/MultilineMethodCallIndentation` offenses; every other count is identical.

The offense set is **identical** to `pr-1657`'s — a `diff` of the two sorted outputs shows a single line, and it is a line-number shift, not a new offense (`widgets_data.rb:1598` → `:2671`, because this PR added ~1,073 lines of documentation above it). The four offenses the stack introduced over `main` (73 → 77) are all still there:

| File:line at this head | From |
|---|---|
| `spec/lib/common/webui_launcher/catalog_integration_spec.rb:126`, `:132` | #1651 |
| `spec/lib/common/script_scope/gtk_review_fixes_spec.rb:354` | #1653 |
| `lib/common/script_scope/gtk/widgets_data.rb:2671` | #1654 |

All are `Layout/MultilineMethodCallIndentation` and all clear with `rubocop -A`. CI is already red on `main` (73 offenses), so this does not newly break anything — the finding is that the claim has now been made twice, in #1657 and here, and is wrong both times. One `rubocop -A` in this PR would make it true and close the finding for the whole stack.

### 2. 163 documented private methods, none carrying `@api private` — the repo's own style guide requires it

`docs/YARD-STYLE-GUIDE.md` defines four tiers, and for the private tier it is explicit:

> ### Private (Developer Documentation)
> Methods marked `private` or documented with `@api private`.
> **Required tags**: summary line, `@param`, `@return`, `@api private`

Across the 46 changed files I count **163 documented methods sitting below a bare `private`, and `@api private` appears zero times** — not once in the entire PR. Samples: `lib/common/script_scope/gtk/session.rb:223` `#declare_facilities`, `lib/common/frontend_editor.rb:253` `#update_builtin`, `lib/common/authentication/windows_credential_manager.rb:212` `#string_to_wide`.

This matters more than a normal style nit because the tag is not cosmetic: `@api private` is what keeps those methods out of the generated public documentation. Without it, a sweep whose stated purpose is a published reference will publish 163 internal methods as though they were API. Given the PR is otherwise meticulous about tags, this reads as an oversight rather than a decision.

(Probe attached as `probe-api-private.rb`.)

### 3. Two new tags formalize a contract the stack does not currently honour — `lib/common/script_scope/gtk/session.rb`

```ruby
# @return [true, nil] true when queued, nil for a closed session
...
# @raise [Lich::WebUI::Error] when the session has been shut down
def sync(&block)
```

Both tags describe the *intent* of #1657's F1 fix accurately. Neither describes what the code does, because that fix is incomplete — as I reproduced in my review of #1653/#1657, `enqueue` can return `true` for a job queued into a queue no thread will ever drain, and `sync` can then park for ever on `done.pop` instead of raising:

```
closed=true   worker_alive=false   queue_size=1   sync_result=:STILL_WAITING
```

This is not this PR's bug, and I am not asking it to fix F1. The point is narrower and worth one line of attention: writing a behaviour into a `@return`/`@raise` tag converts it from an implementation detail into a published promise that future callers are entitled to rely on. **Re-check these two tags when F1 is properly fixed** — if the fix lands as suggested (one lock across the closed-check and the push), both tags become true exactly as written and nothing needs changing. If it lands differently, they will need to.

## Nit

### 4. `yard` is not in the Gemfile, so the stats claim can't be reproduced

The repo has a `.yardopts` and a committed style guide, but `yard` appears in neither `Gemfile` nor `Gemfile.lock`, so `bundle exec yard` fails outright:

```
can't find executable yard for gem yard. yard is not currently included in the bundle
```

That makes the PR's headline numbers — "56.9% to 68.6% documented, undocumented methods 1959 to 1288" — unverifiable by a reviewer using the repo's own bundle; you'd need a global `gem install yard`. I am not asking for CI enforcement, since the style guide deliberately rules it out ("**Enforce through PR review** — No CI enforcement"). Adding `yard` to the development group would just let the next reviewer check the claim in one command.

## Open question

- **Are the `@!method` / `@!macro` directives covering the metaprogrammed setters actually correct?** The description says the `def_setter` families and the tree builder's typed components are documented with `@!method` and `@!macro`. Directives like these are the one part of a YARD sweep that a token diff and a signature check cannot validate — they declare methods that do not exist in the source, so nothing in my mechanical checks touches them. **Settles it:** a single `yard doc` run with `yard` installed, eyeballing that the generated `Gtk::Widget` page lists the setters that actually exist. That also closes Nit 4.

## What this PR gets right

**The documentation is accurate where I could check it mechanically.** I extracted every documented method in the 46 files and compared each `@param` name against the real signature, handling multi-line signatures:

```
documented methods checked: 683
@param name mismatches: 0
```

Zero. For a 5,625-line hand-written sweep that is a genuinely good result, and it is the single best evidence that the prose was written against the code rather than around it.

**It documents warts instead of smoothing them.** `Catalog#toggle_favorite` gets `@return [Boolean, nil] the new favorite status; false also when the key does not resolve, nil when the write failed` — which is precisely the return-value ambiguity I noticed in that method while reviewing #1652 and chose not to file. A sweep that surfaces an awkward contract in a tag, rather than writing a tidier tag than the code deserves, is doing the job properly.

**Not one non-ASCII character in 5,625 lines of prose.** The repo enforces ASCII-only source through a custom RuboCop cop, and a hand-written documentation sweep of this size is exactly where smart quotes, em-dashes and arrows creep in. A byte-level scan of every file this PR touches found none, and `Custom/AsciiOnlySource` reports zero offenses across the whole tree. The existing comments' `--` for em-dashes is carried into the new prose consistently.

**Keeping the existing reasoning comments verbatim**, review references included, is the right call. Those comments are the most valuable thing in the stack — they are why the three review passes could tell a scar from a defect — and a sweep that paraphrased them into summary lines would have destroyed that.

**The one code change is disclosed, minimal, and safe.** Splitting two `attr_reader` lines so each carries its own `@return` goes slightly beyond the style guide, which lists `attr_reader` under "Skip (No Documentation Needed)". That is doing more than required, not less, and the token diff proves it changed nothing.

## Process note

The style guide's first principle reads: *"Document as you touch — Add YARD docs when creating or modifying code. **No retroactive documentation sweeps required.**"* This is a retroactive documentation sweep, landing eleventh on a stack that is currently blocked on two Majors in #1657.

That is not a defect and the documentation is welcome. But it adds 46 files of rebase surface above the two PRs that still need to change, and those changes are in `session.rb` and `webui_launcher.rb` — both heavily documented here. Landing the F1 fix first, or rebasing this on top of it afterwards, will be less work than doing it the other way round.

## Coverage notes

**Ran:** full `bundle exec rspec` at this head (8,717 examples, 0 failures, 1 pending — identical to `pr-1657`, confirming no behaviour change); full `bundle exec rubocop` (77 offenses, exit 1) with a `diff` against the `pr-1657` and `main` offense sets; and three mechanical doc checks — token-stream comment-only verification (46 files), `@param`-versus-signature (683 methods), and `@api private` tier compliance (163 private methods). All three probes are attached.

**Did not run:** `yard` itself, so the coverage-percentage claim and all `@!method` / `@!macro` directives are unverified; the jsdom client harness (`npm` absent); any Windows-only code. I also did not read all 5,625 lines of added prose for factual accuracy — the mechanical checks cover tag/signature agreement, not whether each summary sentence describes its method correctly. Spot checks in the areas I knew well from reviewing the rest of the stack (session, catalog, launcher, options) were accurate.

PR #1660, opened 2026-09-18
feat(all): Map.boundary_rooms_for(ids), the perimeter of any set of rooms

Nisugi:
`Map.boundary_rooms_for(ids)`: the perimeter of any set of rooms, on both the GS and DR map classes.

`CreatureTemplate#boundary_rooms` had the only copy of this walk and it was bound to the creature's own rooms. The walk moves to `MapBase` as a class method over any enumerable of room ids; the template hands it its rooms and keeps the memo.

## Definition change

The walk is now forward-only: every room an exit of the set leads to, minus the set. The old version also scanned every room on the map to add rooms with a one-way edge *into* the set. That pass is gone, for two reasons:

- `wayto` is player movement. A room you can arrive from but cannot step out to is not an edge you can hold, hunt or avoid.
- The scan cost the size of the map on every call. The new walk costs the size of the set, so it is safe to call as a room list is being edited.

Nothing in lib or scripts consumed the old definition. eohunter's geography computes incoming and outgoing boundaries itself, as two lists. If an inbound direction is ever wanted it belongs on this method as a `direction:` keyword backed by a reverse-edge index, not as a second name.

## Checks

Map spec: the five-room topology asserts the both-ways neighbour is a boundary and the one-way-in room is not, a `Set` with an unmapped id returns a sorted unique list, and an empty set returns empty. The creature spec now checks delegation and memoization rather than re-testing the walk. RuboCop clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated creature boundary-room detection to include only rooms reachable through exits from the creature’s rooms.
  - Excludes rooms connected only by incoming, non-traversable paths.
  - Boundary-room results are now consistently unique and sorted.
  - Unknown room IDs and empty room sets are handled safely without producing boundary rooms.


mrhoribu (approved):
# PR Review: feat(all): Map.boundary_rooms_for(ids), the perimeter of any set of rooms (elanthia-online/lich-5#1660)

**Reviewed:** 4 files, +54/-24, at `2338e198` on 2026-09-21. Based on: full clone (worktree checkout of the PR head against `main`), with the whole repo and its history available.
**Verdict:** Approve
**Summary:** Moves the perimeter walk out of `CreatureTemplate#boundary_rooms` into `MapBase.boundary_rooms_for(ids)` as a class method usable by any room set on both GS and DR map classes, and narrows the definition from bidirectional to forward-only (exits only). The change is small, self-contained, has no other callers in lich-5, dr-scripts, or scripts to break, and is backed by tests that actually exercise the new boundary condition (the dropped one-way-in room).

## Blockers
None.

## Major
None.

## Minor
None.

## Nits

### 1. `outside` is misnamed before the subtraction — `lib/common/map/map_base.rb:565`

```ruby
outside = inside.flat_map { |id| (self[id]&.wayto || {}).keys.map(&:to_i) }
(outside - inside).uniq.sort
```

At the point `outside` is assigned, it still contains every raw exit target, including ids that are themselves inside the set (e.g. two rooms in the set point at each other). The name reads as already-filtered. Something like `edges` or `reachable` would save a reader the trip to line 566 to confirm it isn't filtered yet. Purely cosmetic — the logic on the next line is correct.

## Open questions
None. The one thing worth double-checking before merge — whether any script outside this checkout calls `boundary_rooms` or the old bidirectional shape — is answered: `boundary_rooms` was added in the same feature wave 10 days earlier (`f86104f4`, 2026-09-07) and never shipped with a consumer in dr-scripts or scripts (both searched, no hits), so there's nothing depending on the removed reverse-edge pass.

## What this PR gets right

- **Correct blast-radius reasoning, verified.** `boundary_rooms_for` lands in `MapBase::ClassMethods`, which both `map_gs.rb` and `map_dr.rb` pull in via `include MapBase` — so the "on both the GS and DR map classes" claim in the PR body is literally true, not aspirational, even though only the GS `CreatureTemplate` consumes it today.
- **The behavior change is exercised, not just asserted.** `spec/lib/common/map/map_base_spec.rb` builds a 6-room topology that specifically includes a room with only a one-way edge *into* the set (room 5) and asserts it's excluded — that's the one line in the PR description that could have silently regressed into "no functional change," and the test would catch it if it did.
- **`creature_spec.rb` was correctly re-scoped**, from re-testing the walk to testing delegation and memoization (`have_received(:boundary_rooms_for).once`) against a stubbed `Map`. That's the right division of labor now that the walk itself lives and is tested in `map_base_spec.rb`.
- **`self[id]&.wayto || {}` cleanly folds two nil checks** (unmapped id, nil `wayto`) that the old code handled as two separate `or next` / `|| {}` branches — genuine simplification, not just relocation.
- Docstring on `boundary_rooms_for` states the cost model (`O(set)` vs. the old `O(map)`) as a design fact, which is exactly the kind of non-obvious rationale worth commenting.

## Coverage notes

- Full local worktree of the PR head against `main`; not diff-only.
- Ran `bundle exec rubocop` on all 4 changed files: clean, matches the PR's claim.
- Ran the two directly affected spec files (`map_base_spec.rb`, `creature_spec.rb`): 238 examples, 0 failures.
- Ran the full repo suite (`bundle exec rspec`): 8027 examples, 0 failures — no ripple effects elsewhere in lich-5.
- Grepped dr-scripts and scripts (full checkouts) for `boundary_rooms` and `bigshot` (the perimeter-list naming homage in the comments): no script consumes `boundary_rooms`/`boundary_rooms_for` directly, and `bigshot.lic` doesn't reference either — confirms the PR body's "nothing in lib or scripts consumed the old definition" claim.
- Ran a CodeRabbit pass (`coderabbit review --agent --base main --committed`) against the checkout: 0 findings across all 4 changed files, consistent with this review.
- Hand-traced both new spec cases (`boundary_rooms_for([1,2,3])` and `boundary_rooms_for(Set[3,7,1])`) against the implementation to confirm expected outputs are actually correct, not just self-consistent with the code.
- Did not test DR's map class directly (no DR-side consumer exists yet to exercise it), but `boundary_rooms_for` is generic over any `MapBase` includer and the GS-side test coverage exercises the same shared class method.

PR #1661, opened 2026-09-20
feat(gs): expose exact server creature health from crtrStatus

veredox-hue:
## Summary

- retain the exact `health` and `maxhealth` values GemStone recently began supplying in `crtrStatus` snapshots
- put those values into `health` and `max_health` instead of adding a parallel `server_*` API
- make the existing `current_hp`, `max_hp`, and `hp_percent` methods prefer a complete server pair when `maxhealth` is positive
- preserve exact negative health in `health` while retaining the established non-negative `current_hp` contract
- gate legitimate `0/0`, omitted, partial, or malformed data onto the existing template/default and observed-damage fallback
- keep the inferred damage tracker running while server data is active so fallback remains available without retaining a stale exact value
- document the public API, fallback behavior, real feed examples, and the raw-capture evidence behind the snapshot policy

## Captured feed examples

These tags are reproduced from raw GemStone server captures.

A living silver-eyed sleek onyx panther:

```xml
<crtrStatus exist="357970513" health="260" maxhealth="260" challenging="1"/>
```

A dead scaly burgee with damage beyond zero:

```xml
<crtrStatus exist="360586159" health="-16" maxhealth="340" hostile="1" dead="1" prone="1"/>
```

A noncombat black-necked hooded toucan with legitimate zero values:

```xml
<crtrStatus exist="356889321" health="0" maxhealth="0" inferior="1"/>
```

## Compatibility and fallback

The existing HP pathway now uses the server as its preferred source of truth when both values are valid and `maxhealth > 0`. Scripts already using `current_hp`, `max_hp`, `hp_percent`, `low_hp?`, `dead?`, Coup de Grace eligibility, or target selection do not need to migrate to new method names.

`health` retains the exact signed server value, including negative overkill. `current_hp` clamps that value at zero, preserving its prior non-negative contract and keeping `dead?` correct for negative health. A `0/0` pair is retained in `health`/`max_health` but does not activate the server-backed combat pathway, so noncombat entities are not marked dead solely because of those values.

The observed-damage tracker continues running while server values are active. If a later complete snapshot omits or invalidates either value, the exact fields clear and the existing methods resume the continuously maintained inferred model. They deliberately do not reuse the last exact server reading, which would be stale, or rewrite `damage_taken` from snapshots, which could double-count hits because status and combat messages arrive as separate feed events. The fallback value can therefore differ at the source transition; if valid attributes return, the existing methods immediately prefer them again.

## Observed feed behavior

The snapshot policy was checked against 186 raw capture files totaling approximately 11 GB and containing 2,352,662 `crtrStatus` tags. Of those, 52,154 supplied health data:

- `health` and `maxhealth` were always supplied together
- all supplied values were complete signed or unsigned decimal integers
- 49,040 consecutive same-creature updates retained both attributes
- 83 same-creature transitions went from absent to present during the rollout
- no same-creature transition went from present to absent

The observed feed therefore supports treating the pair as part of the complete snapshot once enabled. Clearing a future omission is also the safe rollback behavior: the established HP methods fall back instead of treating stale data as the latest exact server value. If no new `crtrStatus` tag arrives at all, Lich has no event from which to clear the last received value; that limitation is documented.

## Test plan

- `bundle exec rspec spec/lib/common/creature/creature_base_spec.rb spec/lib/common/xmlparser_crtr_status_spec.rb spec/lib/gemstone/creature_spec.rb` — 157 examples, 0 failures
- relevant GemStone/Common/DragonRealms compatibility specs — 1,242 examples, 0 failures
- `bundle exec rspec` — 8,037 examples, 0 failures
- targeted RuboCop on all touched Ruby files — no offenses

The parser spec setup also explicitly resets creature configuration and stubs `begin_room_objs`; those two lines repair pre-existing cross-example isolation so the parser integration file runs reliably on its own.


mrhoribu:
@veredox-hue Are you able to resolve the merge conflict on your branch? That should allow for us to run the other PR checks that haven't ran yet. Also can you provide a sample XML of the updated `crtrStatus` feed in the PR description and also in the docs writeup? That should help document the change and help instruct future lookups on this commit once merged.

mrhoribu (changes requested):
# PR Review: Expose exact creature health from crtrStatus (elanthia-online/lich-5#1661)

**Reviewed:** 6 files, +194/-1 at `6acba9f2` on 2026-09-20 (PR branch `codex/expose-crtrstatus-health`, based on `main@4ed650ba`).
**Based on:** full clone (isolated fresh clone + fetched PR ref, kept separate from a local checkout that had unrelated uncommitted work on the same files), plus the PR's own test/rubocop run and a CodeRabbit pass.
**Verdict:** Request changes
**Summary:** Adds `health`/`max_health`/`health_percent` to the shared `CreatureBase` mixin so GemStone (and, structurally, DragonRealms later) can read the exact HP values GemStone recently started putting on `<crtrStatus>`, deliberately kept separate from the existing inferred-HP API. Solid, well-tested, and honestly documented — but it can't merge as-is (stale base, real conflict with `main`), and one design choice is worth a second look before it ships.

## Blockers

### 1. PR is based on a stale `main` and has a real, unresolved merge conflict — `lib/common/creature/creature_base.rb`

GitHub reports this PR as `CONFLICTING`, and it's not a false alarm. The PR branches from `main@4ed650ba` (a CHANGELOG-only commit), but `main` has since picked up #1617 ("Initial Lich state update and detachable-client split GS/DR"), which reworked the exact same region of `InstanceMethods` — replacing `created_at`-based registry aging with a `@last_seen_at`/`touch_seen` mechanism, added right where this PR inserts its `@health`/`@max_health` ivars and `health`/`max_health`/`health_percent` methods, and adding a `touch_seen` call as the first line of `sync_crtr_status`, which collides with this PR's `@health = strict_crtr_integer(...)` insertion at the same spot.

I confirmed this isn't cosmetic by actually rebasing the PR branch onto current `main` in an isolated clone:

```
Auto-merging lib/common/creature/creature_base.rb
CONFLICT (content): Merge conflict in lib/common/creature/creature_base.rb
Auto-merging lib/gemstone/creature.rb            (clean)
Auto-merging spec/lib/common/creature/creature_base_spec.rb   (clean)
Auto-merging spec/lib/gemstone/creature_spec.rb  (clean)
```

Only `creature_base.rb` conflicts, in exactly two hunks: the `initialize_status_tracking`/`touch_seen` block and the top of `sync_crtr_status`. The changes are not logically incompatible — both sides just touch the same lines — so the fix is mechanical: rebase onto `main`, keep `@last_seen_at`/`touch_seen` from main's side and `@health`/`@max_health` from this PR's side in `initialize_status_tracking`, and keep both `touch_seen` and the new `@health =`/`@max_health =` lines (in either order) at the top of `sync_crtr_status`. Everything else in the diff (gemstone/creature.rb, both spec files) auto-merges cleanly. Re-run the full suite after resolving, since main has moved a fair amount since this branch was cut.

## Major

### 2. The exact HP values are wired up but never used — `current_hp`/`dead?`/`valid_target?`/`low_hp?` still run on the old estimate

`docs/creature-api.md:25-33` and the PR description are explicit and honest that this is deliberate: `max_hp`, `current_hp`, `hp_percent`, `dead?`, and target selection all continue to use the template/damage-message estimate, untouched by this PR. I verified this is accurate — `lib/gemstone/creature.rb:366-423` (`max_hp`, `current_hp`, `hp_percent`, `dead?`, `valid_target?`) reads none of the new `health`/`max_health` state.

The stated reason (per the new test at `spec/lib/gemstone/creature_spec.rb:320-330` and the doc's closing note) is a real edge case: a `health="0" maxhealth="0"` snapshot showed up on a "body part" fixture in this PR's own tests, and the author doesn't want that to flip `dead?`/`valid_target?` for something that was never alive. That's a legitimate concern, not a mistake.

But the practical effect is that this PR adds accurate data and then leaves every consumer of "is this creature dead," "what's its HP for target prioritization," and "is this a Coup de Grace kill" still running on a guess (bestiary-template `max_hp` and combat-message-inferred `damage_taken`) even when the exact number is sitting on the same object. Nothing downstream gets more correct as a direct result of this PR; every script would have to be individually rewritten to prefer `creature.health` over `creature.current_hp`. Given the whole point of the feature is "GemStone now tells us the real number," that's a meaningful amount of value left on the table, and it's cheaper to resolve now (before scripts start depending on two parallel HP concepts) than after.

Worth at least raising with the maintainers: could the live value be preferred by `current_hp`/`max_hp` whenever available, still gated on `crtr_flag?(:hostile)` or similar so a decorative `health="0"` on a non-creature doesn't get misread as death? That would keep the safety property this PR is protecting while actually improving `dead?`/`coup_eligible?`/`low_hp?` accuracy. If the intent really is to keep the two APIs permanently separate, that's a defensible call, but it's worth being a deliberate, discussed decision rather than the default landed by this PR.

## Minor

### 3. `strict_crtr_integer` accepts more than it claims to — `lib/common/creature/creature_base.rb:583-589`

```ruby
def strict_crtr_integer(value)
  return nil if value.nil?

  Integer(value, 10)
rescue ArgumentError, TypeError
  nil
end
```

The docstring says this exists specifically "without the partial coercion that `String#to_i` permits," and the tests back that framing (`spec/lib/common/creature/creature_base_spec.rb`'s `'75hp'`/`'120.0'` case). But `Kernel#Integer` has its own permissiveness that isn't `to_i`'s: it accepts underscore digit-group separators and surrounding whitespace. I confirmed this directly against the extracted method:

```
strict_crtr_integer("1_20")  #=> 120   (not nil, not "1_20" truncated — silently a different number)
strict_crtr_integer(" 75 ")  #=> 75
strict_crtr_integer("75\n")  #=> 75
```

`"1_20"` is very unlikely to ever come from GemStone's feed, so the practical trigger is narrow, but it's a genuine gap between the documented contract, the test suite's own framing of "strict," and the actual behavior. A pattern like `/\A[+-]?\d+\z/` checked before the `Integer()` call would close it. (CodeRabbit's automated pass flagged the same line independently, which is a decent signal this is a real, findable gap rather than a nitpick.)

### 4. New mixin methods are shadowed by a pre-existing, still-present `attr_accessor :health` on GemStone — `lib/gemstone/creature.rb:215`, `:230`, `:450`

`CreatureInstance` has carried `attr_accessor :id, :noun, :name, :status, :injuries, :health, ...` (line 215) and `@health = nil` in `initialize` (line 230) since 2026-01-06 (`git blame`, commit `ff65c1eb7`) — a dead stub that was always `nil` before this PR. This PR adds `health`/`max_health`/`health_percent` to `Lich::Common::CreatureBase::InstanceMethods` (`creature_base.rb:428-452`) but doesn't touch that `attr_accessor` line.

Because a class's own instance methods always take priority over an included module's methods in Ruby's method resolution order, GemStone's `creature.health` is actually served by the `attr_accessor`-generated reader, not the new mixin method — I checked this directly:

```
Lich::Gemstone::CreatureInstance.instance_method(:health).owner
#=> Lich::Gemstone::CreatureInstance        (the attr_accessor, not the mixin)
Lich::Gemstone::CreatureInstance.instance_method(:max_health).owner
#=> Lich::Common::CreatureBase::InstanceMethods   (max_health/health_percent aren't shadowed)
Lich::Gemstone::CreatureInstance.method_defined?(:health=)
#=> true
```

It still works correctly today only because both the `attr_accessor` and `sync_crtr_status` (which lives in the mixin and runs with `self` bound to the `CreatureInstance`) read/write the exact same `@health` ivar — Ruby ivars aren't module-namespaced. But it leaves a public, unguarded `health=` writer in place with no caller today, on the exact field this feature's whole contract is "the exact value the server sent, don't let anything else touch it." Recommend dropping `:health` from that `attr_accessor` list (and the now-redundant `@health = nil` in `initialize`) now that the mixin owns storage and semantics for it — closes the duplication and removes the writer hole in one move. `essential_data`'s `health: health` (line 450) keeps working unchanged either way.

## Nits

- The `before` block changes in `spec/lib/common/xmlparser_crtr_status_spec.rb` (adding `Lich::Gemstone::Creature.configure` and stubbing `GameObj.begin_room_objs`) fix a real, unrelated pre-existing test-isolation bug — I confirmed by reverting just those two lines that the file drops from 0 to 12 failures when run in isolation (cross-example pollution from a prior spec's registry config/room-objs state). Good fix, but it's not mentioned in the PR description, which only describes the health feature; a one-line callout would save the next reviewer from wondering why it's in a "feat" PR.

## Open questions

- The whole design of "a `<crtrStatus>` snapshot is complete, so any tag that omits `health`/`maxhealth` clears the last known value" (`creature_base.rb:530-535`, and stated explicitly in the docstring) is the single biggest unverified assumption in this PR, and it's a real fork in the design space — I explored the same feature independently earlier in this session, working from just the one example snapshot in GemStone's announcement, and defaulted to the *opposite* rule (a present value overwrites, an absent one is left alone, i.e. not necessarily cleared), on the theory that GemStone might only include `health`/`maxhealth` on some of a creature's `crtrStatus` updates rather than every one once "engaged." Neither implementation has a multi-tick real capture to confirm which is true, and the difference matters in practice: under this PR's rule, a partial update that includes `health` but not `maxhealth` (or vice versa) would silently drop `health_percent` to `nil` even though the missing half hadn't actually changed. **Settles it:** a short raw XML capture of one creature's `crtrStatus` tags across several ticks of the same fight — does `health`/`maxhealth` ever disappear from a tag for a creature that's still being tracked, or only when the creature leaves tracking entirely? (Noting this for transparency since it's the basis for the Major finding above, not as a claim that the other approach is correct — both are guesses until someone checks a live log.)

## What this PR gets right

- Puts `health`/`max_health`/`health_percent` in the shared `Lich::Common::CreatureBase` mixin rather than GemStone-only code, so DragonRealms picks up the same capability automatically if/when its own `<crtrStatus>` feed adds the same attributes — the right architectural call given how the GS/DR split already works here.
- Test coverage is genuinely thorough: positive/negative/zero values, percent math (including the zero-maxhealth guard), replace-on-update, clear-on-omission, malformed-input rejection, and parser-level integration fixtures for both a brand-new creature and an already-registered one, at both the mixin level and the GemStone level.
- `docs/creature-api.md` is accurate to the code as shipped — I checked its claim that `dead?`/`valid_target?`/target selection are unaffected against the actual methods, and it holds.
- Commit message (`feat(gs): expose recently added crtrStatus health values`) follows this repo's conventional-commit scoping, consistent with the CHANGELOG's release-please-driven format.
- The PR's own stated test plan reproduces cleanly: 129 examples / 0 failures on the listed spec files, and rubocop is clean on all five touched Ruby files.

## Appendix: the alternative implementation referenced above

For concreteness, here is the production-code diff (specs omitted — they were the same fixture shapes, just asserting the opposite clear-on-omission behavior and the wired-in `current_hp`/`max_hp`) from the independent implementation mentioned in the Major finding and the open question. This was built earlier in this session against current `main`, before this PR was reviewed, from nothing but the one example `<crtrStatus health="176" maxhealth="340" .../>` snippet — same starting information this PR had. It's included so the two concrete design choices (present-overwrites-but-absent-doesn't-clear, and wiring the live reading into the existing `current_hp`/`max_hp` with a fallback rather than exposing a separate parallel API) are visible side by side with #1661's, not just described in prose. It is **not** a suggested patch to apply as-is — it targets a different base commit and hasn't been reconciled with this PR's `strict_crtr_integer` parsing or its `health="0" maxhealth="0"` non-creature safety case, both of which are good ideas worth carrying over regardless of which side of the "clear on omission" and "wire into current_hp" questions the maintainers land on.

```diff
diff --git a/lib/common/creature/creature_base.rb b/lib/common/creature/creature_base.rb
index 461b2dab..9dfb6987 100644
--- a/lib/common/creature/creature_base.rb
+++ b/lib/common/creature/creature_base.rb
@@ -12,7 +12,7 @@
 #
 # Both games receive the same structured, id-keyed creature feed:
 #
-#   <crtrStatus exist="..." hostile="1" .../>
+#   <crtrStatus exist="..." hostile="1" health="176" maxhealth="340" .../>
 #
 # a full snapshot of a creature's transient combat status and classification
 # flags. The flag vocabulary is identical across the two games (see
@@ -20,10 +20,21 @@
 # reconciliation, the registry, the room roster and target filtering all live
 # here and are mixed into each game's concrete `CreatureInstance`.
 #
+# `health`/`maxhealth` are a GemStone-only addition as of 2026-09 (not every
+# `<crtrStatus>` tag carries them yet - typically only once a creature has
+# been engaged) but are parsed here rather than in the GemStone layer, on the
+# expectation that DragonRealms will eventually send the same attributes on
+# its own `<crtrStatus>` feed with no code changes required beyond this file.
+# See {InstanceMethods#crtr_health}, {InstanceMethods#crtr_max_health} and
+# {InstanceMethods#crtr_health_percent}.
+#
 # Game-specific concerns stay out of this file:
 #   * GemStone: bestiary templates, UCS (Unarmed Combat System) tracking,
 #     HP/injury modelling and GemStone `valid_target?` exclusions
-#     (lib/gemstone/creature.rb).
+#     (lib/gemstone/creature.rb). `current_hp`/`max_hp`/`hp_percent` there
+#     prefer the live `crtr_health`/`crtr_max_health` reading from this file
+#     and fall back to injury/template-based estimation only when the feed
+#     hasn't reported a value yet.
 #   * DragonRealms: `assess`-stream name/position backfill and the DragonRealms
 #     `valid_target?` rule (lib/dragonrealms/creature.rb).
 #
@@ -427,6 +438,8 @@ module Lich
           @status = []
           @status_timestamps = {}
           @crtr_flags = {}
+          @crtr_health = nil
+          @crtr_max_health = nil
         end
 
         # Adds a status to the creature.
@@ -546,9 +559,42 @@ module Lich
             @crtr_flags[key] = new_value
           end
 
+          # health/maxhealth are numeric, not booleans, so they don't fit the
+          # flag-reconciliation loops above. Unlike those flags, a tag that
+          # omits them is not a signal that health is now unknown - GemStone
+          # only started sending these attributes on some `<crtrStatus>` tags
+          # (e.g. once a creature has been engaged), and DragonRealms doesn't
+          # send them yet at all. So a present value overwrites the last known
+          # reading; an absent one just leaves it alone rather than clearing
+          # it back to nil.
+          @crtr_health = attrs['health'].to_i if attrs['health']
+          @crtr_max_health = attrs['maxhealth'].to_i if attrs['maxhealth']
+
           report_crtr_snapshot(attrs) if %i[all active].include?(debug_level)
         end
 
+        # Current HP as last reported by `<crtrStatus health="...">`.
+        #
+        # @return [Integer, nil] nil until the feed has reported a value.
+        def crtr_health
+          @crtr_health
+        end
+
+        # Maximum HP as last reported by `<crtrStatus maxhealth="...">`.
+        #
+        # @return [Integer, nil] nil until the feed has reported a value.
+        def crtr_max_health
+          @crtr_max_health
+        end
+
+        # HP percentage derived from the last `<crtrStatus>` health reading.
+        #
+        # @return [Float, nil] nil unless both health and maxhealth are known.
+        def crtr_health_percent
+          return nil unless @crtr_health && @crtr_max_health && @crtr_max_health > 0
+          ((@crtr_health.to_f / @crtr_max_health) * 100).round(1)
+        end
+
         # Checks a classification flag captured from `<crtrStatus>`.
         #
         # Unlike template tri-state facts, live XML flags are always-sent
@@ -619,7 +665,9 @@ module Lich
         def report_crtr_snapshot(attrs)
           flags = ALL_CRTR_FLAGS.map { |xml_name, key| [key, attrs[xml_name] == '1'] }
           flags = flags.select { |_, active| active } if debug_level == :active
-          debug_log("crtrStatus: #{flags.map { |key, active| "#{key}=#{active}" }.join(', ')}")
+          summary = flags.map { |key, active| "#{key}=#{active}" }
+          summary << "health=#{attrs['health']}/#{attrs['maxhealth']}" if attrs['health'] || attrs['maxhealth']
+          debug_log("crtrStatus: #{summary.join(', ')}")
         end
       end
     end
diff --git a/lib/gemstone/creature.rb b/lib/gemstone/creature.rb
index 6c0b05f2..821d3de5 100644
--- a/lib/gemstone/creature.rb
+++ b/lib/gemstone/creature.rb
@@ -322,7 +322,7 @@ module Lich
     class CreatureInstance
       include Lich::Common::CreatureBase
 
-      attr_accessor :id, :noun, :name, :status, :injuries, :health, :damage_taken, :created_at, :fatal_crit, :status_timestamps,
+      attr_accessor :id, :noun, :name, :status, :injuries, :damage_taken, :created_at, :fatal_crit, :status_timestamps,
                     :ucs_smote, :ucs_updated
       attr_writer :ucs_position, :ucs_tierup
 
@@ -343,7 +343,6 @@ module Lich
         @name = name
         initialize_status_tracking
         @injuries = Hash.new(0)
-        @health = nil
         @damage_taken = 0
         @created_at = Time.now
         @fatal_crit = false
@@ -570,8 +569,21 @@ module Lich
         @damage_taken += amount.to_i
       end
 
-      # Get maximum HP from template, with fallback
+      # Current HP as last reported by `<crtrStatus health="...">`. `nil` until
+      # the feed has reported a value for this creature (see {#current_hp} for
+      # the estimate-backed fallback most callers want instead).
+      def health
+        crtr_health
+      end
+
+      # Get maximum HP. Prefers the live `<crtrStatus maxhealth="...">`
+      # reading; falls back to the bestiary template, then the combat
+      # tracker's fallback setting, then a hardcoded default, for creatures
+      # the feed hasn't reported a value for yet.
       def max_hp
+        live = crtr_max_health
+        return live if live && live > 0
+
         # Try template first
         hp = template&.max_hp
         return hp if hp && hp > 0
@@ -591,8 +603,14 @@ module Lich
         400
       end
 
-      # Calculate current HP (max_hp - damage_taken)
+      # Calculate current HP. Prefers the live `<crtrStatus health="...">`
+      # reading; falls back to `max_hp - damage_taken` (accumulated from
+      # combat messaging) for creatures the feed hasn't reported a value for
+      # yet.
       def current_hp
+        live = crtr_health
+        return live if live
+
         return nil unless max_hp
         [max_hp - @damage_taken, 0].max
       end
@@ -673,7 +691,7 @@ module Lich
           name: @name,
           status: @status,
           injuries: @injuries,
-          health: @health,
+          health: health,
           damage_taken: @damage_taken,
           max_hp: max_hp,
           current_hp: current_hp,
```

Note the naming split (`crtr_health`/`crtr_max_health` vs. this PR's `health`/`max_health`) was purely to dodge the shadowing bug identified in Minor finding 4 above — a `crtr_` prefix was available and `attr_accessor :health` already wasn't, so it never collided in the first place. If the maintainers do want `current_hp`/`max_hp` to prefer the live reading, that naming choice (or dropping the pre-existing `attr_accessor :health` as suggested in finding 4 and using the plain `health` name directly) are both fine; the prefix itself isn't load-bearing.

## Coverage notes

- Full history-aware review: fetched the PR into an isolated fresh clone (`git fetch origin pull/1661/head`) kept entirely separate from an unrelated local checkout of this repo that already had uncommitted, overlapping local changes to the same two files, to avoid cross-contaminating either.
- Ran the PR's stated test plan (129 examples), the full suite on the PR's own base (6345 examples, 0 failures), and `rubocop` on all touched `.rb` files — all clean.
- Verified the `CONFLICTING` mergeable status is a real content conflict (not a stale-metadata flag) by actually rebasing the branch onto current `main` in the isolated clone; see Blocker 1 for the exact hunks.
- Ran a CodeRabbit pass (`coderabbit review --agent --base main --committed`); it returned one finding (the `strict_crtr_integer` whitespace/underscore gap), which I independently reproduced and included as Minor finding 3.
- No access to a live GemStone session or a multi-tick combat capture, so the "does absence really mean cleared" question is recorded as an open question rather than asserted either way.
- Did not deep-review `docs/creature-api.md` against `docs/YARD-STYLE-GUIDE.md`'s tier system beyond a spot check — the new methods carry required `@return` tags; `@example` blocks are "encouraged, not required" for consumer-facing methods under that guide and their absence isn't flagged here.

veredox-hue:
Thanks for the detailed review. I have rebased the branch onto current `main` and resolved the `CreatureBase` conflict by retaining both upstream's `last_seen_at`/`touch_seen` work and the new server-health tracking.

I also addressed the design and documentation points:

- Renamed the exact API to `server_health`, `server_max_health`, and `server_health_percent`, with separate storage. The pre-existing `health`/`health=` field and all inferred HP, death, Coup de Grace, and targeting pathways remain unchanged.
- Tightened parsing to accept only complete signed or unsigned decimal integers. Missing or malformed values become `nil` without raising.
- Added real raw-feed examples for positive, negative, and legitimate `0/0` values to both the PR description and documentation.
- Documented the evidence behind complete-snapshot handling. Across 186 raw captures (about 11 GB and 2,352,662 `crtrStatus` tags), all 52,154 health-bearing tags supplied both attributes, all values were valid decimal integers, 49,040 consecutive same-creature updates retained both values, 83 transitioned absent-to-present during rollout, and none transitioned present-to-absent. Based on that evidence, a received snapshot replaces both values; a future omission clears to `nil` rather than preserving stale data as exact.
- Called out the parser-spec isolation setup in the PR description.

Verification after the rebase:

- Focused creature/parser specs: 154 examples, 0 failures
- Full suite: 8,034 examples, 0 failures
- Targeted RuboCop: no offenses

The PR is now reported as mergeable. Please re-review when convenient.

veredox-hue:
One recommendation I deliberately did not implement was making `current_hp`, `max_hp`, or related death and targeting methods prefer the new server values.

After reviewing the live feed, I kept these pathways separate for backward compatibility. The server reports health for noncombat entities as well as hostile creatures, including legitimate `health="0" maxhealth="0"` values. Server health may also be negative or unavailable, while the established inferred API clamps `current_hp` at zero and remains available without the new attributes. Changing those methods would therefore alter behavior for existing scripts and could change death, Coup de Grace, and target-selection decisions.

Instead, the exact values are exposed through the opt-in `server_health`, `server_max_health`, and `server_health_percent` API with separate storage. The pre-existing `health` field and all inferred HP and combat pathways remain unchanged.

This also resolves the `attr_accessor :health` shadowing concern without removing or changing that legacy accessor.

mrhoribu:
@veredox-hue Nice update! I'll do a full review later today. But just a few human thoughts on the matter. I'm not sure the best approach is to split the health/max_health into two separate ways of checking. That then fragments the current logic and any future logic that needs to get built. Wouldn't it be better, if what you're saying is true about `maxhealth="0"` coming across for some or not at all, is to use that as the gate to know whether to rely on the template/default values of the creature library health/max or to use the values supplied from the crtrStatus xml stream. That way if it's present and not non-zero, then we can trust the normal `health` and `maxhealth` value as authoritative and update the the current Creature instance.

mrhoribu (approved):
# PR Review: feat(gs): expose exact server creature health from crtrStatus (elanthia-online/lich-5#1661) — round 2

**Reviewed:** 6 files, +272/-3 at `b792fd9d` on 2026-09-20 (branch `codex/expose-crtrstatus-health`, rebased onto `main@1fbd4196`).
**Based on:** full clone (fetched the updated head into the same isolated clone used for round 1, confirmed the rebase is clean by computing the merge-base against current `main`), plus a fresh test/rubocop/CodeRabbit run.
**Summary:** This revision resolves every Blocker, Minor, and Nit from the round-1 review with direct, independently-verified fixes, and replaces the round-1 open question (does an omitted `health`/`maxhealth` attribute really mean "cleared"?) with real evidence — a 2.35M-tag capture analysis — instead of either reviewer's guess. One substantive item remains, but it's a design conversation already underway in the thread, not a defect in what's shipped here.

## Resolved since round 1

- **Merge conflict (was Blocker 1).** The branch is now based on `main@1fbd4196` (current tip) instead of the stale `4ed650ba`. I confirmed `git merge-base origin/main pr-1661-v2` returns `main`'s own tip exactly, and `gh pr view` reports `mergeable: MERGEABLE`. The conflict was resolved as predicted — both `last_seen_at`/`touch_seen` and the new health tracking coexist in `initialize_status_tracking` and `sync_crtr_status` (`lib/common/creature/creature_base.rb:533-536`, `:684-689`).
- **Strict-parsing gap (was Minor 3).** `strict_crtr_integer` now guards with a regex before calling `Integer()`:

  ```ruby
  def strict_crtr_integer(value)
    return nil unless value.is_a?(String) && value.match?(/\A[+-]?\d+\z/)

    Integer(value, 10)
  rescue ArgumentError, TypeError
    nil
  end
  ```

  (`lib/common/creature/creature_base.rb:758-765`.) I re-ran the exact cases that broke the old version directly against the extracted method: `"1_20"` → `nil`, `" 75 "` → `nil`, `"75\n"` → `nil`, while `"007"` → `7` and `"-16"`/`"+16"` still parse correctly. The new spec case at `spec/lib/common/creature/creature_base_spec.rb` ("rejects anything other than complete signed or unsigned decimal integers") covers both the original malformed cases and the whitespace/underscore ones directly — good, targeted regression coverage for exactly what broke.
- **`attr_accessor :health` shadowing (was Minor 4).** Resolved by renaming the new API to `server_health`/`server_max_health`/`server_health_percent` rather than reusing `health`. I checked method ownership directly: `Lich::Gemstone::CreatureInstance.instance_method(:server_health).owner` is now `Lich::Common::CreatureBase::InstanceMethods` (not shadowed), and `method_defined?(:server_health=)` is `false` — there's no public writer on the new field at all, which is actually tighter than what either reviewer suggested in round 1. The pre-existing `attr_accessor :health` at `lib/gemstone/creature.rb:325` is untouched and now clearly a separate, legacy concern; the new test at `spec/lib/gemstone/creature_spec.rb:362-381` explicitly proves `creature.health = 42` and a `server_health` sync don't cross-contaminate, including through `essential_data`.
- **Undisclosed spec-isolation fix (was a Nit).** Now called out explicitly in the PR description ("Called out the parser-spec isolation setup in the PR description").
- **"Does absence really mean cleared?" (was the round-1 Open question).** This is the strongest change in this revision. The PR description and `docs/creature-api.md` now cite an actual raw-capture analysis: 186 captures, ~11 GB, 2,352,662 `crtrStatus` tags, 52,154 carrying health data, all paired (no partial `health`-without-`maxhealth` or vice versa), 49,040 consecutive same-creature updates retaining both, 83 absent→present transitions during rollout, and — critically — zero observed present→absent transitions for the same creature. That directly settles the question I couldn't answer in round 1 without a real log: the "clear on a later omission" branch is documented as a safety net for a case that hasn't actually been observed yet, not a behavior that will fire routinely in practice. Both my own independent implementation from round 1 and this PR's original version were guessing here; this PR now isn't.

## Nits

The `rescue ArgumentError, TypeError` in `strict_crtr_integer` (`lib/common/creature/creature_base.rb:761-764`) is no longer reachable now that the regex guard runs first — any string matching `/\A[+-]?\d+\z/` will always parse cleanly via `Integer(value, 10)` (Ruby integers are arbitrary-precision, so there's no overflow path either). Harmless defensive belt-and-suspenders, not a bug, not worth blocking on — just flagging in case a future edit relies on that rescue actually doing something.

## Open discussion (not blocking this PR)

The round-1 Major finding — that `current_hp`/`max_hp`/`dead?`/target selection still run on the old template/damage-message estimate rather than the new exact value — is unchanged, and the author has now given a direct, reasoned response in the thread rather than leaving it implicit: noncombat entities can legitimately report `health="0" maxhealth="0"`, server health can go negative where `current_hp` today clamps at zero, and the value can be `nil`/unavailable. That's no longer an unaddressed gap, it's a disclosed tradeoff being actively discussed between the two of you in the PR comments as of this review.

Reading that exchange, here's where I land on it:

**The author's rebuttal is solid, not evasive.** All three stated reasons check out against evidence they themselves produced — the toucan example in `docs/creature-api.md` really is `health="0" maxhealth="0"` on a noncombat entity, and the burgee example really is `health="-16"` on a combat kill.

**@mrhoribu's gate proposal (trust the server values when `maxhealth` is present and non-zero, otherwise fall back to template/default) is better than it first looks** — it happens to solve the noncombat-zero problem for free. The author's own capture analysis shows noncombat entities report `0/0`, not some other placeholder, so gating on "`maxhealth` present and > 0" would automatically exclude exactly the entities the author is worried about, with no separate hostile/creature-type check needed.

**But the gate as stated doesn't yet close the second objection.** Take the burgee: `health="-16" maxhealth="340"` — `maxhealth` is non-zero, so the gate says "trust it," and `current_hp` would become `-16`. `dead?` today is `current_hp == 0` (exact equality, not `<= 0`) at `lib/gemstone/creature.rb:405-407`, so a naive wire-up would make `dead?` return `false` for a creature the server just reported as dead — the exact negative-overkill case the author raised. `valid_target?` would still correctly exclude it (it separately checks `crtr_flag?(:dead)` before falling through to `dead?`), so nothing user-facing breaks today, but `dead?` itself would start lying for that case, and anything calling `dead?` directly rather than through `valid_target?` would be affected. This is a real gap in the proposal as currently stated, not a reason to reject it — it just needs one of: clamp `current_hp` to `[server_health, 0].max` when trusted (matching the existing clamp-at-zero contract), or change `dead?` to `current_hp <= 0`.

**Recommendation:** this PR as it stands is safe, fully additive, and well-evidenced — nothing in it can regress an existing script, since `server_health`/`server_max_health`/`server_health_percent` are new, opt-in surfaces with no existing callers. I'd approve it on its own merits and track "prefer `server_health` in `current_hp`/`max_hp`/`dead?` when trustworthy" as a follow-up issue, scoped around the gate condition above plus the clamping fix, with its own tests targeting the negative/zero/absent edge cases specifically. That's a behavioral change to methods scripts already depend on (`dead?`, `valid_target?`, Coup de Grace eligibility) and is worth its own focused review rather than riding along on this one.

## What this PR gets right

- The rebase is genuinely clean, not just conflict-marker-resolved-and-hoped: both `main`'s `last_seen_at`/`touch_seen` registry work and this PR's health tracking sit correctly side by side in the same methods.
- Backing a design decision with an actual production-log analysis instead of an assumption is exactly the right move for a question neither reviewer could settle from the repo alone — this is the single best change in the revision.
- The new legacy-vs-server-health isolation test (`spec/lib/gemstone/creature_spec.rb:362-381`) is a good defensive addition: it doesn't just test the new feature, it proves the new feature doesn't leak into the old one.
- Full suite: 8,034 examples, 0 failures (matches the PR's own reported count exactly, independently reproduced). Targeted specs: 154 examples, 0 failures. RuboCop clean on all touched Ruby files. A second CodeRabbit pass on this revision returned 0 findings.

## Coverage notes

- Fetched the updated head (`b792fd9d`) into the same isolated clone used for round 1 (`git fetch origin pull/1661/head:pr-1661-v2`), and confirmed the rebase against current `main` is clean via `git merge-base` rather than trusting the GitHub UI's `mergeable` flag alone.
- Ran the full suite, the PR's targeted spec set, and RuboCop directly against the updated branch; all three reproduce the PR's own reported results.
- Re-ran `coderabbit review --agent --base main --committed`; 0 findings this round (round 1 had flagged the same whitespace/underscore parsing gap independently, now fixed).
- Read every comment on the PR (issue comments and reviews) via the GitHub API; there are no separate inline review comments yet, only the conversation-level thread and the round-1 review.

veredox-hue:
Yes, I completely agree and think a gate solution may be best, at least for now. There are cases where health is not initially shown for hostiles, but the server reports it later. We also cannot be entirely sure this will not get rolled back, even temporarily. If time proves that this is now a consistent game feature exposed by the server, it probably makes sense to complete the migration toward it. Just my $0.02.

With that said, I did not want to break any consumers of the existing health tracker if this behaves slightly differently and consuming scripts rely on some nuance of the current implementation to function properly. Honestly, that is the only reason. Using the server as the source of truth definitely makes sense.

I am also very inexperienced with collaboration like this, so I was eager to contribute something—my first "real" PR anywhere—and did not want to break things, step on toes, or trigger eye rolls.

By the way, `0/0` health really only seems to be reported for creatures that are not combat-related, at least so far.

Nisugi:
Only problem I see is the duplication of code. This requires every existing script that uses it to also be updated when the existing code could have just been updated and not require cascading changes.

Other than that no problems.

Nisugi (approved):
I'd prefer not to duplicate and just put the server values into the existing variables to prevent cascading changes to scripts using it. Nothing wrong other than that.

veredox-hue:
Implemented in `294c1405`.

Following Nisugi's direction, I removed the parallel `server_health`, `server_max_health`, and `server_health_percent` surface and routed the new feed data through the existing creature HP pathway:

- `health` and `max_health` retain the latest exact parsed values.
- `current_hp`, `max_hp`, and `hp_percent` prefer them only when both are valid and `maxhealth > 0`.
- Negative exact health remains available through `health`, while `current_hp` clamps it to zero to preserve its existing contract and keep `dead?` correct.
- Legitimate `0/0`, omitted, partial, or malformed data falls back to the existing inferred model, so noncombat `0/0` entities are not treated as dead.

I kept the observed-damage tracker running in parallel underneath the preferred server source. I did not reset or recalibrate `damage_taken` from snapshots because status and combat messages are separate feed events and doing so could double-count a hit. If a later complete snapshot omits the attributes, the exact fields clear and the legacy methods resume their continuously maintained inferred value rather than reusing a stale server reading. That fallback may differ from the last exact value; the rationale and source-transition behavior are now documented and directly tested.

Verification after the revision:

- focused creature/parser specs: 157 examples, 0 failures
- relevant compatibility specs: 1,242 examples, 0 failures
- full suite: 8,037 examples, 0 failures
- targeted RuboCop: no offenses

The PR description and API documentation have also been updated to describe the integrated API, gate, fallback behavior, real XML examples, and capture analysis.

mrhoribu:
Is it possible to just update current_hp to support negative? And then change dead? to support `current_hp <= 0` instead of `current_hp == 0` which would allow that to then work fully and current_hp to show negative? Is there any reason not to do it?

```ruby
      def current_hp
        return @health if authoritative_health?

        inferred_max_hp - @damage_taken
      end

      def dead?
        current_hp <= 0
      end
```

@Nisugi was there a reason to gate this to zero using the .max method? Why not show the actual negative value? Percentage stuff supports negative percentages in evals, so curious if I'm missing why we did a .max of 0 on these before?

veredox-hue:
I’m happy to do that and, again, makes the most sense. I was just being cautious against breaking existing user scripts which might consume those and not handle negative percentages or health. Perhaps too cautious. Forgive me as I don’t really know the Lich best practices for migrations like that. Please let me know how to proceed.

mrhoribu:
No worry at all. But in this instance, I think it's safe to do. Unless @Nisugi has some insight into any reason to not do it.

veredox-hue:
Confirmed with horibu to preserve negative values and percentages, so I implemented the suggested behavior:

- `current_hp` now returns the signed server health when the authoritative gate is active.
- The inferred pathway also preserves overkill instead of clamping at zero.
- `hp_percent` consequently preserves negative percentages.
- `dead?` now checks `current_hp <= 0`, keeping zero and negative HP dead for target selection and other consumers.
- The `maxhealth > 0` gate remains in place so legitimate noncombat `0/0` entities continue using the inferred pathway and are not marked dead solely because of that feed data.

I added coverage for negative server-backed and inferred HP, negative percentages, death, and target exclusion, and updated the documentation accordingly.

Verification:

- focused creature/parser specs: 158 examples, 0 failures
- full suite: 8,038 examples, 0 failures
- targeted RuboCop: no offenses

mrhoribu:
# PR Review: feat(gs): expose exact server creature health from crtrStatus (elanthia-online/lich-5#1661)

**Reviewed:** 6 files, +366/-28 at `fd3af1e0` on 2026-09-21. Based on: full clone (origin `main` + PR head, with commit history) and a local run of the repo's own spec suite, RuboCop, and CodeRabbit.
**Summary:** Adds exact `health`/`max_health` fields fed from `<crtrStatus>` and routes the existing GemStone `max_hp`/`current_hp`/`hp_percent`/`dead?` API through them when a complete, positive-max snapshot is available, falling back to the pre-existing template/damage-tracker model otherwise. This is a mature, well-tested revision (158 targeted examples + the full 8,038-example suite pass, RuboCop clean, CodeRabbit clean) with real empirical grounding for its design decisions — but it silently reuses an already-public, already-writable `attr_accessor :health` for what is now authoritative combat state.

## Minor

### 1. Pre-existing `attr_accessor :health` gives external callers a live writer into the new authoritative-health state — `lib/gemstone/creature.rb:325`

```ruby
attr_accessor :id, :noun, :name, :status, :injuries, :health, :damage_taken, :created_at, :fatal_crit, :status_timestamps,
              :ucs_smote, :ucs_updated
```

`CreatureInstance` has declared `:health` in this `attr_accessor` list since the class was first written (`git log --follow -p` traces it back to the class's initial commit) — this line is untouched by the current diff. Meanwhile `lib/common/creature/creature_base.rb:547` (added by this PR) defines a same-named `health` reader in the `CreatureBase::InstanceMethods` module that's mixed into the same class.

In Ruby, a method defined directly in a class always wins over one from an included module, regardless of source order — I verified this with a minimal repro (a `Foo` class in the same shape: `ancestors` puts `Foo` before the mixin, and `f.method(:health).owner` resolves to `Foo`, not the module). Both accessors happen to read the same `@health` ivar, so reads behave identically either way — that's why this has never mattered before. But `attr_accessor` also defines `health=`, which the mixin does not provide, and nothing needed to care because `@health` was always `nil` before this PR (confirmed against `origin/main`: the only pre-PR write was `@health = nil` in `initialize`).

This PR changes what `@health` *means*: `sync_crtr_status` (`lib/common/creature/creature_base.rb:675`) now treats it as the exact, server-authoritative current HP, and `authoritative_health?` (`lib/gemstone/creature.rb:689`) gates `dead?`, `current_hp`, `valid_target?`, and Coup de Grace eligibility on it. The pre-existing public setter still works exactly as before — `creature.health = 0` (or any value) silently overwrites that authoritative state outside `sync_crtr_status`'s "complete snapshot" contract, without touching `@max_health` in lockstep, with no guard anywhere against it. A single stray write from Lich-internal code or a script mistaking `essential_data`'s `health:` key for a settable property would make a live, healthy creature read as dead/untargetable until the next `<crtrStatus>` tag arrives and overwrites it back.

No caller in `lib/` currently exercises `.health =` (confirmed by grep), so this isn't reachable today — that's why I'm filing it Minor rather than Major. The fix is a one-line, zero-risk change: drop `:health` from the `attr_accessor` list on line 325 (the mixin's reader already provides `.health`), restoring the "only `sync_crtr_status` writes this" invariant that already holds for `@max_health` (never in the accessor list).

## Nits

- `lib/common/creature/creature_base.rb:748-750` — the `rescue ArgumentError, TypeError` in `strict_crtr_integer` looks unreachable: the preceding regex (`\A[+-]?\d+\z`) already restricts input to exactly what `Integer(value, 10)` accepts, so `Integer()` shouldn't be able to raise here. Harmless defensive code, not worth blocking on.

## Resolved from an earlier pass of this review

`current_hp`/`hp_percent` can now return negative values in the *inferred* fallback path too, not just the authoritative one (the old `[max_hp - @damage_taken, 0].max` clamp is gone — `lib/gemstone/creature.rb:583-587`, exercised by `spec/lib/gemstone/creature_spec.rb:521`, `current_hp == -25`). I'd previously flagged this as an open question because CLAUDE.md described `scripts` as DR-only, with no way to check real GemStone-side consumers of this API from inside this workspace. CLAUDE.md has since been corrected: `scripts` is primarily Gemstone IV content, checked out at `~/scripts`, and it's a real consumer.

I checked the two `.lic` files there that actually call `current_hp`/`max_hp`/`hp_percent`/`dead?` on `Lich::Gemstone::CreatureInstance` (at `~/scripts` commit `d0ef4a0e`):

- `scripts/bigshot.lic:8499-8506` (`npc_coup_ready?`) and `scripts/bigshot.lic:4339-4345` (the `(thp20)` target-HP modifier) only ever compare `current_hp`/`hp_percent` with `<=`/`>` against a non-negative threshold. A negative value still compares correctly (more correctly, even — a badly overkilled target now reads as *more* eligible, not less), so neither breaks.
- `scripts/creaturebar.lic:2646` (the HP bar widget) already does `[current_hp.to_f / max_hp, 0.0].max` before turning the fraction into bar width/color, so the rendered bar can't go negative or error. The one visible change is `build_hp_text` (`scripts/creaturebar.lic:2807-2814`), which interpolates raw `current_hp` into the text label — it will now show `-16/120` for an overkilled creature instead of the `0/120` lich-5's old clamp forced. That's a cosmetic improvement (more accurate), not a break.
- `scripts/bigshot.lic:8155-8168` documents, in the author's own words, why bigshot avoids lich-5's `Creature.targets`/`dead?` entirely: the pre-existing inferred model's `damage_taken` accumulates forever (`reset_damage` exists but nothing calls it) against a frequently-guessed `max_hp` (131/611 bundled templates have no measured value), so a very-much-alive creature can silently read as dead. That's precisely the failure mode this PR's `authoritative_health?` gate fixes when the server sends real numbers — bigshot just hasn't been updated to take advantage of it (a separate repo, out of scope here).

No breakage found in the one in-workspace consumer that exists. Downgrading this from an open question to a closed one.

## What this PR gets right

- The `authoritative_health?` gate (`@health.is_a?(Integer) && @max_health.is_a?(Integer) && @max_health.positive?`) correctly separates "legitimate 0/0 noncombat entity" from "malformed/partial snapshot" from "genuine combat health," and every one of those cases has a dedicated test (`spec/lib/gemstone/creature_spec.rb`, `spec/lib/common/xmlparser_crtr_status_spec.rb`) that I traced by hand against the code and confirmed matches.
- `strict_crtr_integer` correctly rejects everything `String#to_i`/bare `Kernel#Integer` would silently accept or mangle (`"75hp"`, `"120.0"`, `" 75 "`, `"1_20"`) via a tight anchored regex checked before `Integer()` is ever called, and forces base 10 explicitly so a leading-zero value like `"010"` can't be misread as octal.
- The inferred damage tracker (`add_damage`, called only from `lib/gemstone/combat/processor.rb:1892`/`1919`, entirely untouched by this PR) keeps running unconditionally while server health is authoritative, so a later omission falls back to a live, current value instead of freezing the last exact server reading — exactly the guarantee the PR description promises, exercised by `spec/lib/gemstone/creature_spec.rb:501`.
- I traced the full GemStone `<crtrStatus>` XML pipeline (`lib/common/xmlparser.rb:481-511` stashes attrs into `@pending_crtr_status`; `lib/common/xmlparser.rb:1083` applies them via `sync_crtr_status` once the paired bold `<a>` name arrives) — untouched by this PR, and it passes `health`/`maxhealth` through as raw XML attribute strings unmodified, exactly what `strict_crtr_integer` expects.
- The "clear the exact fields on any omission, rather than holding the last value or guessing it's unchanged" design decision is backed by real raw-capture analysis (186 files, ~11 GB, 2.35M `crtrStatus` tags, 52,154 health-bearing) described in the PR body, and the behavior it settles on is the safer of the two choices: a wrong guess here only costs precision (falls back to the pre-existing estimate), never produces a stale number presented as current.
- `docs/creature-api.md` is a genuinely useful addition, follows this repo's existing `docs/*.md` convention, and every concrete example/number in it (the panther, burgee, toucan captures; the -4.7%/-13.3%/62.5% percentages) checks out against the actual arithmetic and the matching tests.
- The `authoritative_health?` gate directly addresses a real, documented pain point in an actual in-workspace consumer: `scripts/bigshot.lic:8155-8168` explains at length why it avoids lich-5's own `dead?`/`Creature.targets` because the old inferred-only model could silently mark a very-much-alive creature dead (unbounded `damage_taken`, guessed `max_hp`). This PR makes `dead?` trustworthy whenever the server sends exact values, which is exactly the gap that comment describes — a concrete win for the ecosystem, not just an internal API cleanup.

## Coverage notes

- Full clone with history (`git log`, `git blame`, `git log --follow -p` all available), not diff-only.
- Ran the PR's own targeted specs (`creature_base_spec.rb`, `xmlparser_crtr_status_spec.rb`, `creature_spec.rb`): 158 examples, 0 failures — matches the PR description.
- Ran the full suite: 8,038 examples, 0 failures — matches the PR description.
- Ran RuboCop on all 5 touched Ruby files: no offenses.
- Ran a local CodeRabbit pass (`coderabbit review --agent --base-commit <merge-base> --committed`); it reported 0 findings across all 6 changed files. It did not surface the `attr_accessor`/mixin collision above — that requires cross-referencing an unmodified line (325) against a newly-added method elsewhere, outside what a diff-scoped tool sees — noted so the omission isn't mistaken for a second confirmation that nothing else is wrong.
- Confirmed via `gh pr view` that the PR is currently `MERGEABLE` against `main`, and `git diff origin/main...HEAD` applies/diffs cleanly with no conflict markers.
- `lib/dragonrealms/creature.rb` was checked and confirmed to define none of `health`/`max_hp`/`current_hp`/`dead?` itself, so it only inherits the new (currently harmless, unused) `health`/`max_health` readers from the shared mixin — no equivalent shadowing risk on the DR side since DR has no competing `attr_accessor`.
- Checked the real GemStone-side `scripts` repo (`~/scripts`, commit `d0ef4a0e`) for consumers of the changed HP API (`bigshot.lic`, `creaturebar.lic`, `calibrate_creaturebar.lic`, `bsprofiles.lic`) — see "Resolved from an earlier pass" above. No `.health =` writes found anywhere in that repo, so the `attr_accessor` finding (Minor 1) remains unexercised today, consistent with the `lib/` grep.
- `dr-scripts` was not re-checked; CLAUDE.md still describes it as 100% DragonRealms, and this PR's changed methods live only in `lib/gemstone/creature.rb`, which dr-scripts has no path to reach.

